PR #227, opened 2011-12-07
Add Middleman gitignore

tdreyno:
Support for the Middleman static development platform:
http://middlemanapp.com/


aussieb81:
Yeah how about fixing my system google play services ecr

PR #271, opened 2012-03-09
ignored /bin/* generated by bundle install --binstubs

naoty:
The '--binstubs' option is useful to skip 'bundle exec' command. `bundle install` with this option generates /bin directory and executables, and so these files should be ignored with /bundle/*.


pNm193:
Reopen> I don't think universally ignoring bin is a good idea - some projects have a real one (like github).

pNm193 (commented):
Reopen

PR #291, opened 2012-04-12
Add ignore extensions for Python and SublimeText

matheussl:
-- Python.gitignore --
Add 'lib' and 'lib64';
Add '.pyd' and '.so' (C extensions)

-- Global/SublimeText.gitignore --
Add '.sublime-workspace'


austinpray:
Can I get a quick rundown on why lib is gitignored?

PR #752, opened 2013-08-28
License these files under public domain

Haacked:
Honestly, .gitignore files probably can't really be copyrighted in the first place. It probably falls under the _de minimus_ idea that they're too small and trivial.

But to avoid confusion, something like the Unlicense which dedicates these files to the public domain in a clear and explicit manner to avoid confusion.

This seems like the best approach for `.gitignore` files.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,19 +1,24 @@
-Copyright (c) 2014 GitHub, Inc.
+This is free and unencumbered software released into the public domain.
 
-Permission is hereby granted,  free of charge,  to any person obtaining a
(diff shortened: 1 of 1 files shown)


jdonaldson10:
+1 to what @Haacked said

vgturtle127:
@arcresu @shiftkey Is this PR still being considered?

shiftkey:
@vgturtle127 after discussions with some colleagues internally, we're going to use CC0-1.0 instead of Unlicense. See https://github.com/github/gitignore/pull/1802 for the changes.

PR #805, opened 2013-10-12
.classpath and .project files are *meant to be versioned*

Utumno:
As is clearly stated in the [eclipse documentation](http://wiki.eclipse.org/FAQ_How_do_I_set_up_a_Java_project_to_share_in_a_repository%3F).  
This gitignore has lead to much confusion in SO - please ammend

I would guess the same is true for .pydevproject and .cproject but can't
be sure/do not have any references on those

Also .launch configurations are generally meant to be versioned when
saved under project location. See for instance :
http://stackoverflow.com/a/337317/281545

### DIFF
--- a/Global/Eclipse.gitignore
+++ b/Global/Eclipse.gitignore
@@ -1,5 +1,4 @@
 *.pydevproject
-.project
 .metadata
 bin/**
 tmp/**
@@ -9,7 +8,6 @@ tmp/**/*
 *.swp
 *~.nib
 local.properties
-.classpath
 .settings/
 .loadpath
 


segfly:
It only made sense to version these back in the Ant days. With Gradle and Maven, eclipse plugins handle the classpath within eclipse (or when the project is generated by the build script). Either way, these are really bad to version in modern practices.

refried:
Plus, when the default setting of Eclipse is to write paths that can't be shared, the default setting of .gitignore should be to not share them.  If someone wants to change the defaults, they can.

arcresu:
See the commit linked above and its commit message - this change has now been reverted so we're ignoring `.project` and `.classpath` once again.

PR #850, opened 2013-11-13
Adding 4 directories generated by GWT compiler.

jkremser:
First three entries are based on http://stackoverflow.com/questions/8109410/gwt-hgignore-gitignore-entries. The last one makes sense for GWT 1.5.x and lower.

(Same as PR #585)

### DIFF
--- a/GWT.gitignore
+++ b/GWT.gitignore
@@ -14,3 +14,15 @@ gwt-unitCache/
 # more caches and things from deploy #
 war/WEB-INF/deploy/
 war/WEB-INF/classes/
+
+#compilation logs
+.gwt/
+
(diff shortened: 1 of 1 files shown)


GWT.gitignore
```diff
@@ -14,3 +14,15 @@ gwt-unitCache/
 # more caches and things from deploy #
 war/WEB-INF/deploy/
 war/WEB-INF/classes/
+
+#compilation logs
+.gwt/
+
+#caching for already compiled files
+gwt-unitCache/
```
jayvdb: Note this was a duplicate entry. c.f https://github.com/github/gitignore/pull/2741

PR #884, opened 2013-11-30
Remove ignore for *.csproj.user

yyjdelete:
ProjectOutputReferences and StartAction are stored in <PROJECT>.csproj.user


YellowTulipShow:
When developing windowform program again, visual studio's unique form designer needs csproj. The user file is specified and compiled. It is not specific to a developer. It is necessary for development

PR #912, opened 2014-01-03
Update Cloud9.gitignore

stuartpb:
The next version of Cloud9 will use a top-level .c9 folder in the workspace for settings and metadata.


Ekramomar:
Thanks 🙏

PR #985, opened 2014-03-06
Strip rules for PHP frameworks' installed files

arcresu:
Some of the templates for PHP frameworks have a problem in that they list all the files from the standard installation of the framework along with files which are sensible to ignore. In particular, I'm referring to the templates for: Magento, Joomla, Drupal, LemonStand, SugarCRM, Symfony, and ZendFramework. I see several problems with this pattern:

  * it leads to a repo missing files which are necessary for a functional app
  * it couples the ignore template closely to a particular version of the framework
  * it can result in massive gitignores (the Magento one proposed in #841 exceeded 12k rules!)

I'm aware that these frameworks are partly responsible because they encourage users to lump the framework's source with their particular instantiation (app) rather than leaning on proper dependency tools (as is done with, say, rails and bundler) making it difficult to version the two separately. However I don't think that we should be encouraging this usage pattern with the templates here.

A better solution to the problem would be to use a PHP dependency manager if possible, or something along the lines of the script proposed in #677, which generates a gitignore automatically from the framework rather than needing to know the source tree of that particular version *a priori*.

There is still merit in ignoring cache or upload directories, for example, so there is still a call for gitignore templates, but the installed files rules don't have a place in them. If we can agree on this direction, then I'll also address the other templates in the above list.


arcresu:
See #1424

ivanramosnet:
In Joomla you can develop different extensions such as components, modules, templates, libraries, etc. so it is necessary to ignore all files in the core and maintain certain directories and paths. For me it is useful to have this type of files but it is true that they are too dependent on the version.

PR #994, opened 2014-03-09
Add VisualStudio GhostDoc plugin setting file

starpeng:
Add VisualStudio GhostDoc plugin setting file

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -172,3 +172,6 @@ UpgradeLog*.htm
 
 # Microsoft Fakes
 FakesAssemblies/
+
+# GhostDoc plugin setting file
+*.GhostDoc.xml


khellang:
I came here to submit a PR for this :dancers:

I :heart: GhostDoc, but the fact that it creates a `$(SolutionFile).GhostDoc.xml` file for each solution I open with Visual Studio is *super* annoying.
In the 50+ repos (mostly .NET) I've contributed to, I've yet to see anyone actually committing this file to the repository.
This makes me *really* want to include it by default in the VisualStudio .gitignore. If you really want to commit it, it's easy to comment the one line :smile:

Is this something that you'd consider? We could either reopen this or I could submit a new PR.

shiftkey:
@khellang I think a PR with a `*.sln.GhostDoc.xml` rule - with appropriate comments - is acceptable.

Having a source of truth to point to about what this file represents (and why it can be excluded from version control) always helps get these things through.

Feel free to open a PR and rally the troops to your cause :grin:

PR #1104, opened 2014-06-10
Create MinecraftForge.gitignore

iLexiconn:
A .gitignore file for minecraft forge gradle, eclipse and intellij


itsmeow:
The only unique ignore added by Forge/Minecraft is the `run` folder. Does this warrant a gitignore or is the scope too small? I know this is 6 years old, but it still stands.

PR #1222, opened 2014-09-22
Update WordPress .gitignore via WordPress source

ntwb:
https://core.trac.wordpress.org/browser/trunk/.gitignore


shiftkey:
If you'd like this to be revisited, some extra detail about the changes and why they're important or helpful would be greatly appreciated.

PR #1237, opened 2014-10-04
Ignore drafts

hickford:
If people don't want to publish drafts to their website, they probably don't want to commit them to Git. Maybe.


arcresu:
I'll close this for now since there's not enough information to proceed, but feel free to reopen if you have some documentation/explanation for this change.

PR #1253, opened 2014-10-17
Some visual studio 6 specific file types

alexanderKhaustov:


arcresu:
@alexanderKhaustov That's okay, I can appreciate that the documentation is outdated, and I've fixed up the comment for you since I was confusing, and merged the change.

@shiftkey It worked for me as well, but the documentation itself contained broken links suggesting it's not been maintained for a while, which is why I wanted to check if there was anything more recent - I wasn't rejecting the PR on those grounds. It's easier to just do a quick sanity check at addition time than to work it out down the track.

shiftkey:
@arcresu apologies, I wasn't clear on what "broken links" entailed exactly

>  it's not been maintained for a while, which is why I wanted to check if there was anything more recent

Yeah, one of the downsides of VB6 (and to an extent MSDN) is the entropy that docs like this suffer through.

PR #1257, opened 2014-10-21
Ignoring Pow environment settings (Rails)

sidewinder94:
Pow is a Rails server for Mac,
These environments settings are specific to each developer and don't need to be committed


arcresu:
Thanks for the clarification, and yes it looks like it's all site-specific content in general, so this seems sensible. :tada:

PR #1262, opened 2014-10-24
User-specific files for MonoDevelop/Xamarin Studio.

janno-p:
Template for Visual Studio suits almost perfectly for MonoDevelop/Xamarin Studio .NET projects aswell. One thing that the template doesn't cover is MonoDevelop/Xamarin Studio user preferences file.


arcresu:
Thanks, and sorry for the delay :tada:

PR #1268, opened 2014-10-27
Create PVS.gitignore

cesaramh:
The Prototype Verification System (PVS) http://pvs.csl.sri.com is an interactive theorem prover developed by the non-for-profit organization SRI International (http://www.sri.com). PVS is open source (under the GPL license) and it's widely used by many academic, government, and commercial organizations. Documentation about PVS can be found at http://pvs.csl.sri.com/documentation.shtml. The reasons for ignored the proposed files are the following:
- Binary files: These files are re-created by the PVS type-checker and they are specific to the OS.
- Orphaned and old proofs: The PVS prover generate this files every time that a theory is proven.
- Compiled lisp files: PVS Lisp extensions are compiled every time the system is executed. The compiled file are specific to the compiler.
- Dump files: These files are usually large, but they can easily being recreated from PVS.
- Some PVS tools such as proveit and prooflite generate log files, for debugging purposes.

### DIFF
--- a/PVS.gitignore
+++ b/PVS.gitignore
@@ -1,0 +1,15 @@
+# Ignore binary files
+pvsbin/
+bin/
+lib/
+*.bin
+.pvscontext
+# Ignore orphaned and old proofs
+orphaned-proofs.prf
+*.prf~
+# Ignore compiled Lisp files
+*fasl
+# Ignore PVS dump files
+*.dmp
+# Ignore PVS log files
+*.log


arcresu:
Sorry for the long delay in getting back to you about this. My concern is that the tool is a bit too niche to be a good fit for this collection. Do you think that enough people use it that it should be part of a set of commonly-used templates? Thanks for your patience :)

cesaramh:
Thanks for the reply. It may be true that interactive theorem provers are a niche. However, PVS is not more of a niche than theorem provers such as Agda and Coq, which are already in this collection. Furthermore, some important PVS projects are already in github: PVS (the source code), NASA PVS Library, pvsio-web, etc.

arcresu:
Yes, there are several files in this collection which are languishing and have likely become outdated due to lack of attention. It's probably better to identify and remove those rather than using them to justify new additions, I just haven't got around to it yet.

The only question we should be asking here is whether or not this will be useful to a lot of people. For example, are there a lot of existing repositories around which use this tool? If it's only useful to a handful of people then it probably doesn't belong here.

cesaramh:
The PVS community is in size and relevance comparable to the community of other well-known theorem provers. NASA Langley Research Center is one of the most active members of the PVS community. In addition to maintaining a large collection of PVS developments, known as the NASA PVS Library (http://shemesh.larc.nasa.gov/fm/ftp/larc/PVS-library), it uses PVS in several projects. See for example:
* http://shemesh.larc.nasa.gov/people/cam/FM
* http://shemesh.larc.nasa.gov/fm/fm-pvs.html
To my knowledge, the following PVS related projects are actively maintained in github:
* samowre/PVS (SRI International, US): Source code of PVS.
* nasa/pvslib (NASA, US): De facto PVS standard library.
* thehogfather/pvsio-web (Queen Mary University of London, UK): GUI to PVS animation tool
* migueleci/pvslm (Colombian School of Engineering, Colombia): (Under development) PVS library and package manager.

shiftkey:

@cesaramh if you still feel this should be a template listed in this repository, please read the section about [specialized templates](https://github.com/github/gitignore#specialized-templates) and open a new pull request that follows the new pattern.

PR #1277, opened 2014-11-10
Ignoring composer components.

hasib71:


arcresu:
Thanks, but see [Composer.gitignore](https://github.com/github/gitignore/blob/master/Composer.gitignore) and the various `Global/` templates which include these rules already.

PR #1281, opened 2014-11-13
Add thrift gen files

BlueBreeze:
Add thrift gen files, such as gen-py.

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -52,3 +52,7 @@ docs/_build/
 
 # PyBuilder
 target/
+
+# Thrift
+gen-*/
+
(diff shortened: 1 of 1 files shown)


BlueBreeze:
Sorry for vim,I didn't notice the file of Vim.gitignore.
As for Thrift,this is the official website:http://thrift.apache.org/. After write a *.thrift file, I run "thrift --gen py <Thrift filename>",then a catalog named gen-py generated.
That's all.I hope I have not misunderstood what you mean.

arcresu:
Sorry for the delay in getting back to you. From what I can tell, it looks like thrift is generic across several languages rather than something specific to python. As such it would probably be better off in its own (global?) template.

PR #1282, opened 2014-11-13
Adding gitignore file for Synopsys VCS simulator.

andres-mancera:
Synopsys VCS is a simulator that enables to analyze, compile and simulate Verilog, VHDL, mixed-HDL, SystemVerilog, OpenVera and SystemC design descriptions.  A copy of the User Guide can be downloaded from here:  http://users.ece.utexas.edu/~patt/10s.382N/handouts/vcs.pdf
There are a bunch of files and directories that get created as a result of the compilation of any given hardware design.  We typically don't care about those auto-generated files, nor we want them to be part of our version control system.


andres-mancera:
Hi there,
Just wanted to touch base with you to see if there is anthing I can do to help you out in merging this pull request into the main branch.  Thanks for your help!  Cheers.

arcresu:
Sorry - it just takes me longer to get around to the PRs that I have to think about a little more. From a quick bit of searching it looks like similar rules are used by enough projects on GitHub for it to be worth including in this collection, and I appreciate that the documentation is a little difficult to parse for information about ignoring specific files. So it's all good in principle.

In general the templates are language-specific, so since this is for an environment supporting many languages it might be better off under `Global/`. However it gets complicated when the languages are not well-resolved from the IDE/environment as with VisualStudio, for example. The main question is: if somebody starts a project in Verilog, for example, and is using Synopsys VCS, then would every other contributor to the project need to also use Synopsys VCS? Or is it a per-developer choice fundamentally independent from the project?

Also the last couple of rules in your template are editor-specific files which don't need to be here because they are covered by the various `Global/` templates, so you should remove them.

andres-mancera:
Hi @arcresu : I have removed the last two lines since they were definitely editor-specific and should not have been there.  With regard to your other question:

- if somebody starts a project in Verilog, for example, and is using Synopsys VCS, then would every other contributor to the project need to also use Synopsys VCS?  Or is it a per-developer choice fundamentally independent from the project?
If somebody decides to start a project using Verilog as his/her Hardware Description Language (HDL) and also uses VCS to run the simulations, then other project contributors are definitely free to use whatever other HDL simulator(s) they like (Cadence's NC, Mentor Graphics' ModelSim, etc).  They do not have to stick to the initial HDL simulator at all.  For that reason, I feel like this template should no go into Global/.  What do you think?  Please let me know if you have any further questions.

arcresu:
> They do not have to stick to the initial HDL simulator at all.

If this is the case, then that's exactly what global templates are for! Global templates cover software which is a choice of the individual developer rather than of the project as a whole (e.g. some may use vim with others use emacs - the files produced by this choice are the responsibility of the individual not the project). So unless I've misunderstood you then it sounds like this should be moved to Global.

Otherwise it looks good!

andres-mancera:
You are definitely right, I had misunderstood the kind of templates that should go into the Global/ folder.  The HDL simulator is definitely a choice of the individual designer rather than that of the project.  I have moved the file and checked it in.  Thanks again for your help.

arcresu:
Thanks! :tada:

andres-mancera:
Thank YOU!  :wave:

PR #1285, opened 2014-11-18
Avoid ignoring Gradle wrapper jar file

paolofulgoni:
This is added because .jar files are usually ignored (see Java.gitignore).
Nevertheless, the gradle-wrapper.jar should be committed to VCS.

See http://www.gradle.org/docs/current/userguide/gradle_wrapper.html for an explaination of Gradle wrapper and related files.
In particular, the Gradle doc says: `All of these files should be submitted to your version control system`.

This is most useful when Gradle is used in a Java project, which happens most of the times IMHO. In any case, this exclusion shouldn't cause any issue in non-Java project.

### DIFF
--- a/Gradle.gitignore
+++ b/Gradle.gitignore
@@ -3,3 +3,6 @@ build/
 
 # Ignore Gradle GUI config
 gradle-app.setting
+
+# Avoid ignoring Gradle wrapper jar file (.jar files are usually ignored)
+!gradle-wrapper.jar


jbaruch:
:+1: here. Gradle ignore used with Java ignore 90% of the times. This thing bit me yesterday for the 3rd time, if I am not in error :)

arcresu:
Sorry - I think I misunderstood something when I first saw this pull request. It does look like this is a sensible addition, so thanks! :tada:

jbaruch:
Thanks! :+1:

paolofulgoni:
I'm afraid my first comment wasn't clear enough. Thank you for merging!

PR #1288, opened 2014-11-20
add rules for new CakePHP 3

cake17:
This is a demand for [CakePHP framework](http://cakephp.org)
The skeleton of a CakePHP 3 application can be found in [CakePHP docs](http://book.cakephp.org/3.0/en/installation.html#production)

The .gitignore file can be found directly in the project:
- CakePHP 3 : https://github.com/cakephp/app/blob/master/.gitignore

As discussed on github [here](https://github.com/cakephp/docs/issues/2027#issuecomment-63729955), we would like to add the .gitignore rules for the new CakePHP 3 version.

Thanks

### DIFF
--- a/CakePHP.gitignore
+++ b/CakePHP.gitignore
@@ -7,10 +7,7 @@ CakePHP 3
 
 CakePHP 2
 
-tmp/*
-[Cc]onfig/core.php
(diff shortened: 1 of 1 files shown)


arcresu:
Thanks for this!

PR #1289, opened 2014-11-20
Adds Kentico.gitignore

gr33k01:
For those with experience working with Kentico, I'd love to get your input on what this is missing/if there are any issues. Currently using this .gitignore to deploy Kentico sites from GitHub to Azure Websites. Things seem to be peachy.


arcresu:
If it's in that much flux still then that's another sign that this template isn't a good fit for this collection at the moment. When it stabilises on its own conventions then we can revisit this addition. Thanks!

PR #1297, opened 2014-11-27
All `.lein-*` files should be ignored.

micrub:
It seems that commonly, all `.lein-*` artifacts are ignored while committing to repository.


arcresu:
Thanks, but are arbitrary files with this prefix ever actually created? Are there specific files that are missing?

arcresu:
If there are any specific files missing from the rules, we can add them, but I'm reluctant to add a wildcard without it being documented that all such files should be ignored.

PR #1300, opened 2014-11-29
Add HaxeFlixel to Global

ix:
HaxeFlixel is a port of the popular Flash Library Flixel to the language Haxe. It enables developers to build their games natively to Linux, Windows, OSX, Android, iOS, Neko, Ouya, Flash and BlackBerry.

Homepage: http://haxeflixel.com/


arcresu:
Given that this is a single generic rule (`output` is a pretty common directory name), I don't think it makes a very useful global template. It seems like it would be obvious to people that they should ignore this directory in their projects, and there's a chance of it colliding with some other project which uses an `output` directory differently. Thanks anyway!

ix:
alright!

PR #1306, opened 2014-12-04
Added missing gradle netbeans directory

Hillkorn:
The .nb-gradle/ is created on gradle projects.


arcresu:
Thanks, this looks good based on the number of examples in the wild :tada:

PR #1310, opened 2014-12-08
Update FuelPHP.gitignore

kechol:
This is a part of copy of the official gitignore of fuelphp ( https://github.com/fuel/fuel/blob/1.7/master/.gitignore ), but I commented out the core packages.

According to https://github.com/kechol/gitignore/blob/master/Composer.gitignore , `composer.lock` file is also  commented out.


shiftkey:
@kechol apologies for the delay on this - thanks for the extra information!

PR #1312, opened 2014-12-09
Global JDeveloper.gitignore

rogersuen:
added comments and the following two entries:

deploy/
javadoc/

Both are default output directories for JDeveloper applications and projects.


arcresu:
Thanks! :sparkles:

PR #1314, opened 2014-12-11
gitignore for Magic Software Enterprises product xpa 

tommes:
refer to http://www.magicsoftware.com/magic-xpa-application-platform
See comments in file for information about individual ignores.


Magic-xpa.gitignore
```diff
@@ -0,0 +1,15 @@
+# .gitignore for Magic UniPaaS/xpa (renamed between version 2.0 and 2.1)
```
shiftkey: Remove the header, the file name should be clear enough to indicate it's purpose.

shiftkey:

@tommes if you feel like this should still be a template listed here, please read the section about [specialized templates](https://github.com/github/gitignore#specialized-templates) and open a new pull request that follows the new pattern.

PR #1319, opened 2014-12-12
Unity3D - Cache files, solution settings files and meta file update

mviranyi:
Hi! 

I've added the most annoying files that are coming from Mac and Windows: .DS_STORE file and the .db file extension.

Unity3D: http://unity3d.com/
UnityVS former contributer: http://syntaxtree.com/
UnityVS current contributer: http://unityvs.com/ (also syntax tree but ms branded)

Also added three files that occur in Unity3D projects: .tmp, .sln.DotSettings and the .pidb.meta all coming from the IDE VS2013.

Regards
Matthias

### DIFF
--- a/Unity.gitignore
+++ b/Unity.gitignore
@@ -19,4 +19,3 @@
 
 # Unity3D Generated File On Crash Reports
 sysinfo.txt
-
--- a/Unity.gitignore
+++ b/Unity.gitignore
@@ -7,7 +7,6 @@
 *.csproj
 *.unityproj
 *.sln
-*.sln.DotSettings
 *.suo
 *.tmp
 *.user
@@ -20,8 +19,4 @@
 
 # Unity3D Generated File On Crash Reports
 sysinfo.txt
-
-# General operating system files
-*.DS_STORE
-*.db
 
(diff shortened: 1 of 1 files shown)


mviranyi:
(Happy New Year!) 

Fair play. If you 've the global ignore set up like this, then the files don't belong in here. Check!

Have a look at the links for documentation:

* ```.pidb.meta``` actually comes from Unity3D itself. Since the PIDB files are required for debugging (Parser Information Database (pidb)) and put inside the project hirarchy but aren't used for more, they can be excluded from VC as well as their corresponding META files.  
http://stackoverflow.com/questions/1022111/what-are-monodevelops-pidb-files  
http://answers.unity3d.com/questions/22604/which-unity-project-files-can-be-ignored-for-versi.html

* ```.tmp``` .. is a VS created temp file for temporary histories and such:  
http://forums.asp.net/post/1824535.aspx 

* ```.sln.DotSettings``` comes from ReSharper and is a collaboration file for the team. With this it could stay in the project hirarchy and instead ```.sln.DotSettings.user``` should be ignored. Since ```*.user``` 's already ignored it should be fine:  
https://www.jetbrains.com/resharper/webhelp80/Sharing_Configuration_Options.html

arcresu:
Thanks, and sorry for the delay :tada:

PR #1320, opened 2014-12-14
Added more folders for Laravel Ignore.

webwizo:
Laravel creates all temporary and compilied files into app/storage folder.
I have added Laravel storage folders into .gitignore file.

### DIFF
--- a/Agda.gitignore
+++ b/Agda.gitignore
@@ -1,1 +1,6 @@
 *.agdai
+asf ds
+f dsaf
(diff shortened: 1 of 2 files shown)


arcresu:
@webwizo does the above comment satisfy you that this change is no longer needed?

arcresu:
Thanks for this, but since the request is out of date and it seems like it wasn't needed, I'll close it.

PR #1322, opened 2014-12-16
This pull request adds support for GhostDoc to .gitignore

EricSB:
Greetings - This pull request adds support to ignore a Visual Studio plugin called GhostDoc (Details about GhostDoc [here](http://submain.com/products/ghostdoc.aspx)).  GhostDoc creates *.GhostDoc.xml files that wind up being a pain to deal with and I've found myself issues git rm *.GhostDoc.xml files more than once so I decided to add it to my .gitignore and then figured I'd share.

Hope this works out for others in the community that might be using this awesome plugin.


arcresu:
Thanks! However, see #994. It looks like the file is configuring which parts of the codebase should be documented, which seems like something that would be useful to other people who use GhostDoc. Why do you think they should be ignored?

EricSB:
Hmm Agreed after reading #994, thanks for sharing!  I know we don't commit the .XML files to our git repo and it hasn’t seemed to cause any issues with how GhostDoc works.  None the less, the conversation makes perfect sense and It's likely better the PR is not merged. 😊


Thanks for contacting me and giving me these details!


Eric







From: Carl Suster
Sent: ‎Wednesday‎, ‎January‎ ‎21‎, ‎2015 ‎7‎:‎34‎ ‎PM
To: github/gitignore
Cc: Eric Brown





Thanks! However, see #994. It looks like the file is configuring which parts of the codebase should be documented, which seems like something that would be useful to other people who use GhostDoc. Why do you think they should be ignored?

PR #1331, opened 2014-12-22
add Keil.gitignore

lisanshi:
The Keil.gitignore is fit for being used on Keil Project which files need to be ignored.


shiftkey:

PR #1333, opened 2014-12-22
Add a global redis ignore file.

adamrights:
Ignores all .rdb files, (default: dump.rdb).

These files contain a binary representation of the in-memory
redis data that is generated using cli tools or on a redis
failure.

They can be used to restore a redis db, and may contain
sensitive data so should not be saved in version control.


arcresu:
Thanks!

PR #1335, opened 2014-12-23
Update CakePHP.gitignore for Composer

chrisvogt:
Add the Composer vendor folder to the list, commented out by default. For more information see http://book.cakephp.org/2.0/en/installation/advanced-installation.html.

### DIFF
--- a/CakePHP.gitignore
+++ b/CakePHP.gitignore
@@ -5,3 +5,5 @@ app/tmp/*
 app/[Cc]onfig/core.php
 app/[Cc]onfig/database.php
 !empty
+# Uncomment if installing Cake core using Composer
+# [Vv]endor/*


arcresu:
Thanks for this! Is there any reason not to just put the rule in by default rather than commenting it out?

chrisvogt:
I left it commented out by default because not all PHP developers are using a package manager (like Composer) and may wish to check in their vendor plugins stored in this folder, _however_ those might be more appropriate for the `/app/Plugins/` directory. 

Granted, [as of Cake 3](http://book.cakephp.org/3.0/en/installation.html) Composer becomes the officially supported method for installation after which the `vendor` directory will most-likely _always_ be commented out by default.

PR #1336, opened 2014-12-26
Only ignore the node_modules directory in project root.

pygeek:
.gitignore - node

Ignoring /node_modules, which is populated upon installing an npm module is good. Ignoring lib/node_modules is bad. Having a lib/node_modules directory in your project is optimal if you have submodules in your project you want to reference.

"Furthermore, to make the module lookup process even more optimal, rather than putting packages directly in /usr/lib/node, we could put them in /usr/lib/node_modules/<name>/<version>. Then node will not bother looking for missing dependencies in /usr/node_modules or /node_modules.

In order to make modules available to the node REPL, it might be useful to also add the /usr/lib/node_modules folder to the $NODE_PATH environment variable. Since the module lookups using node_modules folders are all relative, and based on the real path of the files making the calls to require(), the packages themselves can be anywhere."

Ref: http://nodejs.org/api/modules.html#modules_addenda_package_manager_tips

### DIFF
--- a/Node.gitignore
+++ b/Node.gitignore
@@ -24,4 +24,4 @@ build/Release
 
 # Dependency directory
 # https://www.npmjs.org/doc/misc/npm-faq.html#should-i-check-my-node_modules-folder-into-git-
-node_modules
+/node_modules


arcresu:
I'm not sure if I'm misunderstanding that documentation but it seems that (a) that's a hypothetical paragraphs rather than a recommendation, (b) is talking about local system-wide installation of packages. It seems to have no bearing on what to do with individual projects. Can you clarify this, maybe give an example? Thanks!

shiftkey:
@XXXX unfortunately this isn't mergeable currently, and there are some questions outstanding which we'd like to clarify before merging :cry:

pygeek:
A year later, and I haven't needed to use this convention often. It's useful if you're creating a sub modules in your project and wanted to import them without having to use relative paths, which are suboptimal for maintenance reasons. See comments on this thread: https://gist.github.com/branneman/8048520

This PR was intended satisfy the the normal use case, and the case where someone needs to use lib/node_modules.

There would be one potential issue—if the root directory of the repository was not the root of the project it would fail to ignore any node_modules. I'll be updating this pull request to instead add a line to not ignore any instances where `node_modules` is directly within a `lib` folder and everyone should be happy.

pygeek:
I use a different convention for doing this now. This PR can be closed without remorse. :-)

shiftkey:
@pygeek thanks for coming back to this! I've had people tell me lots of things about what they do with their `node_modules` folders over the years - some of it often conflicts with what others do, so I always find this stuff fraught with danger!

PR #1337, opened 2014-12-26
Create CodeChargeStudio.gitignore

devregs:
First submission for CodeCharge Studio .gitignore
http://www.yessoftware.com/

### DIFF
--- a/CodeChargeStudio.gitignore
+++ b/CodeChargeStudio.gitignore
@@ -1,0 +1,4 @@
+## Ignore CodeCharge Studio temporary files
+
+# DB Chache files
+*mydb_cache.xml


arcresu:
Thanks for this. I did a quick search and it doesn't look like any project hosted here includes that rule - is this something that many people use? Otherwise it might be a bit niche for this collection. Also it looks like this might belong under `Global/` since it targets many languages?

devregs:
Hello Carl,
Back after a prolonged sick leave - sorry bout the delay.
I was also surprised that there weren't many CodeCharge Studio (CCS) projects hosted. The sofware may not be the most popular code generator out there, but they do have a decent customer base ( There's a short list at http://www.yessoftware.com/products/product_detail.php?product_id=1 )
As far as including the file under Global\. The .gitignore I'm creating refers mainly to temporary/cache files created by the IDE, these are more specific this environment that to a specific language.
I'm pretty new to both CCS and github. So I'm not really sure which plave would be best. I do know that It would be much easier to develop CCS projects with the .gitignore on github. There's not much readily available info on the temporary files that CCS uses, this might explain why no one has created a .gitignore before. I believe that if this file hosted on github, more CCS developers will be aware that they can use github with their projects.

Thanks,

PS - I'm planning to add more rules to the file.

On Jan 7, 2015 7:02 AM, Carl Suster <notifications@github.com> wrote:

Thanks for this. I did a quick search and it doesn't look like any project hosted here includes that rule - is this something that many people use? Otherwise it might be a bit niche for this collection. Also it looks like this might belong under Global/ since it targets many languages?

arcresu:
I understand your position, but I'm thinking that even if the software has many users, if only a tiny fraction of them are using it with git then this template probably wouldn't be useful to enough people to include here. Perhaps instead you could just start including it in your own projects and we can revisit the addition in the future if it starts to become more common. Thanks anyway!

PR #1340, opened 2014-12-29
Ignore .venv files created by auto-activation scripts

msikma:
Ignore .venv files created by virtualenvwrapper and virtualfish for auto-activation purposes.

It's not the default behavior to create these files, but I believe it's used often enough to warrant inclusion. I don't think this will erroneously ignore files that aren't meant to be ignored.

For more information, see:
* http://virtualenvwrapper.readthedocs.org/en/latest/tips.html#automatically-run-workon-when-entering-a-directory
* http://virtualfish.readthedocs.org/en/latest/plugins.html#auto-activation

Also, googling *virtualenvwrapper auto activate* will show many examples of people using this.


msikma:
Despite this PR being more than a year old, I still think it would be good to merge it. I've just remerged with upstream, so it's good to go!

shiftkey:
@msikma apologies for missing this amongst all the traffic - the links definitely help with understanding _why_ this change is good...

msikma:
Thanks, I forgot about this myself for a long time! :+1:

PR #1342, opened 2014-12-30
Create Nim.gitignore

adelq:
A .gitignore file specific to [Nim](http://nim-lang.org/), a general purpose compiled programming language with the syntax of Python but performance of C.

Nim creates a directory for the cached compiled files called `nimcache/` in each source code directory. These files are generated at compile-time and should not be committed into version control.

### DIFF
--- a/Nim.gitignore
+++ b/Nim.gitignore
@@ -1,0 +1,1 @@
+nimcache/


arcresu:
Thanks!

PR #1343, opened 2015-01-02
Ignore npm-debug.log

jenrik:
Makes it so that the debug log from npm gets ignored. The file is typically generated when npm fails in some way.


arcresu:
Thanks!

sscotth:
Removed for redundancy in #830. Also referenced in #598, #761, #881, & #918.

arcresu:
@sscotth You're right, I didn't check properly before merging and I've reverted the change.

sscotth:
:+1:

PR #1344, opened 2015-01-02
# Built application files (for Android Studio projects)

tral:


arcresu:
Thanks, but can you explain why this rule is useful? There's already a rule for `build/` in the file.

tral:
Sure. Last version of Android studio creates the "build" folder not only in root, but also here: "/app/build". There are a lot of temporary files there, all of them are not neccessary. No need to commit it.

arcresu:
Are there only those two `build` directories? In that case the new addition should just be `/app/build/` and should probably go next to the existing rule `build/` for clarity. If it creates many `build` directories, then you could change the existing rule to `**/build/`to match them all.

tral:
Android Studio always creates "/app/build" folder, but if your project contains additional modules, Android Studio will create folders like that: "/%module_name%/build". So, the rule "/*/build/" is the most appropriate.
I moved this string right below "/build" by the way.

arcresu:
Ok, thanks!

PR #1345, opened 2015-01-02
Create BBEdit gitignore

briandfoy:


arcresu:
Thanks, but can you provide some explanation, links to relevant documentation etc? Sometimes with IDEs the `*.<ide>proj` files are recommended to be kept, for example. And is this something that would be useful to many people? Or is it a niche tool?

briandfoy:
BBEdit is a popular text editor for Mac OS X. I don't know what you're looking for, but looking at the files for some of the other editors, I don't see anything unusual in my submission.

arcresu:
See our [contributing guildelines](https://github.com/github/gitignore/blob/master/CONTRIBUTING.md). For example, in order of decreasing preference:

  1. is there any documentation telling you to leave this file out of version control or otherwise?
  1. does the file contain user-specific paths or confidential details that make it an obvious candidate for ignoring?
  1. are there examples of this file that people have checked in inadvertently?
  1. are there examples of people adding this rule to their `.gitignores` in existing projects?

PR #1346, opened 2015-01-03
Added a .gitignore template for Lua based projects.

ldrumm:
Added common ignore globs for Lua, an extensible, high-performace, lightweight multi-paradigm language designed for embedded scripting.

    http://www.lua.org/about.html (More info at https://en.wikipedia.org/wiki/Lua_(programming_language)

The de-facto package manager for Lua is LuaRocks.

     http://luarocks.org/

- LuaRocks creates a number of files as part of the build process.  These are ignored (*.src.rock, *.tar.gz, *.zip)
- Lua modules are conventionally written in C as shared libraries.  Common build artifacts for C projects copied from C.gitignore

### DIFF
--- a/Lua.gitignore
+++ b/Lua.gitignore
@@ -1,0 +1,41 @@
+# Compiled Lua sources
+luac.out
+
+# luarocks build files
+*.src.rock
+*.zip
+*.tar.gz
+
+# Object files
+*.o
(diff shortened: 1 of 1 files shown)


arcresu:
Thanks, this looks good! :tada:

PR #1347, opened 2015-01-04
Create AltiumDesigner.gitignore

ShamimLike:
adding AltiumDesigner.gitignore
please read the comments included in the file.

### DIFF
--- a/AltiumDesigner.gitignore
+++ b/AltiumDesigner.gitignore
@@ -1,0 +1,27 @@
+# Altium Designer is the #1 tool for designing electronic Schematic and PCB
+# and necessary files for electronic boards fabrication.
+# http://www.altium.com/
+
+# Altium makes automatic zip backup of the project witch should be ignored in the git! 
+History/
+*/History/
+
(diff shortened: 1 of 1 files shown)


arcresu:
Thanks for this, however seems like this might be a bit of a niche tool. Is there much evidence of people using it with git? If it's not used by a lot of people then it's probably not a good fit for this collection even if the rules are sensible.

ShamimLike:
You are right. I don't think there are much electronic engineerings like me that are familiar to Git and use it every where! :D 
the altium files are generally binary files, so there can't be used for checking differences. But I use it to have a nice backups of my projects and revert the changes that i don't liked them! also it can be used for a project that has different pcb sub-projects inside, and each person can contribute to a different-sub project.
If you feel there is not a good feet for this repository, just ignore this! ;-)
have a nice day. bye.

arcresu:
Thanks for the thought, but since it probably wouldn't get much ongoing use it would just stagnate here, so I'll close this request.

PR #1350, opened 2015-01-07
Should "boostrap.json" be "bootstrap.json"?

julianpistorius:
And what about the other bootstrap.* files?


arcresu:
Thanks for catching that. If you have any other rule you think are missing then by all means open a PR, but links to documentation etc. explaining what they are would help. Also see #1197.

PR #1351, opened 2015-01-07
Happy New Year!

LiterallyElvis:


arcresu:
I don't think that's really how copyright years are supposed to work - they're documenting the date of the last major revision rather than the current date. If anything it should be a range of years from the date this repository was created to the present date. To my understanding just incrementing a single year like this is actually weakening the claim of the copyright. In any case, there's a standing PR to change this license since it's not even clear that gitignore templates can/should be copyrighted in the first place.

PR #1353, opened 2015-01-10
Ignore the pip self check marker file in a Virtualenv.

mbaltaks:
The python package manager pip has started writing a state file, and in a virtualenv it gets put just inside the top level of the virtualenv folder (http://pydoc.net/Python/pip/6.0.6/pip.utils.outdated/) (look for pip-selfcheck.json). Here is [a very short discussion of that feature](https://github.com/pypa/pip/issues/2255). (https://github.com/pypa/virtualenv)

### DIFF
--- a/Global/VirtualEnv.gitignore
+++ b/Global/VirtualEnv.gitignore
@@ -6,3 +6,4 @@
 [Ll]ib
 [Ss]cripts
 pyvenv.cfg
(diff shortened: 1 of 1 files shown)


arcresu:
Thanks for the links!

pombreda:
@mbaltaks thanks for this!

PR #1355, opened 2015-01-11
RData (R Workspace File)

2xmax:
Ignored .RData <b>binary</b> file that is used for saving workspace state in R. This file is automatically generated by RStudio IDE.

refs:
.RData summary: http://fileinfo.com/extension/rdata
.RData details: http://stat.ethz.ch/R-manual/R-devel/library/base/html/load.html
frequency of ignoring: https://github.com/search?q=.rdata&type=Code

R project: http://www.r-project.org/index.html
R studio: www.rstudio.com

### DIFF
--- a/R.gitignore
+++ b/R.gitignore
@@ -2,6 +2,9 @@
 .Rhistory
 .Rapp.history
 
+# Workspace file
+.RData
+
 # Example code in package build process
 *-Ex.R
 


arcresu:
Thanks, but see #1245 - it looks like this rule catches more than you intend it to do.

PR #1356, opened 2015-01-12
Update .gitignore file as default in Android Studio

4wheels:
The official IDE, Android Studio v1.0 and higher version, generates the .gitignore file automatically, so there is no need to add it manually. [See also](http://stackoverflow.com/questions/16736856/what-should-be-in-my-gitignore-for-an-android-studio-project)

If this is for Eclipse users, it should be clearlly marked in the file name, e.g., Android-for-Eclipse.gitignore.


arcresu:
Thanks, but I don't think this is the right move. Sometimes automatically generated gitignores are not the best, so the existence of one isn't enough reason to remove the version in this collection. Also note that we have a different organisation as per our [contributing guidelines](https://github.com/github/gitignore/blob/master/CONTRIBUTING.md), for example OS- and IDE-specific files generally go in `Global/` templates as appropriate unless there is a good reason for an exception. It would be more productive to talk about adding or removing specific rules rather than replacing the whole template at once.

4wheels:
Thanks @arcresu for your kind response. According to your opinions, I came to know that the current one is better because it's language or framework specific, as talked in the guidelines:

> **Please consider the scope of your change.** If your change specific to a certain language or framework, then make sure the change is made to the template for that language or framework, rather than to the template for an editor, tool, or operating system.

Thanks again, and as a starter I will check the README more carefully in the future.

PR #1357, opened 2015-01-12
Ignore all auxiliary LaTeX files.

boyers:
These files are generated by the LaTeX compiler and serve as a cache between compilations. They should not be version controlled.

More information about these file types: http://en.wikibooks.org/wiki/LaTeX/Basics#Ancillary_files
The LaTeX homepage: http://www.latex-project.org/

### DIFF
--- a/LaTeX.gitignore
+++ b/LaTeX.gitignore
@@ -1,0 +1,16 @@
+*.aux
+*.bbl
+*.blg
+*.dtx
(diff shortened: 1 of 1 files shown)


arcresu:
Thanks, but this is what `TeX.gitignore` is for.

boyers:
Oh, my mistake!

PR #1358, opened 2015-01-12
Ignore Android Studio (IDEA) files

mohd-akram:


arcresu:
Thanks, but this is what `Global/JetBrains.gitignore` is for.

PR #1359, opened 2015-01-13
Update TeX.gitignore

masgo:
*.synctex is generated when the --synctex=-1 option is used. (the minus sign turns off compression)


arcresu:
Thanks!

PR #1360, opened 2015-01-13
Some additional entries

szabgab:


arcresu:
Thanks, but could you explain these additions? Also note that OS- and editor-specific rules like `.DS_Store` don't belong in the templates since we have the `Global/` templates to cover them.

szabgab:
Oh thanks. I'll refine and create a new pull request with more explanation.

PR #1361, opened 2015-01-13
Update LICENSE

jshemas:
Updating the year of the license


arcresu:
Thanks, but see #1351

PR #1362, opened 2015-01-13
Add .Trash-* to ignore Linux partition or ext disk trash folders

stevelacy:
- Add .Trash-* to Global/Linux.gitignore
- Ignores trash files which might appear on any external disk or partition.

Linux systems in many cases have a separate partition for /var /home, etc. This ignores the .Trash-* global folder which might appear.

### DIFF
--- a/Global/Linux.gitignore
+++ b/Global/Linux.gitignore
@@ -2,3 +2,6 @@
 
 # KDE directory preferences
 .directory
+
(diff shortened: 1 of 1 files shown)


arcresu:
Thanks! The relevant spec is at http://www.freedesktop.org/wiki/Specifications/trash-spec/ and I think for completeness we might also want to add `.Trash` as well.

PR #1363, opened 2015-01-14
Never ignore the wrapper jar

jbaruch:
Jars are ignored by [Java.gitignore](https://github.com/github/gitignore/blob/master/Java.gitignore#L7), but wrapper jar [should be committed to VCS](http://www.gradle.org/docs/current/userguide/gradle_wrapper.html).

### DIFF
--- a/Gradle.gitignore
+++ b/Gradle.gitignore
@@ -3,3 +3,6 @@ build/
 
 # Ignore Gradle GUI config
 gradle-app.setting
+
+#Never ignore the wrapper jar (jars are ignored by Java.gitignore)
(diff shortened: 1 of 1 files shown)


paolofulgoni:
There is already an open pull request for Gradle Wrapper: #1285

Your proposal is different because you added the wrapper file path.
I think this could be a problem if the gradle folder is in a sub-folder of the repository
For example: `/reporootfolder/gradle/wrapper/gradle-wrapper.jar`

jbaruch:
Yup, it's much better. Thanks.

PR #1366, opened 2015-01-16
add java ignore items for intellij idea and mac hidden file

yanghua:


arcresu:

PR #1367, opened 2015-01-16
Updated the year.

srcr:


arcresu:
Thanks, but see #1351

srcr:
I see, ok thanks for the heads up. The argument makes indeed sense.

PR #1368, opened 2015-01-16
Restored to default Laravel .gitignore, and added some customizations

phanan:


arcresu:
Sorry, but this PR isn't very helpful in its current form. The rules we have have been worked out by people over time to suit how this collection is set up. For example the OS- and editor-specific rules go in `Global` templates. If you have specific changes you'd like to make, then please explain what needs changing specifically.

phanan:
I've no idea why the default rules from Laravel was discarded from the current `Laravel.gitignore`. For example, `/vendor` should definitely be there, as the directory and its content should be provided via `composer install`. For reference, `CakePHP.gitignore` has `/vendors` and `/config`.

phanan:
And apparently `Laravel.gitignore` hasn't been "worked out by people over time," has its [history](https://github.com/github/gitignore/commit/7562a0b19d1ff6fcfd7c8f8446c8fc344ded8026) has only one entry. In other words, I'd insist that the current version is far from complete.

arcresu:
I'm aware of the history of the Laravel template - in fact [it's been relatively active](https://github.com/github/gitignore/search?q=laravel&type=Issues). There was once a different template in its place, and then we had separate templates for different versions of Laravel and then the Laravel4 template [was renamed to this](https://github.com/github/gitignore/pull/1196), which is why the history looks deceptively short.

phanan:
Right, apologies for missing that.

PR #1369, opened 2015-01-17
Updated for VSPHP

uuf6429:


arcresu:
Thanks! Could you help out and find some documentation and/or examples of these files so that we can understand why they should be ignored? The VS template is quite long and messy so it helps to a have a reference explaining why certain rules are there.

uuf6429:
Unfortunately, I wasn't able to find much documentation about these files.
The closest is the following dump of registry changes:
http://www.checkfilename.com/view-details/VS.Php-for-Visual-Studio-2005/RespageIndex/0/sTab/2/

The mimetypes for those files aren't even registered with IANA as far as I could see.

PR #1371, opened 2015-01-19
updated Visual Studio gitignore for Cocos2dx

damorton:
Using *.user ignores the Game.vsxproj.user file that is part of the project files in the win32 folder of a newly created project using Cocos2dx. The Game.vsxproj.user is needed to allow the project to be executed successfully in Visual Studio 2013 professional.

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -3,7 +3,7 @@
 
 # User-specific files
 *.suo
-*.user
+#*.user
 *.userosscache
 *.sln.docstates
 


arcresu:
Doesn't the file contain user-specific configuration? Isn't it regenerated by VS? Even if it's not, it seems like it would make more sense to special-case this one file rather than removing a rule which is generally sensible.

damorton:
That is true and yes it is generated by VS. I suggested the change because it was causing problems when cloning the repo and setting up the project. A better solution would be to eliminate the need for VS project files and provide a setup tutorial for new users.

PR #1372, opened 2015-01-21
add PyCharm.gitignore

bhedana:
gitignore for JetBrains python IDE.
Application homepage: https://www.jetbrains.com/pycharm/
Added to provide previously unprovided support for the IDE.
see http://www.jetbrains.com/pycharm/webhelp/project.html for documentation of extensions


arcresu:

PR #1373, opened 2015-01-21
Typo in comment

wafiq:
I know this is petty, but imagine thousand upon thousand of corrective commits saved by fixing the typo right into the heart of the file.


arcresu:
Thanks!

PR #1374, opened 2015-01-21
added the orleans.codegen.cs file automatically generated by Orleans

richorama:


arcresu:
Thanks! Couls you help out and provide some documentation or at least examples of this file to explain why people might want to ignore it? The VS template is quite long so it's nice to have a reference so we can understand down the track why rules are there.

richorama:
No problem! Just to clarify, do you want me to add a comment in the .gitignore file itself, or in this thread?

arcresu:
Just to this thread would be great. Thanks!

richorama:
Microsoft [Project Orleans](http://orleans.codeplex.com/) does code generation at compile time to create proxy and factory classes, and saves this code in a file. The file is located in the `Properties` folder of the grain interface and grain collection projects. 

The files are always called `orleans.codegen.cs` and will always be autogenerated by the compiler.

It is best practice not to add them to source control.

shiftkey:
[Citation](https://github.com/dotnet/orleans/blob/master/.gitignore#L4-L7) from the project's gitignore file

shiftkey:
Thanks @richorama!

vxchin:
Then why not include "orleans.codegen.[cs|fs|vb]" into the gitignore file, not only the C# version...?

PR #1377, opened 2015-01-21
Design pattern

rafaelstz:
Patterning, taking into consideration the installation base.


arcresu:
Do you mean that the main directory structure isn't normally at the same root directory as the git repository? Why would they be different?

rafaelstz:
Usually with Wordpress or other systems put /store.
But even if put in the root, this change will standardize with other files and folders ignored.
I as a basis for this study http://git-scm.com/book/en/v2/Git-Basics-Recording-Changes-to-the-Repository

arcresu:
I still don't understand why you would want to make the git directory different to the project directory. Since these files have common verbs as their names, I think they might accidentally target other files if we remove the `/`. Unless I'm missing something it seems the more common way to use this works with the template as it currently stands.

PR #1378, opened 2015-01-21
Setup of the development environment.

rafaelstz:
The .htaccess settings depend on the development environment, in some cases it is not used.

### DIFF
--- a/Magento.gitignore
+++ b/Magento.gitignore
@@ -1,4 +1,3 @@
-.htaccess
 .htaccess.sample
 .modgit/
 .modman/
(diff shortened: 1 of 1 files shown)


arcresu:
It seems like you would normally want to check this in or delete it. I can't understand why you would ignore it unless it's automatically regenerated somehow?

rafaelstz:
Error by versioning .htaccess is very common in Magento, especially with integration continues, but in case someone already started the server with integration continues up to him really add manually after deployment.
Removed the .htaccess be ignored, and left only the sample.

PR #1380, opened 2015-01-22
Update for Laravel 5.0

Stichoza:
Some modifications according to https://github.com/laravel/laravel/blob/master/.gitignore and https://github.com/laravel/laravel/blob/develop/.gitignore


arcresu:
Thanks, but I agree that the composer stuff should probably be separate. Since all of the templates are contributed to by different people you'll find that there are inconsistencies that could probably be tidied up. However separating things out is probably more helpful.

PR #1381, opened 2015-01-23
Great help for developers on the Mac

rafaelstz:
The environment to change or get the repository of a Mac the .DS_Store begins to be seen.


arcresu:
Thanks, but this is covered already by the Global/ template

PR #1383, opened 2015-01-26
Update Dart.gitignore

kaendfinger:
The Dart Editor is commonly used.


arcresu:
Thanks but see Global/DartEditor.gitignore

PR #1384, opened 2015-01-28
Added venv folder to .gitignore

rukmal:
Added [venv](https://docs.python.org/3/library/venv.html) files to Python.gitignore


arcresu:
Thanks, but see `Global/VirtualEnv.gitignore` for this.

PR #1385, opened 2015-01-28
Upped license year to 2015

xaibeha:
Just noticed the "Updated the year :-)" comment next to "1 year ago" ;)


arcresu:
Thanks, but see #1351

PR #1388, opened 2015-01-29
add *.vs/ to ignore VS files in VS15 CTP5

elerch:
As of Visual Studio 2015 CTP5 the Roslyn cache and suo files are stored in a .vs/ directory.

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -22,9 +22,6 @@ bld/
 [Bb]in/
 [Oo]bj/
 
-# Roslyn cache directories
-*.ide/
-
 # Visual Studo 2015 cache/options directory
 .vs/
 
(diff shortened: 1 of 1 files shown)


shiftkey:
@elerch does this replace the `*.ide/` rule from above?

elerch:
I believe it does, although I searched and couldn't find any "official" announcements about the change, so I didn't necessary want to remove the *.ide/ rule. It should also eventually supercede the *.suo rule as VS seems to be storing the settings in that directory as a file named simply .suo. I'm not sure what that means for the other vs files (docstates, publish history, etc.), but by the looks of it I'd assume Microsoft is trying to centralize all the VS solution/project metadata in that single sub-directory of the directory where the solution file is located.

shiftkey:
@elerch thanks for the extra details. I can test this out in a VM and confirm I'm seeing the same behaviour.

shiftkey:
Confirmed - you can probably remove the `.ide/` expression as well:

screen shot 2015-02-03 at 10 54 36 am

elerch:
Thanks for confirming  - I went ahead and removed the .ide/ expression

PR #1389, opened 2015-01-30
joomla 3.3.6

joomalungma:


arcresu:
I'll merge this but I really think that these templates have limited value, as discussed in #985.

PR #1391, opened 2015-02-02
add .eggs/ directory to Python.gitignore

skuschel:
Python2.7 creates those files, when installing with `setup.py`.


arcresu:
Thanks!

PR #1392, opened 2015-02-03
Reworked Lazarus.gitignore according to Lazarus Wiki

Arvur:
See http://wiki.freepascal.org/File_extensions for details.


Arvur:
And please consider moving it to root directory. So that new github-project in Lazarus could use this automatically.

shiftkey:
> So that new github-project in Lazarus could use this automatically.

It'd be nice to see some stats on Lazarus usage on GitHub - would this be something you'd like to see in our language detection project [linguist](https://github.com/github/linguist)?

Arvur:
> would this be something you'd like to see in our language detection project linguist?
I see Pascal in languages.yml but I have no idea how can I add Delphi and Lazarus.

One can consider them as children of Pascal. Roughly: if .dpr file is present then it is Delphi, if .lpr - Lazarus. But there are projects that contain both file extensions as they can be compiled in both IDEs.

Is there a way to distinguish this?

PR #1393, opened 2015-02-04
Update Python.gitignore

piotrgrundas:


SteveClement:
*.py[cod] already takes care of .pyc -> https://en.wikipedia.org/wiki/Regular_expression

PR #1395, opened 2015-02-05
Add npm and bower package folders for ASP.NET 5

dsplaisted:
ASP.NET 5 (aka "vNext") supports bower and npm packages.  The contents of these packages should probably not be checked in to git.

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -194,3 +194,8 @@ FakesAssemblies/
 
 # Visual Studio 6 workspace options file
 *.opt
+
+#ASP.NET vNext
+bower_components
+**/wwwroot/lib/*
+node_modules


dsplaisted:
@davidfowl @glennc does this look like the right paths to add to the VS .gitignore for ASP.NET vNext?

shiftkey:
Obligatory PSA that checking in `node_modules` is a thing that some in the Node community like to do: http://www.futurealoof.com/posts/nodemodules-in-git.html

dsplaisted:
Is it mostly agreed on in the node community that you should check in modules, or is it just "some in the node community"?

I was working with a developer using ASP.NET.  He checked in everything from the npm and bower packages, but when we tried to check out his changes (on Windows) we got long path errors and only a partial checkout.  This seems like it may be a bug in git on Windows (he checked the changes in on a Windows machine so the original paths weren't too long), but it turned me off even more from checking in packages. :-)

shiftkey:
> Is it mostly agreed on in the node community that you should check in modules, or is it just "some in the node community"?

The author above is part of the core team. Perhaps @aaronpowell has a better gauge of this in the broader community. 

> we got long path errors and only a partial checkout

This is a design problem with NPM (recursive, nested packages) which hits `MAX_PATH`. I've been discussing workarounds for this with others but there's no easy solution at clone time. [This](https://github.com/msysgit/git/commit/8439375065bebbc6ba6850cf164c9b206efe4676) is something I need to try out when I get a chance.

> but it turned me off even more from checking in packages. :-)

Totally understandable.

aaronpowell:
I really see `node_modules` and `bower_components` being treated the same way as we treat `packages`, they don't get checked in.

Regarding the above post @shiftkey linked to, I'm not sure if that is still relevant. If you check out the [npm FAQ](https://docs.npmjs.com/misc/faq) they suggest you don't check them in but use [`shrinkwrap` instead](https://docs.npmjs.com/misc/faq#should-i-check-my-node_modules-folder-into-git):

> Should I check my node_modules folder into git?

> Usually, no. Allow npm to resolve dependencies for your packages.

> For packages you deploy, such as websites and apps, you should use npm shrinkwrap to lock down your full dependency tree:

> https://docs.npmjs.com/cli/shrinkwrap

> If you are paranoid about depending on the npm ecosystem, you should run a private npm mirror or a private cache.

> If you want 100% confidence in being able to reproduce the specific bytes included in a deployment, you should use an additional mechanism that can verify contents rather than versions. For example, Amazon machine images, DigitalOcean snapshots, Heroku slugs, or simple tarballs.

shiftkey:
Thanks @aaronpowell!

aaronpowell:
Party

arcresu:
Just to point out: `node_modules/` and `bower_components/` are both already in the template a few lines up. We can add the other rule though if everyone's happy about that.

glennc:
All we're really talking about is wwwroot/lib then right? wwwroot/lib is where the starter web template is setup to copy bower packages to, so this seems reasonable for that to be in here.

I personally don't check in anything in my wwwroot, not just the lib folder. But I don't think I'm in the majority in that at the moment.

mjrousos:
I also don't check-in anything under wwwroot and would like to see that pushed as a best practice. I'm sure plenty of folks work with scipts and style sheets directly in wwwroot, though, so it's not something we should ignore by default.

wwwroot/lib, though, I think makes sense to add to .gitignore.

shiftkey:
@dsplaisted could you clean out the the NPM and bower values? I'm not hearing much dissent for keeping the `wwwroot` rule though...

shiftkey:
Friendly bump now we've got the shiny new ASP.NET Core name...

shiftkey:

@dsplaisted please let me know if you're interested in working together so this can get merged in.

PR #1396, opened 2015-02-05
ignore /vendor/bundle on Ruby.gitignore

polamjag:
I think `/vendor/bundle` should be ignored like [Rails.gitignore](https://github.com/github/gitignore/blob/master/Rails.gitignore#L21).

Many Ruby developers may do `bundle install --path vendor/bundle` even without Rails.

### DIFF
--- a/Ruby.gitignore
+++ b/Ruby.gitignore
@@ -22,6 +22,7 @@ build/
 
 ## Environment normalisation:
 /.bundle/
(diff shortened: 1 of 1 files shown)


arcresu:
That seems a little unusual but if that's what people do then fair enough. Thanks!

polamjag:
Thanks. You too!

PR #1397, opened 2015-02-05
minted 2.0 makes new files

maieul:


arcresu:
Thanks, but it would be helpful if you could quote/link to something in the documentation to explain this change.

maieul:
http://mirrors.ctan.org/macros/latex/contrib/minted/minted.pdf p. 6 bullet 2

arcresu:
Thanks very much! :tada:

PR #1399, opened 2015-02-05
Update for IntelliJ 14

philipmjohnson:
.idea file shouldn't be ignored in IntelliJ 14. File now follows official JetBrains recommendations.


shiftkey:
@hhariri - does this change to drop `.idea` make sense to you? The timestamp on the linked document is from December, rather than February when this PR was opened.

@philipmjohnson unfortunately this isn't mergeable currently, and the link now requires an account signup to access it :cry:

shiftkey:
Closing in favour of #1475

PR #1402, opened 2015-02-08
Create PSoC.gitignore

dbrwn:
This change adds a gitignore file for Cypress PSoC Creator projects, [see here](http://www.cypress.com/psoccreator/) for more information on PSoC Creator.  [Suggestions from Cypress](http://www.cypress.com/?id=4&rID=76644) were used to create the .gitignore file (it didn't exist before).

### DIFF
--- a/PSoC.gitignore
+++ b/PSoC.gitignore
@@ -1,0 +1,28 @@
+# Ignore the build output directories. 
+
+CortexM3/
+CortexM0/
+Export/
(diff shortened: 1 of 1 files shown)


shiftkey:

@dbrwn if you feel like this should still be a template listed here, please read the section about [specialized templates](https://github.com/github/gitignore#specialized-templates) and open a new pull request that follows the new pattern.

PR #1403, opened 2015-02-09
Update Python.gitignore

janhenke:
Add meta data folder and project files of PyDev IDE (http://pydev.org/) to the gitignore file. Checking in project configuration files from your IDE is generally not a good idea.

The IDE is based on Eclipse, so it creates the standard Eclipse .metadata folder and .project project file. Additionally PyDev writes Python specific project settings to .pydevproject.

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -53,3 +53,8 @@ docs/_build/
 
 # PyBuilder
 target/
+
+# PyDev (Eclipse based)
+.metadata/
+.project
+.pydevproject


arcresu:
Thanks for this, however it seems like these rules should go under a new `Global/PyDev.gitignore` file since they only concern users of that IDE rather than everyone creating a python project. Does that sound reasonable?

janhenke:
It does, at the time I opened this PR that PyDev gitignore file did not exist.

arcresu:
It still doesn't - I meant that you could create it in this PR with these rules :)

janhenke:
I just saw that files are already included in the Eclipse global gitignore. So this PR can be closed.

PR #1404, opened 2015-02-10
Create ABlogCms.gitignore

majosystems:
.gitignore file for a-blog cms.


arcresu:
Thanks, but is this CMS widely used enough to include here? Are there example repositories with similar gitignores? Or documentation explaining these choices?

shiftkey:

PR #1405, opened 2015-02-11
Expand list of root-of-volume dotfiles

occidens:
In addition to `.Spotlight-V100` and `.Trashes`, the following dotfiles
may appear at the root of all OSX volumes (not just external disks):

- `.DocumentRevisions-V100`: auto-save and versions storage
- `.fseventsd`: file system event storage
- `.Temporaryitems`: temp directory used by some applications instead of `/tmp` and `/var/tmp`
- `.VolumeIcon.icns`: storage of custom icon for volume, if set

### DIFF
--- a/Global/OSX.gitignore
+++ b/Global/OSX.gitignore
@@ -8,9 +8,13 @@ Icon
 # Thumbnails
 ._*
 
-# Files that might appear on external disk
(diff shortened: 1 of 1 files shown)


arcresu:
Thanks!

PR #1406, opened 2015-02-11
add IntelliJ ignores

mcandre:
add gitignore patterns for IntelliJ


arcresu:

PR #1410, opened 2015-02-13
Update Joomla.gitignore

adamasantares:
update for Joomla 3.3.6


arcresu:
Thanks, but see #1424. I don't think this is a very useful template to keep including here.

PR #1412, opened 2015-02-17
add dependency-reduced-pom.xml, which is a maven shader thing

dareonion:
e.g., this thing: http://stackoverflow.com/questions/22904573/what-is-the-purpose-of-dependency-reduced-pom-xml-generated-by-the-shade-plugin


arcresu:
Thanks!

PR #1414, opened 2015-02-19
Update Node.gitignore

iamsebastian:
Corrected URL, which linked to the `Should I check my node_modules folder into git?` readup.


shiftkey:
Thanks!

PR #1417, opened 2015-02-24
add OpenFrameworks.gitignore

firmread:
http://openframeworks.cc/
C++ Creative Coding platform

.gitignore file sourced from main OpenFrameworks repo here:
https://github.com/openframeworks/openFrameworks/blob/master/.gitignore

the file should works across all supported platform

### DIFF
--- a/OpenFrameworks.gitignore
+++ b/OpenFrameworks.gitignore
@@ -1,0 +1,109 @@
+#########################
+# openFrameworks patterns
+#########################
+
+# build files
+openFrameworks.a
+openFrameworksDebug.a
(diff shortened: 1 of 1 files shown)


arcresu:
Thanks, but this file isn't a good fit for our collection in its current state. For starters, all the editor-specific stuff belongs in ine various global templates rather than here. Have a look at our [contributing guidelines](https://github.com/github/gitignore/blob/master/CONTRIBUTING.md) and if you'd like to proceed, please open another pull request following those guidelines. For example, it's a lot more helpful if you can explain the rules rather than just copying them.

PR #1419, opened 2015-02-25
Add pattern to match dist-* directories for Haskell

eamsden:


arcresu:
Thanks, but could you explain what this is fixing? A link to documentation or at least examples would help.

shiftkey:
I'm still unclear on what this rule is for, but I count [many results on GitHub for Haskell repos specifying this](https://github.com/search?utf8=%E2%9C%93&q=dist-*&type=Code&ref=searchresults)

Any extra information would be great, but I'm just going to take this in now.

PR #1420, opened 2015-02-26
Add a .gitignore file for IntelliJ family of IDEs

lumeng:
Add a .gitignore file containing ignore statements for meta data of IntelliJ family of IDEs


arcresu:

PR #1422, opened 2015-02-28
Added gnuplot output files patterns to TeX.gitignore

altaris:


arcresu:
Thanks, but it would be great if you could comment with some documentation describing these rules so that we understand where they came from in the future. Also the PR will need updating so it can be merged. Thanks!

shiftkey:
@altaris unfortunately this isn't mergeable currently, and there are some questions outstanding which we'd like to clarify before merging :cry:

I need to clean up stale pull requests at some stage, and will close this out if I don't hear anything before Friday 26th February.

PR #1424, opened 2015-03-01
Add an explanation to reduce PHP framework clutter

arcresu:
Several of the templates consist solely of long lists of files which are installed by default by PHP frameworks. These template really aren't all that helpful because they are brittle in that the list of files is specific to a particular version of the software. Having a template for each version seems overkill, and in most cases you would be better off using `ls` or similar to automatically produce such a gitignore, if that's what you need.

See also #985 for my previous explanation of this problem. If we agree that this is a problem, then I propose to delete the templates which currently follow this pattern. If there is a way to curate a small, useful set of rules about these projects, then people can re-introduce them in new PRs, but the current list of rules seem unsalvageably messy in most cases.
- [Magento](https://github.com/github/gitignore/blob/eaeed89430318d8b04bc4734bd07190aaaeb29bd/Magento.gitignore)
- [Joomla](https://github.com/github/gitignore/blob/2b1fe3e280dbb0a53f5c5906ac0f3f6fd0e44dab/Joomla.gitignore)
- Drupal
- LemonStand
- SugarCRM
- Symfony
- ZendFramework

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -28,6 +28,11 @@ high quality, we request that contributions adhere to the following guidelines.
 In general, the more you can do to help us understand the change you’re making,
 the more likely we’ll be to accept your contribution quickly.
 
+If a template is mostly a list of files installed by a particular version of
+some software (e.g. a PHP framework) then it's brittle and probably no more
+helpful than a simple `ls`. If it's not possible to curate a small set of
+useful rules, then the template might not be a good fit for this collection.
+
 Please also understand that we can’t list every tool that ever existed.
 Our aim is to curate a collection of the *most common and helpful* templates,
 not to make sure we cover every project possible. If we choose not to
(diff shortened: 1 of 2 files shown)


emilv:
How are they brittle? When new files are introduced you can often just add them too to the .gitignore without damaging old versions of the framework. And it does not seem like a huge burden to keep them updated. I would agree with you for aged, unmaintained .gitignore files, but as long as people are willing to maintain them I do not see a reason to exclude them from here.

We use the Magento .gitignore template all the time even for internal projects.

So I do not agree with this PR. It is an unneccessary restriction. If pollution becomas a problem it is entirely possible to just arrange this repo into subfolders.

Kevinheaton86:
'

shiftkey:
@emilv

> When new files are introduced you can often just add them too to the .gitignore without damaging old versions of the framework.

An interesting example of this - Magento v2 came out and is a significant change with Magento v1 - there's three PRs to add a new version, because v2 is incompatible with the layout of v1.

I'm not sure how to go about managing these situations (I'm pushing for contributions for new versions to be incorporated into the one file) but for now I like this guidance at least. I have other suggestions to refine this which I'll PR in due course.

PR #1426, opened 2015-03-01
Added jet

krzykz:
Added jetbrains pycharm's .idea/workspace.xml folder to gitignore, it is strictly system dependant making it useless to git track
PyCharm is a popular enough ide so this should be a welcome change to anyone using it and being tired of manually ignoring this file in each project.

Official site: https://www.jetbrains.com/pycharm/
Docs: https://www.jetbrains.com/idea/help/project.html

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -54,3 +54,6 @@ docs/_build/
 
 # PyBuilder
 target/
+
+# JetBrains PyCharm, WebStorm, etcc..
+.idea/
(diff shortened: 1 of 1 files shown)


krzykz:
Will create another branch since i misread the rules for a pull request.

drothmaler:
You don't need to do so... Just use `Global/JetBrains.gitignore` :-)

krzykz:
But i guess that would be the same thing as i would have to use two gitignores one for python and one for jetbrains. Also, thank you, the Global folder passed completely unnoticed for me :+1:

PR #1427, opened 2015-03-02
Update Python.gitignore

xpdz:
Add the virtualenv default folder.
see https://github.com/kennethreitz/python-guide/blob/master/docs/dev/virtualenvs.rst


arcresu:
Thanks - there's a separate template for virtualenv, though.

PR #1429, opened 2015-03-05
Update Jetbrains.gitignore

ergl:
When users have a folder or package named `out/`, the current gitignore will ignore that folder too, making them update their local `.gitignore` to exclude these folders specifically.

Given that the `.gitignore` file is usually located at the same level as the intellij' `out/` build folder, I think it makes sense to only ignore the top level `/out/` build folder.


arcresu:
Thanks, seems like a good idea to me.

PR #1430, opened 2015-03-06
Added coverage annotation files

Lucretiel:
Coverage creates annotation files by appending ",cover" (with a comma, not a period) to the filename. These should be ignored.

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -41,6 +41,7 @@ htmlcov/
 .cache
 nosetests.xml
 coverage.xml
+,cover
 
 # Translations
 *.mo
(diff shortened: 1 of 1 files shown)


arcresu:
Thanks, is there some documentation or at least examples?

Lucretiel:
Running on OS X 10.10, Python 3.4.2, coverage 3.7.1, in a fresh virtualenv:

```
(.env)$ ls
hello.py
(.env)$ cat hello.py 
print("Hello, World!")
(.env)$ coverage run hello.py 
Hello, World!
(.env)$ coverage annotate
(.env)$ ls
hello.py       hello.py,cover
(.env)$ cat hello.py,cover 
> print("Hello, World!")
```

arcresu:
Thanks for the example. In this case, shouldn't the rule be `*,cover`?

Lucretiel:
Right you are. Fixed.

arcresu:
Thanks!

PR #1432, opened 2015-03-10
Add ".env" to Laravel.gitignore

jakejohns:
Laravel 5 uses vlucas/phpdotenv for env configuration.
http://laravel.com/docs/5.0/configuration#environment-configuration

Laravel 4 used env.php files. 
The new configuration files do not have ".php" extensions

Pull req  #1380 addressed this, but included other things.

Fixes #1380.


arcresu:
Thanks!

PR #1434, opened 2015-03-12
spring mvc gitignore

Dogzhou:
gitignore for java spring mvc project


arcresu:
Thanks! However we like to keep things a bit more modular here. OS-specific (`.DS_Store` and friends), and IDE-specific (`.iws` etc.) rules belong in their respective `Global/` templates. Can you strip all of those out so we just have rules relevant to spring mvc? And then some examples or documentation explaining the rules would be helpful.

shiftkey:
Closing in favour of #1489 as it seems more organized

PR #1437, opened 2015-03-14
add Cargo.lock

marti1125:
Generated by Cargo


arcresu:
Thanks! Could you explain why this file needs to be ignored? If as the name suggests this is something like ruby's `Gemfile.lock` then it might not be a good idea to ignore it.

marti1125:
the reason is after run $ cargo build
it will generate automatic Cargo.lock with this content:

[root]
name = "hello_world"
version = "0.0.1"

arcresu:
Ok, but I assume that this isn't a completely arbitrary file. What is it intended to do?

marti1125:
this can help :D
https://github.com/rust-lang/cargo/issues/315

arcresu:
From the [documentation linked](http://doc.crates.io/guide.html#cargo.toml-vs-cargo.lock) to from the PR you mentioned, it seems that you definitely _do_ want to check this file in unless you're working on a library. I think maybe a better change would be to add the rule commented out with a note to uncomment it if you're developing a library.

shiftkey:
@marti1125 unfortunately this isn't mergeable currently, and there are some questions outstanding which we'd like to clarify before merging :cry:

shiftkey:
Nevermind, this has already been incorporated: https://github.com/github/gitignore/blob/master/Rust.gitignore#L13-L15

marti1125:
ok :D :+1:

PR #1439, opened 2015-03-17
Only exclude folders in the project root directory

VoxelMatt:
Updated Unity.gitignore to only ignore the Library, Temp, Build and Obj folders in the **root** of a project. 

Library, Obj, Build etc are all folder names that might be used *inside* any unity project (but we definitely want to exclude them on the root). For example, obj is a common modelling format, so it might make sense to have a folder of .obj files called 'Obj'.


shiftkey:
It looks like https://github.com/github/gitignore/commit/a4661a2d39f9cd733ff616dbcc2918f959828d76 in #1481 was the same change as this proposal, so I'm going to close this out as resolved.

Thanks for raising it @VoxelMatt!

PR #1440, opened 2015-03-17
Add Elm.gitignore

marcelbuesing:
http://elm-lang.org/

### DIFF
--- a/Elm.gitignore
+++ b/Elm.gitignore
@@ -1,0 +1,2 @@
+elm.js
+elm-stuff/


jasonkuhrt:
Please LGTM

chendrix:
:+1:

Apanatshka:
Complete information:

> The Elm programming language is on [elm-lang.org](http://elm-lang.org). 
> This gitignore ignores build artefacts and downloaded dependencies. The directory name was decided on [here](https://groups.google.com/d/msg/elm-discuss/BNGXIjQpNAI/1c_98tAmr34J). 

(copied from [my PR](https://github.com/github/gitignore/pull/1515), I didn't notice this one)

bbugh:
Shouldn't this re-include `elm-stuff/exact-dependencies.json`, like so? If I understand correctly, that file is significant, much like `Gemfile.lock`.

```
elm.js
elm-stuff/*
!elm-stuff/exact-dependencies.json
```

(but I'm relatively new to Elm and could be wrong)

chendrix:
I think it's a matter of if you're publishing a library or a "final application".

E.g. even the Ruby gitignore tells you 

```
# for a library or gem, you might want to ignore these files since the code is
# intended to run in multiple environments; otherwise, check them in:
# Gemfile.lock
# .ruby-version
# .ruby-gemset
```

If you're writing a final application you may want to check in exact-dependencies.json, a library? nope.

bbugh:
That should probably be specified in the `.gitignore` then, the way the Ruby one does. It's pretty important.

shiftkey:
Closing in favour of https://github.com/github/gitignore/pull/1610 which has some extra rules around the REPL.

PR #1442, opened 2015-03-19
Create KiCAD.gitignore

DanNixon:
This adds a ```.gitignore``` for the [KiCAD](http://www.kicad-pcb.org/) EDA software.

Documentation on the file types is available [here](http://www.kicad-pcb.org/display/KICAD/File+Formats).

Essentially all files that are either temporary files or can be generated directly from a file already under version control (e.g. ```.net``` files are generated from ```.sch``` files) are ignored.


arcresu:
Thanks!

PR #1444, opened 2015-03-22
Provide a gitignore for Dlang (and dub)

Geod24:
This adds a gitignore for the D programming language and one of it's build tool, dub.

### DIFF
--- a/D.gitignore
+++ b/D.gitignore
@@ -1,0 +1,36 @@
+# Gitignore for projects in D using dub.
+# Dlang: http://dlang.org/
+# Dub:   https://github.com/D-Programming-Language/dub
+
+## Binary files ##
+
+# Object files
+*.o
+*.obj
+
+# Libraries
+*.a
+*.lib
+*.so
+*.dylib
+*.dll
+
+# Executables
+*.exe
+
+
+## Dub specific ##
+.dub/
+
+# Result of 'dub test'
+__test__*__
+
+# Documentation artifacts (dub --build=[docs|ddox])
+docs.json
+__dummy.html
+docs/*
+
+# By default, ignore dub.selections.json.
+# You should usually rely on tags in dub.json rather than pinning your project
+# to a certain commit, but if you really need to, remove this line:
+dub.selections.json


D.gitignore
```diff
@@ -0,0 +1,36 @@
...
+
+# Documentation artifacts (dub --build=[docs|ddox])
+docs.json
+__dummy.html
+docs/*
+
+# By default, ignore dub.selections.json.
+# You should usually rely on tags in dub.json rather than pinning your project
+# to a certain commit, but if you really need to, remove this line:
+dub.selections.json
```
jacob-carlborg: `dub.selections.json` should not be ignored. Or rather, it probably should be ignored for libraries but not for applications.

Geod24: It should be by default. That's why I added a lengthy comment to explain that it can safely be removed, but it's addition to source control should be discouraged, because it defeats the purpose of have semantic versions.

jacob-carlborg: You're clearly not aware of the reasons why `dub.selections.json` was added and why it's needed.

The problem is a scenario like this:

In you're project you have specified a dependency on project A with an exact version, "1.2.3". Project A has a dependency on project B, specified as "~>4.5.6". When you're starting developing this project and running "dub" it will download project A with version "1.2.3", then it will download project B with version "4.5.6" as well.

Now, at some later point in time another developer joins your project and clones the project. At the point the developer runs "dub" a new version of project B has been released, "4.5.7". Now dub will download project A with the expected version "1.2.3" and the unexpected version "4.5.7". This can also happen if you're cloning your project on a different computer or a build machine.

Suddenly you have now idea what code you're running and I consider that to be a major security issue. Also I wouldn't trust a single project for a second to not break my code if I updated to a later patch/build version.

We have already gone through this issue in lengthy discussions in the Dub forum.

shiftkey:
@Geod24 apologies for missing this in favour of #1676 - this looks more concise and readable.

Are you interested in resolving the differences? I'm not sure where the discussion on `dub.selections.json` has ended up, I guess that's the only blocker here...

shiftkey:
There's also some interesting feedback from @jacob-carlborg about the `dub.selections.json` rule

shiftkey:
Looks like I went and merged an alternate PR for this https://github.com/github/gitignore/pull/1676 and then some tweaks https://github.com/github/gitignore/pull/2009.

@Geod24 @jacob-carlborg please open a new PR if there's any additional changes you'd like to see added.

PR #1445, opened 2015-03-23
Added .recommenders/ to gitignore

seberik:
Eclipse added the .recommenders/ directory which should be part of the gitignore.


arcresu:
Thanks! Could you link to an example or some documentation so that we have a record of why this change was made?

shiftkey:
Closing in favour of #1593

PR #1447, opened 2015-03-24
astyle

mcandre:
Add an example gitignore for astyle users.


thomaswardiii:
:-1: If you have git merge conflicts and don't clean up properly, .orig files will be created as a result. Showing these may be useful to indicate to somebody such a file exists and their branch needs cleaning up.

mcandre:
Good point. I wonder if the astyle community would mind changing their extension to something else, so that it doesn't conflict with git merge .orig files.

arcresu:
In light of the name clash, I tend to agree that this might cause issues, so I'll close the request, but thanks anyway

PR #1448, opened 2015-03-24
Update Python.gitignore to include .bak files created by 2to3

gambl:
The standard Python distribution's 2to3 tool migrates code from Python 2 to Python 3.
The tool will create files with the extension '.bak' when run with -w argument, which appears to be fairly standard usage.

For example:

`2to3 -w test.py`

will produce:

`test.py.bak`

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -55,3 +55,6 @@ docs/_build/
 
 # PyBuilder
 target/
+
+# 2to3
+*.bak


arcresu:
It seems to me that it would be a bad idea to have these `.bak` files hanging around indefinitely. Once you've verified that the automatic conversion was reasonable, you'd delete the files. They aren't going to keep reappearing. So thanks, but I think it's at best unnecessary and at worst encouraging a bad pattern to make this addition.

PR #1451, opened 2015-03-25
Update to NetBeans .gitignore nbproject folder

theWhiteFox:
When I commit a NetBeans 8.0.2 HTML5Application nbproject/private/ does not ignore files project.properties and project.xml.
My edit nbproject/* is cleaner and ignores the nbproject folder and all folders and files within - including project.properties and project.xml

### DIFF
--- a/Global/NetBeans.gitignore
+++ b/Global/NetBeans.gitignore
@@ -1,4 +1,5 @@
-nbproject/private/
+### NetBeans ###
+nbproject/*
 build/
 nbbuild/
 dist/


thomaswardiii:
Is this what is desired in general? Do a majority or all users of NetBeans (me included) want to always have projects tracked by git ignore all NetBeans project folder files?

This would be great for those that have multiple editors in the same organization, but perhaps not for ones that have standardized on NetBeans. The organization I work for indeed has multiple editors (vim, emacs, NetBeans, PhpStorm, and some others) so this would be fine for me. In fact I use a similar line in my global ignore. But I think the purpose here is to stop files that are of very minimal or no use to others working on the same project

shiftkey:
@theWhiteFox is there some official resources about whether this whole folder can be safely ignored?

theWhiteFox:
@shiftkey I have stopped using NetBeans. 
I am now using [vscode](https://code.visualstudio.com/) to build HTML5 apps.

PR #1455, opened 2015-03-26
Remove phpunit from symfony file

nunorafaelrocha:
PHPUnit configuration files are essential to run continuous integration tests on platforms like travis.

### DIFF
--- a/Symfony.gitignore
+++ b/Symfony.gitignore
@@ -26,10 +26,6 @@
 /web/bundles/
 /web/uploads/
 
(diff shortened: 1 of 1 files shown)


arcresu:
Thanks for this. Can you provide an example or documentation just so we have something to refer to if this comes up again in the future?

shiftkey:
@altaris unfortunately this isn't mergeable currently, and there are some questions outstanding which we'd like to clarify before merging :cry:

PR #1456, opened 2015-03-26
Update VisualStudio.gitignore

stevelydford:
Added line to ignore *.ncrunchsolution files

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -102,6 +102,7 @@ _TeamCity*
 # NCrunch
 _NCrunch_*
 .*crunch*.local.xml
+*.ncrunchsolution
 
 # MightyMoose
 *.mm.*


shiftkey:
As per the [documentation](https://www.ncrunch.net/documentation/considerations-and-constraints_using-ncrunch-with-source-control) the official advice is to include these files in source control:

> It's recommended that the last two files listed in the above table (`.v2.ncrunchsolution` and `.v2.ncrunchproject`) be stored inside a source control system to be shared with team members. These files contain settings that are critical to being able to build projects and run tests, and as such they will be needed by anyone seeking to use NCrunch with the solution.

stevelydford:
Fair enough. Should have checked the docs first, my bad.

shiftkey:
@stevelydford all good, I generally defer to whatever the tools have to say in their docs

PR #1457, opened 2015-03-27
Included senchatouch project gitignore

PradipShrestha:
Can be helpful while creating projects using sencha framework.

### DIFF
--- a/SenchaTouch.gitignore
+++ b/SenchaTouch.gitignore
@@ -1,0 +1,5 @@
+# All sass cached files inside .sass-cache
+**/.sass-cache
+
+# All temporary build files
+**/temp


murrple-1:
I'm in agreement with this being added, but is it not true that archive/ and build/ should also be added?

http://www.sencha.com/forum/showthread.php?261050-Git-.gitignore-file-for-Sencha-Touch-2.1.x

ALSO! Should bootstrap.js, bootstrap.json and build.xml also be added, as they are generated? They can be regenerated via `sencha app refresh`, at least in my limited understanding of the Sencha ecosystem?

PradipShrestha:
@murrple-1: I agree on that. I just wanted to initialize ignore for that. I can include those if you want or if you want to add please add.

shiftkey:
@PradipShrestha thanks for the contribution! Unfortunately this seems like a minor template and doesn't feel like a good fit for this repository.

PR #1458, opened 2015-03-27
Update gitignore to exculde Mac DS_Store file

CharlesCai930:
Update gitignore to exculde Mac DS_Store file

### DIFF
--- a/Global/Eclipse.gitignore
+++ b/Global/Eclipse.gitignore
@@ -34,3 +34,6 @@ local.properties
 
 # TeXlipse plugin
 .texlipse
+
+# Mac File
+.DS_Store


ergl:
You can already find that in the [`Global/OSX.gitignore`](https://github.com/github/gitignore/blob/master/Global/OSX.gitignore)

RogerThiede:
If using OS X and Eclipse simultaneously, you can combine the two templates for Global/OSX.gitignore and Global/Eclipse.gitignore.  Since the Eclipse tool does not have a dependency on OS X and since .DS_Store file is not created by Eclipse, these may be better off staying as separate templates.

arcresu:
Yes, this rule belongs in the global template. Thanks anyway.

PR #1461, opened 2015-03-30
adding the .orig files to the ignore list

G3N7:
This is to keep noise from using source tree to a minimum.

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -174,6 +174,9 @@ Backup*/
 UpgradeLog*.XML
 UpgradeLog*.htm
 
+# SourceTree
+*.orig
+
 # SQL Server files
 *.mdf
 *.ldf


RogerThiede:
Would this pull request be specific to VisualStudio.gitignore, or would it represent a new SourceTree.gitignore for users of that tool?

arcresu:
As has been pointed out in #1447, this rule could create a problem with git's merging. I think that for that reason we should hesitate to add this rule to such a common template without good reason. Thanks anyway :)

G3N7:
out of curiosity, how could it create a problem with git's merging?

RogerThiede:
@G3N7, it masks a problem that could occur after a failed merge or merge that couldn't be automatic.  ```.orig``` files would be created as the developer manually merges lines.  These are files created by git, not the third party tool SourceTree.  If the developer doesn't clean these files up properly, and they were in the gitignore, the issue that these files still exist would be masked.  If they were not in the gitignore, it would be very clear that their working directory needs cleaning up.

G3N7:
OK, thanks for the clarification.  Its more of a workflow issue than you ever actually wanting these to end up in the VCS, you just want to ensure they show as outstanding issues.

PR #1463, opened 2015-03-31
Add DNX project.lock.json to ignore file

anurse:
I'm 95% sure this is correct, but adding @davidfowl and @lodejard to confirm.

Users may want to check the lock file in on release branches but to do so they need to a) "lock" it using a specific command (`dnu lock` or something) and b) can override the ignore file for the specific add operation.


shiftkey:
Looks like this was superseded by #1505

PR #1465, opened 2015-04-02
Update Maven.gitignore

tgkprog:
Else sub projects (main dir has sub dirs with own pom files - sub modules) targets are not ignored.

### DIFF
--- a/Maven.gitignore
+++ b/Maven.gitignore
@@ -1,4 +1,5 @@
 target/
+**/target/
 pom.xml.tag
 pom.xml.releaseBackup
 pom.xml.versionsBackup


dserodio:
:+1:

shiftkey:
LGTM

arcresu:
Actually I don't think that this was necessary - the addition is redundant with the existing rule as in #1508

shiftkey:
@arcresu #1508 is about the `Android` template - is that the right PR?

arcresu:
Yes, my point was that the other PR shows that in gitignore syntax the rule `some_directory/` matches at any depth, not just the root level. This is the same scenario here.

shiftkey:
@arcresu ah, right. I'll open a PR to revert this. Apologies for the noise.

PR #1466, opened 2015-04-03
Add IDEs to Java.gitignore

phase:
This ignores generated files from IntelliJ & Eclipse which every Java project doesn't want.


arcresu:
Thanks, but we keep things modular by putting these in `Global/` templates.

PR #1467, opened 2015-04-03
Ignore .virthualenv not .virtualenv

Abizern:
This change was introduced in 3bb4e51184066f5cf60daa029c31e7f52398353b, where it was marked as changed due to a typo. However, that is the correct name of this particular virtual environment. See https://hackage.haskell.org/package/virthualenv

### DIFF
--- a/Haskell.gitignore
+++ b/Haskell.gitignore
@@ -6,7 +6,7 @@ cabal-dev
 *.chs.h
 *.dyn_o
 *.dyn_hi
-.virtualenv
+.virthualenv
 .hpc
 .hsenv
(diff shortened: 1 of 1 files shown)


arcresu:
Thanks, I didn't realise! However I notice on that package page that the package is deprecated. Since the templates here are predominantly used for _new_ projects, would it perhaps make more sense to remove this rule altogether?

Abizern:
@arcresu Good point. I've updated the PR. Thanks.

arcresu:
Thanks! :sparkles:

PR #1473, opened 2015-04-09
pre-emptively separate default ignore rules from user-defined ones

katrinleinweber:
Just an idea I got while using GitHub's Win & Mac clients. Adding ignore rules via the context menu of staged files simply appended them to the end. Since there is no pre-labelled separator, it was sometimes not clear whether I had added a certain ignore rule, or whether it came pre-defined. Adding this or a similar "footer" to the file would ensure that the pre-defined ignore rules above don't get confused with use-defined ones.

### DIFF
--- a/R.gitignore
+++ b/R.gitignore
@@ -11,3 +11,5 @@
 # produced vignettes
 vignettes/*.html
 vignettes/*.pdf
+
+# Custom ignore rules


arcresu:
Thanks, but I think if this were to happen then it would need to be a change made in the Windows and Mac clients rather than in these templates. In any case, the templates are called templates because we'd expect that you would edit them, removing irrelevant things etc. They're a starting point rather than an immutable boilerplate.

katrinleinweber:
OK. Emailed this suggested to GitHub's support, and it apparently made the to-be-considered list.

PR #1474, opened 2015-04-10
Update the year of LICENSE

Necior:


arcresu:
Thanks, but see #1351

PR #1475, opened 2015-04-14
Update JetBrains.gitignore according to the official knowledge base article

LazarusX:
The current JetBrains.gitignore file contains several entries which are suggested to be shared, such as the project .ipr file and the .iml module files, according to [the official knowledge base article](https://intellij-support.jetbrains.com/entries/23393067). This commit fix this issue.


shiftkey:
This seems to be a more complete version of #1399

shiftkey:
@LazarusX unfortunately this isn't mergeable currently, and the link now requires an account signup to access it :cry: But it looks promising, judging by the documentation. Are you able to update this?

LazarusX:
Hi @shiftkey,

I just resolved the conflicts and this is now mergeable. The documentation is now redirected [here](https://intellij-support.jetbrains.com/hc/en-us/articles/206544839) and does not require login to access. You can check it out to verify the changes.

Entries such as `.idea/sqlDataSources.xml` and `.idea/dynamic.xml`, are not mentioned in the documentation. I tend to leave them in the `.gitignore` because they look sensitive and I am not 100% sure that they can be shared.

shiftkey:
> The documentation is now redirected [here](https://intellij-support.jetbrains.com/hc/en-us/articles/206544839) and does not require login to access.

Great, thanks!

> ... I tend to leave them in the .gitignore because they look sensitive and I am not 100% sure that they can be shared.

This is fine - with the change to track `.idea` by default it's good to exclude those known files again.

PR #1476, opened 2015-04-16
Android studio stuff

InderKumarRathore:


arcresu:
Thanks, but we keep IDE- and OS-specific studd in separate templates under `Global/`. I believe these rules are all covered there.

PR #1479, opened 2015-04-17
Rework PrestaShop.gitignore

julienbourdeau:
As a PrestaShop core developer I reworked our .gitignore file last year (https://github.com/PrestaShop/PrestaShop/commit/d6956ebe081e864ba262a91b54bbd6207a8834cd). I realized recently that the one provided by GitHub is an old version of our file.

The file in this repo was written to maintain the core, then we ignore files that a normal user will need to track (and _vice versa_).

Here are some explainations about my modifications ('we' refers to the PrestaShop core team) :

- All `index.php` files, except the one under the root directory, are redirections to the root folder. We used to version them because of the licence at the top of the file.
- `cache/` should be ignored completely (of course).
- `robots.txt` should be versionned. As a user, you will edit it and want to keep his changes. We ignored it to keep a clean version for releases.
- PrestaShop will generate multple sitemaps if you have multistore or multilanguage features activated. All sitemaps will be prefixed so I changed the pattern to `*sitemap.xml`
- A user will want to version his modules to track his changes. Only `config*.xml` files needs to be ignored since they are generated if needed.

If you need more details please let me know, I'll be happy to explain.

__NOTE__: The previous version kept track of core files, I didn't change this behavior.


arcresu:
Thanks!

PrestanceDesign:
@julienbourdeau I notice that `cache/` causing problem for gitignore `classes/cache/`.
I create this PR #1693 to fix it.

PR #1481, opened 2015-04-18
Made it so the folders only get ingored in the root

Tandoori:
With the current setup these folders are ignored in the whole repo, but they're only automatically generated in the project root so it makes sense to only ignore the ones in the root.

I made this change because one of the assets in my project uses a folder called 'Library', and I was wondering why it never uploaded that folder.

It might also make sense to extend this 'root only' change to VS/MD and crash report lines, as I believe those can only be placed in the root.


arcresu:
Thanks!

PR #1482, opened 2015-04-20
Ignore pyenv's version marker

nicktimko:
[pyenv][pyenv] is a Python version manager for *nix shells. It allows the Python version and virtual environment to be set locally, per-directory, using a `.python-version` file. Because pyenv environment names are arbitrary and machine+user-specific (the name being in the `.python-version` file), they are usually noise.

[pyenv]: https://github.com/yyuu/pyenv

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -55,3 +55,6 @@ docs/_build/
 
 # PyBuilder
 target/
+
+# pyenv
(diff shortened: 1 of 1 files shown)


shiftkey:
Closing in favour of #1801 which has some more back-and-forth.

PR #1483, opened 2015-04-21
Komodo Edit gitignore

felipsmartins:
It is a gitignore for Komodo IDE/Edit tools.


shiftkey:
I'm going to close this out because I'm not sure how much value this would add.

@felipsmartins if you feel like this should still be a template listed here, please read the section about [specialized templates](https://github.com/github/gitignore#specialized-templates) and open a new pull request with more context.

PR #1484, opened 2015-04-23
Adding a few more files/file extensions/directories for Synopsys VCS.

andres-mancera:
@arcresu : This is just a follow-up to pull request [1282](https://github.com/github/gitignore/pull/1282).  I talked to a Field Engineer over at Synopsys, and he pointed out some files that I had originally missed.  Those have been added in this patch.  I have also added a few comments about each one of the files, file extensions and directories that will be ignored just for documentation purposes.

### DIFF
--- a/Global/SynopsysVCS.gitignore
+++ b/Global/SynopsysVCS.gitignore
@@ -1,23 +1,57 @@
-# Waveforms
-*.vpd
+########## Waveforms #######################################
+# Value Change Dumping (VCD) - IEEE Standard
 *.vcd
+# VCDlus Dumping (VPD) - Synopsys proprietary format
+*.vpd
+# Extended VCD (EVCD) - Dump only port information
 *.evcd
+# Fast Signal DataBase (FSDB)
+*.fsdb
 
-# Binary files
+
+########## Simulation executable file ######################
+# Default name of the simulation executable.  A different
+# name can be specified with this switch (the associated
+# daidir database name is also taken from here)
+#   -o <path>/<filename>
 simv
 
-# Directories used for compilation
-csrc/
-simv.daidir/
 
-# Log files
+########## Intermediate files used for simulation ##########
+# Generated for Verilog top configs
+simv.daidir/
+# Generated for VHDL top configs
+simv.db.dir/
+# Infrastructure necessary to co-simulate SystemC models
+# with Verilog/VHDL models.  An alternate directory may
+# be specified with this switch:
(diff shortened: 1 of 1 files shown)


arcresu:
Thanks for this! While I'm fine with the new additions, I'm not sure about the documentation. In general I prefer to reserve comments for (1) grouping sections of rules, (2) helping a user of the template make a choice about (un)commenting something or (3) justifying a particularly controversial rule. Other documentation is helpful in the PR thread, but probably doesn't need to be in the file. This isn't a blocker, so I'll merge if you'd rather not change - I'll leave it up to you.

andres-mancera:
Thanks for your feedback @arcresu .  The reason why I decided to add those comments is in line with your second point (_helping a user of the template make a choice about (un)commenting something_).  For instance, in [line 36](https://github.com/andres-mancera/gitignore/blob/SynopsysVCS_patch/Global/SynopsysVCS.gitignore#L36) I am ignoring all the "*.log" files.  However, the comment that has been added right above that (starting in [line 32](https://github.com/andres-mancera/gitignore/blob/SynopsysVCS_patch/Global/SynopsysVCS.gitignore#L32)) is meant to let the user know that the resulting log file might not have a "log" extension if the "-l" switch is used during the compilation.  In such a case, it would be the user's responsibility to add the corresponding log filename to his/her own ignore file.  If you agree, please go ahead and merge this one in.

andres-mancera:
Hi @arcresu : Just dropping you a line to remind you about this pull request.  Please let me know if you have any further questions.  Cheers!

arcresu:
The kind of comments I was referring to were things like:

  * `# VCDlus Dumping (VPD) - Synopsys proprietary format`
  * `########## Waveforms #######################################`
  * `# Generated for Verilog top configs`

These add noise without helping somebody to use the template.

See [Rails.gitignore](https://github.com/github/gitignore/blob/master/Rails.gitignore) for an example of a template where the majority of comments are either giving maximally concise information about why a change might need to be made to the default rule, or justifying a confusing rule so people don't keep reverting it. I'd be happier with this template if it were more like that - could you try and clean it up a bit?

andres-mancera:
@arcresu : As per your suggestion, I have removed some of the comments that were adding noise.  Can you please take a look again?  I think this time it should be fine.  :smile:

arcresu:
Thanks, that looks better! :sparkles:

andres-mancera:
:loudspeaker: Thank you!

PR #1485, opened 2015-04-23
Ignore folder containing roslyn temporary files

pmiossec:
"This folder should be excluded from the source control system normally."
http://blogs.msdn.com/b/webdev/archive/2014/06/03/announcing-web-features-in-visual-studio-14-ctp.aspx

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -6,6 +6,7 @@
 *.user
 *.userosscache
 *.sln.docstates
+*.sln.ide/
 
 # User-specific files (MonoDevelop/Xamarin Studio)
 *.userprefs


shiftkey:
@pmiossec I believe this special folder has been renamed to `.vs` as per https://github.com/github/gitignore/pull/1388 - and that's covered on [line 27](https://github.com/github/gitignore/blob/master/VisualStudio.gitignore#L26)

pmiossec:
I will verify tomorrow which CTP of Visual Studio 2015 I have installed (even if I think it is the last one) but the folder was still generated in this CTP.

But perhaps, you're right and I should just wait for the next CTP : https://gyorgybalassy.wordpress.com/2015/03/06/i-asked-for-a-vs-folder-and-the-vs-team-gave-it-to-me/

shiftkey:
screen shot 2015-04-30 at 10 00 48 am

Confirmed the hidden folder in VS2015 is now called `.vs`

PR #1486, opened 2015-04-24
Update npmjs docs link

Gouthamve:
The previous link gives 404

### DIFF
--- a/Node.gitignore
+++ b/Node.gitignore
@@ -23,5 +23,5 @@ coverage
 build/Release
 
 # Dependency directory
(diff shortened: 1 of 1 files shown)


arcresu:
Thanks, but while you've fixed the link, the URL fragment has also changed. It should be this:

    https://docs.npmjs.com/misc/faq#should-i-check-my-node-modules-folder-into-git

(The underscore became a dash.). If you fix that up I'll merge :tada:

arcresu:
Perfect, thanks!

PR #1487, opened 2015-04-24
Add Java APT class path to Eclipse

sveyret:


arcresu:
Thanks! Looks from the [documentation](http://help.eclipse.org/luna/index.jsp?topic=%2Forg.eclipse.jdt.doc.isv%2Fguide%2Fjdt_apt_getting_started.htm) has this to say:
> The factory path is stored in a file named ".factorypath" at the project root, similar to the classpath, and should be treated the same way as the classpath with regard to version control.

PR #1488, opened 2015-04-24
Some WordPress rules

Decadence:
Appended some ignore rules for WordPress

### DIFF
--- a/WordPress.gitignore
+++ b/WordPress.gitignore
@@ -11,3 +11,7 @@ wp-content/cache/
 wp-content/upgrade/
 wp-content/uploads/
 wp-content/wp-cache-config.php
+
+/readme.html
+/license.txt
+
--- a/WordPress.gitignore
+++ b/WordPress.gitignore
@@ -11,6 +11,7 @@ wp-content/cache/
 wp-content/upgrade/
 wp-content/uploads/
 wp-content/wp-cache-config.php
+wp-content/plugins/hello.php
 
 /readme.html
 /license.txt


Decadence:
ff510be
I think these files should be ignored to prevent it from production. Otherwise everybody can see your WP version and uses it bugs or security problems. But you have to update WP on local machine and then update production from the repo. Other way is configure your server to lock down web access to this files.

1d3c565
Just ignore useless demo plugin.

wjramos:
This is a good addition

WordPress.gitignore
```diff
@@ -11,3 +11,8 @@ wp-content/cache/
 wp-content/upgrade/
 wp-content/uploads/
 wp-content/wp-cache-config.php
+wp-content/plugins/hello.php
+
+/readme.html
+/license.txt
```
shiftkey: Could excluding the license file from source control risk subverting the GPL?

>   For example, if you distribute copies of such a program, whether
gratis or for a fee, you must give the recipients all the rights that
you have.  You must make sure that they, too, receive or can get the
source code.  **And you must show them these terms so they know their
rights.**

Decadence: I found this thread
http://wordpress.stackexchange.com/questions/60620/is-it-breaking-the-gpl-if-you-remove-license-txt-before-uploading

Decadence:
Strictly speaking I don't see WP version in license.txt but anyway it's useless in repo and some programmers may want to hide the fact that they use WordPress on their sites.

wjramos:
You are free to remove license.txt (or any other files) for production

Decadence:
So what's about merging then?

arcresu:
Thanks!

PR #1489, opened 2015-04-26
add Spring.gitignore

hknochi:
add an ignore file for SpringFramework Projects

### DIFF
--- a/Spring.gitignore
+++ b/Spring.gitignore
@@ -1,0 +1,54 @@
+# get it from: https://github.com/spring-projects/spring-framework/blob/master/.gitignore
+.gradle
+´build/
+
+# Ignore Gradle GUI config
+gradle-app.setting
+
+# Avoid ignoring Gradle wrapper jar file (.jar files are usually ignored)
+!gradle-wrapper.jar
+
+# Package Files #
+*.jar
+*.war
+*.ear
+
+# Ignore Spring specifc files
+/build
+bin
+build.sh
+buildSrc/build
+integration-repo
+.settings
+/spring-*/build
+.springBeans
+spring-test/test-output/
+/src/asciidoc/build
+target/
+
+# Eclipse artifacts, including WTP generated manifests
+.classpath
+.project
+spring-*/src/main/java/META-INF/MANIFEST.MF
+
+# IDEA artifacts and output dirs
+*.iml
+*.ipr
+*.iws
+.idea
+out
+test-output
+atlassian-ide-plugin.xml
+.gradletasknamecache
+
+# Ignore some other files 
+*.java.hsp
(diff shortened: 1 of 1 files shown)


Spring.gitignore
```diff
@@ -0,0 +1,54 @@
...
+buildSrc/build
+integration-repo
+.settings
+/spring-*/build
+.springBeans
+spring-test/test-output/
+/src/asciidoc/build
+target/
+
+# Eclipse artifacts, including WTP generated manifests
```
shiftkey: We maintain the IDE-specific rules under the `Global` folder as this is easier to maintain when there are new releases to IDEs. Could you remove this section and the IDEA section below?

Spring.gitignore
```diff
@@ -0,0 +1,54 @@
...
+*.iml
+*.ipr
+*.iws
+.idea
+out
+test-output
+atlassian-ide-plugin.xml
+.gradletasknamecache
+
+# Ignore some other files
```
shiftkey: I'd love for these to be documented a bit so that others can understand why these rules exist...

shiftkey:
I'm trying to wrap my head around where this should live:

 - aside from a few specific exceptions, templates at the root represent the defaults for programming languages
 - we use the `Global` folder for collecting IDEs and tools that build upon whatever you - as many of them support multiple languages

From reading the project site (apologies if I make mistakes here, please clarify if I'm wrong at all) Spring is a framework for building enterprise-scale Java apps. So my first instinct is to ask **"could this be done with rules in Java.gitignore?"** (like we do with other framework-esque relationships like Django is to Python) There's a lot of rules here, so perhaps that's not the ideal approach - others might not want or need these rules.

The next feeling that I have is that this is something you would opt-in for later - rather than when creating a new repo. Let's say your company is a Java shop, so when starting a new project you go and create the repository using the `Java.gitignore` template. A bit later, you work out that the team wants to use IntelliJ - so you add the `Global/JetBrains.gitignore` template to your own. **Does that "pick things as you go" flow feel like how you'd adopt `Spring.gitignore`?**

Of course `Global` is an amorphous namespace and doesn't quite feel right for this situation, but perhaps that's a decision from years ago that I need to revisit...

Spring.gitignore
```diff
@@ -0,0 +1,54 @@
...
+# Ignore some other files 
+*.java.hsp
+*.sonarj
+*.sw*
+ivy-cache
+jxl.log
+jmx.log
+derby.log
+argfile*
+pom.xml
```
hdiwan: Why is this line included? Generally, I want my build directive included in source control.

shiftkey:
Closing due to inactivity and pending questions.

PR #1490, opened 2015-04-26
.gitignore for KiCad EDA Software Suite

AlexAbak:
Signed-off-by: Алексей Кляузер <drum@pisem.net>

### DIFF
--- a/KiCad.gitignore
+++ b/KiCad.gitignore
@@ -1,0 +1,4 @@
+# KiCad
+*.bak
+*cache.lib
+*.kicad_pcb-bak


arcresu:
Thanks! Is this tool commonly used? Is there any documentation or are there any examples to show why you've added these rules?

AlexAbak:
Good day!
Yes. This program commonly used. Is an EDA software suite for the creation of professional schematics and printed circuit boards. http://www.kicad-pcb.org/.

I created multiple repositories for its open-source project. These repositories are the files for the program. But I could not find in the drop-down list .gitignore prohibiting the temporary files that the program creates.
These repositories:
* https://github.com/AlexAbak/TritLedSmdHysteresis
* https://github.com/AlexAbak/TritLedSmd
* https://github.com/AlexAbak/TritLed

Later I would find where the documentation describes the temporary files. That list which I added was determined empirically.

shiftkey:
It looks like we have a PR for this file in #1442, and #1604 added to it.

@AlexAbak please open a new PR if there are any additional rules to add to the template.

PR #1493, opened 2015-04-28
Merge the two Python ignore files

gsnedders:
There's only one `C.gitignore` instead of one for each of GCC/Clang/MSVC, so similarly there should be one `Python.gitignore` instead of one for each of CPython/Jython.

This adds the relevant ignore to `Python.gitignore` for Jython compiled Python files, and gets rid of `Jython.gitignore` (which hasn't been touched since it was added).


arcresu:
Thanks! Yes this seems like a good idea.

PR #1494, opened 2015-04-29
add a rule to not ignore directories ending in .[Cc]ache

LunicLynx:
add a rule to not ignore directories ending in .[Cc]ache

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -152,6 +152,7 @@ AppPackages/
 
 # Others
 *.[Cc]ache
+!*.[Cc]ache/
 ClientBin/
 [Ss]tyle[Cc]op.*
 ~$*


arcresu:
Thanks for this! Could you just help us understand why this is causing a problem?

LunicLynx:
We have a C# Project ending in ".Cache", which when created through vs by default creates a Folder of the same Name.
So in my opinion the ignore rule is too General and should only apply to files with a .Cache Extension but not directories.

For reference:
http://stackoverflow.com/questions/23609669/excluding-cache-files-from-svn-pending-changes
http://stackoverflow.com/questions/588077/what-visual-studio-files-should-be-ignored-by-subversion-to-minimize-conflicts

shiftkey:
Can we document this change to explain why it's needed?

LunicLynx:
Sure, if you let me know what you need, I'll be happy to supply it. It is not a required change per se I guess. Just that the reason for the greedy rule that is in the .gitignore at this moment was the exclusion of files with a .Cache (dot cache) Extension and not Folders. And we just so happen to have a Project (and with that a directory of the same name -> Visual Studio default behavior) ending in .Cache (dot cache).

shiftkey:
@LunicLynx - what about something short and sweet like `a greedy rule to ensure folders ending in .[Cc]ache are not ignored`?

LunicLynx:
Like so?

shiftkey:
@LunicLynx sorry, I wasn't clear. I was thinking of a comment in the `.gitignore` file itself - like we've done with the NuGet syntax further up in the file:

```

# Others
*.[Cc]ache
# a greedy rule to ensure folder names ending in .[Cc]ache are tracked
!*.[Cc]ache/
...
```

LunicLynx:
This request can be closed
i submitted this one instead: https://github.com/github/gitignore/pull/1503

PR #1496, opened 2015-04-30
Add Visual Studio Code .gitignore

Decadence:
Add .gitignore for new cross-platform code editor by Microsoft released yesterday (don't confuse with Visual Studio).
Download: https://code.visualstudio.com/ (may be unavailable sometimes)
Info: http://techcrunch.com/2015/04/29/microsoft-shocks-the-world-with-visual-studio-code-a-free-code-editor-for-os-x-linux-and-windows/

Ignore: editor settings folder.


shiftkey:
Any pointers to docs for what this `.settings` file is (and why it should be ignored)?

dschenkelman:
@shiftkey used to store workspace settings. They are editor preferences for a particular user in that particular workspace. It might make sense to store some (e.g. ES version) not other (e.g. font size).

Since there are a lot that are editor specific it seems to make sense to ignore.

shiftkey:
@dschenkelman would it be easier to just fold this into our existing `VisualStudio.gitignore`? Or does that  not feel right given you might be doing web development?

dschenkelman:
In my particular case I'm not using it from VS. It might make sense to have it separate here and also in the `VisualStudio.gitignore`.

shiftkey:
Yeah, I'm okay with having a new ignore template - there's a lot of things that VS Code won't support, so bringing over all those rules doesn't make sense...

Decadence:
>> Any pointers to docs for what this .settings file is (and why it should be ignored)?
I didn't find any docs because editor preview released just yesterday.
Ignored because it stores user specific preferences.

I don't think it should be merged into Visual Studio rules cause it's not really Visual Studio (not even express version).


P.S. some info (from https://code.visualstudio.com/Docs/customization):

We provide two different scopes for settings:

- User these settings apply globally to any instance of VSCode you open
- Workspace these settings are stored inside your workspace in a .settings folder and only apply when the workspace is opened. Settings defined on this scope overwrite the user scope.

shiftkey:
@Decadence thanks for the extra details. I'll try and get a contact on the team to confirm that this should be ignored by default.

arcresu:
Also, it looks like this should probably be a `Global/` template.

Decadence:
@arcresu I didn't find any mentions about it while using editor or serfing doc, what does it serve for?

Decadence:
@arcresu You're right. I moved it in second commit.

Decadence:
@arcresu What's about merging or I should do something else?

Decadence:
And silence...

arcresu:
@Decadence I believe @shiftkey is in a better position to help with this PR so I was intending to leave it to him, and as per his last comment he's following it up with the team. I just noticed the thing about `Global/` is all.

shiftkey:
Still haven't heard back from the team on this. Better to ask for forgiveness than permission, I guess...

Decadence:
@shiftkey ahaha)
@arcresu I thought you are the person from team which he contacted)

PR #1497, opened 2015-05-02
Add bower_components to Node.gitignore

pine613:
Hi.
I'm using Node.js to development in back-end and front-end.

Bower is very popular package manager for front-end.
So, I think that it is good if add a dependency directory of Bower.

The `bower_components` directory is Bower's dependency directory.
In [Bower official repository](https://github.com/bower/bower/blob/master/.gitignore), it has been added in `.gitignore`.

Thank you.


shiftkey:
Closing in favour of #1601

PR #1498, opened 2015-05-02
Windows.gitignore additions

Sduibek:
Added entries for:
System Volume Information directory
Recovery directory
pagefile
hiberfile
the two alternate versions of Recycle Bin directory

### DIFF
--- a/Global/Windows.gitignore
+++ b/Global/Windows.gitignore
@@ -7,6 +7,8 @@ Desktop.ini
 
 # Recycle Bin used on file shares
 $RECYCLE.BIN/
+$Recycle.Bin/
+RECYCLER/
 
 # Windows Installer files
 *.cab
@@ -16,3 +18,15 @@ $RECYCLE.BIN/
 
 # Windows shortcuts
 *.lnk
+
+# System Restore data & restore points. (Also: Windows Backup, Volume Shadow Copy, Content Indexer, CHKDSK logs)
+System Volume Information/
+
+# System Recovery boot image
+Recovery/
+
+# Virtual Memory file
+pagefile.sys
+
+# Hibernation data
+hiberfil.sys


arcresu:
Thanks, but again I don't understand why the case insensitivity is necessary, and if you're adding new rules, it would be helpful to have some explanation and ideally documentation links.

Sduibek:
Hello! I apologize for the delay in responding to this.

Does git/GitHub ignore case to begin with?

Some of the current .gitignore templates do have declarations that specify case-insensitivity with the [], so I assumed case-insensitivity in such declarations was useful.

On Windows at least, it's possible for the same type of file that we want to ignore to exist in different cases, such as "Desktop.ini" and "desktop.ini" which can both exist.

Regarding documentation, the most relevant/credible is probably:
http://social.technet.microsoft.com/wiki/contents/articles/4490.default-root-folders-in-windows-7.aspx

Others:
https://www.raymond.cc/blog/about-recycler-and-system-volume-information-folder-in-xp-and-vista/
http://blogs.msdn.com/b/oldnewthing/archive/2003/11/20/55764.aspx
http://www.theeldergeek.com/system_volume_information_folder1.htm
http://answers.microsoft.com/en-us/windows/forum/windows_7-system/what-is-the-crecovery-folder/a2067cd8-7de2-4871-98a6-2b882050a892

vgturtle127:
@arcresu It looks good to me, I have seen various capitalization and case sensitivities when analyzing hard drives and directories on various version of Windows.

For example, thumbs.db is often lowercase, but sometimes is uppercase depending on what application is the default for running the images in that location.

It sounds stupid and like it's a bug, and it is, but that's how it works. At least somewhat, it seems random to me at times. desktop.ini is another weird one that changes based on the time of day and what color shirt you are wearing. :confused: 

Regardless, I use these entries in all of my repositories, and nothing has broken. :smile: 

Seems OK to merge in my opinion.

shiftkey:
> Does git/GitHub ignore case to begin with?

This depends on how Git is configured: `git config core.ignorecase`

From the docs:

> If true, this option enables various workarounds to enable Git to work better on filesystems that are not case sensitive, like FAT. For example, if a directory listing finds "makefile" when Git expects "Makefile", Git will assume it is really the same file, and continue to remember it as "Makefile".
>
> The default is false, except git-clone[1] or git-init[1] will probe and set core.ignoreCase true if appropriate when the repository is created.

shiftkey:
@Sduibek unfortunately there are now conflicts - are you interested in resolving these?

PR #1500, opened 2015-05-02
MicrosoftOffice.gitignore case-insensitivity

Sduibek:
Added case-insensitivity for temp files.

### DIFF
--- a/Global/MicrosoftOffice.gitignore
+++ b/Global/MicrosoftOffice.gitignore
@@ -1,4 +1,4 @@
-*.tmp
+*.[Tt][Mm][Pp]
 
 # Word temporary
 ~$*.doc*


arcresu:
Thanks, but could you explain why this is necessary? If you're using a case-insensitive filesystem then this is unnecessary, and if you're not on Windows then it's too late for gitignore rules to help you.

Sduibek:
Hello! I apologize for the delay in responding to this.

Does git/GitHub ignore case to begin with?

Some of the current .gitignore templates do have declarations that specify case-insensitivity with the [], so I assumed case-insensitivity in such declarations was useful.

arcresu:
As I understand it, git is case sensitive if the underlying filesystem is case sensitive. If you haven't run into any specific problem with this rule, then it's probably not worth making rules case agnostic just in case.

Sduibek:
Okay, fair enough. I will go ahead and close this.

Thank you.

PR #1501, opened 2015-05-04
add yii2 php framework default .gitignore file

su-xiaolin:


arcresu:
Thanks, but it seems like this is essentially composed of existing templates. We prefer to keep things modular and avoid duplication unless there's a good reason. Thanks anyway!

PR #1502, opened 2015-05-05
Add gitignore file for dotnet application

diegoleonardo:


shiftkey:
For some platforms which are IDE-focused we transitioned from using language-based templates (which often duplicated the same things) into a unified template.

We originally had ignore templates for VB and C#, but these were unified under the [Visual Studio](https://github.com/github/gitignore/blob/master/VisualStudio.gitignore) template a while ago.

If there's things are missing from there, please submit a PR to update that file.

PR #1503, opened 2015-05-06
add rule to keep track of directories ending in .Cache

LunicLynx:
More Information
http://blogs.msdn.com/b/msbuild/archive/2008/02/11/what-s-up-with-xxx-sln-cache.aspx


shiftkey:
Thanks!

PR #1507, opened 2015-05-07
Added .RData and *.Rproj rules to R.gitignore

rohit-gupta:
.RData files are usually saved at the end of an R session and contains the workspace image. (https://stat.ethz.ch/R-manual/R-devel/library/base/html/save.html)

*.Rproj is the extension used by R Studio projects


arcresu:
Thanks, but see #1245 and #1194 for why these rules were removed.

PR #1508, opened 2015-05-07
Android: Remove redundant pattern for matching build directories

baudm:
build/ already matches all directories named 'build' in the repository,
regardless of level. Gradle can have more than two levels of project nesting.
However, /*/build/ matches only the 'build' directories of 2nd-level projects.
Thus, the first pattern is more appropriate than the second.

### DIFF
--- a/Android.gitignore
+++ b/Android.gitignore
@@ -15,7 +15,6 @@ gen/
 # Gradle files
 .gradle/
 build/
-/*/build/
 
 # Local configuration file (sdk path, etc)
 local.properties


arcresu:
Thanks, but are you sure about that? I haven't checked, but I seem to recall that if there is a slash anywhere in the pattern, git starts anchoring it to the top of the project. It would be nice to double check this first.

baudm:
Yes I'm sure. /build would match a file or directory at the root, while
build/ would match a directory at any level. /*/build/ would match a
directory at the second level while /**/build/ would match a directory at
any level, which is essentially the same as build/.

baudm:
Pertinent excerpts from the [gitignore manpage](http://git-scm.com/docs/gitignore):

> If the pattern ends with a slash, it is removed for the purpose of the following description, but it would only find a match with a directory. In other words, foo/ will match a directory foo and paths underneath it, but will not match a regular file or a symbolic link foo (this is consistent with the way how pathspec works in general in Git).

...

> A leading slash matches the beginning of the pathname. For example, "/*.c" matches "cat-file.c" but not "mozilla-sha1/sha1.c".

If you're still in doubt, you could check it yourself. :)

baudm:
@arcresu any updates on this? I've already quoted the pertinent excerpts from the gitignore manpage; all info needed is there. And as I said before, this is really easy to test. Just create a new Git repo, add a bunch of "build" directories on different levels, and observe how each pattern works.

arcresu:
Sorry for the delay - I haven't had much time to process PRs here lately. Yes, you're right about this rule. Thanks for confirming, I just couldn't quite remember how it worked. :tada:

PR #1509, opened 2015-05-07
Update Python.gitignore

NamPNQ:


shiftkey:
Interesting - the [advice](https://www.jetbrains.com/idea/help/project.html#d1369943e149) from JetBrains seems to be to keep this directory in source control:

> The `.idea` directory contains a set of configuration files (`.xml`). Each file contains only a portion of configuration data pertaining to a certain functional area which is reflected in the name of a file, for example, `compiler.xml`, `encodings.xml`, `modules.xml`.

> Almost all of the files contain information core to the project itself, such as names and locations of its component modules, compiler settings, etc. **Thus, these files may (and should) be kept under version control.**

NamPNQ:
I think many dev don't want keep it, because someone don't use pycharm, why we need keep it? And as i known, in .idea had project config (this is for one user, and no need apply to all person using this source). This is my point :D

arcresu:
Thanks, but all the jetbrains stuff is kept in `Global/JetBrains.gitignore`, so we should focus our efforts on that. There are already some PRs to update that template, so we can carry on the discussion in #1475 and #1399, which I haven't had a chance to look at yet.

PR #1510, opened 2015-05-07
Android: Add Android Studio support

j796160836:
Update Android.gitignore to Add Android Studio support


arcresu:
Thanks, but we prefer to keep the templates modular where possible. In this case you'll find most of these rules under `Global/JetBrains.gitignore` and the various operating system templates under `Global/`.

PR #1511, opened 2015-05-07
update VisualStudio.gitignore file : ignore visual studio 2015 Roslyn…

nahidf:
… ignored Compiler engine temporary files


shiftkey:
As of VS15 CTP6 these temporary files now live under the `.vs` directory, which we've already added (see line 26).

PR #1512, opened 2015-05-07
Ignore common OS generated files

YavorIvanov:


shiftkey:
These feel like they should be added to the OSX ignore file, right?

arcresu:
Exactly, these are best kept in `Global.OSX.gitignore`. Thanks anyway!

PR #1514, opened 2015-05-10
sync up xcode related gitignore files

bbodenmiller:
Xcode, Obj-C, & Swift updates. Also comment to help prevent fragmentation.

Some details based on #1221

### DIFF
--- a/Global/Xcode.gitignore
+++ b/Global/Xcode.gitignore
@@ -1,4 +1,12 @@
+# Xcode
+#
+# gitignore contributors: remember to update Global/Xcode.gitignore, Objective-C.gitignore & Swift.gitignore
+
+## Build generated
 build/
+DerivedData
+
+## Various settings
 *.pbxuser
 !default.pbxuser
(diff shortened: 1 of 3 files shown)


shiftkey:
Is it worth having a discussion about unifying these ignore files? We did this a while ago for `VisualStudio` - which was a combination of `CSharp` and `VisualBasic` languages - and I can't help but feel this is heading down that path...

arcresu:
@shiftkey See comments on 5bd4330 for why we have the current state of these particular templates. I think it needs fixing, but people have strong feelings here.

bbodenmiller:
Can we get these changes merged? Given the comments on the other thread I don't see a merge of the templates being feasible in the near future.

shiftkey:
@bbodenmiller I'll give @arcresu a chance to chime in if has any issues with it, but I'm :thumbsup: with this change.

arcresu:
Thanks, yeah this looks like a good change.

PR #1516, opened 2015-05-11
Ignore temp Hypothesis test things

hugovk:
https://github.com/DRMacIver/hypothesis
https://hypothesis.readthedocs.org

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -43,6 +43,7 @@ htmlcov/
 nosetests.xml
 coverage.xml
 *,cover
+.hypothesis/
 
 # Translations
 *.mo


hugovk:
cc @DRMacIver

DRMacIver:
Seems eminently reasonable to me, yes. Checking these in is unlikely to ever be a good idea.

hugovk:
@shiftkey Hi, do you think this could be merged? Thanks!

shiftkey:
Do you have a link to what this file represents and why it can be safely excluded from version control?

hugovk:
@shiftkey I couldn't find it documented in its docs https://hypothesis.readthedocs.org/en/latest/, but perhaps that's because it's for its own local temp stuff.

---

It's ignored by its own project: https://github.com/DRMacIver/hypothesis/blob/bb66deacc831426294d163ef3b69060df4b54104/.gitignore#L6

It's the home directory used for storage:
https://github.com/DRMacIver/hypothesis/blob/69f84cb31e09b88c609140ad5ec5c30439f8a8ba/src/hypothesis/settings.py#L51-L67

Which is used for caching:
https://github.com/DRMacIver/hypothesis/blob/d5de4a83cf768017556df80f3a5c4c21ce066ab0/src/hypothesis/internal/reflection.py#L337-L357

---

The project author @DRMacIver said earlier in this issue:

> Seems eminently reasonable to me, yes. Checking these in is unlikely to ever be a good idea.

https://github.com/github/gitignore/pull/1516#issuecomment-100817707

If you need more, perhaps @DRMacIver can help.

shiftkey:
@hugovk thanks! I just wanted some details to point to, because we often get people come back weeks/months/years after and want to understand a specific rule (or even challenge it).

DRMacIver:
You're absolutely right that having some documentation about this would be useful. I've created a ticket about that.

hugovk:
Thanks!

Python.gitignore
```diff
@@ -43,6 +43,7 @@ htmlcov/
 nosetests.xml
 coverage.xml
 *,cover
+.hypothesis/
```
kxepal: The problem is that .hypothesis directory may occurs not only at the project root, but in project subdirectories as well (like tests/.hypothesis). This depends on how you run the tests. To ignore these cases you need to remove trailing slash. FYI @DRMacIver

jwg4: I think this is wrong @kxepal . The trailing slash just means that it is a directory. A *leading* slash would mean to only ignore a file with this name in the root directory (or in any case, the directory which contains the `.gitignore`).

kxepal: And you're right: it seems bug in my ide and it thinks that if path contains trailing slash so that rule shouldn't get applied to subdirectories. I should have to check git instead. Sorry for noise, everything is correct here!

PR #1517, opened 2015-05-13
Add extention compiled files python 2.x

Bleno:
In python2 always is compiled files when a module is executed individually for testing.


arcresu:
Thanks, but these are already covered by the rule two lines above :)

PR #1521, opened 2015-05-14
Remove Meteor.gitignore

lezed1:
There are only two entries in this .gitignore, and both are not needed.
- .meteor/local is already ignored by the standard Meteor .gitignore that is included with every project. This is redundant and should b left solely to the standard Meteor .gitignore in case this changes in the future.
- .meteor/meteorite should no longer be in any Meteor project. Meteorite was a package manager for Meteor, but it has been replaced by an official package manager built into the Meteor CLI. This directory should not exist, except on out of date Meteor projects which already have this ignore present.

### DIFF
--- a/Meteor.gitignore
+++ b/Meteor.gitignore
@@ -1,2 +1,0 @@
-.meteor/local
-.meteor/meteorite


shiftkey:
@lezed1 thanks for the details - do you have a link to some supporting docs about this so I can close this out?

lezed1:
Here is the skeleton .gitignore with local already included: https://github.com/meteor/meteor/blob/devel/tools/skel/.meteor/.gitignore

And here is the web interface of the package manager demonstration the install command which no longer uses meteorite: https://atmospherejs.com/meteor/accounts-github

lezed1:
Is there anything else needed to be done for this pull request?

arcresu:
Sorry for the wait; this looks good. Thanks! :tada:

PR #1523, opened 2015-05-15
[Tex] Change *.mtc0 to *.mtc[0-9] *.mtc[1-9][0-9]

dash00:
When using minitoc package, LateX often generates multiple .mtc files such as `*.mtc0`, `*.mtc1`, ..., `*.mtc13`, ... This patch allows one to ignore all *.mtc files from index 0 to index 99 (ignoring ` *.mtc[0-9] ` and ` *.mtc[1-9][0-9] `).

### DIFF
--- a/TeX.gitignore
+++ b/TeX.gitignore
@@ -93,7 +93,8 @@ acs-*.bib
 # minitoc
 *.maf
 *.mtc
-*.mtc0
+*.mtc[0-9]
+*.mtc[1-9][0-9]
 
 # minted
 _minted*


shiftkey:
@dash00 this seems like a good rule to make more generic - do you have any external references to this behaviour for us to point curious users at?

dash00:
There is a very nice explanation for this by Werner in this topic of tex.stackexchange : http://tex.stackexchange.com/a/36874/77285

arcresu:
A nice thing about TeX is that the packages are generally written in a literate code format which compiles into a commentary of the source code, so it's easy to get references for any behaviour. From the [manual](http://texdoc.net/texmf-dist/doc/latex/minitoc/minitoc.pdf) on p. 303:

>`\@tocfile`
The name of the file containing the minitoc is constructed from `\jobname` and a suffix `\@tocfile`, which is `.mtc` (long extensions) or `.M` (short extensions) followed by the absolute number of the minitoc.

```
\if@mtc@longext@
  \def\@tocfile{mtc\The@mtc}%
\else
  \def\@tocfile{M\The@mtc}%
\fi
```

PR #1524, opened 2015-05-15
First .gitignore for Stata files

hcollet:
Although Stata is not the most frequent language of statistical softwares on GitHub, I thought it would be good to have a proper .gitignore file for this kind of project.


shiftkey:
As [Stata](http://www.stata.com/stata14/) seems to be about an IDE-esque environment, could this live under the `Global/` folder as per this [README](https://github.com/github/gitignore#folder-structure) section?

hcollet:
Sure thing, this makes sense as these are editor-specific files.
Thanks

arcresu:
Is there any documentation explaining these rules?

Also note that `Global/` rules are meant to go in a developer's system-wide gitignore rules, so as it stands all images would be ignored in all projects with these rules. It's a bit of a problem with the IDE = Global assumption in how we lay out these templates. Is it necessary to ignore these image extensions at all here?

hcollet:
Agree, I have commented out the exclusion of image files that can be exported from the editor environment with a documentation link. Let me know if that needs some other tweaks.

shiftkey:
@hcollet apologies for the delay - this all seems :gem:!

PR #1525, opened 2015-05-15
Add trailing slash to .sass-cache

LayZeeDK:
Add trailing slash to `.sass-cache` to signify intent of matching directories.

### DIFF
--- a/Sass.gitignore
+++ b/Sass.gitignore
@@ -1,2 +1,2 @@
-.sass-cache
+.sass-cache/
 *.css.map


shiftkey:
As someone who's not familiar with SASS (i was in the LESS camp the last time I did web dev seriously) do you have a link to an explainer for this `sass-cache` folder?

shiftkey:
The current `.sass-cache` rule will catch a directory named the same - is there something that's not working about this current setup?

LayZeeDK:
I am under the impression that it is good practice to mark directories with a trailing slash in `.gitignore`s. Am I wrong?

arcresu:
There are slight subtleties about the trailing slash when it comes to selectively un-ingnoring contents of the directory. In this case I don't think you'd ever want to do that, so it shouldn't make a difference, and the slash documents that it is a directory, so its presence is slightly preferable as far as I see.

I would recommend against mass-changing rules to add or remove slashes, though, because it's not always straightforward. I tend to leave the trailing slash or lack thereof as it was first introduced unless there is a specific problem to address.

chenyunyou:
2506571489@qq.com

arcresu:
I'll merge this because it seems fine, but with the note that in general changing trailing slashes without a specific motivation is probably not very useful and at worst could cause subtle issues.

PR #1526, opened 2015-05-16
Removed the extra "/" from the directory ignore.

haridas:


shiftkey:
Is this a recent change in Rust or just an oversight?

haridas:
Yes, the new Rust-1.0's cargo build tool doesn't required this "target" folder. I was playing around with a fresh project after the new Rust stable release.

arcresu:
See #1563

PR #1528, opened 2015-05-19
Added gitignore for Biicode - C and C++ Dependency Manager

TheKnarf:
https://www.biicode.com/

### DIFF
--- a/Biicode.gitingore
+++ b/Biicode.gitingore
@@ -1,0 +1,3 @@
+# Biicode
+
+bii/


arcresu:
Thanks, but could you please explain why this folder should be ignored, linking directly to documentation? Is this tool widely used, enough to justify its inclusion in this collection?

TheKnarf:
It ignores the `bii` folder which is filled with local dev settings.

I can't seem to find git mentioned in their documentation, so the closest reference I can link to is [this](http://stackoverflow.com/questions/29527482/what-to-put-in-gitignore-when-using-bii-with-travis)

Biicode is one of the only dependency managers for C++ and it's becoming more and more popular within the C++ community.

shiftkey:
Closing due to net impact. @TheKnarf I'd recommend submitting it as an optional rule for specific existing templates, rather than a standalone template.

PR #1529, opened 2015-05-19
Removing bower_components and fixing typo

sayedihashimi:


shiftkey:
@sayedihashimi I've found [this advice](http://addyosmani.com/blog/checking-in-front-end-dependencies/) about it, but do you have a more official URL handy for us to point to?

sayedihashimi:
Adding @madskristensen and @DamianEdwards who were more involved in the discussion regarding this than I was.

sayedihashimi:
th

barrytang:
There are basically two workflows we want to support going forward:
 
1.      People reference ./wwwroot/lib/bower_components directly (this is also how our default template is configured too)
2.      People use build steps to copy files under bower_components to ./wwwroot 
 
We don’t want to special case bower_components for SCC since for workflow #1, people would want to check in ./wwwroot/lib/bower_components.

balachir:
@shiftkey, does Barry's comment  address your concerns?

shiftkey:
@balachir :thumbsup:

hsn95:
..

ternjed:
hi guys I would like to learn as much about what is going on and how I we meaning my kids as this is the reason I am doing this although this is very exciting and one day when I can make a game that my children will play and tell there friends that my dad made this I can not wait thanks for this to all the pioneers who layed the ground work that's all for now

samuel-velazquez:
Secund tool sbyte soget in body take only setup so var stage info tecnology process

PR #1530, opened 2015-05-20
ignore OAuth2 token

katrinleinweber:
Created for example when using https://github.com/jennybc/googlesheets


arcresu:
Thanks! Looks like the [relevant documentation](https://github.com/jennybc/googlesheets/blob/8471679c621c8b5d43679cf4b9598cba2f6a3560/internal-projects/13_how-httr-caches-tokens.md) indicates it includes secrets so this should be ignored.

PR #1531, opened 2015-05-23
Add CMakeScripts directory

emlai:
This folder contains user-specific scripts used by some build systems, at least Xcode.

### DIFF
--- a/CMake.gitignore
+++ b/CMake.gitignore
@@ -1,5 +1,6 @@
 CMakeCache.txt
 CMakeFiles
+CMakeScripts
 Makefile
 cmake_install.cmake
 install_manifest.txt


shiftkey:
@emlai do you have a link to what this folder represents (whether from the CMake docs or from elsewhere)?

emlai:
@shiftkey Unfortunately there seems to exist no official documentation on this folder.

After some googling it seems that it's only used by CMake's Xcode generator. According to [this post](http://inkscape.13.x6.nabble.com/cmake-next-steps-td4973562.html) it "holds various helper functions". The files inside this folder contain user-specific paths, so I automatically ignored it in my project.

I initiated a [StackOverflow question](http://stackoverflow.com/questions/30423793/what-is-the-purpose-of-cmakescripts) to find a definitive answer on this.

emlai:
@shiftkey Update: Someone [answered](http://stackoverflow.com/a/30437827/3425536) my question on SO and according to them, `CMakeScripts` contains "internal scripts to hack around Xcode's make files". For example, in the CMake source code a variable that stores that path is called `CurrentXCodeHackMakefile`.

arcresu:
Thanks for following up on this! It looks like this is something you'd want to ignore

PR #1533, opened 2015-05-27
Added Maven .gitignore model

lossurdo:
Just to contribute and, based on Java.gitignore, also ignore Maven target folder.


shiftkey:
I'm not sure I understand - should we be updating [this](https://github.com/github/gitignore/blob/master/Maven.gitignore) template?

lossurdo:
My mistake! I couldn't find this template before so I created one. Please ignore my pull request.

PR #1534, opened 2015-05-28
Add several downloaded files that should not be incorporated and remove .SRCINFO

marcelkorpel:
As of version 4.0.0 the Arch User Repository is build out of one git repository per package. The .SRCINFO files need to be included in those.

Moreover, I added several extensions of files that are regularly downloaded by PKGBUILD scripts; these shouldn't be included in git repositories of package sources.

### DIFF
--- a/ArchLinuxPackages.gitignore
+++ b/ArchLinuxPackages.gitignore
@@ -1,5 +1,8 @@
 *.tar
 *.tar.*
+*.jar
+*.exe
+*.msi
 *.zip
 *.tgz
 *.log
--- a/ArchLinuxPackages.gitignore
+++ b/ArchLinuxPackages.gitignore
@@ -6,9 +6,5 @@
 *.log.*
 *.sig
 
-# AUR metadata
-.AURINFO
-.SRCINFO
-
 pkg/
 src/


Zopieux:
I second that. I was about to open a similar PR!

shiftkey:
@marcelkorpel @Zopieux any links to help us (and others) understand the changes better?

marcelkorpel:
https://wiki.archlinux.org/index.php/Arch_User_Repository#AUR_4

Zopieux:
> When making changes to the repository, make sure you always include the PKGBUILD and .SRCINFO in the top-level directory.

...
> the .SRCINFO must exist for every commit

shiftkey:
Thanks for those details - I also found [this link](https://patchwork.archlinux.org/patch/2513/) which indicates `.AURINFO` contains the same contents as `SRCINFO`, but it named differently to indicate it's relevant to the AUR.

marcelkorpel:
.SRCINFO is just the new name of .AURINFO; currently, the former is used by the AUR to parse metadata (to prevent parsing/executing possibly malicious bash scripts on the server).

PR #1535, opened 2015-05-29
Add JavaScript gitignore template.

hotoo:
* `node_modules/` is dependencies of node.js and io.js, by npm.
* `spm-modules/` and `sea-modules/` is dependencies of spm.

dependencies will ignored in git.

### DIFF
--- a/JavaScript.gitignore
+++ b/JavaScript.gitignore
@@ -3,5 +3,5 @@
 node_modules/
 
 # spm. see http://spmjs.io/ and https://github.com/spmjs/spm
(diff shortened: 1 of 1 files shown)


arcresu:
Thanks, but given that there is already a `Node.gitignore`, I'm not sure that this is necessary. If `spm` is only used in when node is not, then it would seem weird to put them in the same template, and if they are used together, perhaps it could go there.

zzz6519003:
@arcresu node is not js

PR #1538, opened 2015-05-31
Added solution & project ncrunch files

MikeGoldsmith:
Both *.ncrunchsolution & *.ncrunchproject are not required to be source controlled.

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -106,6 +106,8 @@ _TeamCity*
 # NCrunch
 _NCrunch_*
 .*crunch*.local.xml
+*.ncrunchsolution
+*.ncrunchproject
 
 # MightyMoose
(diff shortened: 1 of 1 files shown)


shiftkey:
@MikeGoldsmith do you have a link to the NCrunch docs so we can link this discussion back to the tool itself?

MikeGoldsmith:
After re-reading the NCrunch documentation it seems I had misunderstood the meaning behind the ncrunchsolution * ncrunchproject files. They are valid for inclusion in source control and should not be added to the .gitignore.

Reference: https://www.ncrunch.net/documentation/considerations-and-constraints_using-ncrunch-with-source-control

shiftkey:
@MikeGoldsmith no problem, and thanks for the link!

PR #1539, opened 2015-06-01
Added exclusion of captures folder

Numan1617:


shiftkey:
Unfortunately this isn't mergeable currently, and I'd love to see some supporting links on why users should have this rule in their gitignore file. 

PR #1540, opened 2015-06-01
Ignoring: Android Studio local files

gbazilio:


hiralin:
+1

ryanseys:
Easier to just ignore:

```
.idea
*.iml
```

But yes, I agree that Android Studio should be added to the ignored files.

arcresu:
Thanks, but this is why we have `Global/JetBrains.gitignore`

ryanseys:
I'd suggest also having these in Android Studio considering they come with the default .gitignore that Android Studio creates.

arcresu:
The `Global/` rules are intended to go in your global ignore file so that you wouldn't have to add them in each project. That's the way we've divided up this repository.

PR #1541, opened 2015-06-04
ignore PyCharm project file

zhanglongqi:


arcresu:
Thanks but see `Global/JetBrains.gitignore`

zhanglongqi:
OH.... Thanks. but how can we select this when we create new repository?
Is this the only way that we download and add it to our project?

arcresu:
You can't, but the idea is that with git those kind of rules belong in your global ignore list. Have a look at the gitignore man page.

PR #1542, opened 2015-06-05
Document Vim file patterns

barraponto:
I'm a heavy vim user, but those patterns were not obvious to me.
So I thought documenting them would be a nice addition.
I guessed what the patterns were from the git blame :8ball:


shiftkey:
This is great! Thanks!

PR #1543, opened 2015-06-05
Add scrapy stuff to .gitignore

barraponto:
The `.scrapy` is used for caches, should just be ignored.


shiftkey:
I gather we're talking about this project, right? http://scrapy.org/

barraponto:
Yes, we are.

shiftkey:
@barraponto excellent, thanks for confirming!

PR #1544, opened 2015-06-07
Update Java.gitignore

balu213:
Added the list to be ignored for a java project like the build folder, .class generated files and the project files generated by the eclipse editor


arcresu:
Thanks, but we keep Eclipse rules under `Global/Eclipse.gitignore`

PR #1546, opened 2015-06-08
DerivedData and xcuserdata are folders; *.xcuserstate is always inside

Coeur:
DerivedData and xcuserdata are folders
*.xcuserstate is always inside xcuserdata folder

### DIFF
--- a/Global/Xcode.gitignore
+++ b/Global/Xcode.gitignore
@@ -4,7 +4,7 @@
 
 ## Build generated
 build/
-DerivedData
+DerivedData/
 
 ## Various settings
 *.pbxuser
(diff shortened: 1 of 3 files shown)


Global/Xcode.gitignore
```diff
@@ -15,9 +15,9 @@ DerivedData
...
 !default.perspectivev3
-xcuserdata
+xcuserdata/
 
 ## Other
-*.xccheckout
 *.moved-aside
-*.xcuserstate
+*.xccheckout
+*.xcscmblueprint
```
shiftkey: This one seems to be new, but I'm not going to hold this up as it seems very specific.

shiftkey:
Thanks!

PR #1548, opened 2015-06-09
Sonar .gitignore file

rvacaru:
When running "Get Issues from Sonar runner" from NetBeans (using the radar sonar plugin) a {project_dir}/.sonar directory is created.
This directory is also created when running the maven sonar:sonar goal in preview mode analysis.

### DIFF
--- a/Sonar.gitignore
+++ b/Sonar.gitignore
@@ -1,0 +1,2 @@
+#Sonar generated dir
+/.sonar/


shiftkey:
As this seems to be a NetBeans-specific plugin, it should live under the [NetBeans.gitignore](https://github.com/github/gitignore/blob/master/Global/NetBeans.gitignore) template.

rvacaru:
After further investigating I found out the {project_dir}/.sonar directory is the default value for the 'sonar.working.directory' optional parameter, as you can read from the SonarQube documentation below:
http://docs.sonarqube.org/display/SONAR/Analysis+Parameters

This directory is used to store the analysis that you don't want to store in the sonarqube server. These so called preview analysis can be performed using maven goals, ant tasks or manually by using the sonar runner, which do not depend by the IDE. So I don't think this is related to Netbeans after all.

rvacaru:
Could you give me a feedback?

shiftkey:
> Could you give me a feedback?

I'll try my best, but I'm still rather unfamiliar with the Sonar runner.

I'm trying to find the best place to put this rule. A new file for just one rule feels excessive, so I'd love to understand whether there's a more suitable home for this, so that others are going to have this rule without having to come and find this new file and add it to their project.

For example, while you mention that this isn't specific to the NetBeans IDE, [this repository](https://github.com/SonarSource/sonar-runner) mentions a number of related technologies in it's README:

> Common library used by many Scanners (SQ Scanner, SQ Scanner for **Maven**, SQ Scanner for **Gradle**, SQ Scanner for **Ant**, ...) and by SonarLint. Used to programmatically run SQ analysis.

I'd love to understand more about how Sonar is being used in your setup, so I can hopefully propose a better home for this rule.

shiftkey:
Closing due to impact. @rvacaru I'd recommend submitting this against the related existing templates, rather than creating a new standalone template.

PR #1551, opened 2015-06-12
add py ignore file

0312birdzhang:


barraponto:
Those are already ignored by the Python `.gitignore`.

arcresu:
Yes, that's right - we should leave these for the python template unless there's a strong reason to add them here as well.

PR #1552, opened 2015-06-12
Add PDB

linouxis9:


shiftkey:
Can you elaborate a bit on why this should live in the C template?

We have a Visual Studio template which includes this rule.

shiftkey:

PR #1553, opened 2015-06-13
Added the "cache" folder entry from "App_data"

mrflo:
The cache folder is generated and don't need to be commited

### DIFF
--- a/Umbraco.gitignore
+++ b/Umbraco.gitignore
@@ -6,6 +6,7 @@
 **/App_Data/Logs/
 **/App_Data/[Pp]review/
 **/App_Data/TEMP/
+**/App_Data/[Cc]ache/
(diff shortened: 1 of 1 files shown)


shiftkey:
@mrfloo any pointer to Umbraco docs about these rules, so we can point users to them to understand more about the setup?

mrfloo:
Hi @shiftkey,
I could'nt find a nice documentation on this. This folder is caching all the content since version 7.1.1 
As explained in this article about Azure but that's happen outside of azure as well: http://www.orbitalgamestudios.com/blog/umbraco-cache-size-azure/

shiftkey:
Closing in favour of #1661 which has this rule (and others) and is currently under review.

PR #1554, opened 2015-06-16
Added idea to python gitignore

yershalom:


arcresu:
Thanks but see `Global/JetBrains.gitignore`

PR #1555, opened 2015-06-16
To ignore gcov note file and data file.

kimizhou:


arcresu:
Thanks but see `Gcov.gitignore`

PR #1559, opened 2015-06-18
Add Dropbox.gitignore

tjrana:
Folders that are synced with Dropbox can contain user-specific files and caches. Dropbox.gitignore adds some Dropbox gitignore files based on the following Dropbox Support articles:
https://www.dropbox.com/help/328
https://www.dropbox.com/help/145

### DIFF
--- a/Global/Dropbox.gitignore
+++ b/Global/Dropbox.gitignore
@@ -1,0 +1,4 @@
+# Dropbox settings and caches
+.dropbox
+.dropbox.attr
+.dropbox.cache


arcresu:
Thanks, but could you help us understand why dropbox and git would ever be interacting? I don't think that syncing git repos over dropbox is a particularly good idea since it could corrupt the repo.

vgturtle127:
@tjrana @arcresu 

I think the reasoning behind it is that he wants to sync larger files or specific files he wants to backup that might not be necessary for whatever application he is using it for.

An example being a large movie file or 3D models or something. If he adds those to his .gitignore, that would theoretically work, but it may add his cache files to the repo without him even seeing them.

To prevent that, he wants to add the Dropbox cache to a .gitignore, and by doing so it will make sure no files from his Dropbox are put into his Git repository, but if he wants to use services interchangeably, they should play nice.

I hope I explained that well enough. :smile: 

I think this sounds alright, especially if other people are using Dropbox and Git together.

tjrana:
@arcresu 
I agree with @vgturtle127's explanation. For those who do use Dropbox and Git together, a .gitignore can surely help a lot.

vgturtle127:
@arcresu @tjrana And there is no possible loss of data or anything, so I see no harm in it.

shiftkey:
These rules are specific enough to Dropbox to be considered safe to opt-in for.

I have reservations about about mixing the two together, but good luck!

PR #1560, opened 2015-06-19
Create Bitrix Framework .gitignore

mmjurov:
http://www.bitrixsoft.com/
Ignored core files which shouldn't be under vcs

### DIFF
--- a/Bitrix.gitignore
+++ b/Bitrix.gitignore
@@ -1,9 +1,20 @@
-bitrix/*
-!bitrix/templates
-!bitrix/components
-bitrix/components/bitrix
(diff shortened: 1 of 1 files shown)


arcresu:
Thanks, but could you provide some documentation to explain the specific rules you're proposing here?

shiftkey:
@mmjurov apologies for the delay in replying to this.

I had a lot of these templates requested, but I couldn't quite figure out the right place for them. I ended up settling on #2857, and you'll see the Bitrix template can now be found at `community/PHP/Bitrix.gitignore`.

PR #1561, opened 2015-06-23
Added Android studio Project files

dmsherazi:


arcresu:
Thanks, but these rules are all for specific IDEs and operating systems. We keep those rules in the corresponding `Global/` templates.

PR #1563, opened 2015-06-24
allow nested rust projects

ddavison:


arcresu:
Thanks, but could you explain what this change is trying to fix?

ddavison:
sure -

Consider the following project structures:

## Structure 1
```
$ ls
project1/
  .gitignore
  src/
  target/
  Cargo.toml
  Cargo.lock
```

## Structure 2
```
$ ls
project2/
  .gitignore
  nested_project1/
    src/
    target/
    Cargo.toml
    Cargo.lock
  nested_project2/
    src/
    target/
    Cargo.toml
    Cargo.lock
 ```

In Structure #1, the current gitignore template works.
In Structure #2, the current gitignore template does **not** work.

With this fix, Structure #1 and Structure #2 will both work.

arcresu:
Thanks for that. I don't use rust, so I don't know, but I would have thought the normal thing would be to have separate repositories for each project. Are they commonly nested like this?

ddavison:
maybe not, but it doesn't hurt :) like I said - it fits both cases

arcresu:
It does hurt in that it would cause any directory anywhere in your project with the name `target` to be ignored, which could be unexpected and hard to notice on the off-chance it happens. I'd prefer to assume one project per repo and keep the rules specific, unless there's a common pattern of usage that makes this unreasonable. Thanks anyway!

PR #1565, opened 2015-06-24
Add support for the `stack` build tool

athanclark:


shiftkey:
@athanclark do you have a link that explains what this folder represents and why it can be safely ignored from version control?

athanclark:
Actually, I should'v just used the source's:
https://github.com/commercialhaskell/stack/blob/master/.gitignore

:)

shiftkey:
The rule on the source repo has the leading `/`, and this change doesn't. Should they be the same?

athanclark:
I think my version would be more similar to what is already available in
this repo

shiftkey:
Not sure whether these other changes need to be incorporated. thoughts @snoyberg?

snoyberg:
Only .docker-sandbox is stack-specific, the others are not. Not all users of stack will need a .docker-sandbox directory, only those using the Docker integration.

manny-fp:
The current version of stack no longer creates a `.docker-sandbox` (only pre-release beta versions did).  That data has been moved inside `.stack-work`.

I also don't think `Dockerfile` or `target/` should be in the default ignore list for Haskell -- those are specific to the stack project itself, not generally to any project that uses stack.  I'm not sure about `TAGS` either (@chrisdone, looks like you added that one to stack's .gitignore, care to weight in?).

chrisdone:
TAGS comes from Emacs's location information about symbols for go-to definition support. Should not be tracked in Git.

shiftkey:
@athanclark it looks like there's some outstanding feedback here about the changes - and also a merge conflict. Are you interested in addressing those, or should I just close this out?

shiftkey:
Closing due to conflicts and inactivity. @athanclark please review the latest file and see if there are any changes that need to be incorporated.

PR #1566, opened 2015-06-25
Include UltraEdit backup files

adamzuckerman:
UltraEdit is a text editor that I use in place of Notepad. It will create backups of files before saving the changes to disk. The default extension is .BAK.

While this modification was made for UltraEdit, other text editors also perform automatic backups.

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -203,3 +203,6 @@ FakesAssemblies/
 
 # Visual Studio 6 workspace options file
 *.opt
+
+# UltraEdit backup files
+*.bak


arcresu:
This doesn't seem to be related to VisualStudio. If you'd like to make a new template for UltraEdit under Global/, then go ahead, but it would be great if you could consider whether this rule if widely useful enough to add. Thanks!

adamzuckerman:
While my specific need is for UltraEdit, there are plenty of other editors used by .NET developers who also have to add the line to ignore the .BAK files.

Am I missing something when it comes to .gitignore? Can you have multiple .gitignore files in one directory that automatically get merged together?

Creating a separate .gitignore for just one line seems philosophically wrong.

alexander-87:
There's only one `.gitignore` file per repository, but you can configure a global file on your system too.

As @arcresu hsa suggested, perhaps adding the `.bak` extension to a new `UltraEdit.gitignore` file under the [global](https://github.com/github/gitignore/tree/master/Global) folder might be better.

This already contains several other text-editor specific files, such as:

* [Notepad++](https://github.com/github/gitignore/blob/master/Global/NotepadPP.gitignore)
* [SlickEdit](https://github.com/github/gitignore/blob/master/Global/SlickEdit.gitignore)
* [SublimeText](https://github.com/github/gitignore/blob/master/Global/SublimeText.gitignore)
* [TextMate](https://github.com/github/gitignore/blob/master/Global/TextMate.gitignore)

Also, there are likely several other patterns for UltraEdit that could be added to the file.

For posterity, here's a few links with info on the `.gitignore` file:

* [**GitHub Help:** Ignoring files](https://help.github.com/articles/ignoring-files/)
* [Git - gitignore Documentation](http://git-scm.com/docs/gitignore)

tomstilljay1:
Backitup

shiftkey:
I think this addition could live under `Global\Windows.gitignore` as there are [many programs](https://en.wikipedia.org/wiki/Bak_file) which create files ending in this format.

PR #1568, opened 2015-06-26
Ignoring the AppCode ide files.#

terwer:
When you use Jetbrains AppCode ,there is a .idea forder,it has nothing to do with the iOS projects,I suggest ignore this forder while commit.


arcresu:

PR #1569, opened 2015-06-27
allow NuGet.exe, needed for package restore

petrhaus:
NuGet.exe should get included in the changeset, as it is needed for restoring NuGet packages.

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -146,6 +146,8 @@ publish/
 !**/packages/build/
 # Uncomment if necessary however generally it will be regenerated when needed
 #!**/packages/repositories.config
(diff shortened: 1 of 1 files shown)


shiftkey:
@csharpfritz do you have any opinions on this?

csharpfritz:
NuGet should not be included in source control as it is part of the local software development kit.  Visual Studio takes care of this for you with the [automatic package restore feature](http://docs.nuget.org/consume/package-restore) and NuGet will be available as part of the toolkit on other platforms that .NET compilation is enabled on in the near future.

shiftkey:
Thanks for the extra context @csharpfritz!

PR #1571, opened 2015-06-28
License these files under public domain

Bjakes1950:
Honestly, .gitignore files probably can't really be copyrighted in the first place. It probably falls under the _de minimus_ idea that they're too small and trivial.

But to avoid confusion, something like the Unlicense which dedicates these files to the public domain in a clear and explicit manner to avoid confusion.

This seems like the best approach for `.gitignore` files.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,19 +1,24 @@
-Copyright (c) 2014 GitHub, Inc.
+This is free and unencumbered software released into the public domain.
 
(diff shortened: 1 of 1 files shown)


arcresu:
Thanks, but it's not necessary to copy an existing issue (#752). I'd like to see that one get fixed as well.

PR #1573, opened 2015-06-28
Update Android.gitignore

wujingchao:


arcresu:
Thanks, but the IDE rules are already in templates like `Global/JetBrains.gitignore`

PR #1576, opened 2015-06-29
An OSX Gitignore

fdizile:
A Gitignore tailored to OSX


marijnvdwerf:
An OS X global ignore file is already included: https://github.com/github/gitignore/blob/master/Global/OSX.gitignore

arcresu:
Thanks, but that's right - we already have a template for this.

PR #1579, opened 2015-07-02
Ignore NCrunch temporary files.

jamiehumphries:
While [NCrunch](http://www.ncrunch.net/) is running it creates little temporary files in the pattern `nCrunchTemp_*`. These are automatically cleaned up, but a badly timed `git status` or `git add -A` can catch them momentarily.

Unfortunately I can't find any official documentation that discusses these files, but here are a few examples of public GitHub repos were people have added an ignore rule like this:

- https://github.com/TinyBlueRobots/NancyFs/blob/master/.gitignore#L63
- https://github.com/cloudfoundry-incubator/if_warden/blob/master/.gitignore#L68
- https://github.com/peterkeating/nancy-boilerplate/blob/master/.gitignore#L46

and some mention of the files on NCrunch forums:

- http://forum.ncrunch.net/yaf_postst1304_Don-t-start-builds.aspx
- http://forum.ncrunch.net/yaf_postst1093_NCrunch-cannot-build-project-that-VS-can.aspx

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -106,6 +106,7 @@ _TeamCity*
 # NCrunch
 _NCrunch_*
 .*crunch*.local.xml
+nCrunchTemp_*
 
 # MightyMoose
 *.mm.*


shiftkey:
ping @remcomulder - a couple of questions before I merge this:

 - can I get a :thumbsup: that this is intended behaviour and not a bug?
 - and if that's the case, what's the best way to get this reflected [in the docs](http://www.ncrunch.net/documentation/considerations-and-constraints_using-ncrunch-with-source-control)?

remcomulder:
NCrunch generates these temporary files to work around the limitations of MSBuild while loading projects.  Under normal operation, they only exist for a couple of seconds and are usually immediately removed.  Sometimes a critically timed crash or source control operation can lead to the files being noticed.  They should not be submitted to source control.  I've just now updated the NCrunch documentation accordingly.

jamiehumphries:
Cool, thanks @remcomulder! That docs link then: http://www.ncrunch.net/documentation/considerations-and-constraints_using-ncrunch-with-source-control

shiftkey:
@remcomulter fantastic, thanks for the speedy response!

PR #1580, opened 2015-07-03
django gitignore

mossplix:
Django gitignore


shiftkey:
I think this would be a better fit under the `Python` template:

https://github.com/github/gitignore/blob/master/Python.gitignore#L51-L52

shiftkey:
Can you please look over the existing rules in `Python.gitignore` and incorporate the Django-specific changes there? A new standalone isn't necessary here.

PR #1505, opened 2015-05-06
adding artifacts and project.lock.json for asp.net 5

sayedihashimi:
I've added the following to ```VisualStudio.gitignore``` to support ASP.NET 5

```
# ASP.NET 5
project.lock.json
artifacts/
```

```project.lock.json``` is created by default whenever a restore operation is executed. It should be ignored by default.

The ```artifacts``` folder will contain build output, and likely should never be checked in.

_Note: I'm on the team that works on Visual Studio for ASP.NET_

PR #1582, opened 2015-07-04
[PROPOSAL] annotated gitignore files

shiftkey:
This is a work-in-progress, based off the discussions in [this gist](https://gist.github.com/shiftkey/bf37c810b726f046bd22) around tackling some of the issues with maintaining a large suite of gitignore templates.

### The Problem

Over time, the context provided in gitignore templates is lost into the aether. By providing some structure to these files, hopefully we can improve this situation.

There's other parties who depend on this repository, so I'm happy to work through this in a methodical and collaborative way, being the guinea pig while we experiment with this.

### The Goals

First off, thanks to everyone who participated in that gist, in particular @joeblau and @arcresu - who have much more context on this issue than myself. I've taken that feedback and turned that into a first step down this path.

I want to enrich our existing templates with metadata and add some simple structure, without breaking compatibility (that is, it should remain consumable as a gitignore template) or introducing any new tooling.

I believe this will help on a few fronts:

 - **maintainers:** will have extra context available to them, when reviewing proposed changes
 - **contributors:** will be able to view context alongside template changes, without having to navigate a long history of PR discussions. Hopefully contributors will also be able to discuss and clarify this context over time as well, but this depends on things like adoption.
 - **users:** will have the context around a template available at consumption time, so they can understand why a template is configured as such a way, and be empowered to make changes because they understand _why_ a rule exists 

### The Proposal

This PR makes some changes to the `VisualStudio` template as a testbed for what data we need to capture. As I've been doing this, a bunch of concepts have crystalized for me, which I'll formalize in the `CONTRIBUTING` document alongside this (it also feels very much like a grammar syntax is the end result, but let's see how the discussion evolves).

I'm using the `VisualStudio` template as a starting point for a couple of reasons:

1. It's one that I'm very familiar with, so taking on the challenge to provide more context is more achievable than choosing a template (or stack) I'm not as familiar with.
2. I've had various discussions with people in the community about recent template changes (like #1505), so it'll be a good test case for how feasible this idea is.

I'll write up some further comments about this format in this PR.


shiftkey:
cc @MichaelMarner @bkeepers @M-Zuber @mikemcquaid who also commented on the gist

VisualStudio.gitignore
```diff
@@ -1,16 +1,18 @@
...
-# User-specific files (MonoDevelop/Xamarin Studio)
+#
+# new location for Visual Studio 2015
+.vs/
+#
+# specific to MonoDevelop and Xamarin Studio
 *.userprefs
 
-# Build results
+# build output folders generated by Visual Studio
```
shiftkey: This is the simplest structure I could come up with, and can be described as (my regex-fu is rusty, apologies in advance):

```
section: 
  context+ 
  rule+

context: comment+;
comment:  '#' ~[\r\n]* CRLF;
rule: ~[\r\n]* CRLF

CRLF
 : '\r'? '\n' 
 | '\r'
 ;
```

TL;DR:
 - each section should have at least one line of context
 - you can have multiple rules in a section

VisualStudio.gitignore
```diff
@@ -22,24 +24,37 @@ bld/
 [Bb]in/
 [Oo]bj/
 
-# Visual Studio 2015 cache/options directory
-.vs/
-
-# MSTest test Results
+# MSTest
+# a test framework from Microsoft
+# remove: if you are not using the MSTest runner to run your tests
```
shiftkey: So @arcresu mentioned "actions" in the gist as a way to call out when rules can be ignored or removed. There was a proposal for `enable-if` and `remove-if` to annotate this, but after wrestling with a couple of examples I found myself settling on this format.

- I experimented with `-when` here as well as `-if` but this lead to me writing the context in a strange way. So I dropped it altogether to just focus on the words associated with `remove`.
- I don't yet have an example of the dual to this yet (a rule that you should enable conditionally) but initial feels on this make me think it should be `include` rather than `enable` - probably just me wrestling with semantics.

VisualStudio.gitignore
```diff
@@ -22,24 +24,37 @@ bld/
...
 [Tt]est[Rr]esult*/
 [Bb]uild[Ll]og.*
 
-# NUNIT
+# NUnit
+# an open source test framework for .NET
+# site: http://www.nunit.org/
+# remove: if you are not using the NUnit test runner to run your tests
```
shiftkey: This is an example of gathering up rules related to a specific tool - . It feels a bit promotion-y, but also feels like a good way for users to explore the ecosystem better.

shiftkey: Also, this should be `url` to be consistent with other usages.

VisualStudio.gitignore
```diff
@@ -1,16 +1,18 @@
-## Ignore Visual Studio temporary files, build results, and
-## files generated by popular Visual Studio add-ons.
-
-# User-specific files
+# temporary files generated by Visual Studio specific to the local user
```
shiftkey: This is an example of gathering up rules into a group. I'll grammarize this later, but the TL;DR is:

 - context to explain the group
 - before each group of rules, you could specify a `url` or `remove` group
 - newline comment between sections for readability

VisualStudio.gitignore
```diff
@@ -1,16 +1,18 @@
...
+# contains profile information and other metadata
+# remove: if no longer targeting Visual Studio 2013 or earlier
 *.suo
 *.user
 *.userosscache
 *.sln.docstates
-
-# User-specific files (MonoDevelop/Xamarin Studio)
+#
+# new location for Visual Studio 2015
```
shiftkey: This is an example of a rule which most users don't currently need (this version of Visual Studio is in RC) but having it enabled currently is good for future-proofing things. As such, it doesn't have an `include` field defined.

VisualStudio.gitignore
```diff
@@ -22,24 +24,37 @@ bld/
...
 # DNX
+# summary: the environment for cross-platform .NET development
+# site: https://github.com/aspnet/dnx
+#
+# locks the package references for a project to specific versions
+# url: https://github.com/aspnet/Home/wiki/Lock-file
+# remove: if you would like to lock your dependencies to a specific version
+#         when committing and prevent package restore from selecting a
+#         more recent version
 project.lock.json
```
shiftkey: This change was submitted in #1505 but a [twitter discussion](https://twitter.com/adamralph/status/616854162735779840) yesterday about the change left things unclear. So here's my attempt at enriching the rule rather than continuing the saga of enabling or disabling this rule.

I've provided :

 - some comments (line 50) about what this file represents
 - a `url` to more information
 - a `remove` rule, which explains the situation when you should be disabling this rule

shiftkey:
cc @adamralph @fekberg @damianh @akoeplinger @filipw who were all bugging me on Twitter about this yesterday.

M-Zuber:
This looks very nice.
The `remove`/`include` sections is something I have wanted for a long time.
Overall it really changes the feel from a random list of file extensions to a document that is self expressing.
 It would be very interesting to see the grammar that develops and to see what tooling rises out of that.

mikemcquaid:
@shiftkey This is neat :+1: I think the key will be having a clear test which fails if it's not in the right format (and that people can run locally using just a `gitignore` checkout).

joeblau:
I'm currently in the process of re-writing gitignore.io's generator in golang.  Do you think people will want to see a documented output file or undocumented as default when someone types `gi visualstudio`?  I assume that the default uncommented lines in these templates are fine and would prefer and undocumented default, but I would like to get some feedback from others if we're going to move forward with this format.

shiftkey:
> Do you think people will want to see a documented output file or undocumented as default when someone types `gi visualstudio`?

I think if you're using a service like gitignore.io you're just interested in getting the contents, rather than the document. Perhaps a option like `--verbose` for those who want to get the annotations as well?

M-Zuber:
I would definitely want an option like `--verbose` as I use gitignore.io to get the template and then make my changes, and having the extra information would be very useful

shiftkey:
As part of getting some focus this year, I'm dropping this. Happy to help someone else if they want to pick it up.

adamralph:
If it's closed then it's unlikely anyone will ever pick it up...

shiftkey:
@adamralph ok then

fekberg:
@shiftkey Any reason for not merging the `VisualStudio` changes (separately) and keeping the proposal?

shiftkey:
@fekberg merge conflicts mostly

adamralph:
@shiftkey oh sorry, this is a PR, I thought was an issue. Another option would be to close this, but copy out the proposal into an issue.

shiftkey:
@adamralph issues have been disabled for this repo :cry:

adamralph:
:footprints: :gun:

mikemcquaid:
@shiftkey FWIW I think it's completely reasonable to close this out.

shiftkey:
If someone else in here wants to take some time out cleaning up the `VisualStudio.gitignore` template, it'd be :heart:. But as I said above I'm putting some of the big-picture stuff I have in mind on hold for the moment...

adamralph:
@shiftkey this raises an argument for enabling issues, since then you could open one and mark it up for grabs. Very few people are ever going to search closed pull requests for way to help out.

shiftkey:
> this raises an argument for enabling issues, since then you could open one and mark it up for grabs. 

I think this is a bigger problem that won't necessarily be solved by enabling issues on the repository, but I'll think about it. Meanwhile, the branch is still open for reference.

PR #1584, opened 2015-07-06
Ignore PyCharm working files

jedberg:
PyCharm is one of the most popular Python editors in use today, so I wanted to make sure we ignore it's working files, which keep track of things like which files you have open and window placement.


arcresu:
Thanks, but see `Global/JetBrains.gitignore` for this

PR #1586, opened 2015-07-08
update Visual studio gitignore for qt add-in

AlexiaChen:
this VisualStudio.gitignore have no ignore files for qt-vs-addin. such as moc_*.cpp qrc_*.cpp ui_*.h

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -204,3 +204,10 @@ FakesAssemblies/
 
 # Visual Studio 6 workspace options file
 *.opt
+
+#Qt Add-in for Visual Studio
+GeneratedFiles/
+*.moc
+moc_*.cpp
+qrc_*.cpp
+ui_*.h


shiftkey:
There's a couple of details I need before closing this out:

 - a link to the QT add-in for VS as an intro
 - a link to an explanation for what these entries represent, and why they should be ignored in source control

AlexiaChen:
1.Here is the wiki about QT add-in for VS :
https://wiki.qt.io/Visual_Studio_Add-in

2. 
moc_*.cpp---------http://www.qtcentre.org/threads/22644-what-is-a-moc-file
moc files are just generated files by Meta-Object-Compilor of Qt. they are similar to *.obj files for temporary during building time.

qrc_*.cpp-------------here is a link to the resource system of qt: http://doc.qt.io/qt-5/resources.html
qrc files are just  generated files by Qt-Resource-Compilor  for temporary during building time.

ui_*.h------------------here is a link to the User Interface Compiler (uic) http://doc.qt.io/qt-4.8/uic.html
ui_*.h  files are just generated files by UIC.exe for  temporary during building time. it represent the layout of User Interface in C++ 

GeneratedFiles/------------This is a folder generated by Qt VS add-in. moc,qrc,ui files are all under this folder

shiftkey:
> `GeneratedFiles/`
> This is a folder generated by Qt VS add-in. moc,qrc,ui files are all under this folder

So does this mean `GeneratedFiles/` will include all the other rules?

AlexiaChen:
Reply:"So does this mean GeneratedFiles/ will include all the other rules?"

Yes,it will include all the other files I mentioned above.

shiftkey:
Okay cool, so can we drop the other rules here and just leave `GeneratedFiles/` as the only addition?

I'm not sure how I feel about linking to an SO answer which explains this rather than some official documentation, but this is a good summary:

http://stackoverflow.com/a/18930469/1363815

@arcresu thoughts?

AlexiaChen:
Yes,SO answer which explains this rather than some official documentation. I agree. 

OKay,We can drop the other rules here and just leave GeneratedFiles/ as the only addition,I think it is a good idea.  :+1:

arcresu:
Official documentation is better because it somewhat constitutes a contract, but it's not realistic to expect any official docs about this stuff to exist for every tool. SO answers or even examples of file contents are a lot better than nothing when there's no obvious official source.

shiftkey:
> OKay,We can drop the other rules here and just leave GeneratedFiles/ as the only addition

I'd be happy with that.

shiftkey:

@AlexiaChen please let me know if you're interested in working together so this can get merged in.

AlexiaChen:
@shiftkey  Yeah, I am here, I was interested in working together as always.

AlexiaChen:
@shiftkey  Maybe, I need to repost a new PR for 

> OKay,We can drop the other rules here and just leave GeneratedFiles/ as the only addition

?

AlexiaChen:
this PR need to delete some extra rules, we just leave only one rule of "GeneratedFiles/"

shiftkey:
@AlexiaChen that sounds good!

AlexiaChen:
@shiftkey   soooo?  so I need to repost a new PR for this as I mentioned above?

shiftkey:
@AlexiaChen yes please!

AlexiaChen:
@shiftkey   OK. Thank you from my heart

PR #1589, opened 2015-07-12
Create Crystal.gitignore

keplersj:
Taken from the Crystal compiler's project init feature. Gitignore created by @flaviut.
### What is it?

[Crystal](http://crystal-lang.org/) is a language derived from Ruby, compiled for the LLVM. Because it's only syntactically similar to Ruby there are some differences which should be taken into account.
### What is being ignored?

Files created by the compiler in a normal session are being ignored here. `.deps` is where the compiler pulls in dependencies declared in a project's `Projectfile`. `libs` is similar is that it is external code downloaded later. `.crystal` is contains the LLVM bytecode created by the compiler at compilation. `doc` is well documentation. And `.deps.lock` is comparable to a `Gemfile.lock`.
### Why?

Crystal is a fairly new language which has been recently gathering a following. Thought it would be nice for new Crystal projects on GitHub projects to be created with a proper gitignore.

### DIFF
--- a/Crystal.gitignore
+++ b/Crystal.gitignore
@@ -1,0 +1,10 @@
+/.deps/
+/libs/
+/.crystal/
+/doc/
+
+
+# Libraries don't need dependency lock
+# Dependencies will be locked in application that uses them
+# Uncomment the line below if you are creating a library.
+#/.deps.lock


bcardiff:
:+1: LGTM for the time being. 
Things might change a little when dependency system get more mature towards 1.0.

Thanks @k2b6s9j 

NB: The submitted .gitignore is consistent with the built in `crystal init` command which uses https://github.com/manastech/crystal/blob/master/src/compiler/crystal/tools/init/template/gitignore.ecr

veelenga:
Since [0.8.0](https://github.com/manastech/crystal/releases/tag/0.8.0) release Crystal use shards to resolve dependencies and .gitignore need to be refreshed:

`.deps` -> `.shards`
`.deps.lock` -> `shard.lock`

keplersj:
This PR has gone outdated and superseded by #2309. However, considering that this is my oldest sitting Pull Request at 2 years and 5 days it seems unlikely that GitHub will take in new language gitignores. 😒  That being said I've been using @joeblau's [gitignore.io](https://gitignore.io) personally and it works great.

Vendetta26 (commented):
`re-usable`

PR #1590, opened 2015-07-13
Adding optional ignore proposal for wwwroot folder

JimiC:
There are use cases where you have gulp/grunt tasks that create the static files of a project from bower_components, images, scripts folders.

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -24,6 +24,8 @@ bld/
 
 # Visual Studio 2015 cache/options directory
 .vs/
+# Uncomment if you have tasks that create the project's static files in wwwroot
+#wwwroot/
 
 # MSTest test Results
 [Tt]est[Rr]esult*/


VisualStudio.gitignore
```diff
@@ -24,6 +24,8 @@ bld/
 
 # Visual Studio 2015 cache/options directory
 .vs/
+# Uncomment if you have tasks that create the project's static files in wwwroot
```
shiftkey: @sayedihashimi is this something we should be calling out? How do you feel about the words here?

JimiC:
As additional info here is what the default ASP.NET 5 template generates in project.json.
```  
"exclude": [
    "wwwroot",
    "node_modules",
    "bower_components"
  ],
```
Hovering over `exclude` says "Glob pattern to indicate the default files to be excluded from processing".
`node_modules` is included in gitignore. I know that `bower_components` was excluded for a reason (yeah I read the article). So this leaves us to decide the fate of `wwwroot`.

sayedihashimi:
I think there will be a good number of people excluding wwwroot so I'm OK with this.

shiftkey:
@sayedihashimi thanks! I'll take this in now.

PR #1592, opened 2015-07-15
Update visualstudio.gitignore

arindamnayak:
Added *.sql file to include for # SQL Server files section.

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -187,6 +187,7 @@ UpgradeLog*.htm
 # SQL Server files
 *.mdf
 *.ldf
+*.sql
 
 # Business Intelligence projects
 *.rdl.data


arindamnayak:
@arcresu , I have added one more file type for visual studio type projects, [This MSDN link] (https://msdn.microsoft.com/en-us/library/ms171969(v=vs.90).aspx) explains, how these files can be created in such projects. gitignore should ignore such file types. So I have changed this file and created pull request.

shiftkey:
@arindamnayak I'm not clear on where in that link it explains *why* the file should be ignored. Also, this impacts everyone who creates `*.sql` files, not just those generated by Visual Studio.

arindamnayak:
@shiftkey , my intention was to ignore those created by VS, since it will also ignore those SQL files , custom created ( now, I realize this), so we I taking the pull request back. We can close this. Sorry, for making confusion here.

shiftkey:
@arindamnayak no problem!

PR #1594, opened 2015-07-16
Update Android.gitignore

arindamnayak:
We should ignore iml and ideal files for android projects. Those are created by android studio. This link - http://stackoverflow.com/questions/16736856/what-should-be-in-my-gitignore-for-an-android-studio-project, says we should ignore them. Feel free to comment here.


arcresu:

PR #1595, opened 2015-07-16
Added OS Generated Files

anton-280:
.gitignore Operation System Garbage


shiftkey:
These are covered in the [Global/OSX.gitignore](https://github.com/github/gitignore/blob/master/Global/OSX.gitignore) and [Global/Windows.gitignore](https://github.com/github/gitignore/blob/master/Global/Windows.gitignore)...

PR #1598, opened 2015-07-22
Update CVS.gitignore

rdalkire:
I added and tested another leading asterisk to deal with the CVS directories further down the tree.

### DIFF
--- a/Global/CVS.gitignore
+++ b/Global/CVS.gitignore
@@ -1,4 +1,4 @@
 /CVS/*
-*/CVS/*
+**/CVS/*
 .cvsignore
 */.cvsignore


shiftkey:
Just so I'm clear:  we should match with *any* directory named CVS found in the repository?

rdalkire:
Well I needed to match that way, because I needed to avoid all those directories from CVS without actually deleting them.  I assume anyone else coming from the same use-case, (where CVS is part of work requirements but you also want to use git, without tracking the CVS metadata in git) would also need that.

As an aside: Since my posting, I have found that egit (built into Eclipse 4.4.2 Luna, and based on jgit) does not seem to work correctly for these CVS patterns, and indeed it has a [lot of other problems](https://bugs.eclipse.org/bugs/buglist.cgi?query_format=advanced&short_desc_type=allwordssubstr&short_desc=&classification=Technology&product=EGit&long_desc_type=allwordssubstr&long_desc=&bug_file_loc_type=allwordssubstr&bug_file_loc=&status_whiteboard_type=allwordssubstr&status_whiteboard=&keywords_type=allwords&keywords=&bug_status=NEW&bug_status=ASSIGNED&bug_status=REOPENED&emailtype1=exact&email1=&emailtype2=substring&email2=&bugidtype=include&bug_id=&votes=&chfieldfrom=&chfieldto=Now&chfieldvalue=&cmdtype=doit&order=Reuse+same+sort+as+last+time&field0-0-0=noop&type0-0-0=noop&value0-0-0=) as well.  Git Extensions works fine though, so I'm just using that instead.  Maybe a little OT, but maybe someone will stumble across this and find it useful after I spent so much time on it.

shiftkey:
@rdalkire thanks for the extra info!

PR #1599, opened 2015-07-22
Ignores for Visual Studio Lightswitch

sfhardman:
There are a few Lightswitch for VS2013 build output files/folders that should be excluded, but aren't covered by the existing VisualStudio.gitignore.  I haven't found any documentation on the subject, but when adding a Lightswitch project to TFS version control through Visual Studio these items are excluded automatically.


shiftkey:
@sfhardman thanks for submitting this. Are there links to some documentation around what these files represent, and why they can be ignored in version control?

shiftkey:
@sfhardman the reason why I ask is that it'd be nice to get some extra context in here for reference.

shiftkey:
cc @estiller for feedback

estiller:
Hi,
I ran some tests using the default Visual Studio .gitignore file, and it seems like adding these couple of lines solves the issue:
_Pvt_Extensions/
GeneratedArtifacts/

The "_Pvt_Extensions" folder contains files which are extracted from lightswitch packages at runtime, and "GeneratedArtifacts" contains, well, generated code files which are created at design/compile time. These two folders should not be checked into source control. It seems like all other files are already covered.

shiftkey:
@estiller :heart: thanks for taking time out to review

@sfhardman do you think we can condense these down into the two rules, and comment on what they represent?

estiller:
After some additional thought, I think it would be best to stick with @sfhardman commit as he suggested. It is more "secure", ignoring less files.

sfhardman:
Thanks for looking into this.  I think ModelManifest.xml and generated.parameters.xml are also needed

ModelManifest.xml is generated at build and holds checksums of the lsml files.  Visual Studio excludes it by default when adding to TFS (I can't find any official citation for this though)

generated.parameters.xml is generated when publishing a Lightswitch project.  Visual Studio also excludes it when adding to TFS (it contains the publishing connection strings, server names etc.).

estiller:
@sfhardman Yes, I agree. That's what I meant in my second comment (https://github.com/github/gitignore/pull/1599#issuecomment-123963248). I think your commit is spot-on.

terrajobst:
@bethmassi, you might want to provide some feedback here :-)

VisualStudio.gitignore
```diff
@@ -206,3 +206,12 @@ FakesAssemblies/
...
 # Visual Studio 6 workspace options file
 *.opt
+
+# Visual Studio Lightswitch build output
+**/*.HTMLClient/GeneratedArtifacts
+**/*.DesktopClient/GeneratedArtifacts
+**/*.DesktopClient/ModelManifest.xml
+**/*.Server/GeneratedArtifacts
+**/*.Server/ModelManifest.xml
+**/*.Server/generated.parameters.xml
```
philonordlund: I can't confirm that this file should be ignored, and the VS 2015 default .gitignore does not mention it. You may want to omit it for now.

VisualStudio.gitignore
```diff
@@ -206,3 +206,12 @@ FakesAssemblies/
 
 # Visual Studio 6 workspace options file
 *.opt
+
+# Visual Studio Lightswitch build output
```
philonordlund: "LightSwitch" (capitalization)

philonordlund:
Hi all, I'm not Beth but I'm on the LightSwitch team at Microsoft. I left a couple comments, but otherwise the changes look great to me. Thanks for contributing!

VisualStudio.gitignore
philonordlund: (specifically, the generated.parameters.xml file is the one I'd omit)

sfhardman:
Ok, I've updated the PR omitting generated.parameters.xml and correcting casing of LightSwitch.  Thanks all

shiftkey:
Thanks for all the feedback everyone!

PR #1600, opened 2015-07-23
update Magento 1.9.2.0

thiagorock22:
Magento 1.9.2.0 included the follow paths:

dev/
app/bootstrap.php
lib/Net

### DIFF
--- a/Magento.gitignore
+++ b/Magento.gitignore
@@ -87,7 +87,7 @@ lib/Cm/
 lib/Credis/
 lib/Magento/
 lib/Net/
-dev/
+dev/*
 LICENSE_AFL.txt
 LICENSE.html
 LICENSE.txt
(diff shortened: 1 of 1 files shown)


Magento.gitignore
```diff
@@ -85,6 +86,8 @@ lib/Zend/
 lib/Cm/
 lib/Credis/
 lib/Magento/
+lib/Net/
+dev/
```
mattjanssen: Why aren't we starting all git ignore lines with a leading slash? There's a good post on what happens without leading slashes.

In your commit, any directory named `dev` would be excluded. Even if it was `/in/another/directory/dev/` which is probably unintended.

martinvw:
Ignoring dev like this would not allow you to add your own tests, which would be a good practice.

shiftkey:
Superseded by #1746

PR #1601, opened 2015-07-24
Ignore Bower Components

deacons2016:
Added Ignore Bower Components

### DIFF
--- a/Node.gitignore
+++ b/Node.gitignore
@@ -25,3 +25,6 @@ build/Release
 # Dependency directory
 # https://docs.npmjs.com/misc/faq#should-i-check-my-node-modules-folder-into-git
 node_modules
+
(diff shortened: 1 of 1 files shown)


shiftkey:
@deacons2016 this isn't currently mergeable due to other PRs, but I've had a discussion with various other Bower users and they seem interested in making this the default.

I'd love to see a comment to this effect here so it's clear to others:

> If you aren’t authoring a package that is intended to be consumed by others (e.g., you’re building a web app), you should always check installed packages into source control.

Also, linking to the blog post discussing this in detail would be :heart: - https://addyosmani.com/blog/checking-in-front-end-dependencies/

shiftkey:
Closing in favour of #2157

PR #1603, opened 2015-07-26
Do not ignore imported assemblies

theoxuan:


shiftkey:
Not being familiar with Unity, is there a compelling reason to track binary assemblies by default? Any supporting links or discussions would be greatly appreciated...

theoxuan:
It's just a matter of directories. Binary assemblies under the directory of **Assets/** should not be ignored by default.

shiftkey:
@theoxuan any other links discussing this, which I can point other curious to?

Unity.gitignore
```diff
@@ -14,6 +14,9 @@
 *.pidb
 *.booproj
 
+# Do not ignore imported assemblies
+!*.dll
```
shiftkey: > Binary assemblies under the directory of **Assets/** should not be ignored by default.

So this rule could be made more specific?

theoxuan:
@shiftkey Hello here is the official documentation about Unity3D folder structures. 
[https://unity3d.com/learn/tutorials/modules/beginner/architecture/folders-in-version-control](url)

shiftkey:
@theoxuan that's great, thanks for sharing!

I'm still not quite clear on why we're interested in tracking these `*.dll` assemblies by default though...

theoxuan:
@shiftkey A huge bunch of third-party plugins are used in Unity projects. Some of them are source codes while others are assemblies. By default, all assemblies in Assets/ folder should not be ignored, but those in Libray/ and Temp/ folders.

JaySoyer:
Just wanna chime in and mention that @theoxuan is correct.  We need to track dll files under the Assets/ folder only.  However, in regards to the PR...I'm not sure we need to explicitly mention to not ignore dll files since it's already providing the correct behavior as written.

shiftkey:
@JaySoyer thanks for following up. 

@theoxuan as mentioned above, there doesn't seem to a rule here which is preventing assemblies from being tracked in a Git repository - do we need to do this?

PR #1604, opened 2015-07-27
Add KiCad ignore for .bck and .kicad_pcb-bak

Caerbannog:
These are two other filetypes that KiCad creates before overwriting something.

### DIFF
--- a/KiCAD.gitignore
+++ b/KiCAD.gitignore
@@ -3,6 +3,8 @@
 # Temporary files
 *.000
 *.bak
+*.bck
+*.kicad_pcb-bak
 
 # Netlist files (exported from Eeschema)
 *.net


shiftkey:
@Caerbannog do you have a link to what this files represent and why they can be (or should be) ignored in version control?

Caerbannog:
Each time a schematics / layout file is saved, the previous contents are moved instead of deleted. Here are the lines responsible for this behavior for each extension:
https://github.com/KiCad/kicad-source-mirror/blob/master/pcbnew/files.cpp#L636
https://github.com/KiCad/kicad-source-mirror/blob/master/eeschema/libedit.cpp#L416

shiftkey:
@Caerbannog thanks!

PR #1605, opened 2015-07-31
Update VisualStudio.gitignore

LostBalloon1:
As stated in the comment, added the ignore for the ".vs" folder which as of VS2015 contains the user preference files that should not be versioned.


shiftkey:
This has already been added - see #1388

PR #1608, opened 2015-08-03
ignore apple desktop service store file

Martin91:


arcresu:
Thanks, but this is in `Global/OSX.gitignore`

Martin91:
@arcresu got it, thanks for your reply!

PR #1609, opened 2015-08-04
Adding more files to ignore when using minitoc

habi:
As per http://tex.stackexchange.com/q/258515/828


shiftkey:
The link works for me. Thanks!

PR #1610, opened 2015-08-04
Create Elm.gitignore

jpatomic96:
.gitignore for Elm projects. 

Ignores installed packages (elm-package.json) and elm-stuff/, as these files will be generated once elm-package has been run for a .elm file. This is also beneficial where dependencies have changed, and an install package is no longer required.

elm-repl generated files have also been ignored. Most of the time, elm-repl manages to delete these files automatically, but there are scenarios where these files are not deleted on the closing of elm-repl

### DIFF
--- a/Elm.gitignore
+++ b/Elm.gitignore
@@ -1,0 +1,5 @@
+# elm-package generated files
+elm-package.json
+elm-stuff/
+# elm-repl generated files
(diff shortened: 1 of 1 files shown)


shiftkey:
I'm happy to take this in as a starting point. Thanks @jpatomic96!

PR #1611, opened 2015-08-05
Add .generated.swift to Swift gitignore

AndrewSB:
`.generated.swift` files are pre built swift files that often cause unnecessary merge conflicts.

### DIFF
--- a/Swift.gitignore
+++ b/Swift.gitignore
@@ -26,6 +26,7 @@ xcuserdata
 ## Obj-C/Swift specific
 *.hmap
 *.ipa
+*.generated.swift
 
 # CocoaPods
 #


shiftkey:
Hey @robrix can I get a "hell yeah!" here?

robrix:
@AndrewSB: What are they being built by? Is this something Xcode does, or is it specific to your toolchain?

AndrewSB:
@robrix believe its an Xcode thing, haven't seen them before, currently from [R.Swift](https://github.com/mac-cain13/R.swift)

robrix:
@AndrewSB: Per R.Swift’s [installation instructions](https://github.com/mac-cain13/R.swift#installation), this is peculiarity of how R.Swift works, and not a standard Xcode feature. Based on that, I think the ignore rule should be added to repos which use R.Swift, and not to this file.

Thanks very much for the PR, and for sharing that project!

AndrewSB:
Totally understand - I didn't do my research but now that I look at it I see that you're completely right, thanks for taking the time to look at my PR :smile:

robrix:
You’re welcome!

PR #1612, opened 2015-08-05
gitignore file for a basic elm lang project

usharf:
`elm-stuff` is a directory created by `elm-make` in the process of compiling .elm source files. It contains dependencies and artifacts created in the process and should not be checked-in, with the exception of `elm-stuff/exact-dependencies.json`, which the user might want to keep.

ref https://github.com/elm-lang/elm-make

> If you are creating an app or product, you want to keep your dependency ranges very specific. When you build, a file is generated called elm-stuff/exact-dependencies.json which lists all of the packages needed for your project and the exact versions you happen to be using right now. You may want to check this in to version control if you want the same exact thing every time.

### DIFF
--- a/Elm.gitignore
+++ b/Elm.gitignore
@@ -1,0 +1,2 @@
+elm-stuff/*
+!elm-stuff/exact-dependencies.json


Elm.gitignore
```diff
@@ -0,0 +1,2 @@
+elm-stuff/*
+!elm-stuff/exact-dependencies.json
```
shiftkey: I took in #1610 which doesn't have this rule - could you tell me about what this is for any why it should be in version control?

shiftkey: Apologies, I glossed over the quoted text which answers this.

@evancz do you feel this is a good default to have in the template for Elm, to help ensure new projects are setup correctly?

evancz: elm-stuff/ definitely is all build artifacts. I'm not certain about the exact-dependencies.json file, but I think anyone doing production quality Elm would want to keep track of that.

shiftkey:
@usharf unfortunately this isn't mergeable currently, and there are some answers to outstanding changes around this.

I need to clean up stale pull requests at some stage, and will close this out if I don't hear anything before Friday 26th February.

PR #1614, opened 2015-08-08
Create AtmelStudio.gitignore

rubberduck203:
Added a new .gitignore for [Atmel Studio 6](http://www.atmel.com/microsite/atmel_studio6/). 

Atmel Studio is for embedded development, so the build result files are mostly C.
It's built on top of an old version of Visual Studio, so they prefixed VS file extensions with `at`.
Keeping user settings out of the repo applies to everyone in my opinion. 
It also used a similar build output directory structure as Visual Studio, hence the exclusion of `Debug` and `Release` directories.

### DIFF
--- a/AtmelStudio.gitignore
+++ b/AtmelStudio.gitignore
@@ -1,0 +1,22 @@
+## Ignore Atmel Studio temporary files and build results
+# http://www.atmel.com/microsite/atmel_studio6/
+
+# Atmel Studio is powered by an older version of Visual Studio,
+# so most of the project and solution files are the same as VS files,
+# only prefixed by an `at`.
+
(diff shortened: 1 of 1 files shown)


shiftkey:
I'm going to close this out because I'm not sure how much value this would add, given it's roots in Visual Studio.

@rubberduck203 if you feel like this should still be a template listed here, please read the section about [specialized templates](https://github.com/github/gitignore#specialized-templates) and open a new pull request with more context.

rubberduck203:
@shiftkey I still thinks this adds value.
I'm happy to open a new PR with the file in the "community" location, but I don't know what more context you're looking for.

PR #1616, opened 2015-08-10
Added .gitignore template for the Racket langauge (www.racket-lang.org)

eu90h:
The added template is equivalent to the one created by the Racket language's command line utilities (see http://docs.racket-lang.org/pkg/cmdline.html#%28part._raco-pkg-new%29) 

It ignores DrRacket (an editor) autosave files, compiled Racket byte-code, and the OS X .DS_Store directory.

### DIFF
--- a/Racket.gitignore
+++ b/Racket.gitignore
@@ -1,0 +1,10 @@
+# DrRacket autosave files
+*~
+\#*
+.\#*
+
+# OS X specific directory information
+.DS_Store
+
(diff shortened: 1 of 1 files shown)


shiftkey:
@eu90h apologies for the delay in replying to this.

I had a lot of these templates requested, but I couldn't quite figure out the right place for them. I ended up settling on #2857, and you'll see the Racket template can now be found at `community/Racket.gitignore`.

PR #1617, opened 2015-08-10
Ipython Notebook checkpoints

manavm:
Checkpoints are automatically committed when adding Ipython notebooks to Github. This versioning is redundant given that git already performs code versioning to previous checkpoints. Hence, most projects prefer to include these files in the gitignore.

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -56,3 +56,6 @@ docs/_build/
 
 # PyBuilder
 target/
+
+#Ipython Notebook
+.ipynb_checkpoints


jakevdp:
I'd love to see this as well – IPython notebook is a commonly-used tool in the Python world, and I can't think of any situation that you'd want to version-control ``.ipynb_checkpoints``

arfon:
Thanks for this @manavm. It'll take a little longer before this shows up on GitHub.com as it requires a manual deploy.

PR #1619, opened 2015-08-12
Update Android.gitignore

axilleas:
Exclude `.idea/` dir and `*.iml` files from Android Studio.

Several sources indicating this is the right thing:
  * https://code.google.com/p/android/issues/detail?id=156708#c6
  * https://stackoverflow.com/questions/18415541/android-studio-import-project-with-gradle-changes-iml-files
  * https://github.com/owncloud/android/blob/master/.gitignore

---

The link to googlecode is an answer from a Google employee on Android:

> For Gradle-based Android projects, Studio uses build.gradle files as its source of truth and generates the iml files based on the project structure defined in build.gradle files. How they get regenerated is in IDEA's functionality.
>
>It is strongly recommended to not include iml in source control.

### DIFF
--- a/Android.gitignore
+++ b/Android.gitignore
@@ -12,9 +12,12 @@
 bin/
 gen/
 
-# Gradle files
+# Gradle and Android Studio files
 .gradle/
 build/
+*.iml
+.idea
+.navigation/
 
 # Local configuration file (sdk path, etc)
 local.properties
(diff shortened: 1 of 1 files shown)


arcresu:
Thanks, but see `Global/JetBrains.gitignore` for this

axilleas:
Ah sorry didn't see that!

PR #1621, opened 2015-08-16
Adding the Android Studio captures folder

gbaron:
The captures folder is used by Android studio to keep heap dumps and other captures files. These are usually not shared as part of the source file.

### DIFF
--- a/Android.gitignore
+++ b/Android.gitignore
@@ -27,3 +27,6 @@ proguard/
 
 # Android Studio Navigation editor temp files
 .navigation/
+
+# Android Studio captures folder
+captures/


shiftkey:
@gbaron do you have a link to some docs about this feature - so others can follow up on this if they have questions?

gbaron:
@shiftkey The only official mention at the moment is this one:
"Android Studio creates the heap snapshot file with the filename ... in the Captures tab."
From here: http://developer.android.com/tools/studio/index.html

shiftkey:
@gbaron ah well - one can dream. Thanks anyway.

PR #1624, opened 2015-08-22
Add Scheme.gitignore

wikibootup:
There was no file for Scheme ( Only LISP exists ),

so I added basic .gitignore for Scheme


shiftkey:
@wikibootup thanks for contributing this!

PR #1626, opened 2015-08-22
Removed unnecessary newline

damadmai:
2 times newline is for separation of Frameworks like in:
https://www.gitignore.io/api/vim,osx


shiftkey:
Could you tell me a bit about this change?

There's a specific rule here which needs the right set of characters to be valid:

https://github.com/github/gitignore/pull/1626/files#diff-235aa03516afdbe9fc516e7b1bec4395R6

I fear this change has regressed that, but I'd need to pull this down and poke at the specific file contents to confirm my suspicion.

damadmai:
In the original file https://github.com/github/gitignore/blob/master/Global/OSX.gitignore there is an issue with Line Endings: (There should be no CR LF)

osx_line_endings

I tried to fix that on GitHub via the online Editor, but it changed all LF to CR LF.

Now I made a commit, that should be fine.

damadmai:
Sorry, I just messed up the date of my commit, i made a new Pull Request:
https://github.com/github/gitignore/pull/1627
This one can be deleted.

PR #1627, opened 2015-08-24
Removed unnecessary CRLF newline

damadmai:
There is an CR LF which should'nt be there
osx_line_endings

2 times newline is for separation of Frameworks like in:
https://www.gitignore.io/api/vim,osx

### DIFF
--- a/Global/OSX.gitignore
+++ b/Global/OSX.gitignore
@@ -3,7 +3,7 @@
 .LSOverride
 
 # Icon must end with two \r
-Icon
+Icon
 
 # Thumbnails
 ._*


shiftkey:
That `\r` is there for a specific reason  - see the previous PR which touched this rule https://github.com/github/gitignore/pull/804 and the related discussion http://thread.gmane.org/gmane.comp.version-control.git/102266 for context.

> This is well and good, but poses a bit of a problem for ignoring files whose
name ends in \r. **In particular, Mac OS X's icon files are named Icon\r. Yes,
this is sick and annoying on the part of Apple.**

damadmai:
Oh, I'm sorry. Hmm, now i see that I overlooked the comment that statet this...
I would never have thought of something like this. But it seems that everything is possible.
Keep up the good work!

shiftkey:
> I would never have thought of something like this. But it seems that everything is possible.

It's okay, I'd been bitten by this before so had a bunch of context that others probably weren't aware of (through acts of suffering)...

LiangjunDai:
aaa

PR #1630, opened 2015-08-25
Visual Studio Team Explorer needs simple pattern to ignore packages

mgnslndh:
The current .gitignore for Visual Studio works fine from the command line but the Git support in Team Explorer does not recognize the same pattern as Git command line tools on Windows. 

A simpler pattern like "/packages" is required to ignore the folder.

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -150,6 +150,8 @@ publish/
 !**/packages/build/
 # Uncomment if necessary however generally it will be regenerated when needed
 #!**/packages/repositories.config
+# Visual Studio Team Explorer needs a simple pattern to recognize packages as ignored
+/packages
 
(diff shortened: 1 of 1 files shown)


niik:
ping @shana

shana:
@mgnslndh 

Sounds like this is a problem with Team Explorer, not with .gitignore. If Team Explorer is not parsing .gitignore correctly and showing you things you shouldn't be seeing, then patching .gitignore does nothing to actually fix the issue with Team Explorer. You should file an issue via the Report a Bug tool in VS (Help - About - Customer Feedback Options - More Options)

In any case, I'm looking at a project that has packages, and Team Explorer is properly parsing .gitignore and not showing anything in that folder. It doesn't look broken here. What is the broken behavior you're seeing?

PR #1631, opened 2015-08-25
Allow for ignoring other important files and save space

ahmedbaracat:
Ignore build, MacOSX and Windows files

### DIFF
--- a/Unity.gitignore
+++ b/Unity.gitignore
@@ -23,10 +23,3 @@ sysinfo.txt
 # Builds
 *.apk
 *.unitypackage
-
-# MacOSX-specific
-.DS_Store
-
-# Windows-specific
-thumbs.db
-
--- a/Unity.gitignore
+++ b/Unity.gitignore
@@ -19,3 +19,14 @@
 
 # Unity3D Generated File On Crash Reports
 sysinfo.txt
+
+# Builds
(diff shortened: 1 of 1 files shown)


shiftkey:
@ahmedbaracat rather than duplicating OS-specific rules for all projects, these live in https://github.com/github/gitignore/blob/master/Global/OSX.gitignore and https://github.com/github/gitignore/blob/master/Global/Windows.gitignore. 

The build-related changes seem relevant here - any supporting docs around what these files represent and why they can be excluded from version control would be greatly appreciated!

ahmedbaracat:
@shiftkey Thanks for pointing that out.

The main reason for excluding build-related changes is not to hit the file size limit which, I believe, is 100 MB. Normally, games developed with unity, especially 3D ones, have large build files and you can always build the project when you fork/clone/download it :)

**Hope this helps**
APK: "the package file format used to distribute and install application software and middleware onto Google's Android operating system" https://en.wikipedia.org/wiki/Android_application_package

Unity package: https://help.market.envato.com/hc/en-us/articles/202500634-Building-a-UnityPackage

shiftkey:
@ahmedbaracat those details are good enough for me - and the files are Unity-specific, not OS-specific as I first thought...

PR #1635, opened 2015-08-27
'Podfile.lock' in Objective-C gitignore

MinJunKweon:
* If you want to use [Cocoapods](http://cocoapods.org), add `Podfile.lock` in Objective-C gitignore file
   * *But It's comment now*

### DIFF
--- a/Objective-C.gitignore
+++ b/Objective-C.gitignore
@@ -34,6 +34,7 @@ xcuserdata
 # http://guides.cocoapods.org/using/using-cocoapods.html#should-i-check-the-pods-directory-into-source-control
 #
 #Pods/
+#Podfile.lock
 
 # Carthage
 #


shiftkey:
@MinJunKweon is this a recent change? Is there a link to some content explaining this addition (and when you should include it in your ignore file)?

Objective-C.gitignore
```diff
@@ -34,6 +34,7 @@ xcuserdata
 # http://guides.cocoapods.org/using/using-cocoapods.html#should-i-check-the-pods-directory-into-source-control
 #
 #Pods/
+#Podfile.lock
```
shiftkey: From [this link](http://guides.cocoapods.org/using/using-cocoapods.html#should-i-check-the-pods-directory-into-source-control) I can see the official advice is:

> Whether or not you check in the Pods directory, the **Podfile** and **Podfile.lock** should always be kept under version control.

shiftkey:
@MinJunKweon unfortunately this isn't mergeable currently, and this seems to go against the recommended behaviour. I'd love to confirm this, but it's been quiet :cry:

PR #1636, opened 2015-08-27
ignore file for Blender 3D (blender.org) scripts and models

eugenschindler:
Rudimentary ignore definition for Blender 3D (blender.org) scripting and modeling.


shiftkey:
@eugenschindler thanks for the contribution! Unfortunately this seems like a minor template and doesn't feel like a good fit for this repository.

PR #1637, opened 2015-08-28
Added IntelliJ generated files

esoxjem:
Not essential needed to be checked in for version control

### DIFF
--- a/Android.gitignore
+++ b/Android.gitignore
@@ -30,3 +30,7 @@ proguard/
 
 # Android Studio captures folder
 captures/
+
+# Idea
+.idea/
+*.iml


shiftkey:
We track the IntelliJ rules [here](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore#L6) rather than applying them per-project type..

Android.gitignore
```diff
@@ -30,3 +30,7 @@ proguard/
 
 # Android Studio captures folder
 captures/
+
+# Idea
+.idea/
+*.iml
```
shiftkey: This rule is new to me - should we add that here? https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore

esoxjem:
@shiftkey Since most Android Developers would be using Android Studio and not an Android project in IntelliJ, I think these generated files should be added here.

esoxjem:
Also, these are regenerated by Android Studio on various occasions.

shiftkey:
> Also, these are regenerated by Android Studio on various occasions.

I'm not quite clear on this - are we talking about IntelliJ or Android Studio creating these file extensions?

> Since most Android Developers would be using Android Studio and not an Android project in IntelliJ, I think these generated files should be added here.

If there are files related to Android Studio (which is reasonable to consider a default here for Android projects) which should be ignored, I'm :thumbsup: with including them here. But we track all existing IntelliJ rules elsewhere.

esoxjem:
@shiftkey My bad. I should have been more clear. These files are related Android Studio.

drothmaler:
As you can read at [the Android Studio Download Page](https://developer.android.com/sdk/index.html), Android Studio is 
> "Built on IntelliJ IDEA Community Edition, the popular Java IDE by JetBrains."

So the rules should stay in JetBrains.gitignore

shiftkey:
@drothmaler we have a precedent for this in other files, e.g:

https://github.com/github/gitignore/blob/master/Swift.gitignore#L3

@esoxjem is that an adequate compromise here, given the relationship between IntelliJ and Android Studio?

esoxjem:
@drothmaler I'm not quite convinced. Going by that approach, we should split the entire `Android.gitignore` and move to `JetBrains.gitignore`, `Gradle.gitignore` and only keep Android specicifv
@shiftkey Android devs would have to copy both `.gitignore` files or figure out what goes where. Since there's already an `Android.gitignore`, I think things should stay here.

drothmaler:
@esoxjem Yes I think so... :-) The Android.gitignore currently is a real mess. I like the modular way (nearly) all other gitignore files are organized. And as far as I know, it is still possible to use Eclipse and Maven for Android development (instead of Gradle and Android Studio) 

@shiftkey  I don't think this precedent is a good one... Now there are redundancies distributed over 3 different files (Swift, Objective-C and Xcode). I also don't like that... But thats just a personal opinion.

shiftkey:
> Now there are redundancies distributed over 3 different files (Swift, Objective-C and Xcode). I also don't like that... But thats just a personal opinion.

There's lots of history and backstory here, and some of these issues/constraints are specific to the Xcode ecosystem. I'm not sure of a way out of this particular situation, given the relationship between IntelliJ and Android Studio, but let me think more on this...

esoxjem:
Have an `AndroidStudio.gitignore`?

Soulcyber:
This should be added to the gitignore, for me it worked only when i added all 3 those lines

Soulcyber:
@shiftkey If you create a new project with Android Studio 2.0 Preview 2, the .gitignore generated by the system is the following:
```
*.iml
.gradle
/local.properties
/.idea/workspace.xml
/.idea/libraries
.DS_Store
/build
/captures
```
so we should add at least those lines to the gitignore

esoxjem:
I believe @shiftkey's point is that these are already mentioned in JetBrains.gitignore as it covers all IDEs by JetBrains. The purpose of Android.gitignore is to cover files generated by the Android SDK, like dex files, APKs, etc.

shiftkey:
@Soulcyber

> The final gitignore should be a merge of the suggested gitignore generated by the system and this one you already have in the repo

We (the maintainers of this repo) don't have the tooling - or the experience with all the ecosystems combinations that come up - to achieve this, let alone the time to support this (check out how many PRs are needing review). 

For now, our solution is to organize things logically and and ask consumers to craft their gitignore file how they want. If that's still unsatisfactory I'd recommend having a look at https://www.gitignore.io/ which use this source but also let you craft your template quickly...

shiftkey:
@esoxjem @Soulcyber can we discuss the `*.iml` extension in a new PR? I'd love to understand where it's come from and where it should reside...

PR #1638, opened 2015-08-29
Added OS X Specific .DS_Store ignore

duliodenis:
With the majority of ObjC and Swift being done on Mac OS X I have added an ignore for the OS X Specific .DS_Store file in order to make git add . commands easier and less prone to inclusion of this file.

### DIFF
--- a/Objective-C.gitignore
+++ b/Objective-C.gitignore
@@ -27,6 +27,9 @@ xcuserdata
 *.hmap
 *.ipa
 
+## OS X specific
+*.DS_Store
+
 # CocoaPods
 #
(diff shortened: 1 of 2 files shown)


arcresu:
Thanks, but this is why we have `Global/OSX.gitignore`

duliodenis:
Thanks @arcresu - as Swift and especially ObjC are primarily OS X languages I thought this convenience in the suggested template would benefit the majority of github users who have not set-up their OS specific templates.

arcresu:
These templates are structured so that the global rules go under `Global/` and end up in your global gitignore file; we don't want to replicate them in every template. Thanks anyway!

PR #1643, opened 2015-08-31
Create SAS.gitignore

kfinlay:
This excludes binary data formats for SAS. Documentation is here: http://support.sas.com/documentation/cdl/en/hostwin/63047/HTML/default/viewer.htm#n0sk6o15955yoen19n9ghdziqw1u.htm .

### DIFF
--- a/SAS.gitignore
+++ b/SAS.gitignore
@@ -1,0 +1,14 @@
+# Binary data formats
+.sas7baud
+.sas7bdat
+.sas7bvew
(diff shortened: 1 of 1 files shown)


shiftkey:

@kfinlay if you feel like this should still be a template listed here, please read the section about [specialized templates](https://github.com/github/gitignore#specialized-templates) and open a new pull request that follows the new pattern.

PR #1644, opened 2015-08-31
Create Stata.gitignore

kfinlay:
Stata is a statistical software package: http://www.stata.com . This gitignore exclude the binary data format: http://www.stata.com/manuals14/pfileformatsdta.pdf .


shiftkey:
Closing in favour of #1524 which was merged.

PR #1646, opened 2015-09-02
[Typo3] assorted changes

josefglatz:


shiftkey:
@jousch do you have any extra details or links to help explain these changes?

M-arcus:
I'll try to explain some of the changes:

**Commit "[TYPO3] Remove AdditionalConfiguration.php"**
AdditionalConfiguration.php is used for the config-changes that differ from environment to environment. Instead of editing LocalConfiguration.php you can simply switch out the AdditionalConfiguration.php while keeping the main config intact. Quite handy.

**[TYPO3] Add temporary files to .gitignore**
These files enable access to the install-wizard and the install-tool. They should never be comitted, because it could allow access to external users, that have no access to the file-system.

shiftkey:
@M-arcus thanks!

PR #1648, opened 2015-09-02
Reports created by Ranorex test executions should not be tracked.

JohnWashburn:
Ranroex is a large plugin to Visual Studio for GUI test automation, so the usual VS ignores (.bin, .obj, etc.) should ignored with some additional ignores which are specific to a Ranorex Solution.

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -215,3 +215,15 @@ FakesAssemblies/
 **/*.Server/GeneratedArtifacts
 **/*.Server/ModelManifest.xml
 _Pvt_Extensions
+
+# Execution reports created Ranorex
+*.nlog*
+*.rxlog*
+*.rxlog
+*_trace*
+*/*.rxuser
+*.rxuser
+*.rximg
+Report
+Reports
+*/Reports/*


VisualStudio.gitignore
```diff
@@ -215,3 +215,15 @@ FakesAssemblies/
...
+
+# Execution reports created Ranorex
+*.nlog*
+*.rxlog*
+*.rxlog
+*_trace*
+*/*.rxuser
+*.rxuser
+*.rximg
+Report
```
shiftkey: I'm hesitant to include rules which cover general words like this, as it can interfere with others who may inadvertently miss this rule when committing changes.

shiftkey:
Do you have a link to any documentation around these rules and what they represent?

VisualStudio.gitignore
shiftkey: In some other scenarios, rules can be committed but then commented out, so users can opt-in to it, e.g.

```
# Uncomment if necessary however generally it will be regenerated when needed
 #!**/packages/repositories.config
```

This helps to explain when specific rules might be necessary.

shiftkey:

@JohnWashburn please let me know if you're interested in working together so this can get merged in.

PR #1651, opened 2015-09-06
Added NCrunch v2 files to VisualStudio template

dustinmoris:


Nellemandela:
This PR should be dismissed / Closed

https://www.ncrunch.net/documentation/considerations-and-constraints_using-ncrunch-with-source-control

>It's recommended that the last two files listed in the above table (.v2.ncrunchsolution and .v2.ncrunchproject) be stored inside a source control system to be shared with team members. These files contain settings that are critical to being able to build projects and run tests, and as such they will be needed by anyone seeking to use NCrunch with the solution. Because these two files are named according to the NCrunch version, they can exist side-by-side with their equivalents for other NCrunch versions. This means it is possible for a solution to be shared by team members using different versions of NCrunch.

>The remaining files should be placed on an exclusion list and left out of the source control system. These files contain user preference information and cached data that is not ideal for sharing within a team.



It is very common to have multiple projects and test that you dont want ncrunch to execute automatically.

dustinmoris:
Ops that's right! I blindly assumed these files have been missed :/
Thank you for pointing to the official documentation!

PR #1652, opened 2015-09-09
Adding GPG gitignore

huertanix:
GPG private keys should not be checked into source control for the same reasons passwords should not be. This gitignore excludes them from being accidentally checked in.

Project: https://gnupg.org/
Documentation: https://gnupg.org/documentation/index.html


shiftkey:
:thumbsup:

PR #1656, opened 2015-09-11
Add rule for MS PowerPoint temporary files

paolofulgoni:
These files are created by PowerPoint when a file is opened. It's a temporary lock file.
The rule is similar to those for Word/Excel that are already present.

Unfortunately I could not find any official MS page which describe this file, but I've just verified that it is created on my pc using Office 2010.


shiftkey:
The change seems specific enough to not cause issues for others. Thanks!

PR #1658, opened 2015-09-12
Add a standard file for jekyll project

jeblister:
A useful .gitignore file for your standard project jekyll.

### DIFF
--- a/Jekyll.gitignore
+++ b/Jekyll.gitignore
@@ -1,3 +1,44 @@
+# Ignore docs files
+.ruby-version
+
+# Numerous always-ignore extensions
+*.diff
+*.err
+*.orig
+*.log
+*.rej
+*.swo
+*.swp
+*.zip
+*.vi
+*~
+
+# OS or Editor folders
+.DS_Store
+._*
+Thumbs.db
+.cache
+.project
+.settings
+.tmproj
+*.esproj
+nbproject
+*.sublime-project
+*.sublime-workspace
+.idea
+
+# Komodo
+*.komodoproject
+.komodotools
+
+# grunt-html-validation
+validation-status.json
+validation-report.json
+
+# Folders to ignore
+node_modules
+bower_components
 _site/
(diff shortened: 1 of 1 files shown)


Jekyll.gitignore
```diff
@@ -1,3 +1,44 @@
...
+*.orig
+*.log
+*.rej
+*.swo
+*.swp
+*.zip
+*.vi
+*~
+
+# OS or Editor folders
```
shiftkey: The OS-specific files should be covered by the `Global/OSX.gitignore` and `Global/Windows.gitignore` or other editor-specific template.

shiftkey:
cc @benbalter for guidance about two things:

 - do you think this should be a standalone template? or should it be folded up into somewhere else?
 - how the rules specified here match up with what you think a default gitignore should contain for a Jekyll project

benbalter:
I'm not 100% what's being proposed here. It looks like [there's already a Jekyll `.gitignore` template](https://github.com/github/gitignore/blob/master/Jekyll.gitignore), which matches pretty well with I'd expect from a Jekyll project. This pull request seems to add a lot of unrelated cruft/personal preference (e.g., `*.zip`) that aren't necessarily specific to Jekyll.

@jeblister Given that there's already a Jekyll template, albeit conservative, mind sharing a bit more about your motivation for opening up the pull request?

shiftkey:
@benbalter oh nice spot, I thought the PR would have told me *something* about there being a template existing already. My apologies for not noticing that one.

shiftkey:
@jeblister so I think I agree with what @benbalter raised here:

 - these should be additions to the existing Jekyll template in the repository - I'm not sure how that came to be.
 - lots of the general rules here live in other templates (we'd love to keep this file containing *just* the Jekyll-specific rules).

shiftkey:

@jeblister please let me know if you're interested in working together so this can get merged in.

PR #1660, opened 2015-09-15
JetBrains IDEs: add shelf (VCS -> Shelve changes... menu)

quadroid:


shiftkey:
As it's constrained to `.idea` I'm happy to take this in as-is. Thanks!

PR #1661, opened 2015-09-15
Update Umbraco.gitignore

JimBobSquarePants:
Add ImageProcessor cached files and nuget package updates


shiftkey:
I do like the reorganization of core/non-core stuff, but the headers seem excessive. Could we use whitespace to make it a bit cleaner? (see how the [VisualStudio](https://github.com/github/gitignore/blob/master/VisualStudio.gitignore) template organizes things)

cc @shazwazza @aaronpowell for any objections to the new changes

aaronpowell:
Courier is non-core, same with image processor, so it seems odd having them.

aaronpowell:
Not familiar with the aspnet_client folder either

Shazwazza:
Hrm I've never really looked through this file properly ;) certainly a lot of things to clean up, I'll reply inline once I'm in the office today

shiftkey:
@aaronpowell @Shazwazza :heart:

JimBobSquarePants:
ImageProcessor is included with core so it should be there. Can easily clean up the commit to remove the headers.

Shazwazza:
@JimBobSquarePants I'll just add some inline comments on your PR, would be fantastic if you can make all changes so we can have a shiny template :)

Umbraco.gitignore
```diff
@@ -1,12 +1,21 @@
 # Note: VisualStudio gitignore rules may also be relevant
 
+# =========================
+# Umbraco
+# =========================
 # Ignore unimportant folders generated by Umbraco
 **/App_Data/ClientDependency/
```
Shazwazza: Both the **/App_Data/ClientDependency/ and the **/App_Data/ExamineIndexes/ can be removed because that is not actually where those files are stored.

Umbraco.gitignore
```diff
@@ -1,12 +1,21 @@
...
 **/App_Data/ExamineIndexes/
 **/App_Data/Logs/
 **/App_Data/[Pp]review/
 **/App_Data/TEMP/
-Cached/
+**/App_Data/NuGetBackup/
+**/App_Data/FeedCache/
```
Shazwazza: Not sure what this is from ?

Umbraco.gitignore
```diff
@@ -1,12 +1,21 @@
...
+**/App_Data/courier/revisions/
+**/App_Data/courier/cache/
+**/App_Data/umbraco.licensing.log.txt
+**/App_Data/*.udt
```
Shazwazza: Not sure what this is from ?

Umbraco.gitignore
```diff
@@ -1,12 +1,21 @@
...
+**/aspnet_client/
```
Shazwazza: This can be removed. I've seen it generated from very old versions of Umbraco + Visual Studio. It's not part of the core and i haven't seen it for years so shouldn't be in this template.

JimBobSquarePants:
@Shazwazza Ace. I'll Get it updated today. :smile:

shiftkey:
*friendly bump*

JimBobSquarePants:
Sorry for the delay folks, 

Having done a review of several installations I've trimmed the changes back, adding now only the Nuget backup and ImageProcessor folders. I've also removed a few folders (Examine, Client Dependency) since they are no longer in use.

shiftkey:
Thanks!

shiftkey:
@Phyxion `ClientDependency` was mentioned here as not being the correct location https://github.com/github/gitignore/pull/1661#discussion_r41150501 and Courier is considered non-core.

Phyxion:
Thanks :+1:

PR #1662, opened 2015-09-16
Ignore npm debug log file

ghost:


shiftkey:
This should be covered by line 4

https://github.com/github/gitignore/blob/master/Node.gitignore#L4

PR #1663, opened 2015-09-18
Add gitignore file for bazel generated files

durandj:
[Bazel](http://bazel.io) is Google's new build system for compiled
languages. At present there is no documentation on what files are
generated that need to be ignored but these are the ones that I've
noticed. A link to the documentation can be found
[here](http://bazel.io/docs/build-encyclopedia.html). Bazel is a really
awesome tool and it thankfully does not generate my extra files. It is
also incredibly easy to get started with so I imagine that it will
become pretty popular in the near future.

### DIFF
--- a/Bazel.gitignore
+++ b/Bazel.gitignore
@@ -1,0 +1,5 @@
+bazel-bin
+bazel-genfiles
+bazel-out
+bazel-testlogs
+


shiftkey:
If this is a tool rather than a language, should it live under the `Global` namespace?

bramp:
Bazel is a build system, like Maven, Gradle and CMake. Those three all live in the root (not Global). If a project uses Bazel (for example https://github.com/mbrukman/autogen/), then all users of that project will need the gitignore in the project, not installed globally. So I think as is this patch is good.

shiftkey:
@durandj apologies for the delay in replying to this.

I had a lot of these templates requested, but I couldn't quite figure out the right place for them. I ended up settling on #2857, and the Bazel template now lives at `community/Bazel.gitignore`.

Feel free to add any further changes there, and thanks again for the contribution!

PR #1664, opened 2015-09-20
Added .htaccess to concrete5 gitignore

littleibex:
Updated concrete5 gitignore to include htaccess

### DIFF
--- a/Concrete5.gitignore
+++ b/Concrete5.gitignore
@@ -1,3 +1,4 @@
 config/site.php
 files/cache/*
 files/tmp/*
+.htaccess


shiftkey:
@littleibex can you tell me about why this file should be excluded? Is there a specific tool that's creating this?

littleibex:
It's created  when you enable pretty URLs in concrete5. The htaccess file on the server can have a lot of settings in htaccess which I believe you should not be overwritten by the local repository.

shiftkey:
> The htaccess file on the server can have a lot of settings in htaccess which I believe you should not be overwritten by the local repository.

To make sure I understand: typically the `.htaccess` file on the server is not in version control, and you will want whatever's there rather than whatever's created locally?

littleibex:
.htaccess file is not there initially in the project. However, after a particular setting from the CMS is enabled, the .htaccess file is created in the root of the project which then becomes a part of the repository. It's not good to have .htaccess in version control since, more than the project, it's associated to the server or machine where it is hosted.

shiftkey:
@littleibex apologies for the delay - this all seems reasonable. Thanks!

PR #1666, opened 2015-09-22
VisualStudio Template - Added ignore for Paket dependency manager

Nellemandela:
Added ignore as recomended in http://fsprojects.github.io/Paket/getting-started.html


shiftkey:
That link explains it all nicely. Thanks!

PR #1668, opened 2015-09-23
Use https for all Help URLs

jdennes:
Use https for all Help URLs and avoid unnecessary redirects from old Help URLs.


arcresu:
Thanks!

PR #1672, opened 2015-09-25
Rework laravel file to bring it up to date

e4rthdog:
- **Remove vendor directory**. Best practice seems to be that you dont include vendor directory and use composer install to install your preferred libraries (from composer.lock), to your production environment.
- **Add homestead related files**
- **Add node related files**

### DIFF
--- a/Laravel.gitignore
+++ b/Laravel.gitignore
@@ -1,3 +1,15 @@
+#Exclude vendor libraries.
+#Including composer.lock and install using composer install.
+#Comment if you want to include all libraries
+/vendor
+
+#Node related
+/node_modules
+
+#Homestead related
+Homestead.yaml
+Homestead.json
+
 /bootstrap/compiled.php
 .env.*.php
 .env.php


Laravel.gitignore
```diff
@@ -1,3 +1,15 @@
+#Exclude vendor libraries.
+#Including composer.lock and install using composer install.
+#Comment if you want to include all libraries
+/vendor
+
+#Node related
+/node_modules
+
+#Homestead related
+Homestead.yaml
```
shiftkey: Can you tell me a bit about why these files can be safely ignored?

I can't see much details in the docs http://laravel.com/docs/5.1/homestead

aziflaj:
@shiftkey as far as I know (after using Homestead), there are no homestead-related files in any Laravel project. Usually, Homestead is used outside of the project directory

Laravel.gitignore
e4rthdog: Since we commit composer.loc we dont need vendor directory. It will be created via a composer install.

homestead.yaml its created when you npm install , but i havent verified this myself.

shiftkey:
@e4rthdog apologies for not coming back to this sooner. It looks like there's some merge conflicts here - are you interested in resolving those so we can merge this in, or should I close this out?

The comments by @aziflaj about Homestead are also interesting - not sure if you agree or disagree with them.

shiftkey:
Closing due to inactivity and conflicts. @e4rthdog please have a look at the latest template and see if any of these changes still need to be incorporated.

PR #1674, opened 2015-09-25
Don't include gradle.properties in version control

aziflaj:
Based on [Android Best Practices](https://github.com/futurice/android-best-practices) by Futurice, `gradle.properties` should be used to store sensitive data, like keys and passwords.

### DIFF
--- a/Android.gitignore
+++ b/Android.gitignore
@@ -15,6 +15,7 @@ gen/
 # Gradle files
 .gradle/
 build/
+gradle.properties
 
 # Local configuration file (sdk path, etc)
 local.properties


friederbluemle:
The "Android Best Practices" by Futurice contain a lot of useful tips, but some are a bit too opinionated.

It is true that `gradle.properties` *can* be used to store sensitive data. However, it is used by some projects to store version information or jvm options - therefore it should be committed to source control in most cases.

An example that uses `gradle.properties` for jvm options can be seen here: https://android.googlesource.com/platform/tools/build/+/master

shiftkey:
Perhaps this should be included as a commented-out example (with a note to explain why you should care).

friederbluemle:
The default generated `gradle.properties` file is titled `# Project-wide Gradle settings` - it does not mention anything about ignoring this file, and it is present in most Android projects. The [official Gradle Build Environment documentation](https://docs.gradle.org/current/userguide/build_environment.html) also does not recommend to ignore it. Under normal circumstances this file *should* be committed to source control, so I don't think there has to be a commented-out line in `.gitignore`. In the occasional cases where people do use it to store passwords etc, it is not that big of a deal to manually add it to their `.gitignore`.

There are many other ways of passing sensitive information to the build system: Environment variables, command line Gradle properties (`-P` switch), a `gradle.properties` file in the home directory, or the `local.properties` file (which is already ignored).

shiftkey:
Hey @aziflaj, did you have a chance to look at @friederbluemle's feedback about the importance of the `gradle.properties` file?

shiftkey:
Closing due to inactivity.

friederbluemle:
Thanks @shiftkey.

Another example why `gradle.properties` is important and should be committed to source control: With the latest version of the Android Gradle plugin, the task `transformClassesWithDexForRelease` will output a **warning** if the heap space of the Gradle daemon does not have a sufficient size, causing dex to run as a separate process.

```
Running dex as a separate process.

To run dex in process, the Gradle daemon needs a larger heap.
It currently has 1024 MB.
For faster builds, increase the maximum heap size for the Gradle daemon to at least 1536 MB.
To do this set org.gradle.jvmargs=-Xmx1536M in the project gradle.properties.
For more information see https://docs.gradle.org/current/userguide/build_environment.html
```

What I've been doing for projects with sensitive information is to simply create a `release.properties` file, and exclude that in `.gitignore`.

PR #1675, opened 2015-09-25
Ignore vendor directory

aziflaj:
The `vendor` directory is where the app dependencies are stored, after running `composer install` or `composer update`. It shouldn't be added to version control, since it is enough to have the `composer.json` file to resolve dependencies


shiftkey:
Closing in favour of #1672

PR #1678, opened 2015-09-28
Add Otto

justincampbell:
https://ottoproject.io


shiftkey:
This seems a bit trivial to incorporate here (and we don't have a Vagrant template either).

Could you tell me a bit about why you think this should live here?

justincampbell:
There is a Vagrant gitignore in this repo https://github.com/github/gitignore/blob/master/Global/Vagrant.gitignore

This seemed like the correct place to put gitignore conventions for a popular project.

shiftkey:
Ugh, my apologies - I checked the root of the repo and thought that was it.

So do you think this should replace `Global/Vagrant.gitignore`?

justincampbell:
No, Otto is a separate project, which happens to use Vagrant under the hood (in addition to Terraform, Packer, and Consul).

shiftkey:
Okay, I'll take this in and then leave it up to others to refine it. Thanks!

justincampbell:
Thanks!

PR #1679, opened 2015-09-30
add fabric.properties to JetBrains

victorhaggqvist:
Since that is what it should be called now a days according to https://dev.twitter.com/fabric/android/integrating


shiftkey:
Thanks!

PR #1680, opened 2015-10-01
Update VisualStudio.gitignore

reddyjannavarapu:
added for sloution files


alexgwolff:
But without these files visual studio can't open solution. 
these files are necessary

shiftkey:
@reddyjannavarapu `.csproj` and `.sln` files are essential for many different types of VS projects - can you tell me a bit about why you think they should be ignored?

drothmaler:
-1

dariusdamalakas:
Please can someone close this PR, both file types need to be in source control

shiftkey:
@dariusdamalakas I'm not convinced these should be included by default - Visual Studio needs these to manage the solutions and projects a user can create.

dariusdamalakas:
@shiftkey, well that's exactly what i've said :+1:  Thanks for closing this.

PR #1682, opened 2015-10-01
Created gitignore file for the QuartusII IDE from Altera.

espenkn:
This is ignore file is for the Quartus2 IDE from Altera. 

links to IDE SW: https://dl.altera.com/?edition=web

Why you’re making a change
I created this file since i could not find an other file for quartus.

Scope of your change:
Just sharing.

All work is based on information from
links to documentation: http://www.alterawiki.com/wiki/Version_Control (01.10.2015)

Not thoroughly tested!


espenkn:
I did not notice https://github.com/github/gitignore/pull/1193 until now... Guess I better remove this.

espenkn:
I'm reopening this pull request in order to see if there are other users that thinks this should be in the repo.

Global/AlteraQuartusII.gitignore
```diff
@@ -0,0 +1,60 @@
+##list taken from http://www.alterawiki.com/wiki/Version_Control (01.10.2015)
+
+######### Quartus II source files
+
+# project files:
+### project_name.qpf Quartus II project file
```
shiftkey: Are we able to bring these comments closer to the source? I think this is a neat list to curate, but currently it's very verbose and the comments being separated from the rules like this makes it hard to follow.

shiftkey:
I'm going to close this out because of inactivity.

@espenkn if you feel like this should still be a template listed here, please read the section about [specialized templates](https://github.com/github/gitignore#specialized-templates) and open a fresh pull request.

PR #1683, opened 2015-10-02
Add local_settings to Django's ignore

edsondota:
local_settings it's a Django standard and very recommended in Django's community

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -50,6 +50,7 @@ coverage.xml
 
 # Django stuff:
 *.log
+local_settings.py
 
 # Sphinx documentation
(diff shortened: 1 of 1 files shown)


shiftkey:
Do you have a pointer to the Django docs for this (or a quick explainer) for future reference?

edsondota:
In addition to the recommendations in blogs and forums. I have that reference [#1](https://code.djangoproject.com/wiki/SplitSettings#RobGoldingsmethod).

shiftkey:
Sure, that seems reasonable (and user-specific)! Thanks!

PR #1684, opened 2015-10-02
Ignore output from R CMD build and R CMD check

jimhester:
Both rules are restricted to the top level only, so `*.tar.gz` files in data directories for instance will _not_ be ignored.


shiftkey:
@jimhester this all seems reasonable. Thanks for contributing!

PR #1685, opened 2015-10-02
Ignore `.RData` files

jimhester:
N.B. this ignores only the exact file `.RData` generated by default from `save.image()` and `quit()`.  It is _not_ what is referred to in https://github.com/github/gitignore/commit/1f8466164330277fc8c9126ded3cc207ff3baf91 and http://cran.r-project.org/doc/manuals/r-release/R-exts.html#Data-in-packages, which refers only to files with an `*.RData` _extension_.

### DIFF
--- a/R.gitignore
+++ b/R.gitignore
@@ -2,6 +2,9 @@
 .Rhistory
 .Rapp.history
 
+# Session Data files
+.RData
+
 # Example code in package build process
 *-Ex.R
 


b-long:
:+1: LGTM

shiftkey:
Thanks!

arcresu:
This rule was previously removed in #1245. Is that reasoning no longer valid?

jimhester:
It was not valid in #1245, I mentioned this in the commit comment, Writing R extensions is talking about file's with an _extension_ of `.RData`.  This rule only applies to files whose _name_ is `.RData`, which are temporary files used to store the data for an R session (and should always be ignored).

PR #1687, opened 2015-10-04
Ignore FAKE - F# Make cache folder

vbfox:
Recent versions of FAKE uses a cache folder containing a compiled version of all scripts named .fake that should never be commited to git.


shiftkey:
Yes! I encountered this but forgot to raise it upstream. Thanks!

PR #1688, opened 2015-10-07
Created "gitignore" for Unreal Engine 4 projects

chbecher:


UE4.gitignore
```diff
@@ -0,0 +1,53 @@
...
+*.ipa
+*.apk
+
+# Generated by the engine
+*.xcodeproj
+*.sln
+*.suo
+*.opensdf
+*.sdf
+*.sln
```
shiftkey: Duplicate

shiftkey:
This all seems nice. 

Thoughts @arcresu?

arcresu:
LGTM. Just that I usually prefer to alphabetically sort the rules under each sub-heading which helos with catching duplicates and overlapping rules.

shiftkey:
@chbecher looks like we're good to go after addressing this feedback!

chbecher:
I am on it :)

shiftkey:
It looks like there's a few extensions which have been removed - could you shed some light on the motivation for dropping those?

chbecher:
Sorry, that was a misunderstanding with the duplicates. 
I'll add the "duplicates" again and sort them alphabetically under the heading. Right?
I used as example the Unity.gitignore
Do you have a better example?

shiftkey:
Closing in favour of #1719 which has some additional rules.

@chbecher apologies for not returning to this PR in a timely manner.

PR #1691, opened 2015-10-08
Adding Fastlane ignore files

felipeplets:
Fastlane https://fastlane.tools/ is a very useful tool that lets you define and run your deployment pipelines for different environments. It helps to unify apps release process and automate the whole process. fastlane connects all fastlane tools and third party tools, like CocoaPods and xctool.

These two ignore lines are the ones that avoid generated files to be checked in to the repo and also used in the fastlane main example: https://github.com/KrauseFx/fastlane-example/blob/master/.gitignore

### DIFF
--- a/Objective-C.gitignore
+++ b/Objective-C.gitignore
@@ -41,3 +41,10 @@ xcuserdata
 # Carthage/Checkouts
 
 Carthage/Build
+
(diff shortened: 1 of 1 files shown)


shiftkey:
This seems to have been done in #1704

PR #1693, opened 2015-10-09
Fix Prestashop unwanted gitignored directory

PrestanceDesign:
The line `cache/` causing problem for gitignore important class directory and files `classes/cache/*`.

### DIFF
--- a/Prestashop.gitignore
+++ b/Prestashop.gitignore
@@ -8,6 +8,7 @@ config/settings.*.php
 
 admin-dev/autoupgrade/
 cache/
+!classes/cache/
 config/xml/*.xml
 log/
 *sitemap.xml


julienbourdeau:
I didn't realized when I wrote it ! Thank you !

I think it would be better to add a slash at the beginning to specify we only want to ignore the cache folder in the root directory:
```
/cache
```

Like if you have a module about cache or something, there might be a important folder named cache and you don't want to ignore it.

PrestanceDesign:
> I think it would be better to add a slash at the beginning to specify we only want to ignore the cache folder in the root directory

Yes I agree, it's better. I'll update my PR immediatly.

PrestanceDesign:
OK, you've already done. I close this PR.

julienbourdeau:
Hey, I've done it on the PrestaShop/PrestaShop repo, you still need to do it here to update the gitignore provided by github :)
You should update and reopen it.

PrestanceDesign:
I already does in this new PR #1694.
Thx @julienbourdeau for the advice.

julienbourdeau:
Great ! :+1:

PR #1694, opened 2015-10-10
Fix path to ignore only root log, img and cache folder

PrestanceDesign:
As in our exchange here https://github.com/PrestaShop/PrestaShop/pull/4246, this PR fix significant following subfolders ignored :
```
classes/log/
```
```
classes/cache/
```
```
modules/**/img/
```


shiftkey:
Thanks!

PR #1699, opened 2015-10-14
Ignoring build folder inside app module in Android projects

vitorprado:
The `*/build` folder should be ignored because the android structure generate files and outputs inside this folder.


shiftkey:
Thanks!

PR #1700, opened 2015-10-16
added .env.example for projects generated with laravel new command

mjoshy:


shiftkey:
Thanks!

mjoshy:
@shiftkey My mistake! Just now, I cloned a laravel repo and realized it was mistake to ignore .env.example, because without '.env' and '.env.example', it becomes hard to create .env file. I had to get the file from elsewhere paste it into my workspace.

PR #1701, opened 2015-10-16
Update Eclipse.gitignore

SuyashLakhotia:
Added *.class to the .gitignore file for Eclipse.


shiftkey:
This should be covered by the Java template

https://github.com/github/gitignore/blob/master/Java.gitignore#L1

Ideally this template should contain IDE or tooling-specific rules.

PR #1703, opened 2015-10-16
ignore Pycharm project file and OSX dir .DS_Store

xlzd:


arcresu:
Thanks, but see `Global/JetBrains.gitignore` and `Global/OSX.gitignore1 for these.

PR #1707, opened 2015-10-19
Added doctrine 2 orm default cache and proxy locations

jeanCarloMachado:
Many of the ZF2 applications are integrated with Doctrine 2. The default location of Proxies and Caches are the ones specified in the PR. For more details checkout the configurations examples: https://github.com/doctrine/DoctrineORMModule/blob/master/docs/configuration.md


shiftkey:
@jeanCarloMachado this all seems reasonable, thanks for contributing!

PR #1710, opened 2015-10-21
Update Dart.gitignore

ArgentiApparatus:
1. Modify patterns to explicitly match directories

2. Modified patterns to match files and directories created by tools in Dart SDK 1.12:

    1. `pub` patterns:

        * Added patterns matching files listed in https://www.dartlang.org/tools/private-files.html

    2. `dart2js` patterns:

        * Added `*.js`, `*.precompiled.js` removed `*.js_`, `*.dart.js` to match files created by dart2js - see https://stackoverflow.com/questions/20314796/which-files-are-generated-when-executing-dart2js-and-why
        * Added `*.info.json` - see https://www.dartlang.org/tools/dart2js/

### DIFF
--- a/Dart.gitignore
+++ b/Dart.gitignore
@@ -1,18 +1,18 @@
-# Don’t commit the following directories created by pub.
+# Files and directories created by pub
 .buildlog
+.packages
+.project
 .pub/
 build/
-packages
-.packages
+**/packages/
 
-# Or the files created by dart2js.
-*.dart.js
-*.js_
+# Files created by dart2js
+*.js
 *.js.deps
 *.js.map
 
-# Or the files created by dartdoc.
+# Directory created by dartdoc
 doc/
 
-# Don't commit pubspec lock file. (Library packages only! Remove pattern if developing an application package.)
+# Don't commit pubspec lock file (Library packages only! Remove pattern if developing an application package)
 pubspec.lock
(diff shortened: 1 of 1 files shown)


shiftkey:
@ArgentiApparatus thanks for the first link, really explain it well.

Do you have a link summarizing the recommended changes for dealing with dart2js's output? I can only see these rules mentioned from the first link.

```
*.dart.js
*.js_
*.js.deps
*.js.map
```

ArgentiApparatus:
https://www.dartlang.org/tools/private-files.html should be authoritative, but experiments with SDK 1.12 `dart2js` showed that `.js`, .`js.deps` and .`js.map` only are created. Presumably an earlier version of `dart2js` created that `.js_`.

I believe the contents of the Dart tools webpages are on Github, if so I'll look for or create a issue for that page. 

`pub` build creates a `.dart.js` file, but it is in the `./build/ directory`, and hence excluded by the `pub` patterns.

Note I pushed a commit to my branch adding `*.info.json` which I noticed when reading through https://www.dartlang.org/tools/dart2js/

ArgentiApparatus:
A little more research turned up https://stackoverflow.com/questions/20314796/which-files-are-generated-when-executing-dart2js-and-why

I pushed another commit adding `*.precompiled.js`

shiftkey:
@ArgentiApparatus thanks for the extra details. I'm happy to take this in.

ArgentiApparatus:
Thanks!

PR #1711, opened 2015-10-22
Some details for Castalia statistics file

Arvur:
Before XE7 Castalia used to be standalone expert thus I suppose here should be some grounding.

P.S. This is addition to pull request #1303.


shiftkey:
:sparkles: thanks for clarifying this!

PR #1712, opened 2015-10-22
Ignore .publishsettings for Visual Studio

LimeyJohnson:
publishsettings files are certificates used to authenticate with Azure services which are widely used in visual studio.

From https://msdn.microsoft.com/en-us/library/dn495124.aspx : Security Note: The file contains an encoded management certificate that serves as the credentials to administer Windows Azure subscriptions and services. Store this file in a secure location or delete it after you use it

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -142,6 +142,12 @@ publish/
 *.pubxml
 *.publishproj
 
+# Azure .publishsetting file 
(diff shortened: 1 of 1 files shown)


shiftkey:
I gather you found it was already defined further down in the file, but if you'd like to annotate that existing entry with more details about what the file represents I'd :heart: to take in that contribution.

PR #1713, opened 2015-10-23
Added some folders the are "compiled" things

insign:
/vendor - composer modules can be (and probably should) recompiled. Laravel comes with this line by default.
/node_modules - npm modules, same of /vendor.  Laravel comes with this line by default.

/app/storage - Laravel things that can be despised (in staging) like cache, views cache and sessions


shiftkey:
Thanks!

PR #1716, opened 2015-10-25
Commented lines that is not commented.

Lordpeara:
I think that lines are comments, not ignores obviously.

### DIFF
--- a/Global/OSX.gitignore
+++ b/Global/OSX.gitignore
@@ -19,6 +19,6 @@ Icon
 # Directories potentially created on remote AFP share
 .AppleDB
 .AppleDesktop
-Network Trash Folder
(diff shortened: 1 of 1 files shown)


shiftkey:
I believe these are expected to be uncommented, but there's not much detail in the original PR https://github.com/github/gitignore/pull/1015

Lordpeara:
Thanks. I wondered gitignore supports name with spaces

shiftkey:
@Lordpeara I'm kinda indifferent to it - the `\` is generally used to escape special characters like `#` and `!` if the rule needs to use them - spaces aren't really special here...

Lordpeara:
Thanks for review!

PR #1718, opened 2015-10-26
ignore IntelliJ IDEA workspace configuration file

ikidou:
`.idea/workspace.xml` and `.idea/libraries/`  are not necessary just like `local.properties`, if not exists android  studio will be create it.

android studio's default .gitignore file
```
.gradle
/local.properties
/.idea/workspace.xml
/.idea/libraries
.DS_Store
/build
/captures
````

### DIFF
--- a/Android.gitignore
+++ b/Android.gitignore
@@ -31,3 +31,7 @@ proguard/
 
 # Android Studio captures folder
 captures/
+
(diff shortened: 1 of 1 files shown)


shiftkey:
These rules are included under https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore

PR #1464, opened 2015-04-01
Add UnrealEngine.gitignore

0ndorio:
Adds an basic .gitignore for Unreal Engine 4 and the automatically generated Visual Studio project files

project homepage:
    - https://www.unrealengine.com/

documentation related to this first push: 
    - https://answers.unrealengine.com/questions/16787/how-to-setup-git-for-an-ue4-project.html

PR #1719, opened 2015-10-27
Add a .gitignore for Unreal Engine 4

samsheff:
I think it would be really helpful to have a gitignore for Unreal Engine 4 projects. This is the one I have been using - But if there's anything I've overlooked or missed I'd be more than happy to add it. 

Also, I've seen a few PR's that have similar gitignore's (See first link), but nothing is currently in master. Mine is actively maintained (see second link), but I think it would be beneficial and have more reach if we put one in this master list.

Technology Homepage: https://www.unrealengine.com/
Links with more info:
- https://answers.unrealengine.com/questions/228063/gitignore-for-ue4-47.html (Question on UE4 Answers Site)
- https://github.com/samsheff/UE4-Gitignore (My Repo)


shiftkey:
Ugh, I thought we had this already but it turns out https://github.com/github/gitignore/pull/1688 and https://github.com/github/gitignore/pull/1464 didn't get much love and/or feedback.

Lemme have a look over those other PRs and add feedback, as your seems far more complete.

UE4.gitignore
```diff
@@ -0,0 +1,60 @@
...
+*.opensdf
+*.sdf
+*.sln
+
+# Precompiled Assets
+SourceArt/**/*.png
+SourceArt/**/*.tga
+
+# Binary Files
+Binaries/*
```
shiftkey: Some of these rules have been made case-insenstive:

```
[Bb]inaries/
[Dd]erivedDataCache/
[Ii]ntermediate/
[Ss]aved/
```

Do you think that's worth doing here?

UE4.gitignore
```diff
@@ -0,0 +1,60 @@
...
+
+# Builds
+Build/*
+
+# Don't ignore icon files in Build
+!Build/**/*.ico
```
shiftkey: Another rule that might be valuable here:

```
!Build/**/*.png
```

shiftkey:
Okay, I folded the other differences from those earlier PRs into here. Let me know what you think of them.

UE4.gitignore
samsheff: I think it's wortg changing. The only thing is unreal generates them uppercased, and unless they are changed and you have a case sensitive file system (like those used by Linux) it wouldn't matter.

UE4.gitignore
samsheff: I'm not sure about that. The icon was excluded because it would cause windows builds to fail (See https://github.com/samsheff/UE4-Gitignore/issues/1), but I don't know if we should exclude PNGs since they would be just copied as part of the build. It seems too broad to me.

samsheff:
@shiftkey Great! Left some feedback, let me know what you think.

UE4.gitignore
shiftkey: >  I don't know if we should exclude PNGs since they would be just copied as part of the build. It seems too broad to me.

cc @chbecher who proposed this in an earlier

UE4.gitignore
shiftkey: :thumbsup: I'll defer to you as you're far more familiar with the tooling itself than I am

UE4.gitignore
samsheff: @shiftkey I think we should leave that out for now. My understanding of the build folder is that it's all processed code and assets, plus things such as project files generated by the engine. Any PNG's that end up in there would be already in another part of the project.

shiftkey: :thumbsup:

shiftkey:
Only thing I can think of is the file name - should this be `UnrealEngine` or `UnrealEngine4` to make it more clear?

samsheff:
Let's make it `UnrealEngine4`, that way there's definitely no confusion.

shiftkey:
@samsheff I'd rather us handle the different versions of things in the single file - sorry I wasn't clearer on that. Could you change it to `UnrealEngine`? 

The workflow that we've done with other projects like this is that when a new major release occurs:
 - introduce the new rules and annotate them to indicate which version they belong to
 - move deprecated rules out to a new section and indicate they are only specific to a new version

Having multiple files to manage (and update) just means more headaches for everyone. With one file at least we know where everything is, and with good organization in the file we can help users understand the differences between version...

samsheff:
@shiftkey I agree with that completely. I'll update it in a few minutes

samsheff:
@shiftkey Renamed the file! Give it a final look over and let me know if there's any other changes that need to be made. Thanks!

shiftkey:
well-done-sir

BGR360:
I have a few contributions to make to this new Unreal Engine gitignore. Should I fork this repository?

shiftkey:
@BGR360 yes please! You should be able to make the changes on the website (it'll take care of forking for you, and then ask if you want to open a pull request).

BGR360:
Awesome! I will get right on that!

BGR360:
@shiftkey Which website are you talking about?

shiftkey:
@BGR360 this one!

PR #1720, opened 2015-10-27
gitignore for laravel5 and lumen

jerodev:
Created lumen gitignore
Created laravel5 gitignore compatible with the new folder structure.

### DIFF
--- a/Laravel.gitignore
+++ b/Laravel.gitignore
@@ -1,11 +1,14 @@
+vendor/
+node_modules/
+
+# Laravel 4 specific
 bootstrap/compiled.php
+app/storage/
+
+# Laravel 5 & Lumen specific
 bootstrap/cache
+storage/
 .env.*.php
 .env.php
 .env
 .env.example
-
-app/storage/
-node_modules/
-storage/
-vendor/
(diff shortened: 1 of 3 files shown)


shiftkey:
Is there a reason why the existing Laravel file cannot be used?

https://github.com/github/gitignore/blob/master/Laravel.gitignore

jerodev:
i thought it would be better to create a new one because Laravel5 has an entire different folder structure compared to Laravel4.

But indeed, it would be possible to merge them. I will create another commit.

shiftkey:
@jerodev would it be worth grouping the new rules - the ones specific to Laravel 5 - together to help users understand which ones they should care about?

shiftkey:
Also, I'm not familiar with this Lumen project - could you tell me a bit about that and where those rules have come from?

jerodev:
@shiftkey Lumen is a php micro framework. It is actualy a stripped down version of Laravel. http://lumen.laravel.com/

I have removed the lumen gitignore because actualy the folder structure from Laravel5 and Lumen is almost the same and the rules work for both frameworks. I have also added comments to let the users know what rules are for wich version.

shiftkey:
@jerodev that's great, thanks!

shiftkey:
Let me know when this is ready to review - it looks close!

jerodev:
@shiftkey I have tested the gitignore with both a Laravel5 and a Lumen project, and it works great!

PR #1726, opened 2015-10-30
Add custom-built swift libraries

mcandre:
Add patterns for Swift libraries that the user builds

### DIFF
--- a/Swift.gitignore
+++ b/Swift.gitignore
@@ -31,3 +31,11 @@ DerivedData
 # Carthage/Checkouts
 
 Carthage/Build
+
+# Libraries
+#
+# Swift libraries built as artifacts of part of a Swift project
+
+*.swiftmodule
+*.swiftdoc
+*.dylib


Swift.gitignore
```diff
@@ -31,3 +31,11 @@ DerivedData
...
 
 Carthage/Build
+
+# Libraries
+#
+# Swift libraries built as artifacts of part of a Swift project
+
+*.swiftmodule
+*.swiftdoc
+*.dylib
```
shiftkey: While the other two rules are very specific to Swift, this third one seems to be more general. Is there a particular reason why we're needing to introduce this now?

shiftkey:
@mcandre thanks for this, and apologies for not getting to it sooner. Are you interested in resolving the merge conflicts here so we can merge this in?

Also, I'm curious if there's any documentation around that we can point users to, which explains why these files are useful.

shiftkey:

PR #1728, opened 2015-10-31
Keep Visual Studio and XCode project files

BGR360:
The original author excluded XCode .xcodeproj files and Visual Studio .sln files. Personally, in my current project, I would prefer to include the (lightweight) project files in my repo because it allows anybody to clone the repo and then immediately build the project from the IDE. Without the files, the cloner has to open the .uproject file in UE4 Editor and go to File > Generate [Visual Studio, XCode] Project. Sometimes, the user may want to just open the project in the IDE and build the Release version without having to fiddle with the Editor.

This is of course a personal preference/opinion but I believe it would be worth implementing.
If so, lines 37 and 38 should also be noted as being important because XCode stores a bunch of user-specific garbage inside the .xcodeproj folder.


BGR360:
Actually, I'm just realizing that if you were to add/remove code on a Mac and then clone onto a PC, the Visual Studio project would not be updated. So, it would be best to have to refresh the IDE projects every time you clone.

shiftkey:
I'll let @samsheff chime in here about this as I'm not a domain expert.

shiftkey:
@BGR360 unfortunately this isn't mergeable currently, and I was hoping to get some @samsheff to help :+1: this change, but it didn't work out that way :cry:

Can you point to other threads where people have suggested this (and it might be approaching a small consensus of people)? Otherwise perhaps annotating the benefits of disabling this rule might be a easier change to incorporate here...

samsheff:
@BGR360 and @shiftkey Sorry about not chiming in on this guys, been busy and dropped the ball.

Here are my thoughts:
XCode files are generated by older versions of Unreal as build artifacts. Unlike Unity, which requires you compile the project outside of the editor, Unreal doesn't and simply uses them internally as some part of the build pipeline. A similar thing applies to sln files, but since I never use unreal on Windows I cant guarantee the same rules apply in the same way. 

In short, I think they should definitely be added to the .gitignore by default.

shiftkey:
Thanks @samsheff for the feedback! Thoughts @BGR360?

BGR360:
The need to keep these files is not really there. After having used the engine for a while, I have realized it is trivial (and beneficial) to simply regenerate the project files when needed. No need to pull my changes :)

PR #1732, opened 2015-11-02
Add .npm cache directory

jucrouzet:


shiftkey:
Thanks!

jucrouzet:
For further references : npm uses ~/.npm as its cache directory, sometimes (like a case I encoutered which is launching npm from a docker container that uses the project directory as the user's home dir), this directory can be added in git.

PR #1733, opened 2015-11-03
Ignore ASP.Net 5 project.lock.json

Kagamine:
Ignore ASP.Net 5 project.lock.json

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -179,6 +179,7 @@ ClientBin/
 *.publishsettings
 node_modules/
 orleans.codegen.cs
+project.lock.json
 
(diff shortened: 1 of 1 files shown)


shiftkey:
cc @davidfowl @eilon

Eilon:
Hmm it's not necessarily wrong to check it in. For some projects it's the right thing to do because it allows easily locking dependencies. @davidfowl thoughts?

Kagamine:
When the environment changed, the project.lock.json must be generated again, so I think we should ignore it.

shiftkey:
Nevermind, I knew we already talked about this before https://github.com/github/gitignore/pull/1505

PR #1738, opened 2015-11-04
Revert 2bd7d8ec88bc9fad5aca91951826b16118227554

apriendeau:
Reverts github/gitignore#1689

@shiftkey This is bad. Can we revert this please? It is a common go idiom to commit your dependencies. Only [gb](https://github.com/constabulary/gb) uses a manifest file. The most popular is [godeps](https://github.com/tools/godep) right now and they suggest that you commit the dependencies. I think since this controversial. I would suggest that remove this. I know that we commit our dependencies and would like to continue to do so by default. Go 1.5 does not state anywhere that you should not commit them.

### DIFF
--- a/Go.gitignore
+++ b/Go.gitignore
@@ -22,7 +22,3 @@ _testmain.go
 *.exe
 *.test
 *.prof
-
-# GO15VENDOREXPERIMENT flag specifics
-vendor/**
-!vendor/manifest


shiftkey:
cc @enzian

enzian:
Hmm, not entirely true, since gvt actually uses manifests too!! Altough i disagree on the 'commit your dependencies' approach I agree on the argument that it is common to commit dependencies in go! Lets revert it, and I'll keep excluding those directories ;-)

shiftkey:
@enzian @apriendeau :thumbsup:

PR #1740, opened 2015-11-05
Remove nb-configuration.xml from NetBeans ignore

bondolo:
nb-configuration.xml is a project file generated by the IDE. per the comment in the file:

    This file contains additional configuration written by modules in the NetBeans IDE.
    The configuration is intended to be shared among all the users of project and
    therefore it is assumed to be part of version control checkout.
    Without this configuration present, some functionality in the IDE may be limited or fail altogether.

It should therefore not be ignored.


shiftkey:
Thanks!

PR #1741, opened 2015-11-06
add max osx .DS_Store and pycharm .idea folder

twz915:
add max osx .DS_Store and pycharm .idea folder

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -5,6 +5,11 @@ __pycache__/
 
 # C extensions
 *.so
+# mac osx
+.DS_Store
+
+#pycharm
+.idea
 
 # Distribution / packaging
 .Python


cowlicks:
Why should `.DS_Store` be in python gitignore and not a global gitignore?

cowlicks:
See here https://github.com/github/gitignore/blob/master/Global/OSX.gitignore#L1 for .DS_Store 

Maybe a global gitignore would be more appropriate for the pycharm .idea folder

shiftkey:
As @cowlicks mentioned, there are global ignore files for OS X and other IDEs - see https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore for the JetBrains rules.

PR #1743, opened 2015-11-07
Added .DS_Store for mac users

payamrastogi:


shiftkey:
We store OS-specific rules in the global templates, so they don't need to be duplicated:

https://github.com/github/gitignore/blob/master/Global/OSX.gitignore

PR #1746, opened 2015-11-09
Updates for current Magento version.

martinvw:
The newer Magento versions contain a folder 'dev' which is the base for some tests and allows you to add your own tests.

More information about the new tests:
http://magento.stackexchange.com/questions/86078/how-to-use-test-in-dev-folder

Besides some extra libs were added and two additional files.


Magento.gitignore
```diff
@@ -45,8 +45,43 @@ app/etc/local.xml
 app/.htaccess
 app/locale/
 app/Mage.php
+app/bootstrap.php
 /cron.php
 cron.sh
+/dev/tests/functional/etc
+/dev/tests/functional/lib/Magento/Mtf
+/dev/tests/functional/tests/app/Mage
+/dev/tests/functional/testsuites/Magento/Mtf/TestSuite/InjectableTests/3rdParty.xml
```
shiftkey: I gather these are generated tests - could these be simplified into a more wildcard-friendly rule? Or are user tests likely to be put in the same locations?

martinvw: I will take a look at it. I just installed it out of the box and made the assumption that they are there to extend. But I will validate it :+1:

martinvw: Like the name already suggested you can use these xml files to describe test suites. By default the basic.xml is used. But you could optionally change this behavior via the configuration file. And it seems the developers at Magento already give a away some of there own integration test suites.

So yes, this a place where you add your own testsuites. Which make it hard to catch it in a wildcard without catching too much.

PR #1747, opened 2015-11-10
Issue with .gitignore not ignoring node_modules

vmaudgalya:
This fixes the bug that is documented here:
https://github.com/vmaudgalya/redux-timer/commit/30ba26fda235e516fa31b82d33b39455bd177965

The bug is as follows:
Initially, when I created the repository "redux-timer" I initialized the repo with the default .gitignore for Nodejs projects. However, after pushing, .gitignore did not ignore node_modules/ folder and its contents, and thus I ended up pushing node_modules, and ended up having to remove it later. This PR should fix this issue.

### DIFF
--- a/Node.gitignore
+++ b/Node.gitignore
@@ -25,7 +25,7 @@ build/Release
 
 # Dependency directory
 # https://docs.npmjs.com/misc/faq#should-i-check-my-node-modules-folder-into-git
-node_modules
+node_modules/
 
 # Optional npm cache directory
 .npm


shiftkey:
What version of Git are you running? And what OS? I don't believe this is a bug in the syntax of the gitignore file, but perhaps in the implementation of ignoring files in Git:

 - `node_modules` is interpreted as a glob from the directory of the `.gitignore` file, so it should catch everything 
 - `node_modules/` is only different for one scenario - where you want to include a file named `node_modules` and exclude the contents of the folder `node_modules/`

vmaudgalya:
@shiftkey good point, you're right about ```node_modules/``` referring to a single scenario.

I'm using git version 2.4.9 (Apple Git-60) on OS X 10.10.5 (14F1021), Yosemite

JbLoves:
what was the resolve for this issue? My .gitiognore file is not ignoring files as well.

shiftkey:
> My .gitiognore file is not ignoring files as well.

It's hard to say without a good repro - based on my anecdotal experience most of the time it seems like the `node_modules` file is already being tracked before the `.gitignore` is updated, so it needs to be removed manually to ensure it stops tracking this folder.

If there is an bug here I really think it needs to be investigated by the Git team itself - I'm not sure changing this rule is going to address the issue at hand.

vmaudgalya:
When I encountered this bug, it was a one-off bug that was extremely hard to reproduce (trust me, I've tried multiple times). Reflecting over this PR I agree with @shiftkey - I doubt that this is the actual fix for the issue, but perhaps if someone could better document the steps taken to encounter the issue it would be easier to actually find the problem. Tbh, I feel that this might be a [Heisenbug](https://en.wikipedia.org/wiki/Heisenbug).

shiftkey:
@vmaudgalya I wasn't able to reproduce it either. Are we cool to close this out given there's a merge conflict here?

vmaudgalya:
@shiftkey Yup

JbLoves:
I can not reproduce it either.

PR #1751, opened 2015-11-13
Update Android.gitignore add idea folder.

CharonChui:


Soulcyber:
+ 1, without this there will be conflict every time you download a VCS projects in another directory, i suggest to add *.iml too.

shiftkey:
As discussed in #1637, we're tracking all the JetBrains-specific rules in `Global/JetBrains.gitignore` to help with sharing across tooling and reducing duplication.

I'd also love to understand where this new `*.iml` rule came from, and what it's there for. Could you provide a link so I can understand more about this?

shiftkey:
Closing this out because I confused the PRs and there's no `*.iml` rule here.

PR #1752, opened 2015-11-13
add Cargo.lock to gitignore file by default

waynenilsen:
See the [official recommendation](http://doc.crates.io/guide.html#cargotoml-vs-cargolock) 

> * If you're building a library, put Cargo.lock in your .gitignore.
> * If you're building an executable, check Cargo.lock into git.


shiftkey:
Could you put a comment before this line, explaining when you should check this file in? I think I'm happy with making this opt-out for those who are making an executable...

waynenilsen:
is this what you were thinking?

shiftkey:
:thumbsup:

amadalusia:
times have now changed and most gitignores add the cargo.lock back in, so i think this should be reversed.
heres why: https://doc.rust-lang.org/cargo/faq.html#why-have-cargolock-in-version-control

PR #1753, opened 2015-11-13
Add section for pycharm files

nhumrich:


shiftkey:
See [Global/JetBrains.ignore](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore) for these IDE-specific rules

PR #1756, opened 2015-11-16
Adding endfloat.sty auxiliary files

habi:
endfloat.sty produces these two auxiliary files.
From Section 3.2 of its documentation:
> Loading it will have LATEX produce two extra files with .ttt and .fff extensions (for tables and figures, respectively).


shiftkey:
Thanks!

PR #1757, opened 2015-11-16
Delphi: More extensions to ignore

Arvur:
- Rarely used OLE Control eXtension (OCX) is created by the Active Form wizard
- Files generated by "C/C++ Output file generation" compiler setting


shiftkey:
Thanks for organizing these extensions and documenting the `obj` rule!

PR #1441, opened 2015-03-19
Add Cordova.gitignore to use with Adobe Cordova projects

vrunoa:

PR #1061, opened 2014-04-29
Cordova ignores (android) for 3.4

ccoenen:
Not much else to say: Here's what i ignore on a cordova/android project.

PR #1758, opened 2015-11-16
Added .gitignore for @phonegap/@cordova

christoga:
I added new  `.gitignore` for Cordova,
Because there is a lot of people who use cordova and some files to ignore.

### DIFF
--- a/Cordova.gitignore
+++ b/Cordova.gitignore
@@ -1,0 +1,13 @@
+# Mac
+.DS_Store
+
+# Node
+npm-debug.log
+/node_modules
+
+# Cordova
+/platforms
+/plugins
+
+# res
+resources/signing


Cordova.gitignore
```diff
@@ -0,0 +1,13 @@
+# Mac
+.DS_Store
```
shiftkey: This should be removed as it's covered by `Global/OSX.gitignore`

Cordova.gitignore
```diff
@@ -0,0 +1,13 @@
...
+
+# Node
+npm-debug.log
+/node_modules
+
+# Cordova
+/platforms
+/plugins
+
+# res
```
shiftkey: Can this comment not be abbreviated?

shiftkey:
Hey @brianleroux who can I get to have a look over this and help sign off on it?

shiftkey:
@christoga we had a very old PR for this #1061 that seems to be distilled down to just a `/platforms` rule - am I reading that right?

ccoenen:
my approach was, to keep as much as possible. Sometimes, you edit platform specific files (at least, we did), so i'm not too fond of the extensive ignore, there.

ccoenen:
Also, as far as i can tell, platform information will be stored in `platforms/platforms.json`. This Ignore will make it impossible to restore the exact version that went into a build. The same goes for plugins, where the directory-content itself seems to be the state of plugins. So If you ever need to get back to your project or bring another developer in, you'll have no way but to keep a Readme with all the cordova plugins you've been using (and which versions).

ccoenen:
Interestingly, #1441 used very similar patterns, if there is any discussion among Cordova devs, and this has indeed become a best practise, I'd love to join that discussion!

christoga:
Sorry, is this `.gitignore` useless?
If yes, just close this pull request,

All the best,
Christo

ccoenen:
I am not an authority on either Cordova or GitIgnore, I did not mean to sound harsh in my comments.

shiftkey:
#1061 seems to talk about the previous state of Cordova making any sort of `gitignore` template subject to change. I guess we need to work of if it's worth revisiting this.

christoga:
Why didnt you didnt merge this?
There is a lot of people use cordova.

All the best,
Christo

shiftkey:
> Why didnt you didnt merge this?

I'm still reviewing this.

What helps simplify this review process is links to supporting documentation about what should be in version control and what should be excluded - I can't be across every detail of the programming languages and products in use, and this helps others who return to this PR understand _why_ these rules have been used...

libinvarghese:
I would recommend a generic gitignore Cordova template (`Cordova.gitignore`) which would simply assume there is no custom code in the platforms, as the `--save` option while adding a platform & plugin is used to restore their state respectively  - https://stackoverflow.com/a/29972835/7274758

When you do have custom code with Platforms, I would recommend using additional platform specific gitignore templates say `Cordova-android.gitignore` similar to #1061

christoga:
Bro, this PR is dead.

ccoenen:
I believe that "bro" might be the wrong way to address her ;-)
But I concur that this issue hasn't hat an update in years and is at least outdated. She does make a valid point, though. Cordova has evolved a lot in that time, and the practise mentioned on StackOverflow _looks_ like a good starting point for a rule. I can't fully judge this, as I haven't followed cordova as closely in recent versions.

NatoBoram:
Hey, I just wanted to chime in saying that I'm using this `.gitignore` in prod.

shiftkey:
@christoga apologies for the delay in replying to this.

I had a lot of these templates requested, but I couldn't quite figure out the right place for them. I ended up settling on #2857, and you'll see a Cordova template can now be found at `community/JavaScript/Cordova.gitignore`.

PR #1759, opened 2015-11-18
OpenCart system cache and logs location changes

nklatt:
I tried to locate when this changed in OpenCart (https://github.com/opencart/opencart/commits/master) but wasn't able to just looking at the commit messages. But, as of a fresh install of 2.1.0.1, these appear to be the new locations.

### DIFF
--- a/OpenCart.gitignore
+++ b/OpenCart.gitignore
@@ -7,5 +7,6 @@ admin/config.php
 download/
 image/data/
 image/cache/
-system/cache/
-system/logs/
+system/library/cache/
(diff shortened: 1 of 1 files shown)


shiftkey:
@nklatt apologies for not getting back to this earlier. It looks like #1855 is also overlapping with this change, and this isn't currently mergeable. Are you interested in resolving these conflicts, or should we just move the discussion over there?

shiftkey:
Closing due to inactivity and conflicts.

PR #1760, opened 2015-11-18
Added *.DS_Store for Mac users

BGR360:


shiftkey:
We track these in `Global/OSX.gitignore` rather than apply them to every possible scenario.

PR #1761, opened 2015-11-19
Remove *.xcscheme from Global/Xcode.gitignore,Objective-C.gitignore and…

nrbrook:
… Swift.gitignore

Xcode schemes should be held under source control when shared. When not shared they are stored in xcuserdata, so are already ignored. Therefore ignoring *.xcscheme is incorrect, as far as I'm aware.

### DIFF
--- a/Global/Xcode.gitignore
+++ b/Global/Xcode.gitignore
@@ -21,4 +21,3 @@ xcuserdata
 *.xccheckout
 *.moved-aside
 *.xcuserstate
-*.xcscheme
--- a/Objective-C.gitignore
+++ b/Objective-C.gitignore
@@ -22,7 +22,6 @@ xcuserdata
 *.moved-aside
 *.xcuserstate
 *.xcscmblueprint
-*.xcscheme
 
 ## Obj-C/Swift specific
 *.hmap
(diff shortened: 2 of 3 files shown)


shiftkey:
@nrbrook do you have a supporting link which discusses this? I'd love something (semi-)official to point other curious people to...

nrbrook:
There is nothing specific, but some references to sharing schemes in Apple's documentation.

https://developer.apple.com/library/ios/recipes/xcode_help-scheme_editor/Articles/SchemeShare.html
https://developer.apple.com/library/ios/recipes/xcode_help-scheme_editor/Articles/SchemeManage.html

shiftkey:
So sharing this file seems like something that should be opt-in (either explicitly add the file or delete the line from the `gitignore` file) because perhaps people don't want to share these schemes.

Instead of removing this rule, should keep it around and annotate it to explain when you should be committing this file?

nrbrook:
'Sharing' is opt-in – schemes by default are not shared, and are stored in xcuserdata directory, which is ignored already here. But when you enable sharing on a scheme, Xcode moves it to the xcshareddata directory, and it should be committed. As far as I'm aware, xcschemes are only placed in those two locations. As such, the wildcard to ignore all scheme files is unnecessary.

The most popular answer on stack overflow does not ignore *.xcscheme.

http://stackoverflow.com/questions/49478/git-ignore-file-for-xcode-projects

shiftkey:
@nrbrook thanks for the extra context!

PR #1762, opened 2015-11-19
Add ignores for autoscan files

lcorbasson:
Files generated by [autoscan](https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.69/html_node/autoscan-Invocation.html) should be ignored too: they're a `configure.ac` template (`configure.scan`) and logs.
This is done by numerous projects, e.g. https://github.com/google/ios-webkit-debug-proxy/pull/113/files

### DIFF
--- a/Autotools.gitignore
+++ b/Autotools.gitignore
@@ -2,6 +2,10 @@
 
 Makefile.in
 
+/autoscan.log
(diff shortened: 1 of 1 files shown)


shiftkey:
Thanks!

PR #1763, opened 2015-11-20
gitignore for JENKINS_HOME Jenkins settings

samrocketman:
This allows an admin to use git to keep a backup of Jenkins settings without tracking binary artifacts.  Useful for preserving settings during plugin upgrades.

Note: `secret.key` is purposefully not tracked by git.  This should be backed up separately because configs may contain secrets which were encrypted using the `secret.key`.

See also:
- http://jenkins-ci.org/
- https://wiki.jenkins-ci.org/display/JENKINS/Administering+Jenkins

### DIFF
--- a/JENKINS_HOME.gitignore
+++ b/JENKINS_HOME.gitignore
@@ -1,0 +1,19 @@
+#Learn more about Jenkins and JENKINS_HOME directory for which this file is intended.
+#  http://jenkins-ci.org/
+#  https://wiki.jenkins-ci.org/display/JENKINS/Administering+Jenkins
+
+#ignore all JENKINS_HOME except jobs directory, root xml config, and .gitignore file
+/*
+!/jobs
+!/.gitignore
+!/*.xml
+
+#ignore all files in jobs subdirectories except for folders
+#note: git doesn't track folders, only file content
+jobs/**
+!jobs/**/
+
+#exclude only config.xml files in repository subdirectories
+!config.xml
+
+#as a result only settings and job config.xml files in JENKINS_HOME will be tracked by git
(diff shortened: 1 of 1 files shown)


stephengroat:
This has been working well for me. I've been using it to version control configs for different slaves

samrocketman:
Awesome :)

samrocketman:
Is there any interest in merging this?

samrocketman:
@shiftkey mind taking a look at this again?

Stargator:
@samrocketman What about the config-history folder? Should that be ignored by default?

I think anything that is either a binary or contains credentials should be ignore too (next build numbers are optional, which I like)

Is there anything else like config-history that could be useful?

samrocketman:
@Stargator 

config-history folder is added by a Jenkins plugin so should not be included by default.  I don’t typically see it installed though I know people use it.  Should be straight forward to add an entry after getting this one.

binaries are already ignored.

Credentials are strongly encrypted so as long as you back up `secret*` separately from the JENKINS_HOME config.  I’m confident storing `credentials.xml` will be fine.  Jenkins stores credentials in many places beyond `credentials.xml` so it is not possible to separate credentials from configuration.  I’m very familiar with this aspect because I independently discovered Jenkins storing credentials with a weak encryption algorithm in older releases of Jenkins (which has been updated Jenkins 2.0 and beyond).

Note: the default gitignore does not track the secret files which are used in conjunction with `config.xml` to decrypt Jenkins credentials in XML configuration.

samrocketman:
@Stargator I just realized this config does not omit `secret*` even though my personal config does.  I haven’t looked at this in a very long time so will update it.

samrocketman:
@Stargator correction I was confused.  It already does not track `secret*`.

shiftkey:
@samrocketman @Stargator @stephengroat thanks for working through the details!

samrocketman:
Thanks for your time @shiftkey !

PR #1764, opened 2015-11-20
Update CakePHP.gitignore

psantos10:
Ignore all tmp and log files but preserve the directory structure

### DIFF
--- a/CakePHP.gitignore
+++ b/CakePHP.gitignore
@@ -2,8 +2,20 @@
 
 /vendor/*
 /config/app.php
-/tmp/*
+
+/tmp/cache/models/*
+!/tmp/cache/models/empty
+/tmp/cache/persistent/*
+!/tmp/cache/persistent/empty
+/tmp/cache/views/*
+!/tmp/cache/views/empty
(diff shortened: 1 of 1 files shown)


shiftkey:
@psantos10 are there any supporting links or discussions around this change? I'd love to be able to point to something so that others can understand _why_ a change was made...

psantos10:
> As such, make sure the directories logs, tmp and all its subdirectories in your CakePHP installation are writable by the web server user. Composer’s installation process makes tmp and its subfolders globally writeable to get things up and running quickly but you can update the permissions for better security and keep them writable only for the webserver user.
- http://book.cakephp.org/3.0/en/installation.html#permissions

@shiftkey on some systems, without these folders (`tmp` and `log`), the installation will fail due permissions issue. To get thing done, you need to create the folders manually as CakePHP will be incapable to do that.

For this reason, I think that the best way is to keep the `tmp` directory structure and ignore only the logs and temp files.

shiftkey:
@psantos10 :thumbsup: thanks for this!

PR #1765, opened 2015-11-23
Fix ignore rule for Dart packages symlinks

hermanbergwerf:
This fixes the ignore rule for Dart (https://www.dartlang.org/) packages directories. Dart uses a tool called `pub` for dependency management. All dependencies are stored in the repository root in a directory called `packages`. According to https://www.dartlang.org/tools/private-files.html the packages directory which holds the dependencies should be ignored (of course this also makes a lot of sense). Dart makes a symlink to this packages directory in all directories in your repositories. Since Git treats symlinks as files, the trailing slash on this line should be removed so not only the root packages directory but also all symlinks to this directory are ignored.

### DIFF
--- a/Dart.gitignore
+++ b/Dart.gitignore
@@ -6,7 +6,7 @@
 .project
 .pub/
 build/
-**/packages/
+packages
 
 # Files created by dart2js
 # (Most Dart developers will use pub build to compile Dart, use/modify these 


shiftkey:
Should both rules (the folders and the symlink) both be kept around in the interim, for compatibility reasons? Or will the vast majority be on 1.12 already?

hermanbergwerf:
I believe `packages` should both match the directory and the symlink in any directory. Otherwise I'm not an expert on older Dart versions (I just started using it).

shiftkey:
I'll let this sit for a bit and see if anyone else has some thoughts. Shouldn't be too hard to confirm that rule behaviour anyway.

devoncarew:
We used to have `packages` until recently in this file - not sure what motivated that change but I think `packages` _is_ what we want here.

shiftkey:
@devoncarew oh, yes - it's still on line 5

shiftkey:
Okay, I've left this for a while - is it still relevant @hermanbergwerf @devoncarew given Go 1.6 landed recently?

shiftkey:
Closing due to inactivity and conflicts. @hermanbergwerf please review the latest Dart file and see if any other changes need to be incorporated.

hermanbergwerf:
Looks like a new line for > SDK 1.20 has been added now, so we should be fine.

PR #1766, opened 2015-11-24
add new version of NCrunch files to VisualStudio.gitignore

RaphHaddad:


shiftkey:
This seems to contradict the advice from here: http://www.ncrunch.net/documentation/considerations-and-constraints_using-ncrunch-with-source-control

PR #1767, opened 2015-11-26
add octobercms.gitignore

joaonzangoII:


OctoberCms.gitignore
```diff
@@ -0,0 +1,19 @@
+/bootstrap/compiled.php
+/vendor
+composer.phar
+.DS_Store
+.idea
+.env
+.env.*.php
+.env.php
+php_errors.log
```
shiftkey: Could this be distilled down to a single `*.log`?

OctoberCms.gitignore
```diff
@@ -0,0 +1,19 @@
+/bootstrap/compiled.php
+/vendor
+composer.phar
+.DS_Store
```
shiftkey: We use [`Global\macOS.gitignore`](https://github.com/github/gitignore/blob/master/Global/macOS.gitignore) to track macOS-specific rules. Please omit this rule from your template to prevent unnecessary duplication.

OctoberCms.gitignore
```diff
@@ -0,0 +1,19 @@
...
+.idea
```
shiftkey: We use [`Global\JetBrains.gitignore`](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore) to track rules specific to the JetBrains family of IDEs. Please omit this rule from your template to prevent unnecessary duplication.

PR #1768, opened 2015-11-26
Add rule to unignore the maven wrapper jar

thomaseizinger:
If a project uses the maven-wrapper library (https://github.com/takari/maven-wrapper) the .jar file of the maven-wrapper gets excluded if there is a catch-all `*.jar` rule in the .gitignore. 

I think the un-ignore rule for the maven-wrapper fits best into the normal maven `.gitignore`. 
However, I have noticed that the un-ignore rule is ignored if the catch-all `*.jar` rule is applied after the one I add. Is it possible to further customize the rule in a way, so the order in which the rules appear in the `.gitignore` is not important?

### DIFF
--- a/Maven.gitignore
+++ b/Maven.gitignore
@@ -7,3 +7,6 @@ release.properties
 dependency-reduced-pom.xml
 buildNumber.properties
 .mvn/timing.properties
+
+# Exclude maven wrapper
+!/.mvn/wrapper/maven-wrapper.jar


Maven.gitignore
```diff
@@ -7,3 +7,6 @@ release.properties
 dependency-reduced-pom.xml
 buildNumber.properties
 .mvn/timing.properties
+
+# Exclude maven wrapper
+!/.mvn/wrapper/maven-wrapper.jar
```
shiftkey: I'm not quite sure about this rule - it isn't necessary unless a previous rule is hiding the `.mvn/wrapper/` folder.

Am I understanding the intent here?

shiftkey:
> However, I have noticed that the un-ignore rule is ignored if the catch-all *.jar rule is applied after the one I add. Is it possible to further customize the rule in a way, so the order in which the rules appear in the .gitignore is not important?

Unfortunately not - order matters here

Maven.gitignore
thomaseizinger: The ``maven-wrapper.jar`` is excluded through the catch-all rule ``.jar``. However, the usecase of the maven-wrapper is, to avoid the need of having maven installed on a machine, where I want to build a just-checked-out project. Therefore, the ``maven-wrapper.jar`` needs to be excluded from the the ``.gitignore`` in order to be able to check it in along with the source code.

shiftkey: > Usually, one would also add the `Java.gitignore` file to a Java project, which unfortunately contains such a catch-all rule for `.jar`s. 

Thanks, this clears things up a bit. Not having enough context on how projects manage this, I have a couple of general questions (I don't expect to address them as part of this PR):

 - Is it common for people to use `Java.gitignore` without `Maven.gitignore`?
 - Should all the Maven rules from `Java.gitignore` live in `Maven.gitignore`?
 - Is Maven a tool rather than a programming language? If so, it should live under the `Global` namespace...

thomaseizinger: > Is Maven a tool rather than a programming language? If so, it should live under the Global namespace...

Yes. Maven is a build and dependency management tool for the JVM.

> Is it common for people to use Java.gitignore without Maven.gitignore?

I would not say it is common but there are of course situations where one would not want to use Maven. Although I think Maven is still a very popular build/dependency management tool for Java projects.

> Should all the Maven rules from Java.gitignore live in Maven.gitignore?

No, because of the above reason.

thomaseizinger:
Are there any plans on merging this or is something still unclear?

Maven.gitignore
drothmaler: I'll add my 2 cents too :-)

> Is it common for people to use Java.gitignore without Maven.gitignore?

I would say it is very common to use Java.gitignore without Maven.gitignore. 
Maven is just one of many Java Build-Tools (there are also Gradle [which became more and more popular during the last years], SBT [mostly for Scala], Ant, Kobalt and probably more). Also you don't have to use a build-system at all. You can also use the Java-Compiler (javac) directly (even if this is not common). 

Java.gitignore also is not just for "real" Java project, but actually for the Java Platform. You would use it for all projects, that are based on the JVM - also Scala, Groovy, Kotlin, ... projects. 
And then you would add more gitignore (for build-tools, languages, transpilers), depending on your project setup.

ssherwood:
FYI - this is affecting some Spring Boot users that generate from http://start.spring.io/ when using the default build tool (which is the Maven wrapper).  If they then use https://www.gitignore.io/ to help generate their .gitignore file, git will end up excluding the wrapper.jar (which is essential to builds).  In my case the downstream impact was a broken build in my CI tool that took me several hours to figure out.

@shiftkey The solution presented here is quite similar to the existing one for the Gradle wrapper (Gradle is an alternative build tool for Java apps).  You can see the rule applied here: https://github.com/github/gitignore/blob/master/Gradle.gitignore

It looks like the rule presented in this patch is even more explicit than that in Gradle in that it checks the full path of the wrapper.jar so it looks fairly safe to apply.  I'm not an expert on the Gradle wrapper so I don't know if there are more possible places that its wrapper.jar could be located which might be why they left it without a folder (the docs seem to indicate it is a know path: https://docs.gradle.org/current/userguide/gradle_wrapper.html#using_wrapper_scripts).

thomaseizinger:
Thanks for commenting @ssherwood!
The solution I present here is indeed directly inspired from `Gradle.gitignore`. As you have pointed out correctly, this one is actually more restrictive, as it just applies to wrapper jar and nothing else.

Gradle introduced the concept of the wrapper jar to allow users to build their projects without having gradle installed locally. This concept has then been ported for maven. It don't think the will change the location or name of the wrapper jar, as there is absolutely no need for doing so.

PS: I too was stuck with a failing CI build until I realized that the catch-all `jar`-exclude rule excluded the wrapper jar from my repository.

@shiftkey What would have to be done to actually merge this pull request? Or is there still something unclear that needs to be sorted out?

cidpunk:
Thanks for get me involed

thomaseizinger:
@shiftkey Any plans on merging this soon? Are there any issues left which should be discussed / resolved before merging this?

shiftkey:
@thomaseizinger no, just wanted to ensure this was a reasonably safe change

PR #1770, opened 2015-11-29
.gitignore for shopware

itinance:
This contribution provides a full-featured .gitignore for shopware-projects based on Shopware 5 (www.shopware.de)

### DIFF
--- a/Shopware.gitignore
+++ b/Shopware.gitignore
@@ -1,0 +1,48 @@
+# PhpStorm-Project
+.idea/
+
+# Cache and logs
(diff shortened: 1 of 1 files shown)


shiftkey:
I'm going to close this out because the value of maintaining a template like this isn't clear to me, as well as how the rules duplicate with existing templates.

@itinance if you feel like this should still be a template listed here, please read the section about [specialized templates](https://github.com/github/gitignore#specialized-templates) and open a new pull request that follows the new pattern.

PR #1771, opened 2015-11-29
remove harmful line from PlayFramework.gitignore

Dveim:
'project/project' is valid directory for 'project' project-related files. For example, if one configure his sbt build to use scala files directly from 'project/', dependencies like 'project/project/plugins.sbt' will be ignored because of this line.
See http://www.scala-sbt.org/0.13/tutorial/Organizing-Build.html#sbt+is+recursive.


shiftkey:
:thumbsup:

PR #1772, opened 2015-11-29
Update Qt.gitignore

tsteinholz:
Qt generates build files for me that are `build-Projname-WhateverWhatever` so this change continues to support *-build-* folders as well as others.


shiftkey:
:thumbsup:

PR #1773, opened 2015-11-30
Incorrect Dir Path

julien731:
When generating the ignore rules for Composer, an incorrect format is used. The vendor directory added is written `vendor/`, which causes other `vendors` directories to be ignored. The correct format to use is `/vendor/`, which would only ignore the root `vendor` directory.

### DIFF
--- a/Composer.gitignore
+++ b/Composer.gitignore
@@ -1,5 +1,5 @@
 composer.phar
-vendor/
+/vendor/
 
 # Commit your application's lock file http://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file
 # You may choose to ignore a library lock file http://getcomposer.org/doc/02-libraries.md#lock-file


shiftkey:
@julien731 do you have an external link discussing this project structure, so that others can refer to it in the future?

julien731:
Well, I don't have a link to a particular discussion no. This is mostly based on having to edit the `.gitignore` over an over after realizing various dependencies that should be tracked aren't ;)

The closest I can find to a discussion about that is: https://laracasts.com/discuss/channels/general-discussion/gitignore-what-should-be-ignored

Also, as seen on the Composer docs: https://getcomposer.org/doc/faqs/should-i-commit-the-dependencies-in-my-vendor-directory.md

This is not exactly a justification for this particular point, but we can see in #4 that the structure used to track the `vendors` directory **but** ignore the `.git` directories is `/vendor/**/.git`, which shows that Composer dependencies should be considered to be at the root of the project.

shiftkey:
@julien731 those links are fine, thanks for the extra info!

PR #1776, opened 2015-11-30
Add Yii2 gitignore

niks999:
gitignore for [Yii Framework 2.0](http://www.yiiframework.com/) taken from the [official application template](https://github.com/yiisoft/yii2-app-advanced)

### DIFF
--- a/Yii2.gitignore
+++ b/Yii2.gitignore
@@ -1,0 +1,30 @@
+# yii console command
+/yii
+
+# phpstorm project files
+.idea
+
+# netbeans project files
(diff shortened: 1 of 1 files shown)


shiftkey:
I'm going to close this out because the value of maintaining a template like this isn't clear to me, as well as the numerous rules that duplicate existing templates.

@niks999 if you feel like this should still be a template listed here, please read the section about [specialized templates](https://github.com/github/gitignore#specialized-templates) and open a new pull request that follows the new pattern.

PR #1777, opened 2015-11-30
easybook gitignore

juaalta:
gitbook for [easybook, book publishing as easy as it should be](http://easybook-project.org/) taken from the [documentantion](http://easybook-project.org/documentation/chapter-1/publishing-the-first-book.html).

### DIFF
--- a/easybook.gitignore
+++ b/easybook.gitignore
@@ -1,0 +1,2 @@
+# Destination directory of the various publications
+Output


shiftkey:
@juaalta thanks for the contribution! Unfortunately this seems like a minor template and doesn't feel like a good fit for this repository.

PR #1778, opened 2015-12-01
Add 'opendb' file to VisualStudio.gitignore

anurse:
Looks like Visual Studio has started creating a "[Solution].VC.opendb" file when you open a solution with C++ projects. This should be added to the gitignore for VS.


shiftkey:
Seems legit.

PR #1779, opened 2015-12-02
Update OpenCart.gitignore

ffsantos92:
Ignore storage folder (contains logs, cache, uploads, etc.)

This change will make .gitignore compatible with all OpenCart versions.


shiftkey:
Thanks!

PR #1780, opened 2015-12-03
add '.DS_Store'

fanpyi:


shiftkey:
Thanks, but we're using the `Global\OSX.gitignore` to track OS-specific rules.

fanpyi:
ok， Thanks

PR #1781, opened 2015-12-03
Fixes for Magento

mkutyba:
According to #1424


shiftkey:
@mkutyba thanks for the update!

PR #1782, opened 2015-12-03
Update JetBrains.gitignore

breandan:
Gradle library exclusion should be more specific.

https://intellij-support.jetbrains.com/hc/en-us/articles/206827587-How-to-manage-projects-under-Version-Control-Systems

### DIFF
--- a/Global/JetBrains.gitignore
+++ b/Global/JetBrains.gitignore
@@ -21,7 +21,7 @@
 
 # Gradle:
 # .idea/gradle.xml
-# .idea/libraries
+# .idea/libraries/Gradle__*.xml
 
 # Mongo Explorer plugin:
 # .idea/mongoSettings.xml


Global/JetBrains.gitignore
```diff
@@ -21,7 +21,7 @@
 
 # Gradle:
 # .idea/gradle.xml
-# .idea/libraries
+# .idea/libraries/Gradle__*.xml
```
shiftkey: I gather we're talking about this rule:

> XML files under **.idea/libraries** in case they are **generated from Gradle** project

Just wanted to confirm this is the most suitable expression here (me not being familiar with Gradle and how it's generating these files).

breandan:
Correct. The files IntelliJ IDEA generates for Gradle dependencies all start with the prefix `.idea/libraries/Gradle__` and end with `.xml`. However there may be other dependencies *not* managed by Gradle under the same `.idea/libraries/` directory.
image

shiftkey:
Thanks for the extra information. Unfortunately the changes I just merged in #1475 now cause a conflict here. Are you able to resolve this so I can merge in this change?

shiftkey:
Closing due to inactivity and conflicted state. @breandan feel free to resubmit.

cc @citizenmatt - not sure if you wanted to throw in your 2c about this (or find someone on your side to :thumbsup: this change)

citizenmatt:
Just spoken to the dev responsible for Gradle in IntelliJ, and the pattern should be:

```
.idea/libraries/Gradle__*.xml
.idea/artifacts/Gradle__*.xml
```

The first pattern is for IntelliJ Gradle projects, and the second is for enterprise/web Gradle projects.

PR #1783, opened 2015-12-04
Visual studio new database engine

vladon:
New database engine (VS 2015) files


VisualStudio.gitignore
```diff
@@ -80,6 +80,8 @@ ipch/
 *.opensdf
 *.sdf
 *.cachefile
+*.VC.db
+*.VC.VC.opendb
```
shiftkey: We've defined `*.opendb` earlier in the file - should that be more specific? Or is the general rule good enough?

shiftkey: See #1778 for that discussion cc @anurse

VisualStudio.gitignore
```diff
@@ -80,6 +80,8 @@ ipch/
 *.opensdf
 *.sdf
 *.cachefile
+*.VC.db
```
shiftkey: Do you have an external link which explains the addition of this new file? (and perhaps why it's not under `.vs/` :grin:)

VisualStudio.gitignore
vladon: If we can specify a special case, then we must do it instead of common case. The extension is always `.VC.VC.opendb` for the case of new database engine.

VisualStudio.gitignore
vladon: The new VS 2015 experimental feature is described [here](http://blogs.msdn.com/b/vcblog/archive/2015/11/11/new-improved-and-faster-database-engine.aspx), but this does not contain any technical specs (may be because it is still experimental?).

VisualStudio.gitignore
anurse: It seems OK to me to shift to `*.VC.db` and `*.VC.opendb`, I do indeed see that the file always has the `VC` suffix.

I'd like to see if we can find anyone from the VC++ team to help. I can try to put out feelers, but I'm not sure I have the contacts :). Worst case, I'll just email Satya :trollface:

VisualStudio.gitignore
anurse: > and perhaps why it's not under .vs/

\*sigh\* yeah, I'd hoped that the `.vs/` directory would end the proliferation of files...

VisualStudio.gitignore
anurse: Ok, I think I've reached out to the right people, but I kinda just sent an email to somewhere random and am hoping it works ;).

shiftkey: @anurse thanks for reaching out

anurse: Confirmation. `*.VC.db` and `*.VC.opendb` should be correct. Also, we can remove the `*.opensdf` entry I added. @vladon can you add that to this PR?

VisualStudio.gitignore
anurse: Apparently they are planning to move them to `.vs` soon, so, y'know

image

VisualStudio.gitignore
vladon: @anurse Why to remove `*.opensdf`? It is still used (for old database engine in VS 2015). And temporary file with `*.VC.VC.opendb` extension (yes, two subsequent `VC`) is used when solution is opened. I think that all of them must stay in the gitignore for VS.

anurse: Sorry, I meant my earlier `*.opendb` line. You're right, `opensdf` should be left.

shiftkey: Apologies for the delay in getting back to this:

> And temporary file with *.VC.VC.opendb extension (yes, two subsequent VC) is used for new DB engine when solution is opened. 

This should be covered by `*.opendb` on line 79, right?

vladon: @shiftkey If we can be more specific, why not to? It is always `*.VC.VC.opendb` for temporary opendb created by Visual Studio. What if user will create its own `*.opendb` which must be in his repo?

Haacked: @vladon @shiftkey's point is that we _already_ have `*.opendb` in this file. So the addition of `*.VC.VC.opendb` does nothing. Either we remove `*.VC.VC.opendb` or we remove `*.opendb`.

@anurse was the one who added `.opendb` in a393ef8de999946cbb0210c24bb8ba39ffb511c0 so I'm not sure what the best call here is.

vladon: @Haacked My point is that VS 2015 only uses `<ProjectName>.VC.opendb` and `<ProjectName>.VC.VC.opendb` (latter only temporarily while the project is open in IDE) for its new databases. Adding `*.opendb` we simply ignoring *all* files with that extension, it can be user files too. I think it will be best to add `*.VC.VC.opendb` plus `*.VC.opendb` and remove `*.opendb` line.

Haacked: If we add `*.VC.opendb` we won't need to add `*.VC.VC.opendb`. @anurse any objection to removing `*.opendb`?

vladon:
Guys, please, add this or any other PR with *.VC.db line. There is VS 2015 Update 2 released yesterday with this database format by default.

shiftkey:
@vladon apologies for the delay, and thanks for the reminder!

PR #1785, opened 2015-12-04
Add Swift Package Manager to Swift.gitignore

natestedman:
This adds lines relevant to the [Swift Package Manager](https://github.com/apple/swift-package-manager) to `Swift.gitignore` - the directories `Packages` and `.build`. This is the new, official package manager for Swift code.

Matches the style used for Carthage - build directory disabled by default, checkouts directory (`Packages`) optionally disabled. This matches what the [official examples do](https://github.com/apple/example-package-dealer/blob/master/.gitignore), although actually *running* `swift build` will leave a dirty working tree, since they haven't checked in `Packages` already.

### DIFF
--- a/Swift.gitignore
+++ b/Swift.gitignore
@@ -27,6 +27,12 @@ xcuserdata
 *.hmap
 *.ipa
 
+# Swift Package Manager
+#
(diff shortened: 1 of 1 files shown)


shiftkey:
Seems legit - thanks for pointing to the official examples!

PR #1787, opened 2015-12-04
Script for adding all files to global config

theconsummate:
This script adds all gitignore files to the global config of git.


shiftkey:
Thanks, but I'd rather keep this repository focused on the templates. Neat idea though!

PR #1788, opened 2015-12-05
Updated to Joomla 3.4.5

camigreen:
Updated the ignores needed for Joomla 3.4.5.  

http://www.joomla.org/3/en/

Updated because the last update was 10 months ago and the last version was 3.3.6.


shiftkey:
Thanks!

PR #1789, opened 2015-12-06
Update to include iMPACT and Core Generator files

stephengroat:
I know the Xilinx ISE has been deprecated, but I still work on old Virtexs and these files really should be included in the ignores.

iMPACT is the JTAG programming utility for the ISE and Core Generator is the proprietary core generation tool.


shiftkey:
Seems reasonable. Thanks!

PR #1791, opened 2015-12-08
Based on a template from jmoody

Oddj0b:
A Ruby UI Automation test framework

[Website](http://calaba.sh/)

[Calabash Documentation](http://developer.xamarin.com/guides/testcloud/calabash/)


shiftkey:
Given this seems to be rules for a tool, rather than a language, I feel this should live under `Global` - which is where we put our IDE or OS-specific templates, so that others can add them to existing projects alongside whatever language they have chosen.

Does that seem reasonable?

Oddj0b:
@shiftkey Yes it does, sorry for misunderstanding the project structure.

shiftkey:
Thanks!

PR #1795, opened 2015-12-09
Add RubyMine specific file to Ruby.gitignore

ferbsx:
[RubyMine](https://www.jetbrains.com/ruby/) is a Ruby and Rails IDE, used very commonly in companies, which automatically makes this file, that you'd want to ignore :smile:

### DIFF
--- a/Ruby.gitignore
+++ b/Ruby.gitignore
@@ -15,6 +15,9 @@
 .repl_history
 build/
 
+# Specific to RubyMine:
+/.idea
+
 ## Documentation cache and generated files:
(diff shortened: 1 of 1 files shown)


shiftkey:
Thanks, but this is tracked in `Global/Jetbrains.gitignore` along with all the other IDEs that JetBrains ship. 

This file is intended to be IDE-agnostic.

ferbsx:
Oh, Great! Should've looked into that :smile: .. Anyway thanks for the quick response :hibiscus:

PR #1796, opened 2015-12-09
Racket ignore files

mpkh:
`.bak` as temp files in root and `compiled` as default build folder


shiftkey:
Closing in favour of #1616

PR #1797, opened 2015-12-10
Unity.gitignore modyfying, 

tim-hub:
add `/[Bb]uilds/` and  `/Assets/AssetStoreTools*` `ExportedObj/` 3 dictionaries and `*.svd` svd files to be ignored

	modified:   Unity.gitignore

- `Builds`: Some one like to use `builds` folder not `build` folder to hold builds
- `AssetStoreTools`: Some developers install this to develop asset for unity asset store, this should be ignored
- `ExportedObj` and `svd` files are from VS


shiftkey:
Thanks!

PR #1798, opened 2015-12-11
Tex: Ignore files generated by pdfcomment package

JelteF:
The pdfcomment package generates *.upa and *.upb files when compiling LaTeX files. These should be ignored as well.


shiftkey:
Thanks!

PR #1799, opened 2015-12-13
Added .DS_Store to Objective c and swift gitignorefiles

Johnykutty:
Added Mac OS X settings file .DS_Store to Objective c and swift gitignore files


shiftkey:
With Swift no longer being OS X only, I'd prefer we keep these OS-specific rules in `Global/OSX.gitignore`

PR #1800, opened 2015-12-13
Bugfix - Typos in comments

jutaz:
This PR fixes few typos found in comments.

> Please only modify one template per pull request. This helps keep pull requests and feedback focused on a specific project or technology.

Considering that this PR does not actually touch **any** ignore paths, I think it is OK for this to modify just few of these files.


shiftkey:
This is **phenomenal** stuff @jutaz. Thanks for taking the time to find these and fix them up!

PR #1803, opened 2015-12-14
Ignore BundleArtifacts folder

belyansky:
Visual Studio creates folder named as `BundleArtifacts` during generation of app package for Windows Store. This folder can be ignored.


shiftkey:
Seems legit. Thanks!

PR #1804, opened 2015-12-16
Added OS generated files.

nikhilthombare:
Added OS generated files in gitignore file for Ruby.


shiftkey:
Thanks, but we use `Global/OSX.gitignore` to track these files, to help consumers compose their overall `.gitignore` files...

PR #1805, opened 2015-12-16
Python

bjut-hz:
Python IDE Pycharm creates .idea/ directory automatically when you use it to write code. So I add .idea/ item in Python.gitignore.


bjut-hz:
I think any IDE configuration file should not be added in VCS. Everyone in a team may use different one.

shiftkey:
> Everyone in a team may use different one.

This is why we have all the JetBrains-specific rules under `Global/JetBrains.gitignore`

bjut-hz:
OK, Thank you for reading.

PR #1808, opened 2015-12-21
Ignore auxiliary files from \usepackage{cprotect}

habi:
[cprotect](https://www.ctan.org/tex-archive/macros/latex/contrib/cprotect?lang=en) tracks its contents in an auxiliary file, this is safe to ignore.


shiftkey:
Thanks!

PR #1809, opened 2015-12-24
Create C#.gitignore

Jarvin-Guan:
gitignore file in C#

### DIFF
--- a/C#.gitignore
+++ b/C#.gitignore
@@ -1,0 +1,19 @@
+# User-specific files
+*.suo
+
+# Publish Web Output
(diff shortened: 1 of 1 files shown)


shiftkey:
We merged the `VisualBasic` and `CSharp` ignore templates (because they were so IDE-specific and so very similar) into a single `VisualStudio.gitignore`. Check it out!

Jarvin-Guan:
OK,Thank you

Jarvin-Guan:
OK,Thank you

PR #1810, opened 2015-12-24
Ignore ncrunch project and solution files

dariusdamalakas:
This ignores [NCrunch](http://www.ncrunch.net/) solution and project files. These files are auto-generated, and IMO should not be part of solution. Without these files, solution will still load, but NCrunch will be disabled.

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -111,6 +111,8 @@ _TeamCity*
 _NCrunch_*
 .*crunch*.local.xml
 nCrunchTemp_*
+*.ncrunchproject
+*.ncrunchsolution
 
 # MightyMoose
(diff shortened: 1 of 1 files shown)


shiftkey:
Someone raised this in #1766, and I defer to the [official documentation](http://www.ncrunch.net/documentation/considerations-and-constraints_using-ncrunch-with-source-control) for NCrunch:

> It's recommended that the last two files listed in the above table (.v2.ncrunchsolution and .v2.ncrunchproject) be stored inside a source control system to be shared with team members.

dariusdamalakas:
Thanks, I was not aware of that recommendation.  I'll keep ignoring those files manually then.

PR #1811, opened 2015-12-25
KiCAD: ignore more temp files and generated data

Caerbannog:
- New temporary file patterns : `*~`, `_autosave-*`, `*.tmp`, `.DS_Store`

- New *Bill Of Materials* file extensions : `.xml`, directly generated from the schematics files (`.sch`, not in gitignore). Then official scripts like *bom2csv.xsl* generate the actual BOM as `.csv`.
Doc: http://docs.kicad-pcb.org/en/eeschema.html#creating-customized-netlists-and-bom-files

### DIFF
--- a/KiCad.gitignore
+++ b/KiCad.gitignore
@@ -18,6 +18,3 @@ _autosave-*
 # Exported BOM files
 *.xml
 *.csv
-
-# Other system files
-.DS_Store
--- a/KiCad.gitignore
+++ b/KiCad.gitignore
@@ -5,9 +5,19 @@
 *.bak
 *.bck
 *.kicad_pcb-bak
+*~
+_autosave-*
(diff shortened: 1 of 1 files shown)


KiCad.gitignore
```diff
@@ -5,9 +5,19 @@
...
 
 # Autorouter files (exported from Pcbnew)
 .dsn
+
+# Exported BOM files
+*.xml
+*.csv
+
+# Other system files
+.DS_Store
```
shiftkey: This is an OSX specific rule, which we track in `Global/OSX.gitignore` alongside others. Can you drop this?

Caerbannog:
@shiftkey : done.

shiftkey:
Thanks!

PR #1814, opened 2015-12-27
Remove `build/` suppression

AArnott:
The authoring of NuGet packages quite often include `build` folders, and when this is .gitignore'd it's very often the source of bugs in files that don't get checked in.

I've never seen a Visual Studio project that builds to a `build` folder (`bin` is the default name). As this is a Visual Studio template file, and we have real projects that include `build` folders that include source code, I recommend we remove suppression of this folder.

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -17,7 +17,6 @@
 [Rr]eleases/
 x64/
 x86/
-build/
 bld/
 [Bb]in/
 [Oo]bj/


AArnott:
CC: @kovalikp, @terrajobst

jchannon:
Agree, I just came across this. 

Also there seems to be some consensus that we will have this folder going forwards https://gist.github.com/davidfowl/ed7564297c61fe9ab814

terrajobst:
I agree with @AArnott. I used to use `build` as the folder for my build output, but that was before I was exposed to more build trees. Virtually all trees use `bin`.

shiftkey:
@AArnott @jchannon thanks for the additional :thumbsup:s

bbodenmiller:
What about @rmunn's comments at https://github.com/joeblau/gitignore.io/issues/234 about `build/` being used with F#?

AArnott:
He mentions that it's not in the fsharp .gitignore file either. given the interest is specific to F#, I recommend he send a PR to add it to that file.

PR #1815, opened 2015-12-28
Ignore ApplicationInsights.config for Microsoft Azure

anderslundsgard:
The feature Application Insights in Microsoft Azure is cool but the config file ApplicationInsights.config should not be in version control.

According to https://azure.microsoft.com/sv-se/blog/upcoming-name-change-for-windows-azure/ Windows Azure is also renamed to Microsoft Azure.

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -160,6 +160,9 @@ csx/
 ecf/
 rcf/
 
+# Windows Azure ApplicationInsights config file
+ApplicationInsights.config
+
 # Windows Store app package directory
 AppPackages/
 BundleArtifacts/
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -152,15 +152,15 @@ publish/
 # Uncomment if necessary however generally it will be regenerated when needed
 #!**/packages/repositories.config
 
-# Windows Azure Build Output
+# Microsoft Azure Build Output
 csx/
 *.build.csdef
 
-# Windows Azure Emulator
+# Microsoft Azure Emulator
 ecf/
 rcf/
 
-# Windows Azure ApplicationInsights config file
+# Microsoft Azure ApplicationInsights config file
 ApplicationInsights.config
 
 # Windows Store app package directory


shiftkey:
@anderslundsgard do you have any links to documentation around this, so that others can understand this requirement?

anderslundsgard:
Yes, here is a good introduction to ApplicationInsights http://docs.asp.net/en/latest/fundamentals/application-insights.html
The docs don't explicit tells that you should not share the config file but I guess others like me don't like to share the <InstrumentationKey> that is found in ApplicationInsights.config.

shiftkey:
@anderslundsgard and for reference, can you share what sort of stuff is in your created `ApplicationInsights.config` file (with the contents sanitized wherever neessary)? I'm not really familiar with AppInsights, so it's mostly for my benefit :grin:

It might also be worth submitting a PR to those docs to explain the `.config` file further - and that it can be safely ignored - but I won't hold this up further.

cc @ardalis for a :thumbsup: on this PR

anderslundsgard:
The element I got stuck on was the InstrumentationKey GUID that is your unique key to send statistics of site usage to Azure.
```
<ApplicationInsights xmlns="http://schemas.microsoft.com/ApplicationInsights/2013/Settings">
    ...
    <InstrumentationKey>xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx</InstrumentationKey>
</ApplicationInsights>
```

shiftkey:
@anderslundsgard thanks, I'll get this in and if others decided they have differing opinions we can revisit

anderslundsgard:
Great, thanks!

gitFurious:
This change breaks [continuous deployment from github through the Azure portal](https://azure.microsoft.com/en-us/documentation/articles/web-sites-publish-source-control/#Step75).

InstrumentationKey isn't a secret. In fact, [it's used on client side web pages](https://www.visualstudio.com/en-us/get-started/insights/app-insights-asp-net).

anderslundsgard:
Ok, I have not experienced that (yet) but then i totally agree with you @gitFurious.

shiftkey:
@gitFurious thanks for bringing this to our attention. I'd gladly take a patch to revert this before the damage goes any further...

PR #1816, opened 2015-12-31
Update Eclipse.gitignore

rosolkodv:
Remove .gradle because it is not eclipse specific and already added in Gradle.gitignore.


shiftkey:
Seems legit :thumbsup:

PR #1818, opened 2016-01-03
Add TeX precompiled preamble and xypic

dopefishh:
A precompiled header(`latex -ini`) compiled creates a `.fmt` file.
To speed up xypic's drawing, precompiled matrices are created with the `\CompileMatrices` entry in the preamble. This creates `*.xyc` files.


shiftkey:
Thanks!

PR #1820, opened 2016-01-09
ignore intellij idea folder

regadas:


shiftkey:
We track the JetBrains-specific rules in [Global/JetBrains.gitignore](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore) as these are IDE-specific and can be associated with a number of different languages.

PR #1822, opened 2016-01-10
Exclude Latexian temp files for Tex documents

rogierslag:
Latexian adds some additional temporary files, which are not required for building the project. Therefore they are better left out as well

The project itself is no longer maintained by TacoSW, however the application is still widely used.

### DIFF
--- a/TeX.gitignore
+++ b/TeX.gitignore
@@ -154,3 +154,6 @@ pythontex-files-*/
 # endfloat
 *.ttt
 *.fff
+
+# Latexian
+TSWLatexianTemp*


shiftkey:
This seems specific enough, and I can't really find any [supporting docs](http://tacosw.com/latexian/support.php) that might explain it further - @rogierslag do you know of any resources so we can link to them here?

rogierslag:
I couldnt find anything in the docs about this either, but a quick [Google Search](https://www.google.nl/search?q=latexian+TSWLatexianTemp) mostly found other `.gitignore` files with this exclusion

shiftkey:
@rogierslag :+1:

PR #1823, opened 2016-01-13
Add Android Studio generated files

MrOutis:
Everything in .idea/ and all .iml files should not be included in git. They
are generated by Android Studio when opening a Gradle project.  They will
be different for each person's Android Studio, so if they are in git, it'll
be an endless source of git conflicts.

### DIFF
--- a/Android.gitignore
+++ b/Android.gitignore
@@ -30,3 +30,7 @@ proguard/
 
 # Android Studio captures folder
 captures/
+
+# Android Studio generated files
+.idea/
+*.iml


lostunknow:
Pilla

shiftkey:
For reference, we're using `Global/JetBrains.gitiginore` to track the rules for all their IDEs and there's some nuances that are mentioned in #1475 which reflect the official documentation for IntelliJ.

MrOutis:
Thanks for the feedback, guys

PR #1825, opened 2016-01-15
add IntellijIDEA.gitignore

lingyfh:


shiftkey:
See https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore

PR #1828, opened 2016-01-18
Added gitignore for Django

yask123:
See this: http://stackoverflow.com/a/3719679/5752691

### DIFF
--- a/Django.gitignore
+++ b/Django.gitignore
@@ -1,0 +1,10 @@
+.installed.cfg
+bin
+develop-eggs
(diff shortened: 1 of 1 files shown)


yask123:
Closing this,  as I think [this] (https://github.com/github/gitignore/blob/master/Python.gitignore) would take care of it.

shiftkey:
@yask123 there's a section for Django-based rules in that file, feel free to add anything there that's not already covered by the other sections...

PR #1830, opened 2016-01-21
Rename VisualStudio.gitignore to Global/VisualStudio.gitignore

jchannon:
This moves the VS gitignore to the Global folder where the other editor gitignore files are


shiftkey:
I'm not sure how I feel about this:

- this came about because it was a bunch of :shit: work maintaining `CSharp` and `VisualBasic` templates - see #492 - and having them be out-of-sync
- it still trips a few people up, who then go and try and submit a new `CSharp` ignore file for example, but on the most part things are pretty manageable
- yes, the file is now not named after a language, so moving this into `Global` is technically the right thing to do
- I fear taking that change means more :shit: work for me to tell people that it exists and lives in a different spot, because lots of people don't understand the `Global` thing and apply IDE-specific rules to existing language templates
- all of this is unrelated to how things appear in the GitHub gitignore lookup - that's an external process

jchannon:
No worries. Just saw it was technically in the wrong place. Happy to close
the PR.

Good use of of [image: :shit:] 😃 (ooh look, shits and giggles)

PR #1834, opened 2016-01-23
Add Visual Studio Code project settings folder

AlexZeitler:
Ignores individual [Visual Studio Code](https://code.visualstudio.com/) project settings like task configuration and debug configuration.

Why: you might want not to share your personal VS Code project settings.

Description of Settings for VS Code: https://code.visualstudio.com/Docs/customization/userandworkspace

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -234,3 +234,6 @@ _Pvt_Extensions
 
 # FAKE - F# Make
 .fake/
+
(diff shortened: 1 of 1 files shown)


shiftkey:
Thanks, but we've spun out a new [`Global/VisualStudioCode.gitignore`](https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore) to capture these...

PR #1835, opened 2016-01-23
Add .byebug_history to gitignore

pisaacs:
.byebug_history included in default Rails5 setup (https://github.com/rails/rails/blob/master/railties/lib/rails/generators/rails/app/templates/gitignore)


shiftkey:
Thanks!

PR #1836, opened 2016-01-25
Qt: ignore CMakeLists.txt.user

varhub:
Qt.ignore already has QtCreator specific lines and Qt's cache files (named *.user).
Therefore, ignore CMakeLists.txt.user created by QtCreator seems innocuous  (same as .pro.user but for cmake based proyects).


shiftkey:
Thanks!

varhub:
Thanks to you for accept it

PR #1499, opened 2015-05-02
OSX.gitignore additions

Sduibek:
Added __MACOSX folder and missing descriptions.

PR #1837, opened 2016-01-27
Add folder attributes

vgturtle127:
Added some folder attributes from a previous PR.

@shiftkey

### DIFF
--- a/Global/OSX.gitignore
+++ b/Global/OSX.gitignore
@@ -1,8 +1,10 @@
 # Folder attributes
-__MACOSX/
 .DS_Store
 .AppleDouble
 .LSOverride
+
+# Resource Forks
+__MACOSX/
 
 # Icon must end with two \r
 Icon


vgturtle127:
@shiftkey Actually, I can see some from PR #1499 that we could add. Should I?

shiftkey:
Ugh, this still looks like it's changing the line endings :sob:

vgturtle127:
@shiftkey Oh, crap! I don't have any global configs, and I used the GitHub editor!

Should I try locally?

shiftkey:
@vgturtle127 I guess that's the only option then :(

vgturtle127:
@shiftkey What are the line endings that won't mess it up, and should I close this and start over? :laughing:

shiftkey:
I think I need to review this change outside of the PR view, something doesn't feel right...

vgturtle127:
@shiftkey I know I suggested this earlier, and I think you said something about it, but why couldn't we add a .gitattributes to the repo? That would automatically normalize all the files to the native OS encodings...

vgturtle127:
@shiftkey Actually, never mind. I just messed another project up by adding the .gitattributes file...

PR #1838, opened 2016-01-27
exclude intellij files

NTCoding:


shiftkey:
Thanks but see [Global/JetBrains.gitignore](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore) for the IntelliJ rules and #1475 for some updates which target the latest release of IntelliJ.

PR #1840, opened 2016-01-29
Create Magento2.gitignore

rafaelstz:


shiftkey:
@rafaelstz I'd rather not introduce a new template that differs from an existing template only by version - are we able to include this changes in the original `Magento` template?

rafaelstz:
Magento 1 doesn't was discontinued and is a totally different platform, not an update (e.g. from Wordpress 3 to 4), has many people that download and use this version yet.
To take my conclusion I used the static files and folders that are generated in each install and not the Magento 2 Core files.

shiftkey:
@rafaelstz apologies for the delay in replying to this.

In #2857 I proposed a pattern where the latest supported version of a framework lives in the main project, and older versions are archived in the `community/` subdirectory.

Please submit any improvements to Magento v2 to the main file, or move it to a `community/PHP/Magento2.gitignore` if it needs to be updated to a v3 version.

PR #1841, opened 2016-02-01
Update Haskell.gitignore

fabianhjr:


Haskell.gitignore
```diff
@@ -1,16 +1,23 @@
-dist
-cabal-dev
+*
+!*.*
+!*/
+### Above combination will ignore all files without extension ###
```
shiftkey: Can you tell me a bit why this should be the default behaviour?

fabianhjr:
When working with several different Haskell Source Codes, several executable are created; Executables are the only files that commonly do not have an extension. (ELF/Mach-O)

Though on second thought I have been hasty and forgotten about the no extension HACK/CONTRIBUTE/README/et al.

I apologize.

PR #1842, opened 2016-02-02
add pycharm projecct file, '.idea'

green23:
When using pycharm, it will have a folder named '.idea'.


shiftkey:
Thanks, but we keep all of the JetBrains IDE rules under `Global/JetBrains.gitignore` - and there's some recent changes to what's important to exclude in `.idea` - see #1475 for more info.

PR #1844, opened 2016-02-03
update Eagle.gitignore for .lck

icylogic:
details:
http://www.cadsoftusa.com/eagle-pcb-design-software/software-version-news/software-version-news-in-v7/
http://www.cadsoftusa.com/fileadmin/journalist/Documents/Release_Notes_v7.5_en.pdf

### DIFF
--- a/Eagle.gitignore
+++ b/Eagle.gitignore
@@ -38,3 +38,7 @@ eagle.epf
 *.info
 
 *.eps
+
+# file locks introduced since 7.x
+*.lck
+


shiftkey:
Do you have a link that explains what these `*.lck` files represent and why they can be ignored safely?

stephengroat:
An ``.lck`` is a type of file lock, implemented by EAGLE. When a lock exists, the following message(s) show up: http://www.ke5fx.com/lock_bugs.png

shiftkey:
@stephengroat thanks!

PR #1845, opened 2016-02-05
Do not ignore gradle wrapper's jar file

tokuhirom:
It's required for bootstrapping.

### DIFF
--- a/Java.gitignore
+++ b/Java.gitignore
@@ -7,6 +7,7 @@
 *.jar
 *.war
 *.ear
+!gradle/wrapper/gradle-wrapper.jar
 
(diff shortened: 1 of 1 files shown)


vorburger:
+1 this would be good. It also goes naturally together with my https://github.com/github/gitignore/pull/1977 for a better .gitgnore for Java Gradle projects.

drothmaler:
-1 there already is a [Gradle.gitignore](https://github.com/github/gitignore/blob/master/Gradle.gitignore) for this purposes. But it might be nice to make the `gradle-wrapper.jar` there a bit more specific.

shiftkey:
Closing due to inactivity, pending feedback and the existing `Gradle` template.

PR #1846, opened 2016-02-06
matlab: ignore octave session info

mcandre:


shiftkey:
@mcandre do you have a link to details about this product or why this rule is helpful to include, for those who want to get more information?

mcandre:
This rule is helpful to avoid checking in junk files in version-controlled matlab/octave projects.

shiftkey:
@mcandre ok

PR #1847, opened 2016-02-07
Remove redundancy in Global/Vim.gitignore

nloveladyallen:
Anything matched by `*.un~` will also be matched by the last rule, `*~`.


shiftkey:
Indeed it will. Thanks!

PR #1851, opened 2016-02-09
Add PyCharm

leegom:
PyCharm (A Python IDE): https://www.jetbrains.com/pycharm/
Add PyCharm into the global gitignores directory.


shiftkey:
Thanks, but we're using `Global/JetBrains.gitiginore` to track the rules for all their IDEs.

We also refreshed it recently in #1475 to reflect the official documentation.

PR #1854, opened 2016-02-10
Added Library Mappings and VCS Root 

RobertoNovelo:
These two files are automatically modified each time the project is opened (specially in different OSs), and do not affect the project if ignored.


shiftkey:
Thanks!

PR #1855, opened 2016-02-11
updated system cache path

aamnah:
The cache folder paths are different in OpenCart 2.x

### DIFF
--- a/OpenCart.gitignore
+++ b/OpenCart.gitignore
@@ -7,7 +7,7 @@ admin/config.php
 download/
 image/data/
 image/cache/
-system/cache/
+system/storage/cache/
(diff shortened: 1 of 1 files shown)


shiftkey:
@aamnah is there any backwards-compatibility concerns here? Or would you expect most new repositories to be running OpenCart 2.x?

shiftkey:
There's also some other changes in #1759 which overlap here. Are those of interest to you here?

aamnah:
I expect most new repositories to be running OpenCart 2.x :smirk: But that's naive i guess. Please include #1759 , it includes more paths than mine. :+1:

PR #1856, opened 2016-02-11
work around intellij gradle wrapper refreshing bug

mcandre:
Work around https://youtrack.jetbrains.com/issue/IDEA-116898 by offering to gitignore the gradle wrapper once it's safely in version control.


shiftkey:
Thanks for making it opt-in - and for linking to the original issue!

leventov:
Probably should be reverted because https://youtrack.jetbrains.com/issue/IDEA-116898 is now fixed

PR #1858, opened 2016-02-12
Add Flask instance folder

Harrison-G:
Contains private configuration. Equivalent to Django's local_settings.py.

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -53,6 +53,9 @@ coverage.xml
 *.log
 local_settings.py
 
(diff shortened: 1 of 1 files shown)


shiftkey:
Relevant docs: http://flask.pocoo.org/docs/0.10/config/#instance-folders

> The instance folder is designed to not be under version control and be deployment specific. It’s the perfect place to drop things that either change at runtime or configuration files.

shiftkey:
Thanks @Harrison-G!

PR #1859, opened 2016-02-12
Fix capitalization of KiCad

Harrison-G:
According to their website, the correct capitalization is KiCad, not KiCAD.


shiftkey:
Thanks!

PR #1861, opened 2016-02-13
Update Android.gitignore

ChArslan:


shiftkey:
Can you tell me a bit about why this rule should be added, and if there's any supporting links or documentation that we can point curious users to?

ChArslan:
All resources file used in Android with .xml extension like make a layout for your App 
[](http://www.tutorialspoint.com/android/android_resources.htm) @shiftkey

ChArslan:
[](http://www.tutorialspoint.com/android/android_resources.htm)

shiftkey:
@ChArslan I'm still not clear on the files related to this change - are there files named `.xml` (no wildcard, no specific name) which should be ignored in an Android project?

rafaelloab:
It doesn't make sense, there are files like AndroidManifest.xml and many other xml files related to the View component of the app and they are essential. Hardly someone would not versionate those files.

shiftkey:
@ChArslan I'm still concerned with the same things that @rafaelloab has raised - XML files should be version controlled in most scenarios. If there are scenarios were an XML file can be safely ignored, it should be a specific rule with a comment to explain why this can be done.

shiftkey:
Closing due to conflicts and inactivity. @ChArslan please have a look at the latest file and see if anything needs to be incorporated.

PR #1862, opened 2016-02-15
Ignore some Windows Store package related files

SaschaNaz:
Ignores _pkginfo.txt and Package.StoreAssociation.xml. Those files can always be regenerated by the IDE.

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -166,9 +166,11 @@ rcf/
 # Microsoft Azure ApplicationInsights config file
 ApplicationInsights.config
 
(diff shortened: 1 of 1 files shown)


shiftkey:
@SaschaNaz 

> Those files can always be regenerated by the IDE.

Is that the recommended behaviour here - rebuild in VS?

SaschaNaz:
@shiftkey No special recommendation but creating an app package automatically rebuild the files.

shiftkey:
@SaschaNaz thanks, that works for me!

PR #1863, opened 2016-02-15
[VisualStudio] ignore visual studio 2010 browser info

heikopanjas:
The file extension for Visual Studio 2010 browser information files (*.bsc) is missing from VisualStudio.gitignore. I've inserted it right after *.sbr as this comes from the same tool.

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -56,6 +56,7 @@ artifacts/
 *.pgd
 *.rsp
 *.sbr
+*.bsc
 *.tlb
 *.tli
 *.tlh


shiftkey:
@heikopanjas do you have any MSDN links that might explain what this tool is for? I'd love to annotate the source here to indicate the two are related - and how they're related.

heikopanjas:
@shiftkey: Sorry, I was wrong when I said both file types come from the same tool. But they are definitely related: On msdn they say that .bsc files (https://msdn.microsoft.com/library/95ws11cy(v=vs.100).aspx) are cross reference databases for C++ symbols. They are created from .sbr files (https://msdn.microsoft.com/en-us/library/z6hhww5f(v=vs.100).aspx).

shiftkey:
@heikopanjas that's okay, and thanks for the extra details.

Do you think it's worth moving these two rules (`*.bsc` and `*.sbr`) down to line 78 - under the `Visual C++ Cache Files` section - where we have the `*. ncb` rule?

heikopanjas:
That would be ok in a way, but .ncb files are created by the IDE. And .sbr/.bsc files are create by the compiler itself. Just like .obj files.

shiftkey:
@heikopanjas I was reading these sentences, which made me feel like they were related (emphasis mine):

> In a C++ project, the .ncb file contains information about the symbols that are used in the project. The .ncb file enables Class View and the Object Browser to locate the definition of a symbol. **However, to support browsing to references of the symbols, you must create a .bsc file for the project.**

I'm not _really_ fussed about the relationship between the IDE and the compiler here, but these all feel relevant to C++ projects and browsing symbols - hence grouping them like that.

heikopanjas:
You are absolutely right. After all, these files are supporting IDE features.

shiftkey:
@heikopanjas are you able to extract those three rules into their own group with a comment? I'd love to document more of these "other" rules for future reference...

heikopanjas:
@shiftkey Sorry for my late reply. I will take the three file extensions and add them to a group named ´´´browser output´´´ with some details. Do you want me to open another pull request?

shiftkey:
@heikopanjas you can do that in this PR

shiftkey:
Closing due to inactivity. @heikopanjas feel free to open a fresh PR if you'd still like to see these changed landed.

PR #1864, opened 2016-02-16
Add "__recovery" folder.

luxidea:


shiftkey:
Citing [this twitter thread(https://twitter.com/andyhtech/status/644217609202868225) to explain what this folder is for:

> Q: We have the _history folder and now we also have __recovery. What's that for?
> A: For ide or system crash, I guess it keeps autosaved copies. When starting up again, detects and restores (same as MsOffice)

So this can be safely ignored.

shiftkey:
@luxidea thanks!

PR #1865, opened 2016-02-18
Adding JetBrains Rider (.NET IDE from JetBrains) specific files

hmemcpy:
This ignores the `.idea` directory that is placed in the root of the solution, and a metadata file `.sln.iml` that is created.


shiftkey:
Not sure a blanket `.idea` rule here fits with [the official docs](https://intellij-support.jetbrains.com/hc/en-us/articles/206544839) but it seems like everyone else just does this.

shiftkey:
Thanks!

PR #1866, opened 2016-02-18
Add a .gitgnore for SourcePawn.

shavitush:
SourcePawn is the scripting language used to compile .sp files, with the outcome of an .smx file, a plugin that SourceMod can run.

- [Introduction to SourcePawn 1.7](https://wiki.alliedmods.net/Introduction_to_SourcePawn_1.7)
- [SourceMod: Half-Life 2 Scripting](http://www.sourcemod.net/)

### DIFF
--- a/SourcePawn.gitignore
+++ b/SourcePawn.gitignore
@@ -1,0 +1,4 @@
+# SourcePawn
+*.dll # extensions
+*.so # extensions
+*.smx # plugins


shiftkey:

@shavitush if you feel like this should still be a template listed here, please read the section about [specialized templates](https://github.com/github/gitignore#specialized-templates) and open a new pull request that follows the new pattern.

PR #1869, opened 2016-02-19
Added a file header to every file to allow for better catting them together

hashhar:
I added file headers like:
```
# ===== filename ===== #
content of file
```
to every file so that operations like,
```
cat /gitignore/Ada.gitignore >> ~/myproj/.gitignore
```
give better looking and more maintainable output.

**I suggest making this a standalone branch of the repo so that people can use whatever they want**


shiftkey:
This seems easy enough to let users tailor for themselves:

```
echo "# ===== Ada.gitignore ===== #" >> ~/myproj/.gitignore
cat /gitignore/Ada.gitignore >> ~/myproj/.gitignore
```

> I suggest making this a standalone branch of the repo so that people can use whatever they want

Maintaining two branches of largely the same content would be a nightmare for me, so I'm :-1:  on this change. And offerings like gitignore.io do a good job of letting you tailor your templates to suit your scenario.

hashhar:
@shiftkey Thanks for your thoughts.

PR #1870, opened 2016-02-22
Update Android.gitignore

hankunpeng:
Add _**.idea**_ folder and  _**.DS_Store**_ file.

### DIFF
--- a/Android.gitignore
+++ b/Android.gitignore
@@ -34,3 +34,7 @@ captures/
 
 # Intellij
 *.iml
+.idea
+
+# Only needed under Mac OS
+.DS_Store


Android.gitignore
```diff
@@ -34,3 +34,7 @@ captures/
 
 # Intellij
 *.iml
+.idea
+
+# Only needed under Mac OS
```
shiftkey: We track this as well as other OSX specific rules under `Global/OSX.gitignore` so things are more composable

Android.gitignore
```diff
@@ -34,3 +34,7 @@ captures/
 
 # Intellij
 *.iml
+.idea
```
shiftkey: I'm also hesitant to add this specific rule as [there's a lot of detail](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore) to the IntelliJ-based IDEs and what they exclude - see https://github.com/github/gitignore/pull/1475 for the discussion about this - and the links to the documentation

shiftkey:

PR #1872, opened 2016-02-22
Cordova folders to ignore

rido-min:
I'm adding some folders to ignore in VS 2015 Cordova Projects

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -21,6 +21,7 @@ bld/
 [Bb]in/
 [Oo]bj/
 [Ll]og/
+[Bb]ld/
 
 # Visual Studio 2015 cache/options directory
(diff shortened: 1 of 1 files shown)


VisualStudio.gitignore
```diff
@@ -21,6 +21,7 @@ bld/
 [Bb]in/
 [Oo]bj/
 [Ll]og/
+[Bb]ld/
```
shiftkey: Is this one also Cordova-specific? I'd love to understand more about where this one is coming from...

rido-min: Yes. Its a temp folder uses by cordova

shiftkey: Can you move it down under the Cordova group then?

PR #1874, opened 2016-02-23
Ignore ncrunch project files

Haacked:
NCrunch will add these to every project including submodules which is annoying. These seem specific to the user so we should ignore these by default.

/cc @shiftkey

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -111,6 +111,7 @@ _TeamCity*
 _NCrunch_*
 .*crunch*.local.xml
 nCrunchTemp_*
+*.ncrunchproject
 
 # MightyMoose
 *.mm.*


shiftkey:
Citation from [the project's documentation](https://www.ncrunch.net/documentation/considerations-and-constraints_using-ncrunch-with-source-control)

> It's recommended that the last two files listed in the above table (_.v2.ncrunchsolution_ and _.v2.ncrunchproject_) be stored inside a source control system to be shared with team members.

Haacked:
Ugh, that's just terrible when working with submodules though. I don't mind the `.ncrunchsolution` file, but the `.ncrunchproject` file litters the submodules making using NCrunch a pain. It assumes the developer has control over every project in the solution.

shiftkey:
This is something I've :boot:ed before because it goes against what the vendor recommends, so I'm hesitant to make it the default behaviour everywhere...

Haacked:
Okie dokie. I'll just go tell the vendor to stop being wrong. :stuck_out_tongue:

PR #1875, opened 2016-02-24
Add *.glsdefs to TeX.gitignore

jutoft:
.glsdefs is generated alongside .gls files when using makeglossaries

### DIFF
--- a/TeX.gitignore
+++ b/TeX.gitignore
@@ -76,6 +76,7 @@ acs-*.bib
 *.glg
 *.glo
 *.gls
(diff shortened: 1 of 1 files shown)


shiftkey:
Here's an discussion from StackExchange which cites removing the `*.glsdef` file as a way to ensure the glossaries are updated:

http://tex.stackexchange.com/questions/180660/generating-glossaries-again-without-deleting-the-old-ones

shiftkey:
Thanks!

PR #1876, opened 2016-02-24
removed elm-package.json

ezekeal:
elm-package.json lists dependencies, so it shouldn't be ignored

### DIFF
--- a/Elm.gitignore
+++ b/Elm.gitignore
@@ -1,5 +1,4 @@
 # elm-package generated files
-elm-package.json
 elm-stuff/
 # elm-repl generated files
 repl-temp-*


shiftkey:
@ezekeal seems reasonable - is there a link to some discussion that we can point people to so they can understand this change?

eeue56:
To give some more background and backing:

`elm-package.json` is the file which contains the information about the current project. It lists the sources directories, dependencies. It's the same as a `package.json` for Node-based projects.

For more information on it, see here - https://github.com/elm-lang/elm-package#preparing-for-publication

ezekeal:
Ha, well It happened on the elm slack channel.

The elm-package.json holds a list of dependencies. 
So that when a user downloads the file they can use the command:
`elm package install`
to automatically install them.

Without saving this file you will get errors for each package until you install them one-by-one, just like in a node pacakge.json.

shiftkey:
For reference, the original PR #1610 had this comment:

> Ignores installed packages (elm-package.json) and elm-stuff/, as these files will be generated once elm-package has been run for a .elm file. This is also beneficial where dependencies have changed, and an install package is no longer required.

I gather this advice has changed, I just wanted a bit more context for people who come looking for this change...

eeue56:
FWIW,

`elm-stuff` is generated when elm-package is run. This is where the downloaded deps live, along with the compiled build artifacts. It should be ignored. It's the same as a `node_modules` folder.

`elm-package.json` is created _if it does not already exist_ when you first run `elm-package install`. After that, it will always exist. It contains your project details and setup. It's the same as a `package.json` file

shiftkey:
@ezekeal @eeue56 thanks for the extra context - it's been a while since I looked at Elm.

zwilias:
@nathanl you should, in fact, ignore all of `elm-stuff` for the time being. 0.19 will have transitive dependencies as part of the main `elm.json` so the confusion should resolve, then.

PR #1877, opened 2016-02-25
Cordova ignore

rido-min:


shiftkey:
@rido-min previously we've had people create a standalone `Cordova.gitignore` file - can you tell me a bit about why you think this should live inside the `VisualStudio.gitignore` template?

shiftkey:

PR #1879, opened 2016-02-25
Update Jetbrains .gitignore dataSources.local.xml

rocketraman:
As of IDEA 14, dataSources.local.xml should be ignored. See https://intellij-support.jetbrains.com/hc/en-us/articles/206544839-How-to-manage-projects-under-Version-Control-Systems and  https://youtrack.jetbrains.com/issue/IDEA-127105.

### DIFF
--- a/Global/JetBrains.gitignore
+++ b/Global/JetBrains.gitignore
@@ -11,6 +11,7 @@
 # Sensitive or high-churn files:
 .idea/dataSources.ids
 .idea/dataSources.xml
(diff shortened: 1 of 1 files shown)


shiftkey:
Thanks! The YouTrack issue also mentions this:

> In IDEA 14 it will be safe to share dataSources.xml and/or ids.

But given it will likely take time for others to upgrade I'm happy to leave that rule in for now.

PR #1880, opened 2016-02-25
Add gitignore for Kotlin

jjst:
[Kotlin](https://kotlinlang.org/) is a JVM-based language developed by the [Jetbrains](https://www.jetbrains.com/) folks that claims 100% interop with Java. The first stable version, [1.0.0, came out in February 2016](http://blog.jetbrains.com/kotlin/2016/02/kotlin-1-0-released-pragmatic-language-for-jvm-and-android/).

I'm not aware of any Kotlin-specific generated files during the build, so because Kotlin has Java interop, generates class files similar to Java, and generally uses the same file formats, I symlinked Kotlin.gitignore to Java.gitignore.

### DIFF
--- a/Kotlin.gitignore
+++ b/Kotlin.gitignore
@@ -1,0 +1,1 @@
+Java.gitignore


jaredsburrows:
+1

drothmaler:
I'm not sure if this adds any value... It just might confuse the user. There are no Symlinks for other JVM-Languages (like Gosu, Cylon, Groovy or Nice) as long as the don't have any extra rules. 

The user might end up with a gitignore-file full of duplicates, if he accidentally imports both (Java.gitignore and Kotlin.gitignore)

jjst:
There are a few precedents: the {Fortran,Clojure}.gitignore are symlinks to respectively {C++,Leiningen}.gitignore. I do get your point, though, especially given the number of JVM-based languages sprawling. I think there's value in symlinking though if only to make it as easy as possible for people using those languages to use [gitignore.io](https://www.gitignore.io/) when using those languages.

Obaied:
So, any updates on the symlink here? It looks good

shiftkey:
Thanks all!

ellet0:
This change might need to be updated, Kotlin now can work independently of JVM. It can work with JS, native platforms, iOS, macOS, Wasm, Objective C, and more

There is a new directory that is used by Kotlin 2.0, which is now independent of Gradle

https://kotlinlang.org/docs/whatsnew20.html#new-directory-for-kotlin-data-in-gradle-projects

PR #1884, opened 2016-02-27
Update Java.gitignore

DUFCLiam:
I would like .java added to it, As that file can be used.

None provided sorry.

### DIFF
--- a/Java.gitignore
+++ b/Java.gitignore
@@ -5,6 +5,7 @@
 
 # Package Files #
 *.jar
+*.java
 *.war
 *.ear
 


shiftkey:
@DUFCLiam this will exclude Java source files from version control, which feels very dangerous. Could you clarify why this is necessary?

DUFCLiam:
I am sorry, My brother had access to my laptop therefore he had access to GitHub sorry for any inconvenience caused.

shiftkey:
@DUFCLiam all good, thanks for confirming!

DUFCLiam:
You're welcome :) @shiftkey

PR #1886, opened 2016-02-29
Stop tracking of VS vshost files

chuckleplant:
Visual Studio generates vshost files automatically, these are not to be tracked.

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -21,6 +21,8 @@ bld/
 [Bb]in/
 [Oo]bj/
 [Ll]og/
+*.vshost.exe.manifest
+*.vshost.exe.config
 
 # Visual Studio 2015 cache/options directory
 .vs/


shiftkey:
I'm curious why this isn't being caught as part of the previous rules in this section - could you tell me a bit about the problem you're seeing?

chuckleplant:
C# console projects for instance generate these within the solution folder. And they're not ignored in the .gitignore file. As far as I understand these are helper files to launch debugging from VS. However they are automatically generated, like the user and database files. 
Shouldn't these be ignored?

shiftkey:
cc @davkean for :eyes: on whether this should be captured in the gitignore template

davkean:
What do you mean by VS generates them in the solution folder? These should be getting generated in bin directory, and already covered by gitignore.

chuckleplant:
Hi there, this is a bit old but maybe in my project those were redirected somewhere else.. can't tell for sure. In any case those should always be ignored even if user selects a different generation folder.

shiftkey:
Closing this one out as unclear. Please reopen if a better repro of this is happening - [VSHost is going away](https://github.com/dotnet/roslyn-project-system/issues/66#issuecomment-269733749) in the future anyway...

PR #1887, opened 2016-03-01
add backup file (.bak)

benny856694:
sometimes, an editor will generate a backup file (.bak) after edit a file, it should be ignored

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -201,6 +201,7 @@ Generated_Code/
 # because we have git ;-)
 _UpgradeReport_Files/
 Backup*/
+*.bak
 UpgradeLog*.XML
 UpgradeLog*.htm
 


VisualStudio.gitignore
```diff
@@ -201,6 +201,7 @@ Generated_Code/
 # because we have git ;-)
 _UpgradeReport_Files/
 Backup*/
+*.bak
```
shiftkey: I think this is a different feature to the VS upgrading files, as per [this thread](https://social.msdn.microsoft.com/Forums/vstudio/en-US/645f21a0-d584-4b0e-9966-a4a7f3ce30c0/disable-the-creation-of-bak-files-in-visual-studio-2008?forum=csharpgeneral)

Could you move this below and add a comment above the rule to indicate it is related to AutoRecovery of files? :heart:

shiftkey:

PR #1888, opened 2016-03-01
Ignore gradle scripts generated by Intellij

haritzmedina:
IntelliJ IDEA if you select gradle wrapper in project creation time it creates two scripts (linux and windows) called gradlew and gradlew.bat, and I think it's not good idea to upload those files.

### DIFF
--- a/Gradle.gitignore
+++ b/Gradle.gitignore
@@ -12,3 +12,7 @@ gradle-app.setting
 
 # # Work around https://youtrack.jetbrains.com/issue/IDEA-116898
 # gradle/wrapper/gradle-wrapper.properties
+
+# Ignore gradle scripts
+gradlew
+gradlew.bat


drothmaler:
`gradlew` and `gradlew.bat`are part of the gradle-wrapper, that **should** be checked in to version control... 
From [Gradle documentation ](https://docs.gradle.org/current/userguide/gradle_wrapper.html#wrapperSimple):
- gradlew
- gradlew.bat
- gradle/wrapper/gradle-wrapper.jar
- gradle/wrapper/gradle-wrapper.properties

"All of these files _should_ be submitted to your version control system. This only needs to be done once. After these files have been added to the project, the project should then be built with the added `gradlew` command. The `gradlew` command can be used exactly the same way as the `gradle` command."

shiftkey:
@haritzmedina the feedback from @drothmaler - as well as the mention in the official documentation - makes me thing this isn't a good change to incorporate.

Let me know if you agree or disagree with his feedback.

haritzmedina:
Yes, I agree with @drothmaler. Thanks for your advices.

PR #1891, opened 2016-03-06
Added nore ignores. Removed.env example file

gaaarfild:

`.env.example` should not be ignored, because it contains exmplaes of actual parameters and on deploy, needs to be copied to `.env` file with appropriate parameter values.

Didn't find a documentation, but I think this is common practice, to use such files.
https://mattstauffer.co/blog/laravel-5.0-environment-detection-and-environment-variables#introducing-php-dotenv

### DIFF
--- a/Laravel.gitignore
+++ b/Laravel.gitignore
@@ -1,7 +1,5 @@
 vendor/
 node_modules/
-.rocketeer/
-.phpintel/
 
 # Laravel 4 specific
 bootstrap/compiled.php
@@ -13,3 +11,6 @@ storage/
 .env.*.php
 .env.php
 .env
+
+# Rocketeer PHP task runner and deployment package. https://github.com/rocketeers/rocketeer
+.rocketeer/
--- a/Laravel.gitignore
+++ b/Laravel.gitignore
@@ -1,5 +1,7 @@
 vendor/
 node_modules/
+.rocketeer/
+.phpintel/
 
 # Laravel 4 specific
 bootstrap/compiled.php
@@ -11,4 +13,3 @@ storage/
 .env.*.php
 .env.php
 .env
-.env.example


Laravel.gitignore
```diff
@@ -1,5 +1,7 @@
 vendor/
 node_modules/
+.rocketeer/
```
shiftkey: These changes seem unrelated to the PR - can you tell me a bit about why they're useful here?

shiftkey:
> Didn't find a documentation, but I think this is common practice, to use such files.

The section further down in [Customizing the example file](https://mattstauffer.co/blog/laravel-5.0-environment-detection-and-environment-variables#customizing-the-example-file) actually covers this nicely:

> That's it! Need 10 variables for each install? Add those 10 variables to your `.env.example` file with sensible (or silly) defaults. This file will get committed to your source control, and then **each new install can start out by running `cp .env.example .env` and then customizing `.env`**.

And you can't do this if the `.env.example` is in version control, so :thumbsup: to this change...

Laravel.gitignore
gaaarfild: Maybe I don't understand the system, but afaik, we don't use different .gitignore files in one folder. So covering some deployment systems in this file is a good way. Or am I wrong?

shiftkey: @gaaarfild I'm fine with handling multiple changes to a file in this PR, I'd just love some context on what these folders represent - and links to the relevant projects in case others want to know more...

gaaarfild: Sorry for my rudeness. This is the package https://github.com/rocketeers/rocketeer

shiftkey: If these rules are related, could you move them out to their own section like this:

```
# Rocketeer - task runner and deployment package
.rocketeer/
.phpintel/
```

shiftkey:
@gaaarfild fantastic, thanks for those changes!

PR #1893, opened 2016-03-07
Add Kobalt gitignore

jaredsburrows:

This is the first `gitignore`.

https://github.com/joeblau/gitignore.io/issues/184#issuecomment-193095027

If this is a new template: 

https://github.com/cbeust/kobalt @cbeust

### DIFF
--- a/Kobalt.gitignore
+++ b/Kobalt.gitignore
@@ -1,0 +1,4 @@
+### Kobalt ###
+.kobalt
+kobaltBuild/
+


drothmaler:
I think there should also be a whitelisting for `kobalt/wrapper/kobalt-wrapper.jar` (like it is done for the gradle wrapper in the [Gradle.gitignore](https://github.com/github/gitignore/blob/master/Gradle.gitignore)). Otherwise the jar would be ignored by the `*.jar` rule from [Java.gitignore](https://github.com/github/gitignore/blob/master/Java.gitignore)

PR #1896, opened 2016-03-08
ignore folder for non-nuget paket dependencies

tpluscode:

Paket can download files from Git(Hub) and the web and puts them in `paket-files` folder

http://fsprojects.github.io/Paket/http-dependencies.html
http://fsprojects.github.io/Paket/github-dependencies.html
http://fsprojects.github.io/Paket/git-dependencies.html

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -235,6 +235,7 @@ _Pvt_Extensions
 
 # Paket dependency manager
 .paket/paket.exe
+paket-files/
 
 # FAKE - F# Make
 .fake/


shiftkey:
Is this a new change? Up until this point I've not heard a need for this to be added...

cc @forki for a quick :thumbsup:

tpluscode:
Nope, been there for ages. I guess it's not used enough. Maybe there's a need for some Twitter PR :)

shiftkey:
I'll let this sit for a bit and see if Steffen will play his BDFL card on this one :grin:

forki:
From paket standpoint this makes sense.

shiftkey:
Thanks @tpluscode and @forki!

tpluscode:
Sorry, so how does this work actually? I've just created a new repo with VS .gitignore and this change wasn't included

shiftkey:
@tpluscode it sounds like VS needs to update the list of templates it has in the box (we do the same with GitHub Desktop)

tpluscode:
My bad, I phrased that wrong. I meant that I created the repository on github.com and selected Visual Studio in the dropdown

shiftkey:
@tpluscode oops, then I guess I need to do the refresh. Apologies for the delay.

PR #1898, opened 2016-03-09
Qt: Insert spaces for consistence

tim-sueberkrueb:

Inserted two spaces to make the comments look consistent. That's it :)


shiftkey:
@tim-sueberkrueb thanks!

PR #1899, opened 2016-03-10
Python: don't track virtualenv files.

e-yes:
Don't track files created by virtualenv.
Typical location are "venv" or "ENV" directories (relatively to project's home). These files depend on environment preferred by particular user. They shouldn't be committed to repository, because 1) it's meaningless and 2) could overwrite or conflict with settings of  another user of the same repository.

http://docs.python-guide.org/en/latest/dev/virtualenvs/
https://virtualenv.pypa.io/en/latest/userguide.html

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -77,5 +77,9 @@ celerybeat-schedule
 # dotenv
 .env
 
+# virtualenv
+venv/
+ENV/
+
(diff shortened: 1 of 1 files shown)


shiftkey:
@e-yes thanks for all the details!

PR #1900, opened 2016-03-11
Ruby: Mac and tools stuff

hboon:

.DS_Store and tags file.

### DIFF
--- a/Ruby.gitignore
+++ b/Ruby.gitignore
@@ -9,6 +9,7 @@
 /test/tmp/
 /test/version_tmp/
 /tmp/
+tags
 
 ## Specific to RubyMotion:
 .dat*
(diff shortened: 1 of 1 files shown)


Ruby.gitignore
```diff
@@ -26,6 +27,9 @@ build/
 /vendor/bundle
 /lib/bundler/man/
 
+## Mac
+.DS_Store
```
shiftkey: We're tracking this rule in `Global/OSX.gitignore` (as well as many others). Please drop this.

Ruby.gitignore
```diff
@@ -9,6 +9,7 @@
 /test/tmp/
 /test/version_tmp/
 /tmp/
+tags
```
shiftkey: I'm not sure what this file represents, and why it can be ignored in version control? Could you elaborate?

hboon:
> tags

These are for `ctags` and compatible tools.

But I completely missed it (and .DS_Store) being included in Global/ already. Sorry. Closing this PR.

PR #1901, opened 2016-03-11
Ruby: ignore more build artifacts specific to RubyMotion

hboon:

RubyMotion build artifacts

### DIFF
--- a/Ruby.gitignore
+++ b/Ruby.gitignore
@@ -14,10 +14,17 @@
 .dat*
 .repl_history
 build/
-vendor/Pods/
 *.bridgesupport
 build-iPhoneOS/
 build-iPhoneSimulator/
+
+## Specific to RubyMotion (use of CocoaPods):
+#
+# We recommend against adding the Pods directory to your .gitignore. However
+# you should judge for yourself, the pros and cons are mentioned at:
+# https://guides.cocoapods.org/using/using-cocoapods.html#should-i-check-the-pods-directory-into-source-control
+#
+# vendor/Pods/
 
 ## Documentation cache and generated files:
 /.yardoc/
--- a/Ruby.gitignore
+++ b/Ruby.gitignore
@@ -14,6 +14,10 @@
 .dat*
 .repl_history
 build/
+vendor/Pods/
+*.bridgesupport
+build-iPhoneOS/
+build-iPhoneSimulator/
 
 ## Documentation cache and generated files:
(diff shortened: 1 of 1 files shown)


shiftkey:
Could you move these into their own section to indicate they are specific to RubyMotion?

Also, are there any links in the RubyMotion docs that explain what these are created for (and why they can be ignored)?

hboon:
It isn't obvious in the diff, but the additions are to an existing RubyMotion section: https://github.com/hboon/gitignore/blob/0ae23acc5b2e763a12bfeebe4fac05e6077b4219/Ruby.gitignore

> Docs

Can't find any official ones unfortunately. But an example somewhere else: https://github.com/dblandin/rubymotion-cookbook/blob/master/text-kit/.gitignore#L5-L7

shiftkey:
@hboon 

> It isn't obvious in the diff, but the additions are to an existing RubyMotion section:

My apologies, indeed I did not look further up the diff.

> Can't find any official ones unfortunately. But an example somewhere else:

That's fine for the last three rules. What about `vendor/Pods/`?

hboon:
> That's fine for the last three rules. What about vendor/Pods/?

My bad. They are the same Pods/ directory for CocoaPods used in Mac/iOS projects. There is indeed a doc for this at https://guides.cocoapods.org/using/using-cocoapods.html under the section "Should I check the Pods directory into source control?"

Similar lines in https://github.com/github/gitignore/blob/master/Swift.gitignore#L38-L44 and 
https://github.com/github/gitignore/blob/master/Objective-C.gitignore#L28-L34

There's more a more detailed comment in those files. Maybe they should be included here too? I have always included the line (uncommented), rather than excluding it by default like in Swift.gitignore and Objective-C.gitignore because Pods/ is huge compared to the actual code in a RubyMotion project. Seems a better default in this case..

shiftkey:
@hboon to be honest, I'd like to go with the conventions that other communities are adopting - would you mind adding in that Cocoapods section like it's done in `Swift.gitignore` and `Objective-C.gitignore` and leave it commented out.

I think that's the last bit, everything else is :gem:

shiftkey:
And for reference, here's the original PR about the rule for `Objective-C.gitignore` - which was then ported over to `Swift.gitignore`

hboon:
There. Thanks for helping out!

shiftkey:
@hboon no no, thank you!

PR #1902, opened 2016-03-12
Ignore Thumbs.db and .DS_Store

moander:
Please?


shiftkey:
We're already tracking these rules in `Global/OSX.gitignore` and `Global/Windows.gitignore` (as well as many others). If you're doing a cross-platform project I'd suggest pulling in both of those into your local template.

PR #1903, opened 2016-03-12
ignore node_modules everywhere as default

aug2uag:

Since practice is to pull node_modules relative to the box, the default config should be for global node_modules ignore to include subdirectories.

Documents relating to the original intent of node_modules, and variance of Javascript compilation across different environments.

### DIFF
--- a/Node.gitignore
+++ b/Node.gitignore
@@ -24,7 +24,7 @@ coverage
 build/Release
 
 # Dependency directories
-node_modules
+/node_modules
 jspm_packages
 
 # Optional npm cache directory
--- a/Node.gitignore
+++ b/Node.gitignore
@@ -3,6 +3,10 @@ logs
 *.log
 npm-debug.log*
 
+# .DS_Store
+*/.DS_Store # 1 level deep
+.DS_Store
+
 # Runtime data
 pids
 *.pid
@@ -23,12 +27,12 @@ coverage
 # Compiled binary addons (http://nodejs.org/api/addons.html)
 build/Release
 
-# Dependency directories
-/node_modules
-jspm_packages
+# Dependency directory
+*/node_modules
 
 # Optional npm cache directory
 .npm
 
 # Optional REPL history
 .node_repl_history
+
--- a/Node.gitignore
+++ b/Node.gitignore
@@ -27,8 +27,9 @@ coverage
 # Compiled binary addons (http://nodejs.org/api/addons.html)
 build/Release
 
-# Dependency directory
+# Dependency directories
 */node_modules
+jspm_packages
 
 # Optional npm cache directory
 .npm


Node.gitignore
```diff
@@ -3,6 +3,10 @@ logs
 *.log
 npm-debug.log*
 
+# .DS_Store
```
shiftkey: This is an OSX specific rule and tracked in `Global/OSX.gitignore` - if you think that rule needs to be changed please open a PR against that file.

Node.gitignore
```diff
@@ -23,12 +27,12 @@ coverage
 # Compiled binary addons (http://nodejs.org/api/addons.html)
 build/Release
 
-# Dependency directories
-node_modules
-jspm_packages
+# Dependency directory
+*/node_modules
```
shiftkey: I'm not clear on this change:

 - `node_modules` will ignore all folders with that name
 - `/node_modules` will ignore any folder at the root of the repository with this name

This change doesn't address either. Could you tell me more about what you're trying to define here?

Node.gitignore
```diff
@@ -23,12 +27,12 @@ coverage
 # Compiled binary addons (http://nodejs.org/api/addons.html)
 build/Release
 
-# Dependency directories
-node_modules
-jspm_packages
```
shiftkey: Please keep this rule around - not everyone is running NPM

shiftkey:
I've been burned by changes introduced by changing the rules around `node_modules` so apologies for being thorough with the review. I'd also love some citations for the change we're considering here, so that I can understand that the community is behind this in some respect.

Node.gitignore
aug2uag: removing `jspm_packages` was not intended, I'll correct this

Node.gitignore
aug2uag: copy that, however isn't it true that only one `.gitignore` is available during the creation of a repository, and in which case OS specific rules may be included?

aug2uag:
np, if no added benefit then please deny merge

Node.gitignore
shiftkey: We separate the OS-specific rules for a couple of reasons:
 
 - easier to maintain standalone OS files in one location than across many files
 - standalone OS files means consumers can tailor their setups more easily (e.g. add Windows and OSX rules to Node)

The template created initially is intended to help you get started - it should definitely be tweaked or extended based on your needs over time...

Node.gitignore
```diff
@@ -24,7 +28,7 @@ coverage
 build/Release
 
 # Dependency directories
-node_modules
+*/node_modules
```
shiftkey: I'm not sure this rule is right, as it now shows the root `node_modules` as untracked (rather than ignored):

```
$ npm init
...
$ npm install express
...
$ echo "*/node_modules" > .gitignore
$ less .gitignore
*/node_modules
$ git status
On branch master

Initial commit

Untracked files:

	.gitignore
	Website/
	community-and-safety/
	node_modules/
	package.json
	ruby-hacking/

```

With the previous rule, any folder named `node_modules` at any level in the repository will be ignored - I think that's what we're going for here, right?

aug2uag:
absolutely that's the desired end.. maybe it was an artifact that nested `node_modules` dirs weren't ignored in my case. if in case you find that reproducible, please feel free to reopen.

PR #1904, opened 2016-03-12
Added 4 lines to user-specific stuff

Jeni4:

Referenced in the documentation as different files containing project specific information

https://www.jetbrains.com/webstorm/help/project.html

### DIFF
--- a/Global/JetBrains.gitignore
+++ b/Global/JetBrains.gitignore
@@ -7,6 +7,10 @@
 .idea/dictionaries
 .idea/vcs.xml
 .idea/jsLibraryMappings.xml
+.idea/.name
+.idea/encondings.xml
+.idea/misc.xml
+.idea/modules.xml
 
 # Sensitive or high-churn files:
 .idea/dataSources.ids


Global/JetBrains.gitignore
```diff
@@ -7,6 +7,10 @@
 .idea/dictionaries
 .idea/vcs.xml
 .idea/jsLibraryMappings.xml
+.idea/.name
+.idea/encondings.xml
```
shiftkey: typo: `encondings` -> `encodings`

shiftkey:
The linked document mentions this:

> All the settings files in the `.idea` directory should be put under version control except the `workspace.xml`, which stores your local preferences. The `workspace.xml` file should be marked as ignored by VCS.

There's also [this document](https://intellij-support.jetbrains.com/hc/en-us/articles/206544839) which covers all the JetBrains IDEs - we've been using that in the past as a canonical source for things...

barrystaes:
@shiftkey I noticed this reference document also suggests ignoring `iml` files for IntelliJ IDEA.

trdarr:
@barrystaes Posting this for future reference because the document has changed, apparently on 13 August but that’s a guess as there’s no obvious way to see its revision history.

> * All the files under .idea directory in the project root except the workspace.xml and tasks.xml files which store user specific settings
> * All the .iml module files that can be located in different module directories (applies to IntelliJ IDEA)

shiftkey:
I've merged in #2096 which means this now has a merge conflict.

Please open a new PR if there are further changes.

PR #1905, opened 2016-03-14
Add AppCode's project settings for Swift and Objective-c

vmalyi:

AppCode's growing popularity make this change needed.

See details about AppCode here: https://www.jetbrains.com/objc/

If this is a new template: 

no

### DIFF
--- a/Objective-C.gitignore
+++ b/Objective-C.gitignore
@@ -49,3 +49,7 @@ Carthage/Build
 
 fastlane/report.xml
 fastlane/screenshots
+
+# AppCode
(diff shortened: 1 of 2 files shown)


shiftkey:
While it's not enabled by default here, we have a separate template for all the JetBrains IDEs under `Global/JetBrains.gitignore` as they have a consistent way of managing the `.idea` folder across all their tools - and some files should be included in version control - see their [official documentation](https://intellij-support.jetbrains.com/hc/en-us/articles/206544839) for more information.

PR #1906, opened 2016-03-15
Ignore .fot .cb .cb2 for TeX

DSIW:

* File which ends by .fot stores information about footnotes and is automatically generated.
* File which ends by .cb or .cb2 are generated too and doesn't have any content on my system.

Unfortunately I couldn't find any documentation about this behavior, but I think it couldn't be bad to ignore these files.

### DIFF
--- a/TeX.gitignore
+++ b/TeX.gitignore
@@ -7,6 +7,9 @@
 *.out
 *.toc
 *.fmt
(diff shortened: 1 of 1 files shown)


shiftkey:
@DSIW this seems reasonable. Thanks!

PR #1907, opened 2016-03-16
Update ignore for fastlane files in Swift

chrisamanse:
Link to documentation of fastlane's updated [.gitignore](https://github.com/fastlane/fastlane/blob/master/fastlane/docs/Gitignore.md#gitignore)


shiftkey:
@chrisamanse thanks!

PR #1909, opened 2016-03-18
Remove a trailing whitepsace

dbburgess:

It's good convention to avoid whitespace at end of line.

http://programmers.stackexchange.com/q/121555
https://www.gnu.org/software/emacs/manual/html_node/emacs/Useless-Whitespace.html


shiftkey:
Thanks!

PR #1910, opened 2016-03-21
Update fastlane links

yanniks:

The old link is broken since Fastlane moved all their sources into one repo.

Fastlane now has a mono repo, the documentation moved into the folder fastlane/docs


shiftkey:
@yanniks thanks!

PR #1911, opened 2016-03-21
add gitignore for openFrameworks

2bbb:
for [openFrameworks](https://github.com/openframeworks/openFrameworks/)

### DIFF
--- a/OpenFrameworks.gitignore
+++ b/OpenFrameworks.gitignore
@@ -1,0 +1,131 @@
+#########################
+# openFrameworks patterns
+#########################
+
+# build files
+openFrameworks.a
+openFrameworksDebug.a
+openFrameworksUniversal.a
+libs/openFrameworksCompiled/lib/*/*
+!libs/openFrameworksCompiled/lib/*/.gitkeep
+
+# apothecary
+scripts/apothecary/build
+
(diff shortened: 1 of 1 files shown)


OpenFrameworks.gitignore
```diff
@@ -0,0 +1,131 @@
...
+*.app/
+*.pyc
+.svn/
+*.log
+*.cpp.eep
+*.cpp.elf
+*.cpp.hex
+
+#########################
+# IDE
```
shiftkey: We have existing templates for various IDEs that are up-to-date, and I don't want to take in changes which duplicate these rules.

Please isolate just the rules specific to OpenFrameworks, and add a comment if there are IDE templates that you think complement this template.

OpenFrameworks.gitignore
```diff
@@ -0,0 +1,131 @@
...
+gradlew.bat
+
+# QtCreator
+*.qbs.user
+*.pro.user
+*.pri
+
+
+#########################
+# operating system
```
shiftkey: We have existing templates for operating systems that are up-to-date, and I don't want to take in changes which duplicate these rules.

Please isolate just the rules specific to OpenFrameworks, and add a comment if there are OS templates under `Global` that you think complement this template.

PR #1471, opened 2015-04-09
Ignore Kile and kBibTeX backup files

lszeremeta:
# Project's homepages
* [Kile](http://kile.sourceforge.net/)
* [kBibTeX](http://home.gna.org/kbibtex/)

# Changes explained

```
# TODO Comment the next line if you want to keep your tikz graphics files
*.tikz
```

Comment if user does not use [knitr](https://github.com/yihui/knitr) and want to keep tikz graphics in the repository

```
## Editors

```

Move editors to editors

```
# Kile
*.backup
```

Kile [automatically creates](https://bugs.kde.org/show_bug.cgi?id=161570) backup (.backup) files.

```
# KBibTeX
*~*
```

kBibTeX also [automatically creates backups](http://lists.unix-ag.uni-kl.de/pipermail/kbibtex/2009-April/000266.html) - filename~, filename~2, filename~3, filename~4...
[Comment about this in kBibTeX source code](http://quickgit.kde.org/?p=kbibtex.git&a=blob&h=f39896e6f6de9dc99cd270dbb2226654f0d4b737&hb=bb70cfdd46f946bbb4afc37aa4eeaff6c43444bb&f=src%2Fparts%2Fpart.cpp) (line: 276)

PR #1913, opened 2016-03-21
More explicit pattern for KBibTeX

lszeremeta:
Pattern for BibTex is now less general.

Referring to discussion: https://github.com/github/gitignore/pull/1471#commitcomment-16788624


shiftkey:
Thanks @lszeremeta (and thanks @duetosymmetry for raising the concerns)!

PR #1916, opened 2016-03-23
Ignoring vignettes/*.R

sainathadapa:
'.R' files are also produced sometimes when building vignettes. These don't need to be committed.

### DIFF
--- a/R.gitignore
+++ b/R.gitignore
@@ -20,6 +20,7 @@
 # produced vignettes
 vignettes/*.html
 vignettes/*.pdf
+vignettes/*.R
 
 # OAuth2 token, see https://github.com/hadley/httr/releases/tag/v0.3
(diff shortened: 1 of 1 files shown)


shiftkey:
@sainathadapa do you have any supporting links which discuss this feature - and the recommendation to not track these files?

sainathadapa:
Vignettes are generally built using either Sweave or Knitr. If a vignette is built with sweave, it will have .Rnw, .Rtex, .Snw or .Stex extensions.  If a vignette is built using Knitr, it will have Rmd extension. So, my understanding was that *.R files can safely be ignored. 

But now that I think about it, it is quite possible that *.R files can be used by some other vignette builder, or they may be 'sourced' while building the vignettes. Hence closing the issue.

PR #1917, opened 2016-03-24
Update Waf for URL and specificity

seanfisk:
- Waf's homepage has moved from Google Code to https://waf.io
- Add specificity to unpacked waflib pattern
- Add support for Windows

- This is [how the lockfile is named](https://github.com/waf-project/waf/blob/waf-1.8.20/waflib/Options.py#L44) in the latest version
- This is [how the unpacked waflib directory is named](https://github.com/waf-project/waf/blob/waf-1.8.20/waf-light#L147) in the latest version, with the [leading dot used for non-Windows platforms](https://github.com/waf-project/waf/blob/waf-1.8.20/waf-light#L153)

Thanks!

### DIFF
--- a/Waf.gitignore
+++ b/Waf.gitignore
@@ -1,4 +1,4 @@
-# for projects that use Waf for building: http://code.google.com/p/waf/
 .waf-*
 .waf3-*
 .lock-*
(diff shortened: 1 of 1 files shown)


shiftkey:
@seanfisk thanks!

PR #1918, opened 2016-03-25
Ignore buckup doctrine entities

krzysiekpiasecki:
Command 'doctrine:generate:entities with option 'backup' generates backup entities. These files should not be shared amongs contributors.s

### DIFF
--- a/Symfony.gitignore
+++ b/Symfony.gitignore
@@ -41,3 +41,6 @@
 
 # Composer PHAR
 /composer.phar
+
+# Temporary files
+*/Entity/*~
--- a/Symfony.gitignore
+++ b/Symfony.gitignore
@@ -42,5 +42,5 @@
 # Composer PHAR
 /composer.phar
 
-# Temporary files
+# Backup entities generated with doctrine:generate:entities command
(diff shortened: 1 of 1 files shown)


shiftkey:
@krzysiekpiasecki any supporting links to this, so others can read up on it?

krzysiekpiasecki:
You mean references ?
- https://github.com/doctrine/DoctrineBundle/blob/master/Command/GenerateEntitiesDoctrineCommand.php
-http://stackoverflow.com/questions/5219937/why-does-doctrine-2-create-entity-php-file

> The commands will automatically create backups of your entities for you, so you will have a new User.php and a backup in User.php~, which you should be careful of if you are using version control and like to git add src/ without thinking too hard about it. You don't want to add those files to source control.

- https://codedump.io/share/Kb77XWSYeB3P/1/doctrinesymfony-entity-generator-and-generating-entity-from-one-table

shiftkey:
@krzysiekpiasecki thanks for those links - the SO question says `prefix` rather than `append` or `suffix` which is different from the last link, but seeing the source code is :heart:...

PR #1919, opened 2016-03-25
Add an exclusion for Visual Studio PublishScripts

pmn:
The `PublishScripts` directory created by selecting "Host in the cloud" on project creation may contain sensitive information. 

This update modifies the Visual Studio `.gitignore` template to exclude those files from source control. 

Screenshot of new project form pointing out the 'Host in the cloud' option

cc @github/windows

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -138,10 +138,15 @@ publish/
 # Publish Web Output
 *.[Pp]ublish.xml
 *.azurePubxml
-# TODO: Comment the next line if you want to checkin your web deploy settings 
+# TODO: Comment the next line if you want to checkin your web deploy settings
 # but database connection strings (with potential passwords) will be unencrypted
 *.pubxml
 *.publishproj
+
+# Microsoft Azure Web App publish settings. Comment the next line if you want to
+# checkin your Azure Web App publish settings, but sensitive information contained
+# in these scripts will be unencrypted
+PublishScripts/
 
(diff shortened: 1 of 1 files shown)


shiftkey:
@pmn can you give an example of the sort of files that will be caught by this new rule - the paths are the interesting bit here - and perhaps the contents of a file as a sample?

pmn:
@shiftkey definitely. An example file that would be caught by this rule is `PublishScripts\AppName.param.dev.json`, which could have contents like this: 

```
{
  "siteName": {
    "value": "YelpForDogs"
  },
  "hostingPlanName": {
    "value": "YelpForDogsPlan"
  },
  "siteLocation": {
    "value": "West US"
  },
  "sku": {
    "value": "Free"
  },
  "workerSize": {
    "value": "0"
  },
  "serverName": {
    "value": "yelpfordogs"
  },
  "serverLocation": {
    "value": "West US"
  },
  "administratorLogin": {
    "value": "yelpfordogsadmin"
  },
  "administratorLoginPassword": {
    "value": "AVerySecureP@ssword"
  },
  "databaseName": {
    "value": "yelpfordogs_db"
  }
}
```

This is an example file generated by the setup wizard, and currently would be committed to source control if the user didn't catch it. There is some information in this file that many users would be uncomfortable having in source control, especially if their project was in a public repository.

shiftkey:
@pmn excellent, thanks for the details :thumbsup:

PR #1920, opened 2016-03-27
Ignored PyCharm files

christoga:
Because I use PyCharm

If this is a new template: 

 - **Link to application or project’s homepage**: [file](https://github.com/christoga/.gitignore/blob/master/Python.gitignore)

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -83,3 +83,6 @@ ENV/
 
 # Spyder project settings
 .spyderproject
+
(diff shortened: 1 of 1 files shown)


shiftkey:
@christoga thanks, but I'd recommend adding the `Global/JetBrains.gitignore` rules to your project as there are some entries under `.idea` that are valuable to share in version control. See #1475 for more information about this.

PR #1922, opened 2016-03-27
Added Visual Studio 2015 .vs folder

vvolkgang:

Visual Studio 2015 moved all the user files to the `.vs` folder, which shouldn't be tracked.

https://visualstudio.uservoice.com/forums/121579-visual-studio/suggestions/6079923-store-project-related-information-in-vs-folder-to

### DIFF
--- a/UnrealEngine.gitignore
+++ b/UnrealEngine.gitignore
@@ -60,5 +60,3 @@ Intermediate/*
 
 # Cache files for the editor to use
 DerivedDataCache/*
-
-
--- a/UnrealEngine.gitignore
+++ b/UnrealEngine.gitignore
@@ -1,3 +1,6 @@
+# Visual Studio 2015 user specific files
+.vs/
+
 # Compiled Object files
 *.slo
 *.lo
@@ -57,3 +60,5 @@ Intermediate/*
 
 # Cache files for the editor to use
 DerivedDataCache/*
+
+


shiftkey:
@vvolkgang not being very familiar with how games using UE are being developed, would there be value in using more of the `VisualStudio.gitignore` template, which has this change?

vvolkgang:
Personally, I just copy paste the "contexts" I see fit for a given project
(in this case I would mix both and be done with that). That being said, the
VS one deals with all sorts of VS ecosystem stuff that isn't going to
happen when you use Unreal with VS ir Xcode, mixing both creates bloated
files which are hard to maintain by newbies (I came here because this was
linked in a Unreal tutorial). Furthermore, there was already some ignores
in place for sln and such.

What I suggest with this is one of the following:

1. Merge the PR, keeping it clean, centralised and easily
explainable/maintainable (I vote on this)
2. Remove all the code IDE specific ignores (VS and xcode), adding a
comment so people know they should probably look at those global ones too.
On Sun, 27 Mar 2016 at 22:01, Brendan Forster <notifications@github.com>
wrote:

> @vvolkgang <https://github.com/vvolkgang> not being very familiar with
> how games using UE are being developed, would there be value in using more
> of the VisualStudio.gitignore template, which has this change?
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/github/gitignore/pull/1922#issuecomment-202144745>
>
-- 
Regards,

Álison Fernandes

shiftkey:
@vvolkgang thanks for the extra info!

PR #1923, opened 2016-03-27
Add dist/ folder to Emacs gitignores

rexim:
Hello there!

`cask package` command by default produces this `dist/` folder with a ready-to-install package.el package, which is generally not something that should exist in a source code repo.

http://cask.readthedocs.org/en/latest/guide/usage.html#cask-package

regards,
rexim


shiftkey:
@rexim thanks!

PR #1924, opened 2016-03-27
Ignore session files

krzysiekpiasecki:
Session files should not be shared.

More on configuring the Directory where Session Files are Saved http://symfony.com/doc/current/cookbook/session/sessions_directory.html


shiftkey:
@krzysiekpiasecki thanks!

PR #1925, opened 2016-03-28
gitignore for ROOT C++ based, used by Experimental Physicists

ravijanjam:
**As of ROOT v 6.xy, files included in the gitignore are put up, which are unnecessary**

If this is a new template: 


shiftkey:

@ravijanjam if you feel like this should still be a template listed here, please read the section about [specialized templates](https://github.com/github/gitignore#specialized-templates) and open a new pull request that follows the new pattern.

PR #1926, opened 2016-03-29
Create C#.gitignore

tchadwick:

This is to add a template file for ignoring common C# project files not wanted in repository

None

If this is a new template: 

None


shiftkey:
Thanks, but we've unified C# and Visual Basic in `VisualStudio.gitignore` rather than manage two essentially-similar templates.

You can read about this change in https://github.com/github/gitignore/pull/492.

PR #1927, opened 2016-03-30
Haskell: Ignore .eventlog files

sachs4:
"EventLog is a fast, extensible event logging framework in the GHC run-time system (RTS) to support profiling of GHC run-time events" (from https://ghc.haskell.org/trac/ghc/wiki/EventLog).

One normally doesn't want to add these files to a repo, because they are only for profiling and debugging purposes.


shiftkey:
@sachs4 thanks!

PR #1928, opened 2016-03-31
Add *.VC.db for Visual C++ intellisense new db

adamwalling:
*.VC.db is the new database used by the sqlite backend to intellisense,
as compared to the old *.sdf SQL Server Compact database

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -80,6 +80,7 @@ ipch/
 *.opensdf
 *.sdf
 *.cachefile
+*.VC.db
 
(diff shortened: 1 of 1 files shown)


adamwalling:
New versions of Visual Studio will be using this *.VC.db format by default instead of the *.sdf format. You can also switch it manually in options.

shiftkey:
Already covered by #1783 which was just merged.

qkrrmsdjr:
https://drive.google.com/open?id=0B6GEM-FIKxd2Rnk3bnA1Zlo1T

PR #1929, opened 2016-04-01
Added .ropeproject directory

geogdog:
Anybody using python could be using the rope refactoring library, so the output directory is missing from this file.

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -83,3 +83,6 @@ ENV/
 
 # Spyder project settings
 .spyderproject
+
+# Rope project settings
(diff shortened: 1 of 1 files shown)


bitmingw:
`.ropeproject` will be automatically created when using vim editor bundled with [python-mode](https://github.com/klen/python-mode), with is a pretty annoying thing.

shiftkey:
Thanks! And this is a bit of the introduction to what this folder contains: https://github.com/python-rope/rope/blob/master/docs/overview.rst#ropeproject-folder

PR #1930, opened 2016-04-01
Ignore NCrunch temporary files for Visual Studio

lbergnehr:

NCrunch produces quite a few cached files for each project. These are ignored by the patterns in this commit.

http://www.ncrunch.net/documentation/considerations-and-constraints_using-ncrunch-with-source-control

If this is a new template: 

**Link to application or project’s homepage**
 http://www.ncrunch.net/

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -250,3 +250,8 @@ paket-files/
 # JetBrains Rider
 .idea/
 *.sln.iml
+
+# NCrunch
+_NCrunch_SOLUTION/*
+SOLUTION.ncrunchsolution.user
+nCrunchTemp_*


shiftkey:
@lbergnehr thanks, but this seems to overlap a bunch of what's already defined https://github.com/github/gitignore/blob/master/VisualStudio.gitignore#L112-L115 - could you unify these two sections?

VisualStudio.gitignore
```diff
@@ -250,3 +250,8 @@ paket-files/
 # JetBrains Rider
 .idea/
 *.sln.iml
+
+# NCrunch
+_NCrunch_SOLUTION/*
+SOLUTION.ncrunchsolution.user
```
shiftkey: This should be caught by `*.user` defined earlier

VisualStudio.gitignore
```diff
@@ -250,3 +250,8 @@ paket-files/
 # JetBrains Rider
 .idea/
 *.sln.iml
+
+# NCrunch
+_NCrunch_SOLUTION/*
```
shiftkey: Is `SOLUTION` a placeholder here? I'm puzzled as to why we've not added this earlier...

lbergnehr:
@shiftkey, you're absolutely right and I feel ashamed I didn't even notice there were NCrunch rules in there already! I think I started using the `.gitignore` and noticed NCrunch files in my project not being ignored, incorrectly assuming they should be added to the ignore rules. However, they were likely just the `*.v2.ncrunchsolution` and `*.v2.ncrunchproject` files, which *should* be added to version control (according to the documentation).

I'm closing this confusing PR.

PR #1932, opened 2016-04-03
ignore more cmake-generated files

mcandre:

`git status` still mentions cmake-generated files.

https://cmake.org/Wiki/CMake_Generating_Testing_Files

Or I can comment these patterns out by default, like the ruby gitignore.


shiftkey:
Closing due to conflicts. However this has been addressed by https://github.com/github/gitignore/pull/2242 and https://github.com/github/gitignore/pull/1960 so no further work is needed :tada:

PR #1936, opened 2016-04-08
[VisualStudio] fixed packages ignore in order to prevent whole-directory deletion on git stash

drauch:

I don't see a reason why to use the asterisk at the end of the packages include:

```
**/packages/*
```

So, I propose to use:

```
**/packages/
```

Using the asterisk has a major downside: on "git stash", the whole directory is deleted, which is not what you want to happen per default.

https://github.com/joeblau/gitignore.io/issues/194

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -153,7 +153,7 @@ PublishScripts/
 # NuGet Packages
 *.nupkg
 # The packages folder can be ignored because of Package Restore
-**/packages/*
+**/packages/
 # except build/, which is used as an MSBuild target.
 !**/packages/build/
 # Uncomment if necessary however generally it will be regenerated when needed


shiftkey:
@drauch I'm not quite sure about this one - the reason why the wildcard character is added is mentioned in 9aa3e1c3a7a5d9a7ecbab67f63eb81e8b4a493b7

> Without the asterisk at the end of the directory's path, an exception for a file within this directory does not work (ex. !packages/repositories.config)

And we have a couple of rules after which use the `!`.

Thoughts?

drauch:
As @jamiehumphries already pointed out in https://github.com/github/gitignore/commit/9aa3e1c3a7a5d9a7ecbab67f63eb81e8b4a493b7 this has the downside that Git not only removes the whole folder on git clean but also on git stash which is very annoying.

We should find some kind of workaround - do we really need the !-rules within the packages folder? E.g., repositories.config should not be included nowadays anyways, see: https://stackoverflow.com/questions/7286261/nuget-repositories-config (starting with NuGet 2.7).

Are there any other important !-rules within the packages folder?

shiftkey:
I can understand how `repositories.config` is no longer needed due to the new Package Restore features since NuGet 2.7, but what about `/build/` - do we still need to keep that around?

drauch:
I'm not sure which type of scenario uses a `/build/` sub directory. I've never encountered one.

shiftkey:
Here's one example: [Microsoft.Bcl.Build](https://www.nuget.org/packages/Microsoft.Bcl.Build) will add a `.targets` file to the `csproj [like this](https://www.nuget.org/packages/Microsoft.bcl.build) and you can see it fail to build if you haven't restored the packages. Perhaps this is also fixed by NuGet 2.7?

drauch:
The build fails anyways, if the packages directory hasn't been restored? What's special about the targets file? Does NuGet restore not work without having those targets already in place? This seems to be a bit weird.

shiftkey:
@drauch I'm not sure, and this is something I'd like to test and verify before merging - simply because I've been burned in the past by hastily merging changes to templates, and lots of people depend on the `packages` folder...

drauch:
@shiftkey : Of course, I understand. However, I can't even find a SO question regarding a build sub folder - do you know by chance which issue has requested the build sub folder to be excluded from the ignore list?

shiftkey:
@drauch the original PR for the build folder is #694 and that links off to http://docs.nuget.org/consume/support-for-native-projects which explains the feature in more detail

drauch:
The packages folder should not be checked in. Package restore is now the default with Visual Studio. This is an outdated assumption in PR #694.

exalted:
I wonder how is this "a problem" _just_ for `VisualStudio.gitignore` but not for any other `*.gitignore` in this repository?

I don't mean this in a contentious manner, rather as a deeper curiosity.

My understanding is that, if the desired outcome is to "ignore" a directory that is known to contain compiled/generated/downloaded binaries, assets, packages, etc., then it makes total sense to ignore that directory like `foo/` rather than like `foo/*`.

In that case, neither `git clean`, and by extension, nor `git stash save --include-untracked` (which AFAIK internally calls `git clean -fd`) would "clean" `foo` directory.

I guess what I am trying to say is: with my arguably limited knowledge on the matter, it _feels wrong_ to me to "ignore" such directories like `foo/*` rather than `foo/` and that includes every other `*.gitignore` file in this repository that matches `grep -RE '/\*$'`.

drauch:
I too haven't found a single counter argument against using "foo/" instead of "foo/*" up to now. I think it is time to merge this :-)

shiftkey:
This is actually a breaking change because of the subsequent rule around including the `build` folder:

```
**/packages/*
!**/packages/build/
```

If you drop the `*` here the subsequent rule is not used and your `build` files aren't tracked.

drauch:
It is indeed a breaking change, but a good one. There is no need for the build sub directory anymore. This renders the Visual Studio .gitconfig as-is useless by default : /

shiftkey:
> There is no need for the build sub directory anymore. 

@drauch if you can get someone from the NuGet team, or point to some NuGet documentation indicating this is no longer required, then we can talk about addressing the asterisk rule above it.

PR #1939, opened 2016-04-11
Added Herokuapp gitignore file.

eric-h0:

Created a .gitignore file for launching apps with Heroku.

index.php file contains entry point for application launched and composer.json file just has an empty {}.

This will make uploading files to github, that are also launched to heroku, cleaner to view by other users.

### DIFF
--- a/Herokuapp.gitignore
+++ b/Herokuapp.gitignore
@@ -1,0 +1,2 @@
+composer.json
+index.php


shiftkey:
@eric-h0 can you provide some links to Heroku or other resources which explain these details further?

eric-h0:
Heroku requires two files to be added to the folder that a user would like to be launched to its own domain.

https://devcenter.heroku.com/articles/getting-started-with-php#prepare-the-app 
First, after initializing folder for Git and ensuring code runs in a static browser, a json file is added to the folder of the application that you would like to launch.
`{}` = composer.json

https://devcenter.heroku.com/articles/getting-started-with-php#deploy-the-app
Secondly, a php file:
`<?php include_once("example.html"); ?>` = index.php
is added that routes that location to be the site's main page.

These two files do not need to be pushed to Github in order to successfully launch to Heroku and I found that .gitignoring them attributed to less files to worry about doing collaboration.

elifoster:
I think that if there were to be a Heroku specific gitignore, only adding PHP related files is strange, since it also supports Ruby, Python, Go, Java, Scala, Clojure, and NodeJS, which might have their own unnecessary files as well. Perhaps these files should be added to the PHP gitignore instead?

shiftkey:
@eric-h0 yep, seconding what @elifoster said above - this template isn't exactly covering all of the scenarios that Heroku supports. I'd rather consumers use the language-specific templates already provided than overlap unnecessarily like we are here...

shiftkey:

PR #1940, opened 2016-04-12
Update Android.gitignore

MaTriXy:

I've added a few missing ignores to the file .

Added missing configurations for Android studio


Android.gitignore
```diff
@@ -32,6 +51,15 @@ proguard/
 # Android Studio captures folder
 captures/
 
+# OS-specific files
```
shiftkey: We're tracking these in `Global/OSX.gitignore` - please drop these...

MaTriXy: ok will update now.

tuxayo:
It seems that part of those changes are covered by the [global JetBrains.gitignore](https://github.com/github/gitignore/blob/c74eb94334463252a15c8f80876847cc580b8dd3/Global/JetBrains.gitignore).

So the real problem is that there doesn't seem to be a clear indication that someone coming just for a .gitignore for Android should also use JetBrains.gitignore (and maybe others??)

MaTriXy:
Android Studio has some similarity with JetBrains. 
When someone wants an Android .gitignore he will most likely use Android studio (at minimum) so having a few overlapping rules is ok as it is usually the only IDE for Android in a specific computer.

PR #1941, opened 2016-04-13
Update VisualStudio.gitignore

RTPeat:

VS2015 .sln.ide folder containing Roslyn Compiler temporary files not being ignored.

http://stackoverflow.com/questions/28144621/visual-studio-2015-should-i-commit-project-sln-ide-files-into-git

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -10,6 +10,9 @@
 # User-specific files (MonoDevelop/Xamarin Studio)
 *.userprefs
 
+# Roslyn Compiler temporary files
+**.ide/
+
 # Build results
 [Dd]ebug/
 [Dd]ebugPublic/


shiftkey:
I think this `.ide` folder was renamed to `.vs` in one of the later updates to VS2015 while it was under development - see http://stackoverflow.com/a/32194175/1363815 for details.

RTPeat:
It might be generating the .vs folder on new projects, but the reason I added the change was for an existing project which after the upgrade to VS2015 update 2 today started including that .ide folder in numerous projects. Adding it to the .gitignore dealt with the folder.

shiftkey:
@RTPeat oh weird, I've been running that version for a while and haven't noticed this. What sort of projects are you seeing this happen to?

RTPeat:
Basically was anything we were using Git for - post updating to Update 2 we were looking at loads of previously untracked files being picked up as changes. The particular project I was looking at had over 5000 new changes picked up. Putting in the previous version of .gitignore got rid of most of them, except for the .ide folder.

RTPeat:
We've had a bit more of an investigate today, and the projects with the problem consistently didn't have a .gitignore before the update, they had a .tfignore instead, that had some exclusions. It appears that in previous versions Visual Studio was making exclusions as per the .tfignore file plus various folders and files that Visual Studio had created without them being defined in the .tfignore. Post update 2 it appears Visual Studio is now using a regular .gitignore.

shiftkey:
I'm pretty sure this should all be pointing to `.vs/` now in the relevant versions of VS.

PR #1942, opened 2016-04-15
Update Rust.gitignore

deepak:
no need to specifically ignore compiled files and executables
as it will be under the target directory

this is as per this [comment](https://github.com/rust-lang/cargo/issues/2565#issuecomment-209474212)
in the rust repo


shiftkey:
Is this a recent change in Rust or do we assume most users will be on the right version?

elifoster:
This has been the case for as long as I have known about Rust, which is at least 2 years ago. Additionally, most Rust projects I've seen and worked on have had gitignores pretty much only consisting of the target directory.

shiftkey:
@elifoster cool - if there's a good example elsewhere we can cite I'm happy to run with this change!

elifoster:
There's the issue in the Rust repo that @deepak linked. There is also all the Rust repositories which ignore solely target:
* [rust-lang/cargo](https://github.com/rust-lang/cargo/blob/master/.gitignore#L1)
* [rust-lang/libc](https://github.com/rust-lang/libc/blob/master/.gitignore#L1)
* [rust-lang/rust](https://github.com/rust-lang/rust/blob/master/.gitignore#L98)
* [rust-lang/crates.io](https://github.com/rust-lang/crates.io/blob/master/.gitignore#L1)
* [rust-lang/rust-by-example](https://github.com/rust-lang/rust-by-example/blob/master/.gitignore#L7)

shiftkey:
@elifoster excellent, thanks!

deepak:
thanks @elifoster and @shiftkey

PR #1943, opened 2016-04-15
Ruby: Ignore .env

elifoster:

This file is used moderately commonly to store environment variables locally. I often find myself adding it to the Ruby gitignore manually in my Ruby applications.

The Python gitignore ignores the same file for the same reason. And before the fact that this file is only used when the dotenv gem is also used, that is the same with Python.

### DIFF
--- a/Ruby.gitignore
+++ b/Ruby.gitignore
@@ -9,6 +9,7 @@
 /test/tmp/
 /test/version_tmp/
 /tmp/
+.env
 
 ## Specific to RubyMotion:
 .dat*
--- a/Ruby.gitignore
+++ b/Ruby.gitignore
@@ -10,6 +10,7 @@
 /test/version_tmp/
 /tmp/
 
+# Used by dotenv library to load environment variables.
 # .env
 
 ## Specific to RubyMotion:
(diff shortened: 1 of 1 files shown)


shiftkey:
For reference, here's a note about this from the `dotenv` README https://github.com/bkeepers/dotenv#should-i-commit-my-env-file

shiftkey:
@elifoster putting aside what we did with the Python template, this comment is interesting:

> Personally, I prefer to commit the .env file with development-only settings. This makes it easy for other developers to get started on the project without compromising credentials for other environments.

elifoster:
Hm, that's interesting. I've never had a need for the .env file to be pushed to version control. Maybe it could be default commented out like many of the other rules in this template?

shiftkey:
> Maybe it could be default commented out like many of the other rules in this template?

I like this idea

elifoster:
I have commented it out. I figured it wasn't really necessary to explain what that file is for, as since it's commented out, it will probably only be used by people who know exactly what they are doing.

shiftkey:
@elifoster what about a brief comment above the line explaining the file's purpose?

elifoster:
@shiftkey There you go :)

shiftkey:
@elifoster thanks!

PR #1945, opened 2016-04-15
Add FlashDevelop and executables to Actionscript

BlueHatbRit:

It's quite common for AIR developers to use FlashDevelop if they're on Windows. The structure of an AIR project generated by FD has a slightly different bin and obj folder naming which isn't on here. There is also no listing of the executables often generated by AIR developers such as: ipa, apk, air files. These have been added these as well under a separate sub-section.

### DIFF
--- a/Actionscript.gitignore
+++ b/Actionscript.gitignore
@@ -2,9 +2,17 @@
 bin/
 bin-debug/
 bin-release/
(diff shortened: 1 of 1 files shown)


shiftkey:
@BlueHatbRit thanks!

PR #1947, opened 2016-04-18
ignore .idea folder generated by PyCharm

zymtech:

ignore .idea folder generated by PyCharm


shiftkey:
Thanks, but the full set of rules for JetBrains' IDEs are maintained in [Global/JetBrains.gitignore](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore). I recommend tailoring your setup by combining both these templates.

PR #1949, opened 2016-04-21
Add nyc test coverage output directory

ro31337:

[nyc](https://github.com/bcoe/nyc#require-additional-modules) is quite popular. And it outputs to `.nyc_output` directory under some circumstances.

This directory is ignored in [nyc itself](https://github.com/bcoe/nyc/blob/master/.gitignore)


shiftkey:
@ro31337 thanks!

PR #1950, opened 2016-04-22
Create Julia.gitignore

hlaaftana:

Provide a template for Julia projects

If this is a new template: 

 - **Link to application or project’s homepage**: http://julialang.org/

### DIFF
--- a/Julia.gitignore
+++ b/Julia.gitignore
@@ -1,0 +1,3 @@
+*.jl.cov
+*.jl.*.cov
+*.jl.mem


2Cubed:
:+1: Great idea.

jiahao:
I would also consider adding `deps/deps.jl`, an autogenerated file for binary dependencies.

shiftkey:
Thanks! I'll take this in and leave it to others to see if they want to refine it further.

PR #1951, opened 2016-04-22
[Autotools] additional generated files

Irfy:


shiftkey:
@Irfy any links to StackOverflow discussions or other documentation about this change?

shiftkey:
Closing this out due to inactivity (and conflicts). @Irfy if you're able to provide more context about the change I'm happy to revisit this PR.

Irfy:
Oh, this was 2.5 years ago... Let me quickly go over my commit:

- `config.log` has been added
- `config.status` has been added
- `config.cache` is generated when the caching feature is explicitly used. Its contents can sometimes go stale, and according to it's [documentation](https://www.gnu.org/software/autoconf/manual/autoconf-2.65/html_node/Cache-Files.html) **_It is wrong to try to distribute cache files for particular system types._** This file is *very* useful to speed up a hierarchy of projects that refer to the same config.cache for cached results, but it never makes sense to keep track of it.
- `Makefile` is always generated from `Makefile.in` which is in turn generated from `Makefile.am`. It makes no sense to leave `Makefile` out, when `Makefile.in` is ignored (correctly), since the latter is the basis for generating the former. It should be without a leading slash, because the Makefile and Makefile.in are typically generated in subdirectories.

Basically it makes a lot of sense to ignore these two files in autotools projects.

The rest of my commit were comments as instructions to autotools users:

- [`AC_CONFIG_AUX_DIR`](https://www.gnu.org/software/automake/manual/html_node/Optional.html) this autoconf directive is used to reduce clutter in the top-level directory of the project -- it allows one to put all auxiliary/helper files/scripts in a separate directory to move them out of the way. These helpers are generated if absent, so it makes sense to ignore the dire specified as the argument to this directive.
- [`Recursive configure`](https://www.gnu.org/software/automake/manual/html_node/Nested-Packages.html) instructions refer to having a hierarchy of autoconf subprojects in a single repository. In that case it makes sense to ignore the autoconf files in all of the subdirs from one root .gitignore by removing leading slashes.
- `Not needed if configuring out-of-tree` is an instruction to the user that building out of tree will not generate listed files in the source directory, so ignoring them is not needed. In that case, omitting them may even be useful, because when they appear in `git status` output, the user will be reminded that s/he made a mistake (GCC mandates building out-of-tree, for example).

So, what would you like to do from here?

shiftkey:
> So, what would you like to do from here?

Let's start fresh - relevant comments in the template and a decent writeup in the PR I think is all I need.

PR #1952, opened 2016-04-23
add Eclipse gitignore

gooree:

Add gitignore file for Eclipse


shiftkey:
@gooree thanks, but we have an existing [Global/Eclipse.gitignore](https://github.com/github/gitignore/blob/master/Global/Eclipse.gitignore) - please add any improvements there...

PR #1954, opened 2016-04-26
Update TeX.gitignore

seungwonpark:

When we use 'thmtools' package for creating 'List of Theorems', auxiliary file '.loe' is created, just as '.toc' for 'Table of Contents' and '.lof' for 'List of Figures'.

Example : 

https://gist.github.com/seungwonpark/759ab6110391bfaa34a066984a7a920e

### DIFF
--- a/TeX.gitignore
+++ b/TeX.gitignore
@@ -141,6 +141,9 @@ sympy-plots-for-*.tex/
 *.pytxcode
 pythontex-files-*/
 
+# thmtools
+*.loe
+
 # TikZ & PGF
 *.dpth
 *.md5


shiftkey:
@seungwonpark I think I understand this change. Could you provide an example or snippet of what's contained in that `*.loe` file, and explain why it's safe to exclude from version control?

seungwonpark:
`*.loe` file is very similar to `*.toc` file. They are created from `*.tex` file each time when compiled, so they doesn't have to be included in version control. Following gist shows the example of `*.loe` and `*.toc` file, and results generated from them.
https://gist.github.com/seungwonpark/0221cd45c6b067a4d2c8dd0612a731e7

shiftkey:
@seungwonpark excellent, thanks!

PR #1955, opened 2016-04-27
Update MicrosoftOffice.gitignore

jgm1986:

 - Microsoft Visio is a tool included on Microsoft Office and it creates temporary files when you are working in a file.

 - [Microsoft Visio File Format](https://msdn.microsoft.com/en-us/library/office/jj228622.aspx)

If this is a new template: 

 - [Microsoft Visio 2016 Trial](https://www.microsoft.com/en-us/evalcenter/evaluate-visio-professional-2016/)

### DIFF
--- a/Global/MicrosoftOffice.gitignore
+++ b/Global/MicrosoftOffice.gitignore
@@ -11,3 +11,6 @@
 
 # PowerPoint temporary
 ~$*.ppt*
+
+# Visio temporary
+*.~vsdx
--- a/Global/MicrosoftOffice.gitignore
+++ b/Global/MicrosoftOffice.gitignore
@@ -12,5 +12,5 @@
 # PowerPoint temporary
 ~$*.ppt*
 
-# Visio temporary
+# Visio autosave temporary files
 *.~vsdx


shiftkey:
@jgm1986  the example link talks about the other extensions from Visio - is `~ vsdx` the right one to ignore here?

jgm1986:
Sorry I have not found any official place where explain the Visio  temporary files extension.

visio temporal file

I have attached an example of my own computer. As you can see, the original Visio extension file is **.vsdx** but the *autosave temporary file* contains the **.~vsdx** extension.

Global/MicrosoftOffice.gitignore
```diff
@@ -11,3 +11,6 @@
 
 # PowerPoint temporary
 ~$*.ppt*
+
+# Visio temporary
```
shiftkey: Can we make this a bit more clearer? If it is related to autosave then perhaps something like this:

```
# Visio autosave temporary files
```

shiftkey:
@jgm1986 thanks, that should be good enough for referring back to later.

Just one comment about the comment above this rule and this should be good to go.

jgm1986:
Sure! How about this? ;)

shiftkey:
:shipit: thanks @jgm1986!

PR #1957, opened 2016-05-01
Add new ignore rules for Eclipse

gooree:

Add new ignore rules for Eclipse


Global/Eclipse.gitignore
```diff
@@ -49,3 +49,15 @@ local.properties
...
+# Java class files
+**/*.class
+**/classes
+**/classes/**
+
+# svn plugin
+**/.svn
+**/.svn/**
+
+# virtual machine crash logs
```
shiftkey: Which sort of virtual machine is this?

Global/Eclipse.gitignore
```diff
@@ -49,3 +49,15 @@ local.properties
 
 # Code Recommenders
 .recommenders/
+
+# Java class files
+**/*.class
```
shiftkey: This rule is captured by our [`Java.gitignore`](https://github.com/github/gitignore/blob/master/Java.gitignore#L1) - the expectation here is that users will grab whatever templates they need based on their usage, so I'd like to avoid duplication wherever possible.

Global/Eclipse.gitignore
```diff
@@ -49,3 +49,15 @@ local.properties
...
+**/classes
```
shiftkey: I'm not sure on these rules - could you tell me a bit about where they come from?

Global/Eclipse.gitignore
```diff
@@ -49,3 +49,15 @@ local.properties
...
+**/classes/**
+
+# svn plugin
+**/.svn
```
shiftkey: We have a precedence here around externalizing other VCSes in their own templates - for example [`Mercurial.gitignore`](https://github.com/github/gitignore/blob/master/Global/Mercurial.gitignore).

We don't have one for Subversion though - do you think that's a suitable compromise here?

gooree: Yeah,may be we'd better make a new gitignore file for Subversion plugin or put them together.






At 2016-05-01 16:59:49, "Brendan Forster" <notifications@github.com> wrote:


In Global/Eclipse.gitignore:

> @@ -49,3 +49,15 @@ local.properties
>  
>  # Code Recommenders
>  .recommenders/
> +
> +# Java class files
> +**/*.class
> +**/classes
> +**/classes/**
> +
> +# svn plugin
> +**/.svn


We have a precedence here around externalizing other VCSes in their own templates - for example Mercurial.gitignore.

We don't have one for Subversion though - do you think that's a suitable compromise here?

Global/Eclipse.gitignore
gooree: Actually,I found that the class files could be anywhere if you change the default output folder settings for specific java project.So do you think we should ignore all these class files wherever they are located in.






At 2016-05-01 16:57:41, "Brendan Forster" <notifications@github.com> wrote:


In Global/Eclipse.gitignore:

> @@ -49,3 +49,15 @@ local.properties
>  
>  # Code Recommenders
>  .recommenders/
> +
> +# Java class files
> +**/*.class
> +**/classes


I'm not sure on these rules - could you tell me a bit about where they come from?

Global/Eclipse.gitignore
gooree: Sorry,I've made a spelling mistake.It should be java virtual machine crash logs.




在 2016-05-01 16:56:20，"Brendan Forster" <notifications@github.com> 写道：


In Global/Eclipse.gitignore:

> @@ -49,3 +49,15 @@ local.properties
>  
>  # Code Recommenders
>  .recommenders/
> +
> +# Java class files
> +**/*.class
> +**/classes
> +**/classes/**
> +
> +# svn plugin
> +**/.svn
> +**/.svn/**
> +
> +# virtual machine crash logs


Which sort of virtual machine is this?

Global/Eclipse.gitignore
gooree: It's not a duplicate rule because the class files could locate in any subdirectories under our java project folder.





At 2016-05-01 16:57:23, "Brendan Forster" <notifications@github.com> wrote:


In Global/Eclipse.gitignore:

> @@ -49,3 +49,15 @@ local.properties
>  
>  # Code Recommenders
>  .recommenders/
> +
> +# Java class files
> +**/*.class


This rule is captured by our Java.gitignore - the expectation here is that users will grab whatever templates they need based on their usage, so I'd like to avoid duplication wherever possible.

shiftkey:

PR #1959, opened 2016-05-04
Add sublime-text-3 gitignore for package control

Javierieh:

Add a .gitignore for Sublime Text 3 package control

https://packagecontrol.io/docs/syncing
https://www.sublimetext.com/3

------------------------------------------------------

This gitignore is used to make a repo with sublime text 3 package preferences stored in
{sublime-text-3}/Packages/User/

https://www.sublimetext.com/3

The .gitignore files are listed here

https://packagecontrol.io/docs/syncing

I added GitHub.sublime-settings to the .gitignore since its a config file added
from a popular github package that contains sensible information about your account

### DIFF
--- a/Global/SublimeText.gitignore
+++ b/Global/SublimeText.gitignore
@@ -12,3 +12,16 @@
 
 # sftp configuration file
 sftp-config.json
+
+# Package control specific files
+Package Control.last-run
+Package Control.ca-list
+Package Control.ca-bundle
+Package Control.system-ca-bundle
+Package Control.cache/
+Package Control.ca-certs/
+bh_unicode_properties.cache
+
+# Sublime-github package stores a github token in this file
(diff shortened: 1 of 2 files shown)


shiftkey:
@Javierieh could you move this file under the `Global` folder? That's where we're collecting our IDE-based templates...

shiftkey:
And as soon as I posted that I realised I should have also checked for myself - there's a [Global\SublimeText.gitignore](https://github.com/github/gitignore/blob/master/Global/SublimeText.gitignore) template.

It doesn't seem to overlap much with what you've defined here, but I'd suggest tweaking that to incorporate these new rules...

Javierieh:
Hey @shiftkey !
Thanks for your quick reply.

I moved the changes to Global/SublimeText.gitignore

shiftkey:
@Javierieh thanks!

PR #1962, opened 2016-05-06
Add .com.apple.timemachine.donotpresent to OSX

skycaptain:

On OSX if an external volume is excluded from being available as a TimeMachine drive, a system file named *.com.apple.timemachine.donotpresent* will be created at the volume's root. Since this is an user choice, there is no need for this file to be included in a git-repo.


shiftkey:
Seems legit. Thanks @skycaptain!

PR #1963, opened 2016-05-06
ignore .dSYM along with .ipa files

evfemist:

Often .dSYM and .dSYM files stored with .ipa files, so they should also be ignored if .ipa ignored.

_Not clear what should be here._

You can get more info what is .dSYM file [here](http://stackoverflow.com/questions/3656391/whats-the-dsym-and-how-to-use-it-ios-sdk)


shiftkey:
@evfemist that SO link helps, thanks!

PR #1964, opened 2016-05-08
Add .swp files

OrkoHunter:

`.swp` files are created by Vim while editing any file. I mainly face this while developing something in Python, so I'm adding it in `Python.gitignore` but should be in all of the files, I believe.


shiftkey:
Thanks, but see [Global\Vim.gitignore](https://github.com/github/gitignore/blob/master/Global/Vim.gitignore) which has this and some other rules.

PR #1967, opened 2016-05-10
Ignore *.DS_Store

Kagamine:

To ignore the .DS_Store file which generated by OSX. Many developers were using Visual Studio in a virtual machine, such as Parallels Desktop. So, their system will generate many .DS_Store files. However, we do not need these files when we commit to the git repository.

/cc @Eilon @shiftkey

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -250,3 +250,6 @@ paket-files/
 # JetBrains Rider
 .idea/
 *.sln.iml
+
+# OSX ds_stores
+*.DS_Store


shiftkey:
Thanks, but in the past I've pushed back on adding OS-specific changes to each template.

We have [Global/OSX.gitignore](https://github.com/github/gitignore/blob/master/Global/OSX.gitignore) for this and the repository is structured so it's easy for users to tailor templates based on their setup - including things like OS and editors.

PR #1968, opened 2016-05-10
Create MASM assembly (ASM) gitignore file

noraj:

there is no ASM gitignore and no MASM in particular

I haven't

If this is a new template: 
- **Link to application or project’s homepage**: https://github.com/shark-oxi/MASM-assembly-beginner/

MASM assembly (ASM) gitignore file
For example under windows keep *.bat (makefiles) and *.asm (ASM source code) and ignore *.obj (Object files) and *.exe (Executables)
There is no no need for linux output because MASM is windows only.

### DIFF
--- a/MASM.gitignore
+++ b/MASM.gitignore
@@ -1,0 +1,5 @@
+# Object files
+*.obj
+
+# Executables
+*.exe


shiftkey:

@noraj if you still feel like this should be a template listed in this repository, please read the section about [specialized templates](https://github.com/github/gitignore#specialized-templates) and open a new pull request that follows the new format.

PR #1969, opened 2016-05-10
Ignore '*.DS_Store' instead of ignore '.DS_Store'

Kagamine:

There are many .DS_Store files in subfolders. We still should ignore these files.

http://stackoverflow.com/questions/18393498/gitignore-all-the-ds-store-files-in-every-folder-and-subfolder

### DIFF
--- a/Global/OSX.gitignore
+++ b/Global/OSX.gitignore
@@ -1,9 +1,10 @@
-.DS_Store
+*.DS_Store
 .AppleDouble
 .LSOverride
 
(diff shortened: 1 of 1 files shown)


shiftkey:
I can't take this change in as-is for a couple of reasons:

 - the lack of context around making this change
 - the line ending changes impact one of the rules - the `# Icon must end with two \r` line

Kagamine:
Moved to #1972

PR #1970, opened 2016-05-10
Create GGTS.gitignore

docahrens:
There is no .gitignore for GGTS specifically, only for Grails.  The ones for Grails did not handle either eclipse/GGTS or the Windows operating system
_TODO_

web-app/WEB-INF/tld/c.tld
web-app/WEB-INF/tld/fmt.tld
suggested by: http://stackoverflow.com/users/451634/benny-neugebauer in this post:
http://stackoverflow.com/questions/470833/suggestions-for-grails-gitignore
_TODO_

If this is a new template: 


shiftkey:
> The ones for Grails did not handle either eclipse/GGTS or the Windows operating system

We have existing templates for Eclipse and Windows under the `Global` directory, and I would recommend composing your project templates from those as well as `Grails` rather than duplicating templates here to add additional maintenance burden.

PR #1972, opened 2016-05-11
Ignore '*.DS_Store' instead of ignore '.DS_Store'

Kagamine:

There are many .DS_Store files in subfolders. We still should ignore these files.

http://stackoverflow.com/questions/18393498/gitignore-all-the-ds-store-files-in-every-folder-and-subfolder

/cc @shiftkey

### DIFF
--- a/Global/OSX.gitignore
+++ b/Global/OSX.gitignore
@@ -1,4 +1,4 @@
-.DS_Store
+*.DS_Store
 .AppleDouble
 .LSOverride
 


shiftkey:
@Kagamine the SO link shows that some people use the `*.DS_Store` rule and others are just happy with `.DS_Store` - I'd love to know a bit more about the format here of these files that are being generated by the OS...

Kagamine:
.DS_Store files will be appeared in any folder, because the file contains the attributes of the folder.

https://support.apple.com/zh-cn/ht1629
http://osxdaily.com/2009/12/31/what-is-a-ds_store-file/

shiftkey:
@Kagamine cool, but is there a different filename convention that isn't covered by the current `.DS_Store` rule? And what do the filenames then look like? That's the gap I'm trying to fill in my understanding here...

Kagamine:
Maybe there is an another one which named `._.DS_Store`. 

http://www.rcseng.ac.uk/publications/docs/annual-report-2009-10/MACOSX/buttons/._.DS_Store/view

But I think `._.DS_Store` was included in `*.DS_Store`.

shiftkey:
@Kagamine I'm still curious why I haven't heard more about this from users, but given this new rule will also catch existing `.DS_Store` files I'm happy to take this in :metal:

Thanks for clarifying the questions I had!

PR #1885, opened 2016-02-28
added dotenv to python gitignore

roll:
**Reasons for making this change:**

`.env` is a common choice to store local environment variables based on `Heroku/foreman` workflow.

**Links to documentation supporting these rule changes:** 

https://github.com/theskumar/python-dotenv

PR #1973, opened 2016-05-12
Add .env to optionally ignored files

maxehmookau:
Many twelve-factor apps use environment variables rather than the built in Rails secrets mechanism to store secrets.
The [dotenv](https://github.com/bkeepers/dotenv) gem is widely used for this purpose and allows environment variables to be loaded from the `.env` file. However, this file should not be committed to source control.

This PR just adds the `.env` file to `.gitignore` so that these secrets aren't accidentally committed. In some rare instances, it may be required that this file is committed to source control so I've added it to the list of files that can be removed if the developer is ok with secrets being committed.

### DIFF
--- a/Rails.gitignore
+++ b/Rails.gitignore
@@ -15,6 +15,9 @@ pickle-email-*.html
 # TODO Comment out these rules if you are OK with secrets being uploaded to the repo
 config/initializers/secret_token.rb
 config/secrets.yml
+
+# dotenv
+# TODO Comment out this rule if environment variables can be committed
 .env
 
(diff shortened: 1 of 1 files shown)


shiftkey:
@maxehmookau could we annotate this rule like we did with the Python PR #1885 so that it's a bit clearer what this is about?

maxehmookau:
@shiftkey Sure, is that better? :)

shiftkey:
@maxehmookau yep, that looks good! Thanks!

maxehmookau:
Great, thanks!

PR #1974, opened 2016-05-12
Ignore Apache Subversion repositories

solomos:
Ignore leftovers after moving from svn to git


shiftkey:
Thanks, but we track VCS-specific rules under the `Global` folder - here's the [SVN](https://github.com/github/gitignore/blob/master/Global/SVN.gitignore) template...

solomos:
I was refering to any SVN leftovers in VS projects.

VS project, more that often, contain the ".SVN" folder that was used before moving to GIT.

PR #1975, opened 2016-05-12
ignore Eclipse Checkstyle Plugin temporary configuration file

vorburger:

http://eclipse-cs.sourceforge.net/ creates temporary .checkstyle files, which should never be in source control.

### DIFF
--- a/Java.gitignore
+++ b/Java.gitignore
@@ -8,5 +8,8 @@
 *.war
 *.ear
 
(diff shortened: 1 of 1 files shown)


drothmaler:
This should go into [Global/Eclipse.gitignore](https://github.com/github/gitignore/blob/master/Global/Eclipse.gitignore)

shiftkey:
I agree with what @drothmaler said here. Aside from that, this change seems reasonable to include there.

PR #1976, opened 2016-05-12
Ignore target/ build/ bin/ output directories

vorburger:

target/ build/ bin/ are very typical output directories in Java projects, which should always be ignored (they are the default directory names for output in Maven, Gradle and Eclipse, respectively).

### DIFF
--- a/Java.gitignore
+++ b/Java.gitignore
@@ -1,4 +1,7 @@
 *.class
+target/
+build/
+bin/
 
 # Mobile Tools for Java (J2ME)
 .mtj.tmp/


drothmaler:
No, there are the specific [Maven.gitignore](https://github.com/github/gitignore/blob/master/Maven.gitignore) [Gradle.gitignore](https://github.com/github/gitignore/blob/master/Gradle.gitignore) and [Global/Eclipse.gitignore](https://github.com/github/gitignore/blob/master/Global/Eclipse.gitignore) rules for that.

shiftkey:
Seconding what @drothmaler says. These are IDE-specific rules, and should be captured there (if not already).

PR #1977, opened 2016-05-12
Ignore .gradle

vorburger:

.gradle/ is the temporary directory for Gradle Wrapper cache downloads. see also e.g. http://stackoverflow.com/questions/17200831/should-the-gradle-folder-be-added-to-version-control

### DIFF
--- a/Java.gitignore
+++ b/Java.gitignore
@@ -8,5 +8,8 @@
 *.war
 *.ear
 
+# Temporary directory for Gradle Wrapper
+.gradle/
+
 # virtual machine crash logs, see http://www.java.com/en/download/help/error_hotspot.xml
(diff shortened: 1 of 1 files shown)


drothmaler:
There already is a [Gradle.gitignore](https://github.com/github/gitignore/blob/master/Gradle.gitignore) for this purpose. Most of the rules inside this repo are modular. There are seperate gitignore files for Java, Gradle, Maven, SBT and so on. You have to combine them according to you need in you local project setup. 

You can also use https://www.gitignore.io to make this a bit easier :-)

shiftkey:
Seconding @drothmaler's feedback - please use `Gradle.gitignore` for any Gradle-specific changes...

PR #1978, opened 2016-05-12
[Java] Ignore Xtend generated files

vorburger:

The xtend-gen/ directory and *._trace & *.xtendbin files are generated by http://xtend-lang.org and should typically not be kept under version control.


drothmaler:
I'm not sure about that... I see Xtend as an extra Language/Transpiler. So might be better to move this to a new `Xtend.gitignore`. (Like Scala or GWT have extra files)

shiftkey:
@vorburger any thoughts on the question that @drothmaler raised?

shiftkey:
Closing as inactive with pending feedback.

PR #1980, opened 2016-05-14
Update .emacs.d 

zonuexe:

This file is missing definition for Emacs standard lisp package and famous packages.

- https://github.com/auto-complete/auto-complete/blob/master/auto-complete.el#L134-L141
- https://github.com/knupfer/haskell-emacs/blob/master/haskell-emacs.el#L48-L51
- http://git.savannah.gnu.org/cgit/emms.git/tree/lisp/emms.el#n286
- https://github.com/jwiegley/emacs-async/blob/master/async-bytecomp.el#L54
- https://github.com/sigma/pcache/blob/master/pcache.el#L61
- https://github.com/kiwanami/emacs-edbi/blob/master/edbi.el#L901
- https://github.com/nonsequitur/smex/blob/master/smex.el#L44
- https://github.com/tkf/emacs-request/blob/master/request.el#L62


Global/Emacs.gitignore
```diff
@@ -7,6 +7,13 @@
 auto-save-list
 tramp
 .\#*
+/games/
```
shiftkey: It's not clear to me where these rules are coming from, and what they represent. Could you elaborate?

shiftkey:
@zonuexe conflicted currently, apologies! :cry:

zonuexe:
@shiftkey I added comment to each lines.  Is this work as you expected?

zonuexe:
@shiftkey Could you review this PR?

shiftkey:
@zonuexe apologies for missing this. The problem with this big PR is that other smaller PRs are easier to review and they land first, causing conflicts again with your branch.

andreineculau:
I bumped into this PR, after finding out that Emacs.gitignore ignores things like the `server` folder or `dist` folder 🤔 

This changeset only makes it worse. `request`, `var`, `games`, `bookmarks`, `tutorial` would all be ignored in repos that have nothing to do with emacs.

It's great to consolidate templates and reuse, but at a much granular level, and at a much stricter level. If the `server` folder would have made it into the vscode-gitignore template, you'd probably have a riot.

aabho:
_ISSUES_

There are many issues identified with this, including but not limited to the few attributed from within an adjacent pull node.

wirecat:
@zonuexe Are you still interested in moving forward with this PR?
As it stands, I see concerns with the large number of additions to the emacs template here. Could we work on reducing the change to a smaller set of changes?

PR #1981, opened 2016-05-15
Update VisualStudio.gitignore

Kagamine:

Kore Build will create a temp folder named `.build`.

https://github.com/aspnet/KoreBuild/blob/dev/template/build.sh#L10

/cc @shiftkey

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -252,4 +252,4 @@ paket-files/
 *.sln.iml
 
 # KoreBuild
-*.build
+*.build/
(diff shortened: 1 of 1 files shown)


shiftkey:
This project seems to be about build scripts, rather than a tool or something more broadly used - does this need to be applied to all templates?

Kagamine:
@shiftkey many projects are using the build script:

github.com/aspnet/mvc

I suppose the .Net Core projects need this, if they are binding to a CI platform.

PR #1983, opened 2016-05-17
Add PyCharm work folder to python ignore

Vic020:
PyCharm usually generate ".idea" folder to record Project info, configuration and etc.

Python Project need to ignore this folder

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -87,3 +87,6 @@ ENV/
 
 # Rope project settings
 .ropeproject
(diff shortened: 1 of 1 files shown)


shiftkey:
Thanks, but [`Global/JetBrains.gitignore`](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore) documents the specific rules for files that should be shared, versus files that should be excluded.

PR #1984, opened 2016-05-17
Merge conflict left over (.orig) files to be ignored

ismusidhu:

Whenever there are merge conflicts and if it is resolved from command line, .orig files are left in workspace which then gets added to source control by mistake. This is a common problem which many teams have.


shiftkey:
We don't do this for other templates, so I suggest you setup a [global gitignore](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore) file to apply this to all your repositories.

PR #1988, opened 2016-05-18
Create Batch.gitignore

jtrent238:

Added Batch

_TODO_

If this is a new template: 

### DIFF
--- a/Batch.gitignore
+++ b/Batch.gitignore
@@ -1,0 +1,2 @@
+# BatchFiles
+*.bat
--- a/Batch.gitignore
+++ b/Batch.gitignore
@@ -1,2 +1,4 @@
 # BatchFiles
 *.bat
+*.cmd
+*.btm


shiftkey:
@jtrent238 can you tell me a bit about the motivation for adding this template?

jtrent238:
@shiftkey I use a lot of batch in my project so this could be helpful, and some other people might find it helpful also.

Arinerron:
@shiftkey I +1 this

Arinerron:
(though I am a Linux user)

gautamkrishnar:
@jtrent238  i am not sure about this. you are actually ignoring the batch script file. Why do you need to ignore that files?

gautamkrishnar:
If you are sure about this then don't forget to add the extensions like :  .`*.cmd` and `*.btm`

stephengroat:
Ignoring Batch files is ignoring the source files if you're a batch programmer (I've had to do some of it, it's rough)

I think that ignoring specific auto-generated batch files is more appropriate for auto-generated batch files from a specific build system or IDE, rather than a separate template

shiftkey:
Closing this out because it doesn't seem clear what problem this template is solving...

PR #1991, opened 2016-05-20
Add PyCharm project settings directory

niraj-dudani:
[Reference](https://www.jetbrains.com/help/pycharm/2016.1/project.html)

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -87,3 +87,6 @@ ENV/
 
 # Rope project settings
 .ropeproject
+
+# PyCharm project settings
+.idea/


shiftkey:
Thanks, but unfortunately it's a bit more complicated than this:

> All the settings files in the .idea directory should be put under version control except the `workspace.xml`, which stores your local preferences. The `workspace.xml` file should be marked as ignored by VCS.

And we have a [Global/JetBrains.gitignore](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore) rule which covers all the JetBrains IDEs (including PyCharm).

PR #1992, opened 2016-05-20
Added *.smod to Fortran.gitignore

jlokimlin:

GCC's gfortran 6 now generates .smod files for Fortran (2008+) submodules

https://gcc.gnu.org/gcc-6/changes.html


shiftkey:
@jlokimlin thanks!

PR #1995, opened 2016-05-24
Update Unity.gitignore

AliHammoud:

Added more useful ignores such as the problematic "Thumbs.db" & its meta for windows + other OS generated files.

Paths to library prefixed with wildcard, many unity projects have a gitignore at the root with different types of projects as sub-directories.

_Example case:_
> ProjectPath
|    > UnityProj
|    > JavaProj
|    > WebProj
|    .gitignore


AliHammoud:
*bump*

drothmaler:
-1 This repo manages gitignore files in a modular way... 
For the OS-Files there are `Global/Windows.gitignore`, `Global/OSX.gitignore` and `Global/Linux.gitignore`. 
And regarding the other additions: As far as I know, Unity is a multiplatform engine, that can be uses from with many different programming languages. So it makes no sense to put the gitignore-rules for all possible combinations in here. The `Unity.gitignore` should only contain files, that are really specific to Unity. 
You should instead combine the existing gitignores, to match your person project setup. http://gitignore.io/ (which uses the rules from this repo) provides a really easy way to do this.

Regarding the generalization of the directory paths: This depends on how tightly coupled the Projects are... I'm not too familiar with Unity, so I'm not sure about this. 
I would probably put this `Unity.gitignore` file directly into the `UnityProj` dir, the `Java.gitignore` in that `JavaProj` dir and so on. Or I would manage the subproject separately, using `git-submodule`, `git-subrepo` or `git-subtree`. My rule of thumb is: If it uses an extra Toolchain/IDE, it should be in an extra (sub)repo.

shiftkey:
:thumbsdown: on including OS-specific files

A couple of questions I have:

 - can you clarify the motivation for making the Unity project folders more generic?
 - I'm not clear on how useful the new sections are - is this worth a separate PR to discuss that further?

AliHammoud:
Oh god, how did I miss that there is a https://www.gitignore.io/

Okay, okay, closing this branch! :))))

PR #1999, opened 2016-05-27
For Mac OS

iFallen:

_TODO_

_TODO_

If this is a new template: 


shiftkey:
Thanks, but we're using `Global/OSX.gitignore` to track these rules...

PR #2002, opened 2016-05-28
Add AppCode .idea folder

Tokuriku:

The number of developers using AppCode is non trivial so their needs should probably be taken into account.

*\* Actual Changes**
The files in the .idea folder are either user-specific, sensitive or high churn files.
They are probably best not committed to the repo.


shiftkey:
The JetBrains IDEs are consistent in their use of the `.idea` folder, and well-documented in terms of what can be shared. I'm going to defer to them.

PR #2004, opened 2016-06-02
Add the PyCharm project settings: .idea/

assem-ch:

PyCharm is a well-known python IDE , and it creates this folder .idea for the settings.

https://stackoverflow.com/questions/5362816/should-the-idea-folder-be-kept-under-source-control


mrburrito:
All of the JetBrains IDE's are covered in the Global file for JetBrains:

PR #2006, opened 2016-06-04
Update Visual Studio gitignore to include vscode

jfaquinojr:
added ignore rule for visual studio code


shiftkey:
Thanks, but see [Global/VisualStudioCode.gitignore](https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore)

jfaquinojr:
thanks for pointing it out.

PR #2008, opened 2016-06-06
white space fix

suszczynski:
it fix also the vim issue related with this file:
Error detected while processing function <SNR>20_WildignoreFromGitignore:
E518: Unknown option

### DIFF
--- a/Global/OSX.gitignore
+++ b/Global/OSX.gitignore
@@ -21,6 +21,10 @@ Icon
 # Directories potentially created on remote AFP share
 .AppleDB
 .AppleDesktop
-Network\ Trash\ Folder
-Temporary\ Items
+Network?Trash?Folder
+
+
+Temporary?Items
+
+
 .apdisk
--- a/Global/OSX.gitignore
+++ b/Global/OSX.gitignore
@@ -3,8 +3,7 @@
 .LSOverride
 
 # Icon must end with two \r
-Icon
-
+Icon^M^M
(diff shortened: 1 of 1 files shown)


shiftkey:
@suszczynski I think this change is affecting the rule defined here:

```
# Icon must end with two \r
Icon

```

The whitespace changes are important to retain here because of the `\r\r` ending.

Could you tell me a bit about the rule that this change is intended to fix?

suszczynski:
There is a small conflict with this plugin: http://www.vim.org/scripts/script.php?script_id=2557 besides that white spaces could also negatively affect the Linux users who have to work with OSX files.

https://github.com/github/gitignore/pull/2008/files#diff-235aa03516afdbe9fc516e7b1bec4395

shiftkey:
For reference, this is the original PR which explains the rule in some more detail: https://github.com/github/gitignore/pull/804

suszczynski:
@shiftkey I'm sorry I mess up a bit with this file my intention was to only change the line no 23-24 to this:
```
Network\ Trash\ Folder
Temporary\ Items
```

shiftkey:
@suszczynski no problem - you're more than welcome to start a fresh PR rather than continuing this one.

suszczynski:
done https://github.com/github/gitignore/pull/2011
btw. there is some problem with the github web UI file editor, I had to fix this file in my system app

PR #2012, opened 2016-06-08
Galaxycoder patch 1

galaxycoder:

Add editor support for AppCode, files in the .idea folder are either user-specific, sensitive or high churn files.
They are probably best not committed to the repo.

[JetBrains Git Ignore](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore)

If this is a new template: 

 - **Link to application or project’s homepage**: [AppCode](https://www.jetbrains.com/objc/)

### DIFF
--- a/Global/AppCode.gitignore
+++ b/Global/AppCode.gitignore
@@ -1,0 +1,5 @@
+# AppCode
+#
+# Ignore files produced by AppCode
+
+.idea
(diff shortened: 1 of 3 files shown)


shiftkey:
Thanks, but the `JetBrains.gitignore` template has more specific rules than just `.idea`.

PR #2016, opened 2016-06-15
Ignore all forms of CMakeLists.txt.user

nocnokneo:
In some cases Qt Creator will create a CMakeLists.txt.user file with a short hash which should also be ignored. For example:

    CMakeLists.txt.user.1fa15d5


shiftkey:
@nocnokneo thanks!

PR #2017, opened 2016-06-15
ignore *.VC.db files

NuLL3rr0r:

In addition to *.VC.opendb, *.VC.db file should be regenerated by Visual Studio automatically. So, it makes sense to ignore *.VC.db file too as it's been ignored in https://github.com/github/gitignore/blob/master/VisualStudio.gitignore

https://github.com/github/gitignore/blob/master/VisualStudio.gitignore

If this is a new template: 

In addition to *.VC.opendb, *.VC.db file should be regenerated by Visual Studio automatically. So, it makes sense to ignore *.VC.db file too as it's been ignored in https://github.com/github/gitignore/blob/master/VisualStudio.gitignore

### DIFF
--- a/UnrealEngine.gitignore
+++ b/UnrealEngine.gitignore
@@ -37,6 +37,7 @@
 *.suo
 *.opensdf
 *.sdf
+*.VC.db
 *.VC.opendb
 
 # Precompiled Assets


shiftkey:
@NuLL3rr0r thanks!

PR #2018, opened 2016-06-16
update android gitignore

banking:

Primary Android.gitignore not contains some gradle temp files,Eclipse setting files,etc.
These commits improve these.

According to my project experience.

If this is a new template: 

  - **Link to application or project’s homepage**:https://github.com/banking/AppConfig/


stephengroat:
I think you need to review some of the other gitignores in this repo. They can be stacked to achieve the same result

For example the [Eclipse.gitignore](https://github.com/github/gitignore/blob/master/Global/Eclipse.gitignore) has everything you need to block Eclipse settings from being checked in

shiftkey:
Seconding what @stephengroat said.

PR #2019, opened 2016-06-16
Update Python.gitignore for vim

cuongnv23:

Ignore all .swp/.swo/~ files created by vim 

Vim's so good and popular. `vim as python ide` returned over 410,000 results. It's worth to be added. Below link also got over 57260 views

http://stackoverflow.com/questions/4824188/git-ignore-vim-temporary-files

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -90,3 +90,5 @@ ENV/
 
 # Vim swp temp files
 *.swp
+*.swo
+*~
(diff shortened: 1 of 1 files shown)


shiftkey:
@cuongnv23 thanks, but we track the Vim rules in [`Global/Vim.gitignore`](https://github.com/github/gitignore/blob/master/Global/Vim.gitignore) and having to maintain it across multiple templates isn't feasible.

We recommend end users tailor their gitignore templates together based on their setup.

PR #2021, opened 2016-06-19
C/C++ .nix executables

RikusW:
Ignore Unix and Linux executables that otherwise would have to be added individually.
It is assumed there is no . in the executable name or path.


RikusW:
It will also ignore Makefiles....

PR #2023, opened 2016-06-21
Added project folder created by Consulo IDE to Unity.gitignore

rpistorello:

Consulo IDE is a fork of the IntelliJ IDEA Community Edition, and the best choice for OS X users since it can't run Visual Studio. With this change the programmer don't need to commit the files generated by Consulo, and the Unity project can be opened in witch IDE the user desire.

**Project’s links:** 

https://github.com/consulo/consulo
https://github.com/consulo/consulo/wiki
https://github.com/consulo/consulo/wiki/Downloads

**Files to be ignored:** 
The files are generated every time the user imports a Unity project to the Consulo IDE. So this is a work for every user/programmer and there is no reason to be commited/pushed.

These are all files generated by the IDE so ignoring the folder will do the job:
/.consulo/.name
/.consulo/compiler.xml
/.consulo/modules.xml
/.consulo/vcs.xml
/.consulo/workspace.xml
/.consulo/copyright/profiles_settings.xml


rpistorello:
Are you sure this is a new template?

shiftkey:
@rpistorello you're right - apologies for the oversight!

PR #2024, opened 2016-06-21
Template for ignore generated code from XText Framework

letientai299:
Add new template for support XText framework.

If this is a new template: 
- **Link to application or project’s homepage**:  https://eclipse.org/Xtext/


letientai299:
I've realized that 06e80e5842ea254d96a45c729dbca3e7e74564c5 is the same with my commit. So why push another commit instead of merging this?

PR #2026, opened 2016-06-24
Only ignore the build directory in the same directory as .gitignore for Gradle

mkobit:

If you have a package with name `com.mkobit.build` all files in that directory will be ignored.

None

If you have a Java package with `build` in it, the current version will always ignore files in that directory.
This change makes it so only the root "build" directory is ignored.

I believe that this assumes .gitignore is in the root of a project.

### DIFF
--- a/Gradle.gitignore
+++ b/Gradle.gitignore
@@ -1,5 +1,5 @@
 .gradle
-build/
+/build/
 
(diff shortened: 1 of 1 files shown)


shiftkey:
@mkobit thanks!

PR #2029, opened 2016-06-29
Update TeX.gitignore

thilaire:

Add support for easy-todo package (that produces .lod file during the compilation)

[https://www.ctan.org/tex-archive/macros/latex/contrib/easy-todo](https://www.ctan.org/tex-archive/macros/latex/contrib/easy-todo)


shiftkey:
@thilaire thanks!

PR #2030, opened 2016-06-29
added prerequisites for C and C++

pablomtz:

With GNU make the preferred way to deal with dependencies is to generate a file of prerequisites (i.e., a makefile) called name.d

https://www.gnu.org/software/make/manual/html_node/Automatic-Prerequisites.html

If this is a new template: 


shiftkey:
@pablomtz thanks!

PR #2034, opened 2016-07-02
Ignore *.pid.lock

ro31337:

Some popular applications (for example, Phusion Passenger) leave `*.pid.lock` file(s). For example, when you type `passenger start` with simple express app listening on port 3000, it leaves the following files:

* `passenger.3000.pid` (removed when passenger stops)
* `passenger.3000.log`
* `passenger.3000.pid.lock`

While `*pid` and `*.log` are ignored, `*.pid.lock` remains unignored. Phusion Passenger is quite popular, and `*.pid.lock` file should be ignored.

It's really easy to confirm this issue by googling the file extension: https://www.google.com/?ion=1&espv=2#q=phusion%20passenger%20pid.lock


shiftkey:
@ro31337 thanks!

PR #2039, opened 2016-07-09
Add *.wrt produced by scrwfile

koppor:

The [LaTeX package scrwfile](https://www.ctan.org/pkg/scrwfile?lang=de) generates `*.wrt` files, which should not be versioned.

http://mirrors.ctan.org/macros/latex/contrib/koma-script/doc/scrguien.pdf, Section "14.2. The Single File Feature", page 297.


shiftkey:
@koppor thanks!

PR #2044, opened 2016-07-14
Ignore terraform module directory

dochang:

The documentation suggests that do not commit `.terraform` directory.

https://www.terraform.io/docs/commands/get.html


shiftkey:
@dochang thanks!

PR #2049, opened 2016-07-25
Update Laravel.gitignore

cyrodjohn:
Ignore Homestead related config files


shiftkey:
These are discussed in #2192, and are documented upstream:

@cyrodjohn thanks!

PR #2053, opened 2016-07-29
Update VisualStudio.gitignore

douglasbruce88:
Added ignores for all .ncrunchproject and .ncrucnhsolution files.

These files will be specific to the user rather than the project.


shiftkey:

> It's recommended that the last two files listed in the above table (`.v2.ncrunchsolution` and `.v2.ncrunchproject`) be stored inside a source control system to be shared with team members.

douglasbruce88:
No problem, must have missed that bit of the guidance!

PR #2055, opened 2016-08-02
added .DS_Store

filipw:

Needed for folders shared between OS X and a Windows VM.


drothmaler:
there is `Global/OSX.gitignore` for that

shiftkey:
Due to the maintenance overhead of ensuring OS-specific changes are synchronized correctly across many templates, this isn't something we've supported in the repository.

Instead, I encourage users to pull in the interesting templates under the `Global` folder to tailor their setup to suit their needs.

PR #2056, opened 2016-08-02
Added NCrunch v2 files

fvilers:

NCrunch v2 adds different files that should be ignored.

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -115,6 +115,10 @@ _NCrunch_*
 .*crunch*.local.xml
 nCrunchTemp_*
 
+# NCrunch v2
+*.ncrunchsolution
+*.ncrunchproject
+
 # MightyMoose
 *.mm.*
 AutoTest.Net/


shiftkey:

> It's recommended that the last two files listed in the above table (`.v2.ncrunchsolution` and `.v2.ncrunchproject`) be stored inside a source control system to be shared with team members.

fvilers:
While I disagree (these files contains specific options that can be different for each developer), I understand your decision.

shiftkey:
> While I disagree (these files contains specific options that can be different for each developer), ...

cc @remcomulder

remcomulder:
The problem with not including the present .v2.ncrunchproject files in your VCS is that these files contain settings that critically control how the projects are built.  If you don't include them, people may need to work through quite a bit of configuration themselves every time they check out from VCS.  This depends entirely on the project itself.. it may well be that these projects can build fine with default settings.

There are some big changes coming in v3 that should help to make the configuration system more flexible in regards to user/shared settings.

PR #2058, opened 2016-08-06
liferay plugin project 

ismail03:
Used in liferay plugin projects development.


shiftkey:
Closing as too specialized. We have existing `Java`, `Maven` and `Eclipse` templates which should cover this scenario.

PR #2063, opened 2016-08-11
Exclude folder for CodeRush settings

pascalberger:
Exclude the folder for [CodeRush for Roslyn](https://visualstudiogallery.msdn.microsoft.com/8a8390ae-1f71-4659-9d8d-5dd56fd8a72e) settings. I assume it is OK to add this to the VisualStudio settings, since already other similar Tools (like eg R#) contained in this setting.


shiftkey:
@pascalberger thanks!

PR #2068, opened 2016-08-16
[Scala] exclude Scala IDE related .cache-main .

sherzodv:
Originating from https://issues.scala-lang.org/browse/SI-9506. Added exclusion for Eclipse as well.

### DIFF
--- a/Global/Eclipse.gitignore
+++ b/Global/Eclipse.gitignore
@@ -49,3 +49,6 @@ local.properties
 
 # Code Recommenders
 .recommenders/
+
+# Scala IDE specific (Scala & Java development for Eclipse)
+.cache-main
--- a/Scala.gitignore
+++ b/Scala.gitignore
@@ -15,6 +15,7 @@ project/plugins/project/
 # Scala-IDE specific
 .scala_dependencies
 .worksheet
+.cache-main
 
 # ENSIME specific
 .ensime_cache/
--- a/Scala.gitignore
+++ b/Scala.gitignore
@@ -15,7 +15,6 @@ project/plugins/project/
 # Scala-IDE specific
 .scala_dependencies
 .worksheet
-.cache-main
 
 # ENSIME specific
 .ensime_cache/


Global/Eclipse.gitignore
```diff
@@ -49,3 +49,6 @@ local.properties
 
 # Code Recommenders
 .recommenders/
+
+# Scala IDE specific (Scala & Java development for Eclipse)
```
shiftkey: If this is already covered by `Scala.gitignore`, this seems to be redundant, right?

sherzodv: Generally right.. Currently most of Scala.gitignore consists of excludes for external tools (build tools, editors, IDEs), this is not an argument to have this in it though.

But I would ask for this redundancy to cover rare corner cases, when for example one opens project without local .gitignore in it.

If you are insisting on keeping Scala.gitignore clean, I'll rework my commits a little.

shiftkey: @sherzodv yes, please move any IDE-specific rules to the relevant configuration files under `Global/`. This prevents duplication and makes maintenance easier.

This is also mentioned in the [`CONTRIBUTING.md`](https://github.com/github/gitignore/blob/master/CONTRIBUTING.md) document:

> **Please only modify one template per pull request.** This helps keep pull requests and feedback focused on a specific project or technology.

sherzodv: @shiftkey, please take a look. I've reverted a change to Scala.gitignore, so there is only a change to Eclipse.gitignore in this PR.

I'll rework Scala.gitignore in another PR: will it be OK to have all moves in a single PR, or should I remove lines from Scala.gitignore in one PR and add them to specific Global sections in other ones?

shiftkey: @sherzodv let's keep changes to one file at a time - feel free to link related PRs so everyone can follow along

shiftkey:
@sherzodv thanks!

PR #2070, opened 2016-08-16
Ignore GNU Global GSYMS database of other symbols

ymartin59:

Support older versions of gnu global, for instance Debian Jessie includes 5.7.1


shiftkey:
@ymartin59 thanks!

PR #2072, opened 2016-08-19
Update Android.gitignore

vsona:

# except mac system file DS_Store

_TODO_

If this is a new template: 

add # except mac system file


shiftkey:
@vsona thanks, but we track all the OS X-specific rules in `Global/OSX.gitignore`

PR #2073, opened 2016-08-19
Fix cache/*/index.php unignored (for git < v2.9)

Klemart3D:

My cache/*/index.php files were not unignored because I'm on git v1.9 and it not unignored files included in a folder excluded. For more explanation with  : http://stackoverflow.com/questions/5533050/gitignore-exclude-folder-but-include-specific-subfolder


shiftkey:
Git 2.0 was released in mid-2014. Please update your environment, as there's lots of improvements and fixes you're missing out on!

PR #2077, opened 2016-08-22
nbactions.xml must not be ignored (?)

cshubhamrao:

This file as documentation (which seems lacking) states is used for calling custom Maven goals from the IDE.
Ignoring this file was a source much confusion in few projects (mapping Run action to custom goals... etc etc) 
Also this file does anything worth ignoring
Added by #145
Exclusion of `nbactions.xml`  from `.gitgnore` should be discussed.

### DIFF
--- a/Global/NetBeans.gitignore
+++ b/Global/NetBeans.gitignore
@@ -3,5 +3,4 @@ build/
 nbbuild/
 dist/
 nbdist/
-nbactions.xml
 .nb-gradle/


shiftkey:
@cshubhamrao any links to discussions or SO threads about this file and it's purpose, in lieu of documentation?

cshubhamrao:
http://stackoverflow.com/questions/10015385

This is the wiki page: http://wiki.netbeans.org/Nbactions.xml

shiftkey:
> I put nbactions.xml into source control BUT there is a caveat: it's internal format can change so if your developers, for any reason, use different versions of NetBeans you could have to remove it because sharing it becomes nasty.

I have some reservations when hearing about things like this, but I'm happy to incorporate this and see who notices.

shiftkey:
@cshubhamrao thanks!

SlowMo24:
We noticed :-) To my understanding this file contains personal settings. It may not be valid for any user and creates noise.

PR #2081, opened 2016-08-25
Added ignore for vscode settings

Bharat-Kalluri:
Modified the python .gitignore file to ignore the visual studio code settings folder(.vscode).


shiftkey:
Thanks, but we have a [`Global/VisualStudioCode.gitignore`](https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore) already.

PR #2082, opened 2016-08-25
[VisualStudio] ignore files after UWP package was created..

bondarenkod:

Because VS  makes a lot of unnecessary files like that - https://dl.dropboxusercontent.com/u/22639368/ShareX/2016/08/2016-08-26_00-56-28.png
after package was created, the vs git ignore contains some ignored files from this list.

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -174,8 +174,10 @@ rcf/
 # Windows Store app package directories and files
 AppPackages/
 BundleArtifacts/
-Package.StoreAssociation.xml
 _pkginfo.txt
+_scale-*.appx
+_language-*.appx 
+Package.StoreAssociation.xml
 
 # Visual Studio cache files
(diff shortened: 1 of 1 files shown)


VisualStudio.gitignore
```diff
@@ -174,8 +174,10 @@ rcf/
 # Windows Store app package directories and files
 AppPackages/
 BundleArtifacts/
-Package.StoreAssociation.xml
 _pkginfo.txt
+_scale-*.appx
+_language-*.appx
```
bramborman: You have to remove the space at the end of this line to make it work :)

shiftkey: Whitespace is important with these rules. @bondarenkod please have a look at this.

shiftkey:
Closing due to inactivity. @bondarenkod let me know if you're still keen to land this and can address the outstanding feedback.

PR #2083, opened 2016-08-27
Addin .vscode/ folder to be ignored.

russhwalker:

Adding vscode folder to be ignored for those of us who also develop using Visual Studio Code. This folder holds workspace settings and options as well as extensions.

https://code.visualstudio.com/docs/customization/userandworkspace


shiftkey:

PR #2085, opened 2016-08-30
Rename OSX.gitignore to macOS.gitignore

johemst:
Apple renames OS X to macOS. Announced @ WWDC 2016


shiftkey:
@johemst good point, thanks!

aheckler:
@shiftkey The GitHub Desktop app for Mac must still be looking for the old filename:

screen shot 2016-09-03 at 6 04 13 am

Noticed that when I went to update my personal dotfiles repo. Reporting here since I don't think the desktop app is open source.

shiftkey:
@aheckler thanks, I'll pass that onto the team

PR #2091, opened 2016-09-05
[ObjC, Swift, Xcode] Make Objective-C & Swift symlinks to Xcode

EtherTyper:

You previously have written `remember to update Global/Xcode.gitignore, Objective-C.gitignore & Swift.gitignore` in all of these duplicate files. I saw that other .gitignore files also behaved as symlinks, so I figured that it would be a nice clean up to make two of them symlinks.

Hope this helps!
The @EtherTyper


Objective-C.gitignore
```diff
@@ -1,62 +0,0 @@
...
-*.moved-aside
-*.xcuserstate
-
-## Obj-C/Swift specific
-*.hmap
-*.ipa
-*.dSYM.zip
-*.dSYM
-
-# CocoaPods
```
shiftkey: It looks like these rules are going to be lost with this change. Is this deliberate?

EtherTyper: @shiftkey I assumed they weren't necessary, as you're instructed to discard them when you overwrite Objective-C.gitignore and Swift.gitignore with Global/Xcode.gitginore anyways. Should we move these extra rules back to Global/Xcode.gitignore?

shiftkey: Sorry, I should have been clearer here. There's rules that exit both in `Swift.gitignore` and `Objective-C.gitignore` that will disappear because they aren't in `Global/XCode.gitignore` - we need to ensure these files are all in sync before we can switch over to symlinks. Make sense?

shiftkey:
Closing this out because of inactivity and unresolved feedback

PR #2092, opened 2016-09-05
Update TeX.gitignore

vbalnt:
Added a line that includes the `auto` folder that Emacs+ AUCTex automatically creates in the gitignore list. 

Emacs + AUCTex creates a folder called `auto` in the local dir. 
This is due to the automatic parsing of the TeX files. 

**About this auto folder** 
https://www.gnu.org/software/auctex/manual/auctex/Parsing-Files.html


shiftkey:
@vbalnt thanks!

PR #2096, opened 2016-09-07
Include shared settings, according to official docs

petermorlion:
Official docs, found here:
https://intellij-support.jetbrains.com/hc/en-us/articles/206544839 (as
of 13 Aug 2016)

jsLibraryMappings.xml, vcs.xml and dictionaries should be included, are not user-specific.


petermorlion:
This is an update of #1904

shiftkey:
@petermorlion thanks!

PR #2098, opened 2016-09-07
Eclipse subprojects ignore

EricXing:

When import a java project that contains sub-projects, eclipse creates `.settings folder`, `.project` and `.classpath` for each sub-project, and they should also be ignored.

The updated pattern can be found in https://git-scm.com/docs/gitignore


shiftkey:
@EricXing any links to StackOverflow discussions or other documentation about this change?

PR #2099, opened 2016-09-07
Ignore maven files of subprojects

EricXing:

When the maven project is a multi-project, each sub-project have a `target` folder. They should be ignored.

The updated pattern can be found in https://git-scm.com/docs/gitignore

### DIFF
--- a/Global/Eclipse.gitignore
+++ b/Global/Eclipse.gitignore
@@ -1,18 +1,18 @@
 
 .metadata
-bin/
+**/bin/
 tmp/
 *.tmp
 *.bak
 *.swp
 *~.nib
 local.properties
-.settings/
+**/.settings/
 .loadpath
 .recommenders
 
 # Eclipse Core
-.project
+**/.project
 
 # External tool builders
 .externalToolBuilders/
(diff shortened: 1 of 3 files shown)


shiftkey:
@EricXing any links to StackOverflow discussions or other documentation about this change?

EricXing:
@shiftkey  [this is](http://stackoverflow.com/questions/2545602/git-ignore-sub-folders) about ignore sub-folders.

drothmaler:
@EricXing this is not necessary... Try it: 

```sh
cd /tmp
mkdir git-subfolder-test
cd git-subfolder-test
git init
mkdir foo
mkdir -p bar/foo
touch foo/test.txt
touch bar/test.txt
touch bar/foo/test.txt
echo "foo/" > .gitignore
git add .
git status
```
--> only `.gitignore` and `bar/test.txt` are listet; content of *both* `foo/`directories is ignored, without `**/foo/`
`**` is only required for more complex ignores like `**/foo/bar.txt`. But simple `file.txt` and `folder/` rules are always matched globally. 

You can also have a look a at the [official gitignore documentation](https://git-scm.com/docs/gitignore#_pattern_format).

drothmaler:
BTW: It seems that you mixed up your pull requests: 
#2097 and #2098
are also contained here

EricXing:
@drothmaler I have tested it. Thank you.

I really mixed up my pull requests. [#2097](https://github.com/github/gitignore/pull/2099) and [#2098](https://github.com/github/gitignore/pull/2099) have been closed.

PR #2103, opened 2016-09-08
Android studio gitignore

shahabhameed:

There should be a dedicated AndroidStudio gitignore file for developers using AndroidStudio to do android development. @aroben this pull is requested in response to your closing comment on #859.  

http://stackoverflow.com/questions/16736856/what-should-be-in-my-gitignore-for-an-android-studio-project
https://gist.github.com/vtanathip/9414323
http://unknownerror.org/opensource/github/gitignore/q/stackoverflow/16736856/what-should-be-in-my-gitignore-for-an-android-studio-project

If this is a new template: 
- **AndroidStudio**: https://developer.android.com/studio/index.html


shahabhameed:
@shiftkey please add a new template label to this pull request.

shahabhameed:
@dzfacts gradle.xml contains the path where the gradle is installed (note: gradle path can be different for different contributors of the project), modules gradle settings, distribution type and external project path settings. 
From the discussion done at [(https://youtrack.jetbrains.com/issue/IDEA-55923)]  the general consensus is that it shouldn't be ignored by default.

AndSync:
Files in folder .idea  are not automatically generated?  Why can't ignore?  I Think even one file is useful ,if everyone generated are the same,the file can be ignored. Am I right?

joeblau:
@shahabhameed - This tempalte is live on [gitignore.io](https://www.gitignore.io) - https://www.gitignore.io/api/androidstudio

Baneeishaque:
I think **vcs.xml, modules.xml, runConfigurations.xml will be useful**. Don't know about misc.xml & codeStyles\Project.xml Files. Even though **misc.xml contains some absolute paths in SvnConfiguration section**.

shiftkey:
@shahabhameed I think `Android.gitignore` is the template you're looking for here. Please submit any additional improvements to that template.

PR #2104, opened 2016-09-09
Ignore .jfm files in Visual Studio

kevcunnane:

.jfm files have started blocking checking for .sqlproj projects in Visual Studio. .jfm files are a new file-type created by the Microsoft ESENT database engine - this is a recent improvement only in Windows 10 Anniversary update. As the SSDT .sqlproj type in Visual Studio uses this engine for .dbmdl metadata files, and this file is locked by the user, this currently blocks users from checking into Git repositories.

This fix helps mitigate this by avoiding checkin of this file type.

An example of this issue is this [StackOverflow post
](http://stackoverflow.com/questions/37704514/visual-studio-2015-database-project-directory-contains-a-file-with-extension-jfm) which shows the issues cause by this, and an explanation of the cause. At present, the file type is not documented by the ESENT team, so there is no official doc covering this.

Disclosure: I work for Microsoft on the SSDT team shipping .sqlproj support into Visual Studio.

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -189,6 +189,7 @@ ClientBin/
 *~
 *.dbmdl
 *.dbproj.schemaview
+*.jfm
 *.pfx
 *.publishsettings
 node_modules/


shiftkey:
@kevcunnane thanks for the extra information!

TheBeardedLlama:
that's nice

I've had a PR waiting to be merged for the past couple of months (since beta of win10 anniversary) to add this extension to the ignore list

at least it's in now...

aabhishek1:
Hi,
I am facing the same issue. 
I have added the *.jfm extension in gitignore file under User-specific files and Other ClientBin/ tag
as mentioned below.

# User-specific files 
*.jfm

# Others
ClientBin/
*.jfm

but it doesn't work. 

Could anyone please let me know what I am missing. 
Any help is highly appreciated.

TheBeardedLlama:
it should work @aabhishek1 

could you find a way to verify your gitignore?

aabhishek1:
Thank You TheBeardedLlama for you reply, could you please let me know I could I verify my gitignore.

TheBeardedLlama:
if you're using VS, then there's a plugin that @madskristensen has developed called .ignore

https://github.com/madskristensen/IgnoreFiles

it will show you what's matching and what's not

Lenka72:
In case someone is looking for this and their .jfm file was not removed after it was added to gitignore run this: git rm --cached *.jfm and it will fix the problem (right-click in the directory and select Git Bash Here and run it).

PR #2105, opened 2016-09-12
added VS code coverage files

filipw:

Automatically ignore VS Enterprise code coverage files.

https://msdn.microsoft.com/en-us/library/dd537628.aspx#Anchor_1

Mentions of the extensions:

> The coverage coloring might be incorrect if the source code has changed since the .coverage file was generated. 

> (...)

> To make results readable as text, choose Export Code Coverage Results. This generates a readable .coveragexml file which you could process with other tools or send easily in mail.


shiftkey:
@filipw thanks!

PR #2106, opened 2016-09-13
Update Erlang.gitignore

ukari:

`/ebin` can't be simplely ignored because `*.app` should be put here.

[Erlang Applications](http://erlang.org/doc/design_principles/applications.html)
> ### 8.4  Directory Structure
> ebin - Contains the Erlang object code, the beam files. The .app file is also placed here.


shiftkey:
@ukari thanks!

PR #2107, opened 2016-09-15
Add support for Cake.Net

RehanSaeed:
Cake.Net creates a tools folder at the root which contains downloaded packages and should not be checked in. See [documentation](http://cakebuild.net/docs/tutorials/getting-started) for reference to the tools folder where NuGet packages are downloaded.

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -264,5 +264,5 @@ paket-files/
 __pycache__/
 *.pyc
 
-# Cake
-tools/
+# Cake - Uncomment if you are using it
+# tools/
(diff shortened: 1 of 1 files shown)


VisualStudio.gitignore
```diff
@@ -263,3 +263,6 @@ paket-files/
 # Python Tools for Visual Studio (PTVS)
 __pycache__/
 *.pyc
+
+# Cake
+tools/
```
shiftkey: I can't help but feel this is too broad a rule, and might catch other things.

For example, FAKE uses a `.fake/` path for it's local cache.

RehanSaeed: I agree it is a possibility. There are a few of these folders in there already e.g. artifacts/ publish/. Cake is now part of the .NET Foundation and growing in popularity. This PR can go either way.

shiftkey: Perhaps we could have this commented out by default, so that users can opt-in to the change...

shiftkey:
@RehanSaeed thanks!

PR #2109, opened 2016-09-15
Added .DS_Store files to Swift and Objective-C gitignore files

Johnykutty:
Since .DS_Store is a hidden system file and act no role in a source code directory, no need to add this to repository.


shiftkey:

PR #2111, opened 2016-09-19
Add .nfs files to Linux.gitignore

jfloff:
.nfs files are created when an open file is removed but is still being accessed

http://serverfault.com/questions/201294/nfsxxxx-files-appearing-what-are-those


shiftkey:
@jfloff thanks!

PR #2115, opened 2016-09-21
Added kernel module compilation results

aarian93p:
Adds Kernel modules compilation results to the ignore list.

### DIFF
--- a/C.gitignore
+++ b/C.gitignore
@@ -36,7 +36,8 @@
 *.su
 
 # Kernel Module Compile Results
-*.mod.c
+*.mod*
+*.cmd
 modules.order
 Module.symvers
 Mkfile.old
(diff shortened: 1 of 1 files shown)


shiftkey:
@aarian93p do you have any supporting links that discuss why these changes are beneficial?

aarian93p:
Yep @shiftkey 
While compiling a kernel module, for each ".ko" file, a ".mod.c" and a ".mod.o" file are made which are made which will finally compile the .ko file and are useless for us cause they are made in the middle of build proccess and used for the rest of it. Thus we should ignore them, for the very same reason that we ignore .o and .ko files. Also .cmd files are temporary too (just like the previous ones). Also the last four are generated in the middle of compilation process and not useful for the developer cause it is mainly depending on the platform blah blah blah.
Hope I could express what I meant to (Bad English :D )

shiftkey:
@aarian93p thanks!

PR #2120, opened 2016-09-27
Don't ignore Python coverage configuration

rouge8:
Python's coverage module is configured in `.coveragerc`, which somehow was previously ignored.

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -44,6 +44,7 @@ nosetests.xml
 coverage.xml
 *,cover
 .hypothesis/
+!.coveragerc
 
 # Translations
 *.mo


Python.gitignore
```diff
@@ -44,6 +44,7 @@ nosetests.xml
 coverage.xml
 *,cover
 .hypothesis/
+!.coveragerc
```
shiftkey: I *think* this is being caught by `.coverage`, right?

If so, can we move this up to there and add a comment explaining why it needs to exist...

rouge8:
> I think this is being caught by `.coverage`, right?

I spent the longest time trying to figure out what was catching this and finally realized it was a typo in my own global gitignore. 😳

PR #2121, opened 2016-09-27
Add vendor folder to Go.gitignore

futabooo:

if use [Masterminds/glide](https://github.com/Masterminds/glide) Vendor Package Management for Golang, the dependencies are exported to the `vendor/` directory

>Go 1.5 introduced experimental support for a "vendor" directory, enabled by the GO15VENDOREXPERIMENT environment variable. Go 1.6 enabled this behavior by default, and in Go 1.7, this switch has been removed and the "vendor" behavior is always enabled.

https://blog.golang.org/go1.7

### DIFF
--- a/Go.gitignore
+++ b/Go.gitignore
@@ -25,3 +25,6 @@ _testmain.go
 
 # Output of the go coverage tool, specifically when used with LiteIDE
 *.out
+
+# external packages folder
+vendor/


shiftkey:
@futabooo thanks, I'm glad to see this land officially!

jamesboswell:
I think this is a mistake to add the vendor folder to the default GitHub  `.gitignore`    

Go doesn't have a central package repository.  Therefore external Go packages can't be referenced in the future  at a specific version the way Node.js packages can (via package.json)   

Developers can "vendor" a package to "freeze it" at a point in time for their project so any future builds will be consistent.   

It should be left to the developer as to whether they want to ignore `vendor/` or not.   This may confuse new/future Go developers

shiftkey:
@jamesboswell I'm neutral on this, but please open a PR and pull in people so you can reach a consensus on this.

futabooo:
@jamesboswell In the case of Glide use, `glide.yml` and `glide.lock` has version information, so I thought it was okay.

kaneshin:
@jamesboswell 
Look https://github.com/github/gitignore/blob/master/Ruby.gitignore#L38-L41 .
I mean, I agree with the PR.

PR #2126, opened 2016-10-06
Add PyCharm project settings

LordDarkula:

Many python programmers use PyCharm, which stores config files in 
a .idea directory located at inside project's root directory. 
These files are PyCharm specific, and should not be tracked 
by source control. 

The following link leads to a page describing PyCharm
project structure in greater detail.

[https://www.jetbrains.com/help/pycharm/2016.1/project.html](https://www.jetbrains.com/help/pycharm/2016.1/project.html)

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -88,3 +88,6 @@ ENV/
 
 # Rope project settings
 .ropeproject
+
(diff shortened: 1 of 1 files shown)


shiftkey:

PR #2127, opened 2016-10-07
Add Buck gitignore values

Qix-:
[Buck](/facebook/buck) is Facebook's build system. There are output files and metadata that should be ignored.

These are taken straight from their repository's gitignore.


joeblau:
@Qix- - This tempalte is live on [gitignore.io](https://www.gitignore.io) - https://www.gitignore.io/api/buck

Qix-:
Thanks :)

PR #2128, opened 2016-10-11
autotools.gitignore: add missing files

stapelberg:

autotools.gitignore was missing some files which are in my source tree after `autoreconf -fi`

See AC_CONFIG_AUX_DIR in https://www.gnu.org/software/automake/manual/html_node/Optional.html


shiftkey:
@stapelberg thanks!

PR #2129, opened 2016-10-11
Added yarn.lock file to Node.gitignore

viatsko:

Yarn is a new package manager for NodeJS released by facebook

https://yarnpkg.com/

### DIFF
--- a/Node.gitignore
+++ b/Node.gitignore
@@ -34,6 +34,9 @@ jspm_packages
 # Optional npm cache directory
 .npm
 
+# Optional yarn lockfile
+yarn.lock
+
 # Optional eslint cache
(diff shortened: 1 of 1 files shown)


dtran320:
I believe `yarn.lock` is actually meant to be checked into version control since it should be reproduced the same way across all machines for the same version of `package.json`:


> All yarn.lock files should be checked into source control (i.e. git or mercurial). This allows Yarn to install the same exact dependency tree across all machines, whether it be your coworker’s laptop or a CI server.

via https://yarnpkg.com/en/docs/yarn-lock

viatsko:
After thinking a bit on it, I must agree with you. Thanks for the feedback

PR #2134, opened 2016-10-12
Add yarn ignored file

morrislaptop:

Yarn may add this file in its workflow. 

https://github.com/yarnpkg/yarn/blob/113fab9698611d6e24f355c1a66e2ad9e7325317/src/cli/commands/install.js#L649


shiftkey:
@morrislaptop thanks!

PR #2138, opened 2016-10-16
Yarn lock file

montyanderson:

_TODO_

_TODO_

If this is a new template: 


drothmaler:
The [Yarn documentation](https://yarnpkg.com/en/docs/version-control) states, that this **should** be checked into Version Control!!
So this is a **-1** from me.

PR #2141, opened 2016-10-17
Removed LaunchSettings.json from being committed

jasongaylord:
Removed Properites/launchSettings.json as others may be using other web servers and ports. Doesn't make sense to include my version.

https://github.com/aspnet/Mvc/pull/3525

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -44,6 +44,7 @@ dlldata.c
 project.lock.json
 project.fragment.lock.json
 artifacts/
+Properties/launchSettings.json
 
 *_i.c
 *_p.c


shiftkey:
@jasongaylord the last comment seems interesting:

>  I'll back out the change once we get upgraded to new tooling.

And that was from November 2015. Does this mean it's not a bug with the tooling, and Working As Intended™?

jasongaylord:
When you say....
> Does this mean it's not a bug with the tooling, and Working As Intended™

Are you asking if the launchSettings.json file should or should not be there? Even in the latest builds, it's there. It also appears in the team's version of .gitignore here: [https://github.com/aspnet/Mvc/blob/dev/.gitignore](https://github.com/aspnet/Mvc/blob/dev/.gitignore)

shiftkey:
> Are you asking if the launchSettings.json file should or should not be there? 

Yeah, based on my reading of the initial issue it sounds like it's not a permanent thing.

@NTaylorMullen can you confirm this is intentional behaviour?  The discussion in https://github.com/aspnet/Mvc/pull/3525 suggests the .gitignore change is a workaround - just wanting to confirm that's no longer accurate.

NTaylorMullen:
If you have customized debugger settings then having the `launchSettings.json` present is necessary. However, some might argue that customized debugger settings should be on a per-user basis (me being one of them) so `.gitignore`ing it is good.

To clarify the linked MVC PR. That was a bug where VS would automatically add a `launchSettings.json` on project open.

shiftkey:
> should be on a per-user basis
> VS would automatically add a `launchSettings.json` on project open

@NTaylorMullen excellent, thanks for the context!

PR #2143, opened 2016-10-19
ignore Visual Studio 6 workspace file

mrs-mrl:

The vbw file doesn't server a purpose besides the last opened files etc. Especially when working in teams this file will basically differ as soon as you open the project and could lead to unnecessary merge conflicts. There's basically no sense in keeping this file.

http://stackoverflow.com/a/1530193
http://stackoverflow.com/a/37311186
https://msdn.microsoft.com/en-us/library/aa241721(v=vs.60).aspx

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -239,6 +239,9 @@ FakesAssemblies/
 # Visual Studio 6 workspace options file
 *.opt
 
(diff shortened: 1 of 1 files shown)


shiftkey:
@mrs-mrl thanks!

PR #2147, opened 2016-10-22
ignore everything

was4444:

Someone like me may want to control all content that indexed.

Force all indexed content meaningfully.


shiftkey:
This is cool and all, but I'd rather focus on supporting the existing ecosystem of languages and tools.

was4444:
@shiftkey, is there any way to use private ``.gitignore`` templates in ``GitHub``.

shiftkey:
@was4444 I don't understand the question. Could you elaborate?

was4444:
@shiftkey, the question probably should not be asked here, but when create a new repo from ``GitHub``, we can select a ``.gitignore``, which seems must in this repo, to create as the initial commit. I just want to use my private ``.gitignore`` file as the initial commit.

PR #2148, opened 2016-10-23
.gitignore file for Xamarin

HeinPauwelyn:

Xamarin is more and more important to make cross platform applications for Android, iOS and Windows.

- https://www.gitignore.io/api/xamarinstudio

- https://www.xamarin.com/

### DIFF
--- a/Xamarin.gitignore
+++ b/Xamarin.gitignore
@@ -1,0 +1,5 @@
+bin/
+obj/
+*.userprefs
+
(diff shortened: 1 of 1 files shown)


shiftkey:
I'd love to tidy up `VisualStudio.gitignore` at some point, in favour of a slimmed-down .NET template. I don't think this is the way to go though. Sorry!

PR #2149, opened 2016-10-23
Add Yarn package manager files

azat-io:
For [Yarn](https://github.com/yarnpkg/yarn) package manager

### DIFF
--- a/Node.gitignore
+++ b/Node.gitignore
@@ -2,6 +2,7 @@
 logs
 *.log
 npm-debug.log*
+yarn.lock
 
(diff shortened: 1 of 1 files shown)


drothmaler:
According to the [yarn documentation](https://yarnpkg.com/en/docs/configuration#toc-use-yarn-lock-to-pin-your-dependencies) `yarn.lock`**should** be checked in, to your source repository!

shiftkey:
Seconding what @drothmaler said - this is an important feature of yarn...

PR #2150, opened 2016-10-23
Nanoc: Fix capitalisation + config filename

ddfreyne:

* Fix capitalisation of Nanoc (_Nanoc_, not _nanoc_)
* Use `nanoc.yaml` rather than `config.yaml` (the former is the default filename, while the latter is deprecated)

* The [Nanoc web site](http://nanoc.ws/)


shiftkey:
@ddfreyne thanks!

PR #2151, opened 2016-10-23
Update ExtJs.gitignore

israelroldan:

There were some files missing from what traditionally should be gitignored

http://docs.sencha.com/cmd/guides/microloader.html#microloader

For full disclosure: I work for Sencha, Inc.


shiftkey:
@israelroldan thanks!

PR #2153, opened 2016-10-26
A new template for Perl 6

briandfoy:

No Perl 6 template exists. Perl 6 uses its own build and installation system, so the Perl 5 template is not useful.

These particular strings aren't documented yet, but I'm working on that.

If this is a new template: 

http://www.perl6.org

Just a start. Perl 6 creates these directories to 
store pre-compiled versions of modules. 

Perl 6 doesn't have a gitignore template yet, and it
doesn't use the same build and installation system as
Perl (5).

### DIFF
--- a/Perl6.gitignore
+++ b/Perl6.gitignore
@@ -1,0 +1,7 @@
+# Gitignore for Perl 6 (http://www.perl6.org)
+# As part of https://github.com/github/gitignore
+
+# precompiled files
+.precomp
+lib/.precomp
+


joeblau:
@briandfoy - This tempalte is live on [gitignore.io](https://www.gitignore.io) - https://www.gitignore.io/api/perl6

JJ:
Any chance this is going to be merged? Perl 6 has, since them, become very popular, with lots of modules and applications in GitHub.

JJ:
Perl 6 is now called [Raku](https://raku.org). Any chance this .gitignore can simply be renamed in production? It's still the same content.

shiftkey:
@JJ I'd recommend renaming the template here as this is the source of truth (and everyone will get the benefit).

JJ:
@shiftkey essentially, making a new pull request, I guess, right?

shiftkey:
@JJ correct

PR #2155, opened 2016-10-26
elisp: add emacs backup files and undo-tree save-history files

troyp:

Emacs backup files and undo-tree save-history files are common files which may occur in any directory by default and should not be included in a repository.

**note:** Undo-tree history files only occur if `undo-tree` is used (and set to save history), but this is very common (and if `undo-tree` is _not_ used, the chance of a filename coincidentally ending in `.~undo-tree` is essentially zero).

- [backup files](https://www.gnu.org/software/emacs/manual/html_node/elisp/Backup-Names.html#Backup-Names)
- `undo-tree` doesn't have documentation (beyond docstrings), but the relevant code is in the function `undo-tree-make-history-save-file-name` in the [undo-tree source](http://www.dr-qubit.org/undo-tree/undo-tree.el)

**If this is a new template:**  _N/A_


shiftkey:
@troyp thanks!

PR #2158, opened 2016-10-28
Use the exact CC-0 license text

benbalter:
The license for this repository isn't being detected because the license file contain additional text, not part of the license. Specifically, it contains a disclaimer between CC and us, the downstream user, disclaiming liability for our use of the license. That text isn't part of the license itself, doesn't need to be included in the license file, and shouldn't be included in the license file if we want GitHub to detect the license.

This PR removes that extra text by using the exact text from choosealicense.com, which GitHub should then detect.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,121 +1,116 @@
-Creative Commons Legal Code
-
 CC0 1.0 Universal
-
-    CREATIVE COMMONS CORPORATION IS NOT A LAW FIRM AND DOES NOT PROVIDE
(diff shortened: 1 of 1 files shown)


shiftkey:
@mlinksva any objections? 😄

mlinksva:
No objection.

Ideally we could detect with or without the license creator (CC) disclaimer, usefully marked as optional by https://spdx.org/licenses/CC0-1.0.html ... but until then (possibly forever), standardizing on the text we've been propagating to others is fine.

PR #2159, opened 2016-10-29
Updated Jetbrains.gitignore

exarus:

The folder .idea/datasource contains files like this:
86e63b6d-632b-48ba-b7e2-c5bb0c00ac60.xml
with some sensitive info about user db connections

Added datasource folder


shiftkey:
@exarus thanks!

PR #2162, opened 2016-10-31
[C] ignore versions of kernel module build results

chaosdefinition:

.tmp_versions is a directory that records version of Linux kernel module build result, should be ignored.

### DIFF
--- a/C.gitignore
+++ b/C.gitignore
@@ -45,6 +45,7 @@
 # Kernel Module Compile Results
 *.mod*
 *.cmd
+.tmp_versions/
 modules.order
 Module.symvers
 Mkfile.old


shiftkey:
@chaosdefinition do you have a link to some documentation or discussions about this, for future reference?

chaosdefinition:
@shiftkey Yes, excerpt from [External kernel modules](https://lwn.net/Articles/69148/):

> During the kernel compile, a list of all files that export symbols is
created in .tmp_versions/ of the kernel source tree. After everything is
compiled, the module versions of all these object files are resolved
with modpost. When an external module is compiled, it ends up being the
only file in .tmp_versions/, and so symbols in other modules would not
be found.

shiftkey:
@chaosdefinition thanks!

PR #2165, opened 2016-11-01
Update to take into account Dart SDK 1.20

ArgentiApparatus:

Dart SDK 1.20 no longer creates `packages` directories

http://news.dartlang.org/2016/10/good-bye-symlinks.html

Dart SDK 1.20 no longer creates `packages` directories. Added comments to `pub` files section to differentiate  what is needed for >= SDK 1.20 and what is additionally needed for older SDK versions.


shiftkey:
@ArgentiApparatus thanks!

PR #2166, opened 2016-11-02
Update of Visual Studio gitignore file.

jvcdk:

Two changelists:
  1) Include link to origin. When I include a .gitignore file in a shared project, I think it is a good idea that other team members have a reference for it's origin.

 2) Add Visual C++ project filters file.

 add 2) http://stackoverflow.com/questions/1826901/should-i-add-vcxproj-filter-files-to-source-control


shiftkey:
@jvcdk thanks!

PR #2169, opened 2016-11-03
Add *.VC.db (New IntelliSense database) for VS2015 Update2 ignore.

Humbunklung:
After the Visual Studio 2015 Update 2 is installed, the solution with VC projects will create a new IntelliSense database, named <solution-name>.VC.db, it replaces the *.sdf file.


shiftkey:
This was already added in #1783

Humbunklung:
OK, thx.@shiftkey

PR #2172, opened 2016-11-04
Delete IPythonNotebook.gitignore

dkav:

The IPython Notebook is now known as the Jupyter Notebook. Jupyter ignore values are already found in the Python.gitignore.

http://ipython.org/notebook.html
http://jupyter.org
https://github.com/github/gitignore/blob/master/Python.gitignore


shiftkey:
@dkav good point! :fire:

PR #2173, opened 2016-11-04
Add .dir-locals.el

robsonpeixoto:

Sometimes we need to create a config to the project

https://www.gnu.org/software/emacs/manual/html_node/emacs/Directory-Variables.html


shiftkey:
@robsonpeixoto thanks!

PR #2180, opened 2016-11-07
ignore Properties/launchSettings.json even if Properties is a subfolder

chucklu:

to support the situation that the Properties folder is a subfolder


shiftkey:
@chucklu good point, thanks!

PR #2181, opened 2016-11-07
Change *.mtc patterns to single *.mtc[0-9]*

Levityx:

Single pattern instead of those 3. 

This pattern encompasses the previous ones, now including the pattern *.mtc0[0-9] that was previously missing but shouldn't happen in practice since LaTeX creates single digits indices instead.


shiftkey:
@Levityx thanks!

PR #2182, opened 2016-11-10
Add .idea/

Pavelyev:
Ignore Rider settings catalog
http://www.jetbrains.com/rider/


shiftkey:

PR #2183, opened 2016-11-12
Go: Revert "Add vendor folder to Go.gitignore"

zchee:
This reverts commit 14740a52c50fabdab1758808e995dcd9766bd76f.

Any Go developers sometimes static vendoring into vendor directory.
Because for the delete vendoring packages repository, reliably build with
vendor packages or etc.
Also, A typical vendoring tool such as `gb`, `gvt` will create `manifest` file on `vendor` directory.

In Go language development, ignoring vendor directory is should be
left to the package authors.(I think)

https://github.com/derekparker/delve/tree/master/vendor
https://github.com/zchee/docker-machine-driver-xhyve/tree/master/vendor

### DIFF
--- a/Go.gitignore
+++ b/Go.gitignore
@@ -25,6 +25,3 @@ _testmain.go
 
 # Output of the go coverage tool, specifically when used with LiteIDE
 *.out
-
-# external packages folder
-vendor/


futabooo:
you have got a point.
It may be better to leave it to developers.

below is my opinion.

>Any Go developers sometimes static vendoring into vendor directory.
>Because for the delete vendoring packages repository, reliably build with
>vendor packages or etc

I think it is not good,It is better to use package management

>Also, A typical vendoring tool such as gb, gvt will create manifest file on vendor directory.

how about this?
```
/vendor
!/vendor/manifest
```

zchee:
@futabooo 
> I think it is not good,It is better to use package management
> how about this?
/vendor
!/vendor/manifest

Yeah, That is our gopher's ideal. like ruby's `gem` or any good package manager.
However, Go vendoring manager tools is not stable yet. Still chaotic(?).

Also, If package manager supports the good vendoring feature, still have depends on the existence of the github repository.
I do not know the thought of other developers, but I guess that is probably concerned about the occurrence of *"npm leftpad"*(at least I am).
Therefore, I did statically vendoring. In other words, including all vendor packages `.go` file to my repository's `vendor` directory.

Well, I always delete it and create a .gitignore, so there is no problem. If you do not like it, I will close it :)
However, it certainly has a `.go` file in the vendor that docker, docker-machine, minikube.

futabooo:
@zchee 
Overall agreement:)
I'd like to ask other people's thoughts, so I do not think you need to close it!

nickng:
Vendoring support is [officially supported as of go1.6](https://golang.org/doc/go1.6#go_command), which means the `vendor` directory is recognised by Go project's `go build` or `go install` command to build Go code.

There are a number of projects, including widely used projects like [docker](https://github.com/docker/docker), which uses the `vendor` directory. Putting `vendor` on GitHub's `.gitignore` will not affect most existing users or those who write their own `.gitignore`, but Go users who create `.gitignore` using GitHub's web interface may be confused by the discrepancies between the official specifications and the behaviour of the default `.gitignore` created by GitHub.

I don't believe this is a matter of personal choice, but should be considered for the consistency between Go project's specification/official tooling support and the expectation of a conservative, default `.gitignore`. Hence I will strongly support leaving `vendor` out of `.gitignore` (i.e. please proceed with the revert).

shiftkey:
Closing due to conflicts. @zchee please have a look at the latest file and see if anything needs to be incorporated.

PR #2187, opened 2016-11-15
Ignore *.ctxt in Java.gitignore

Stevoisiak:
Similar to .class files, .ctxt files are generated every time a project is compiled in BlueJ. Just like .class files, if a .ctxt file is not present when a Java program is compiled, it will be  recreated automatically.

BlueJ has a downloadable extension featured on its website to clear out .class and .ctxt files automatically.
http://www.bluej.org/extensions/extensions.html

### DIFF
--- a/Java.gitignore
+++ b/Java.gitignore
@@ -1,5 +1,8 @@
 *.class
 
+# BlueJ files
+*.ctxt
+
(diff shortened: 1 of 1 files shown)


shiftkey:
@Stevoisiak thanks!

PR #2188, opened 2016-11-15
Create Y86.gitignore

kaveet:
No template exists for Y86 object code generated by Y86 assemblers. The Y86 instruction set is heavily used in educational environments but lacks centralized documentation. Typically an assembler produces assembled object files with extension *.yo as seen in the following YAS instructions and documentation:

https://www.cs.unm.edu/~bradykey/simguide.pdf

If this is a new template: 

https://www.cs.unm.edu/~bradykey/lab4Y86Sim341.html (Original project)

### DIFF
--- a/Y86.gitignore
+++ b/Y86.gitignore
@@ -1,0 +1,2 @@
+# Object File
+*.yo


joeblau:
@kaveet - This tempalte is live on [gitignore.io](https://www.gitignore.io) - https://www.gitignore.io/api/y86

PR #2189, opened 2016-11-16
Replace DNX with .NET Core

HenrikWM:

DNX-name is replaced with .NET Core. Could also be named `dotnet` perhaps.

https://docs.microsoft.com/nb-no/dotnet/articles/core/migrating-from-dnx


shiftkey:
@HenrikWM thanks!

PR #2190, opened 2016-11-17
Add  IntelliJ idea specific for scala

Vic020:
Dear Maintainer,

IntelliJ idea will generate those files automatically. But those files is not useful for scala project.

Please accept this pr.


shiftkey:

PR #2191, opened 2016-11-18
[Windows] encryptable thumbs database

mestremuten:

New Thumbs.db file: _Thumbs.db:encryptable_

### DIFF
--- a/Global/Windows.gitignore
+++ b/Global/Windows.gitignore
@@ -1,5 +1,6 @@
 # Windows image file caches
 Thumbs.db
+Thumbs.db:encryptable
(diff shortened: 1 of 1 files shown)


shiftkey:
@mestremuten interesting - did you find any documentation around this for future reference?

Yousha:
It's same as Thumbs.db but encrypted AND PROTECTED one, which we can't delete, move and even copy it. (Ntfs drives)
It is a "protected" operating system file
created by Windows Explorer to cache images thumbnails when browsing
with explorer.
 It also can carry virus/trojan

PR #2192, opened 2016-11-18
Laravel: ignore homestead config and npm debug.

Arsfiqball:

About the homestead configurations: /.idea, Homestead.json, Homestead.yaml is provided by default in newest laravel 5.3 installation.
About npm log files: laravel-elixir uses npm. If something happen (e.g failed npm install), then npm will create npm-debug.log.

https://github.com/laravel/elixir

### DIFF
--- a/Laravel.gitignore
+++ b/Laravel.gitignore
@@ -1,5 +1,6 @@
 vendor/
 node_modules/
+npm-debug.log
 
 # Laravel 4 specific
 bootstrap/compiled.php


Laravel.gitignore
```diff
@@ -8,9 +8,15 @@ app/storage/
 # Laravel 5 & Lumen specific
 public/storage
 storage/*.key
+/.idea
```
shiftkey: I _think_ this is related to PHPStorm, which we're tracking in `Global/JetBrains.gitignore` and has more specific rules than just this. Can we drop it?

Arsfiqball: Oh yah,

Arsfiqball: It doesn't need to be there, it just shiped with fresh laravel 5.3

Laravel.gitignore
```diff
@@ -8,9 +8,14 @@ app/storage/
 # Laravel 5 & Lumen specific
 public/storage
 storage/*.key
+Homestead.json
+Homestead.yaml
 .env.*.php
 .env.php
 .env
 
+# Logs
```
shiftkey: Could you make this comment clearer that it's related to the use of Larvel Elixir?

shiftkey:
I merged #2049 which covers the Homestead rule. @Arsfiqball please have a look at the comments about the `laravel-elixir` extension.

Laravel.gitignore
Arsfiqball: I made this section to any log files created by system that shouldn't be published.
npm-debug.log for example, created by node.
That file can comes maybe because there was error in laravel-elixir installation.
but it's not related to the app (laravel project).

Arsfiqball:
I decide to put npm-debug.log with vendor and node_modules.
as I said, this file is related to node.js which is used by laravel front end.
So I recomend to add it.

shiftkey:
@Arsfiqball thanks!

PR #2193, opened 2016-11-18
[Magento] basic files that must be ignored

rafaelstz:

The actual other files, such as IDEs files, third-party files, and system files must be added in the gitignore global of the system, not in the project scope.


shiftkey:
@rafaelstz i like this cleanup! Do you have a link to some documentation or project which indicates this is a "good enough" setup for a default Magento project?

shiftkey:
Also, does this replace #1840? Can we just assume people aren't making Magento Classic projects and are using this new layout?

rafaelstz:
Hey @shiftkey, that's the reason because gitignore_global exist, files and folders of the system, IDEs and other. I don't found article or documentation, the base of my conclusion are the files generated during and after the installation, today has the Magento Core files.

Magento 1 doesn't was discontinued and is totally different, has many people that download and use this version yet.

shiftkey:
I'll take this in and see what is left to do about the Magento2 situation.

@rafaelstz thanks!

rafaelstz:
Great @shiftkey , if you have any question please let me know!

PR #2194, opened 2016-11-21
Modified VS Code .gitignore

davormas:
Added extensions.json as exception in .vscode folder. File is used as
extensions recommendation for project.


shiftkey:
@davormas thanks!

PR #2196, opened 2016-11-21
Idea: Add StarterContent to UnrealEngine.gitignore

calben:
A lot of beginners when they make a new project tend to include the StarterContent without realising that it's a huge set of files to add to a repository.
When a repository that should have the StarterContent is downloaded and the content is missing, Unreal will add the content from the local copy.

The pros to adding the StarterContent to the default gitignore are primarily that it will help prevent beginners (and even experienced users) from adding the StarterContent to their repositories, which they will almost never want to do.
The cons are that it's potentially overstepping what a "default" gitignore should do and might result in surprising behaviour if someone does modify the StarterContent (which they shouldn't do, but I'm not the UE4 police).

Personally, I think we should add it because it would save a lot of headaches.
Thoughts?

### DIFF
--- a/UnrealEngine.gitignore
+++ b/UnrealEngine.gitignore
@@ -64,3 +64,6 @@ Intermediate/*
 
 # Cache files for the editor to use
 DerivedDataCache/*
+
+# Default location for StarterContent, engine adds these from local storage
+Content/StarterContent/*


shiftkey:
@calben this seems reasonable. As it's not related to documentation per-se, if you can get some people to :thumbsup: this suggestion I'm happy to take it in...

calben:
@samsheff @NuLL3rr0r @BGR360 @hzor3000 
thoughts?

BGR360:
Great idea! :+1:

samsheff:
I like it as well! Only thing I'd suggest is to add some kind of note mentioning why we did this, because for some projects it might be useful to not have this ignored.

I'm currently at the airport waiting for a flight, so I looked at this from my phone. If I missed something that's probably why! :)

NuLL3rr0r:
Well, in my estimation nobody uses StarterContent in a serious project under git. It's sole purpose is for education, testing or playing around with the engine features so you don't have to look for these stuff when you start out with UE4.

Although the option to add StarterContent is always available at the time of project creation, one might say the choice should be left up to the end-user in the .gitignore. But, as I said rarely anyone uses StarterContent in a serious project. Even if they want to the option to remove it from .gitignore is always available. Furthermore, if I want to use StarterContent in my serious project, I'll rename it to some reasonable name which makes sense.

To avoid confusion for new users I agree with @samsheff if you somehow document it shouldn't cause any truoble at all.

hzor3000:
Given that the Starter Content is copied to the user's project it should not be gitignored by default. If it were a folder reference to a Starter Content inside UE installation and if it were locked for modification by the user then gitignoring it would not be even needed I guess.
I would assume that the Started Content is copied to the project to make that project more portable and make the project in a way disconnected from the version(s) of UE that the user has installed. 
Given all above, and since the user is allowed to modify the version of the Starter Content with meaningful changes (for him/her) we should not gitignore it by default.
Maybe the git integration of UE should warn about the fact that Starter Content is included and have an option to add it to gitignore from its UI.

samsheff:
I think @hzor3000 has a very good point. Regardless of how frequently its used, it still is something that can be required to make the project work, that can't be recreated by the engine itself. Especially since the gitignore is something very out of the way that's often not looked at (especially in game dev), it might cause some issues for people who are unaware we made this change. Definitely something to consider before we merge.

samsheff:
Anyone else have any thoughts on this?

calben:
Oh oops, forgot to respond to that last comment!
Just to clarify, if the StarterContent is missing, UE4 will fix that for you.
At least, last time I checked, UE4 will add the default StarterContent if it is .gitignored.
There will still be surprise behaviour if the user edits something in the StarterContent rather than duplicating the item and editing it in another location, so unless we warn them the user will need to figure out why their changes to StarterContent aren't being saved in their repository... which is bad.

I will look into whether we can have the git integration warn the user before editing the StarterContent, but that might require editing some core UE4 code.
I don't *think* there's currently anything in place for changing read/write behaviour depending on the directory of an asset, but I suspect that would have more uses than just for this.

calben:
Have discussed this some with other Unreal developers.
Instead of adding StarterContent to the Github, I will be exploring making StarterContent a junction handled directly by the engine.
Thanks for the feedback!

PR #2199, opened 2016-11-24
Add missing beamer temporary file extension

Alice94:

The `.pre` extension is missing 

http://mirrors.ctan.org/macros/latex/contrib/beamer/doc/beameruserguide.pdf


shiftkey:
@Alice94 thanks!

PR #2200, opened 2016-11-24
Symfony: Fix rule for backed up entities

Hastegan:

The current rule is not correct.

http://symfony.com/doc/current/doctrine.html#generating-getters-and-setters


shiftkey:
@Hastegan thanks!

PR #2201, opened 2016-11-26
TeX: ignore endnotes *.ent files

thomwiggers:
The `endnotes` package generates `*.ent` files. These should also be
ignored.


shiftkey:
@thomwiggers thanks!

PR #2202, opened 2016-11-26
Ignore wheels directory.

rolando:

As `eggs/` directory, `wheels/` directory may contain compiled wheels.

http://pip.readthedocs.io/en/stable/user_guide/#installing-from-wheels


shiftkey:
@rolando thanks!

PR #2203, opened 2016-11-28
[Global/macOS] update 'Icon\r\r' rule to not depend on special characters

BranicYeti:
First things first, I haven't tested this, so there's a chance I'm wrong about it being a solution. I'm not the contributor to my project with a Mac, so I cannot test it myself.

I merged this file into my .gitignore file, using Sublime Text, which displayed the '\r' character as [CR] as shown here: http://i.imgur.com/bXmRxoe.png
Saving the file and re-opening it removes those markers and replaces it with new lines, and I'm not certain if these new lines are still \r or not. I had a quick google, and came across this StackOverflow answer: http://stackoverflow.com/a/33974028
It appears, according to that answer, that my change does the same thing as the current rule without the chance of being converted to a normal new line marker, which is dependent on the operating system. There's also a chance that this new rule is specific for 'Icon\r', but not knowing anything about macOS, I can't be sure that it's valid since the current rule is for 'Icon\r\r', which means an additional '[\r]' or '?' may be necessary.

I'm putting this here in the hope that someone can test it on my behalf, which would either lead to slightly better cross-platform support for this particular rule, or to me being set on the right path for adding this to my .gitignore file.

### DIFF
--- a/Global/macOS.gitignore
+++ b/Global/macOS.gitignore
@@ -3,8 +3,7 @@
 .LSOverride
 
 # Icon must end with two \r
-Icon
-
+Icon[\r]
 
 # Thumbnails
 ._*


shiftkey:
@BranicYeti I was able to find an instance of this file and tested out this change - unfortunately it didn't seem to work for me (this is with Git 2.11.0 on macOS):

```
$ less .gitignore
Icon[\r]
$ git status
On branch master

Initial commit

Untracked files:

	.gitignore
	"Icon\r"

```

Am I missing something?

BranicYeti:
Like I said, I'm not sure, so take my comments with a pinch if salt.

Regex allows \# in square brackets, so [\r] should match the return. Is the icon file actually called "Icon\r", in which case we would need to escape the backslash? I doubt it, but I really have no idea.

shiftkey:
I've tried a few variations on this but without any luck. I also asked a few fellow Git gurus and couldn't come up with a way out of this.

Thanks for proposing this anyway. Maybe we'll stumble on the right solution eventually...

PR #2207, opened 2016-12-02
Add zip and gzip extensions to Perl.gitignore

sshaw:

ExtUtils::MakeMaker based builds can generate .zip and .gz files

http://perldoc.perl.org/ExtUtils/MakeMaker.html#Distribution-Support

### DIFF
--- a/Perl.gitignore
+++ b/Perl.gitignore
@@ -1,23 +1,34 @@
-/blib/
-/.build/
-_build/
-cover_db/
-inc/
-Build
 !Build/
-Build.bat
 .last_cover_stats
-/Makefile
-/Makefile.old
-/MANIFEST.bak
 /META.yml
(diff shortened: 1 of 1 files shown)


Perl.gitignore
```diff
@@ -18,3 +18,5 @@ nytprof.out
 *.o
 *.bs
 /_eumm/
+/*.zip
```
shiftkey: Could you include a comment before this line to indicate which tools are responsible for this?

I know this hasn't really been done before in this file, but it can't hurt to start annotating things...

sshaw: See my 2nd commit. I annotated as many of the existing entries as I could.

shiftkey: @sshaw fantastic, thanks!

PR #2208, opened 2016-12-06
Existing .gitignore for vim incorrectly matched matched ".swq" through ".swz". 

roubles:

Existing .gitignore for vim is overzealous and incorrectly matched matched ".swq" through ".swz". 

You can see the source code here and see that those file extensions can not be generated:

http://unix.stackexchange.com/questions/326707/vim-what-are-all-the-possible-swapfile-extensions


shiftkey:
@roubles thanks!

PR #2209, opened 2016-12-06
Added pattern for expex forward references

vkusvody:
Linguistic examples package `expex`, with the switch `\gathertags` on, writes forward references to <_mainFile_>`-tags.tex`. It's annoying, because this is a special case of `.tex` that needs to be ignored.

[Expex documentation](http://mirror.aut.ac.nz/CTAN/macros/plain/contrib/expex/expex-doc.pdf), page 62


shiftkey:
@vkusvody thanks!

PR #2210, opened 2016-12-07
[Node] make directory rules explicit

mcandre:

Make patterns more standardized for command line tool parsing

https://github.com/monochromegane/go-gitignore/blob/master/pattern.go#L18-L41

Some .gitignore parsers interpret directory vs non-directory patterns strictly (e.g. monochromegane/go-gitignore), so a `node_modules` pattern would not necessarily match against a real `node_modules` directory. It's generally safe to add a trailing slash to directory .gitignore patterns.

### DIFF
--- a/Node.gitignore
+++ b/Node.gitignore
@@ -28,8 +28,8 @@ coverage
 build/Release
 
 # Dependency directories
-node_modules
-jspm_packages
+node_modules/
+jspm_packages/
 
 # Optional npm cache directory
 .npm


shiftkey:
I'm looking for a canonical source on the gitignore syntax, but the [Git SCM](https://git-scm.com/docs/gitignore) site has this interesting tidbit (emphasis mine): 

> If the pattern ends with a slash, it is removed for the purpose of the following description, but it would only find a match with a directory. In other words, `foo/` will match a directory `foo` and paths underneath it, but will not match a regular file or a symbolic link `foo` (**this is consistent with the way how pathspec works in general in Git**).

This is more for my curiosity than anything else, but I quietly worry about parsers being written where the behaviour differs from how Git reads these files (note the distinction between directories and files/symlinks).

shiftkey:
@mcandre I'll take this in while I'm chasing up my curiosity around parsers...

PR #2214, opened 2016-12-13
Update Xcode.gitignore to include macOS ignore

joihelgi:

Xcode should include os specific ignore... you are not going to use Xcode without macOS

https://github.com/github/gitignore/blob/master/Global/macOS.gitignore


shiftkey:
Thanks, but to keep maintenance of these templates sane I'm going to keep them separate for now.

I'm open to suggestions of alternative approaches.

PR #2215, opened 2016-12-13
ignore compile_commands.json

TennyZhuang:

CMake may generate `compile_commands.json` file with `set(CMAKE_EXPORT_COMPILE_COMMANDS ON)` Option

https://cmake.org/cmake/help/v3.7/variable/CMAKE_EXPORT_COMPILE_COMMANDS.html?highlight=cmake_export_compile_commands


shiftkey:
@TennyZhuang thanks!

PR #2218, opened 2016-12-22
PureBasic

tajmone:
New .gitignore template for the PureBasic language and IDE.

Currently there is no .gitignore template for the PureBasic language.

**Official PureBasic documentation:** 

- http://www.purebasic.com/documentation/index.html

**PureBasic project’s homepage**:

- http://www.purebasic.com/

### DIFF
--- a/PureBasic.gitignore
+++ b/PureBasic.gitignore
@@ -1,0 +1,15 @@
+# PureBasic
+
+# Settings files
+*.pb.cfg
+*.pbi.cfg
+project.cfg
+
+# Shared libraries
+*.so
+*.dylib
+*.dll
+
+# Binary executables
+*.exe
+*.app


tajmone:
PureBasic
=========

PureBasic programming language and IDE.

-   <http://www.purebasic.com/>
-   <http://www.purebasic.com/documentation/index.html>
-   [Wikipedia entry for PureBasic](https://en.wikipedia.org/wiki/PureBasic)

PureBasic is a commercial, cross platform programming language and IDE. It runs on — and compiles to — Windows, Linux and Mac OS X.

> NOTE: I shall use here “Mac OS X” instead of “macOS”, because this is how PureBasic documentation refers to it.

I use PureBasic on Windows, so I might be missing out some file extensions for Linux and Mac OS X related files. The official PureBasic documentation doesn’t provide details in this regard.

But I am sure that all patterns provided for PureBasic’s `.gitignore` file do cover files which are not meant for sharing on Git repository — either compiler productions, or personal settings files.

Files Extensions
----------------

PureBasic source code and project files have the following extensions:

-   `*.pb` — source code file.
-   `*.pbi` — include source code file.
-   `*.pbp` — PureBasic project file.
-   `*.pbf` — PureBasic IDE’s Form Designer form file.

There is no intrinsic difference between `*.pbi` and `*.pb` source code files, except for `*.pbi` being a clue that the file is meant for inclusion by some other file and is not intended for standalone use — ie: the `*.pbi` extension is conventional used for libraries, modules, or for separating procedures declarations and constants definition from the main source, in a fashion similar to `*.h` header files in C.

File with a `*.pbp` extension are PureBasic project files, in XML format.

Settings Files
--------------

Two type of PureBasic settings files can be found along projects:

-   `*.pb.cfg` — per-source settings files: for each purebasic source file.
-   `*.pbi.cfg` — same as above, but for `*.pbi` source files.
-   `project.cfg` — per-folder settings files.

The `*.pbi.cfg` is undocumented, but I’ve tested it.

Compiled Binaries
-----------------

PureBasic uses the PellesC linker (polink); avanced user can invoke PureBasic compiler passing specific directive for polink, therefore it might be possible to also produce binaries with other extensions then those referred to here below — but this would be rare enough to justify excluding them from the `.gitignore` file.

### Compiled Binary Executables

PureBasic can compile to binary executables for Windows, Linux and Mac OS X.

-   `*.exe` — Windows executable.
-   `*.app` — Mac OS X bundled executable

I haven’t had a chance to work with PureBasic under Linux or Mac, but from examining other people examples it seems that Linux binaries have no extension by default. But I might be wrong on this, and missing some patterns.

Also, under Mac OS, there might be other extension patterns I am not aware of.

### Compiled Dynamic libraries

PureBasic can compile to shared libraries for Windows, Linux and Mac OS X.

-   `*.so` — Linux shared object.
-   `*.dylib` — Mac OS X dylib.
-   `*.dll` — Windows DLL.

------------------------------------------------------------------------

Documented References
=====================

PureBasic Files Extensions
--------------------------

PureBasic default file extensions are mentioned in scattered places within the official documentation.

All extensions:

> PureBasic files (`*.pb`, `*.pbi`, `*.pbp`, `*.pbf`) \[…\]
>
> — <http://www.purebasic.com/documentation/reference/ide_tools.html>

`*.pb`, `*.pbi` and `*.pbf` extensions:

> The standard PureBasic file extensions (`pb`, `pbi` `and` `pbf`) are recognized as code files.
>
> — <https://www.purebasic.com/documentation/reference/ide_files.html>

`*.pbf` extension:

> Every form file should be named with the ‘`.pbf`’ extension, so it will be correctly opened by the IDE.
>
> — <http://www.purebasic.com/documentation/reference/ide_form.html>

`*.pbp` extension:

Compiled Binaries Extensions
----------------------------

> **Create executable**
>
> This opens a save dialog, asking for the executable name to create. If the executable format is set to DLL, it will create a DLL on Windows, shared object on Linux and dylib on OS X. When creating an executable on OS X, appending ‘.app’ at the executable name will create a bundled executable with the necessary directory structure, including the icon. If no ‘.app’ is set, then it will create a regular console-like executable.
>
> — <http://www.purebasic.com/documentation/reference/ide_compiler.html>

Advanced compiling options are described here:

— <http://www.purebasic.com/documentation/reference/cli_compiler.html>

Configuration Files
-------------------

> **Save settings to**
>
> This option allows to specify where the compiler options of a source file are saved:
>
> -   **The end of the Source file**
>
>     Saves the settings as a special comment block at the end of each source file.
>
> -   **The file &lt;filename&gt;.pb.cfg**
>
>     Creates a `.pb.cfg` file for each saved source code that contains this information.
>
> -   **A common file project.cfg for every directory**
>
>     Creates a file called `project.cfg` in each directory where PB files are saved. This one file will contain the options for all files in that directory.
>
> -   **Don’t save anything**
>
>     No options are saved. When reopening a source file, the defaults will always be used.
>
> — <http://www.purebasic.com/documentation/reference/ide_preferences.html>

joeblau:
@tajmone - This tempalte is live on [gitignore.io](https://www.gitignore.io) - https://www.gitignore.io/api/purebasic

tajmone:
Thanks for merging it into gitignore.io @joeblau !

Any chances it will also be merged into this repo? 

It's been a while since the PR but didn't get any feedback.

shiftkey:

@tajmone if you still feel like this should be a template listed in this repository, please read the section about [specialized templates](https://github.com/github/gitignore#specialized-templates) and open a new pull request matching the new format.

PR #2219, opened 2016-12-23
[Gradle] build rule should support subprojects

guilhermerigorecio:
The current form does not work with multiple subprojects.

### DIFF
--- a/Gradle.gitignore
+++ b/Gradle.gitignore
@@ -1,5 +1,5 @@
 .gradle
-/build/
+build/
 
 # Ignore Gradle GUI config
 gradle-app.setting


drothmaler:
Hm, but this way we would also ignore all other "build" folders (e.g. if one of your packages is: "com.company.project.build.tools"). So, I think we should at least unignore all `build` folders inside of "src/" folders...

shiftkey:
I'm not sure whether a wildcard term will help or hinder here, as people like to organize their source code in different ways (there's already mentions of `src` and `buildSrc`).

Could we document this rule better to explain that it will only cover `build` at the root, and how to add additional rules to handle subproject scenarios?

shiftkey:

Baneeishaque:
What about this? @ @ @ @shiftkey

PR #2221, opened 2016-12-24
Update Joomla gitignore

brianteeman:
Removed some non core joomla extensions 
Removed some old joomla 2.5 templates
Removed some index.html no longer present
Removed weblinks component, modules and plugins no longer distributed in the core

github.com/joomla/joomla-cms


shiftkey:
Deleting rules, excellent! :fire:

@brianteeman thanks!

PR #2222, opened 2016-12-24
[TeX] delete Leading blanks

yassu:

Template of TeX includes a leading blank.
I think You had better delete it.


shiftkey:
@yassu thanks!

PR #2225, opened 2016-12-27
Update CommonLisp.gitignore

ailisp:

For Clozure Common Lisp, the fast load file is not .fasl or .lisp-temp. It's platform specific. When I use it on Windows, it produce .wx32cl files. On other platforms the suffix is different.

[ccl-doc](https://www.cs.utexas.edu/users/jared/Milawa/Support/ccl/doc/ccl-documentation.html#Platform-specific-filename-conventions)

### DIFF
--- a/CommonLisp.gitignore
+++ b/CommonLisp.gitignore
@@ -1,3 +1,17 @@
 *.FASL
 *.fasl
 *.lisp-temp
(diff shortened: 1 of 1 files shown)


shiftkey:
@ailisp thanks!

PR #2228, opened 2016-12-29
ignore gradle wrapper

dckg:

I don't want to commit an external build tool in my project. 

Gradle Wrapper files: https://docs.gradle.org/current/userguide/gradle_wrapper.html#using_wrapper_scripts

Gradle Wrapper can be ignored completely, however, to ensure that the correct Gradle version is used to build a project, Gradle version can be set in ``build.gradle`` file.
http://stackoverflow.com/a/28473686

For Android Studio, Gradle version is set in ``build.gradle`` by default
https://developer.android.com/studio/releases/gradle-plugin.html

### DIFF
--- a/Gradle.gitignore
+++ b/Gradle.gitignore
@@ -10,5 +10,7 @@ gradle-app.setting
 # Cache of project
 .gradletasknamecache
 
-# # Work around https://youtrack.jetbrains.com/issue/IDEA-116898
-# gradle/wrapper/gradle-wrapper.properties
+# ignore gradle wrapper scripts & binary
+gradlew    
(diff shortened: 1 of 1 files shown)


drothmaler:
Nope definitly not... 
The [Gradle documentation](https://docs.gradle.org/current/userguide/gradle_wrapper.html#sec:wrapper_generation) states: "All of these files _**should**_ be submitted to your version control system."
And this is the way, gradle is usually used... 
If **you** decide to use it in another way, you can simply modify your local `.gitignore` file. But by default, the gradle wrapper should _not_ be ignored.

PR #2229, opened 2016-12-29
[TeX] Add .spl and .soc

koppor:

`.soc` and `.spl` were missing for TeX

- .soc - package [changes](https://www.ctan.org/pkg/changes)
- .spl - documentclass [elsarticle](https://www.ctan.org/pkg/elsarticle) (Elsevier journals)

http://tex.stackexchange.com/a/256803/9075


shiftkey:
@koppor thanks!

PR #2231, opened 2017-01-02
SublimeText: ignore new Package Control certs

acarabott:

Sublime Text's Package Control now has a number of bundles for certificates instead
of the old directory. This ignores them as per comments from Package Control's author.

http://github.com/wbond/package_control/issues/1153
https://forum.sublimetext.com/t/generating-oscrypto-ca-bundle-crt-after-mac-os-install-and-security/23020


shiftkey:
@acarabott thanks!

PR #2232, opened 2017-01-03
[Android] feat: Ignore VSCode Eclipse JDT files

jpventura:

VSCode Java IntelliSense creates some files unrelated to Android
application and should not be added to project history.

Java language on [Visual Studio Code](https://code.visualstudio.com/) is supported via [java-language-server](https://github.com/gorkem/java-language-server),
which internally utilizes [Eclipse JDT](http://www.eclipse.org/jdt/).

  - [Visual Studio Code](https://code.visualstudio.com/)
  - [Language Support for Java(TM) by Red Hat](https://marketplace.visualstudio.com/items?itemName=redhat.java)
  - [Java Language Server](https://github.com/gorkem/java-language-server)
  - [Eclipse JDT](http://www.eclipse.org/jdt/)

### DIFF
--- a/Android.gitignore
+++ b/Android.gitignore
@@ -47,3 +47,9 @@ captures/
 
 # Google Services (e.g. APIs or Firebase)
 google-services.json
+
+# Visual Studio Code
+.classpath
+.project
+.settings
+javac-services.*


friederbluemle:
These files are related to a plugin in Visual Studio Code (which is not an officially supported IDE). If you use use this editor and want to ignore the generated files, I would suggest you add these lines manually to your `.gitignore`, as they won't apply to 99% of Android developers.

shiftkey:
We have `Global\Eclipse.gitignore` which probably already has these rules. I'm :thumbsdown: to duplicating rules because it makes keeping things in sync harder, and I'd recommend tailoring your projects `.gitignore` to the languages and tools you use.

PR #2233, opened 2017-01-03
[Maven] fix comment

tjkemper:
The ! prefix will unignore the file.

The comment was misleading.

N/A

If this is a new template: 

N/A

### DIFF
--- a/Maven.gitignore
+++ b/Maven.gitignore
@@ -8,5 +8,5 @@ dependency-reduced-pom.xml
 buildNumber.properties
 .mvn/timing.properties
 
-# Unignore maven wrapper jar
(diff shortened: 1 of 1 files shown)


dckg:
i find 'exclude' more appropriate

shiftkey:
After thinking about this for a while, I'm not sure either term is better. I like this example from the Gradle template:

```
# Avoid ignoring Gradle wrapper jar file (.jar files are usually ignored)
!gradle-wrapper.jar
```

tjkemper:
@shiftkey made change as per your suggestion.

shiftkey:
@tjkemper thanks!

PR #2236, opened 2017-01-06
VisualStudio: Don't ignore project filters

Armada651:

These filters provide structure to the project tree and are thus important to make sure every team member see the same structure. This is not necessarily a user-specific file unless the team decides that they want to manage the filters per-user.

http://stackoverflow.com/questions/1826901/should-i-add-vcxproj-filter-files-to-source-control

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -8,7 +8,6 @@
 *.user
 *.userosscache
 *.sln.docstates
-*.vcxproj.filters
 
 # User-specific files (MonoDevelop/Xamarin Studio)
 *.userprefs


shiftkey:
Interesting: the same link was used in #2166 where we added it.

Looks like I suggested this change a few weeks ago: https://github.com/github/gitignore/commit/c6a0ceae0863806d829b4b15ef6d29b49294a4cc#commitcomment-20053612

cc @davesalls @jvcdk for additional thoughts.

jvcdk:
Maybe I was a bit too quick and only read the second (non-accepted) answer; maybe I was too quick in reading the accepted answer (not understanding it correctly). Anyway, returning to this, one could not a couple of things
 - The accepted answer favour adding it to SCC.
 - The second answer seem to favour not adding it to SCC (and he *sounds* like it came from somebody inside MS, developing on the MsBuild system - but I did not google it further right now).
 - The accepted answer has upvote of 43 vs. second answer has upvote of 82.

To be honest, I am ok with reverting this commit, if that is what you think is best. Unfortunately I do not work in depth with Visual C++ enough to have a firm opinion myself.

jvcdk:
Ps.: I realized later that when you create a new C++ project in Visual Studio and enable Git-integration, VS will also create a .gitignore file for you. But I forgot whether it included project filter file, and also think it depends a bit on which version of VS you have. But I am not a work right now and cannot check up on it.

shiftkey:
@jvcdk from memory the `.gitignore` file from VS is derived from this source, so getting a formal answer should address it in both places eventually.

Armada651:
@jvcdk Though filter files are specific to VS, the concept of a project structure is a general concept. If you have a project in any other language you would normally also want to commit the project structure and not just the files.

shiftkey:
I asked on Twitter and got a lot of replies back - the consensus seems to be that these should be versioned, as it does affect the layout of the source files in VS.

Here's a demo from @porges:

#### With `.filters` file



#### Without `.filters` file



The goal with this template is to be good for the defaults, so I'm :thumbsup: to ensuring these files are tracked in version control.

PR #2237, opened 2017-01-07
[ADD] Resharper *.DotSettings file

jingkecn:
Add *.DotSettings file to VisualStudio.gitignore.
Reference: http://stackoverflow.com/questions/22234617/do-i-need-to-maintain-resharpers-dotsettings-user-file-in-source-control-tfs
> No, the dotSettings.user file should be excluded from source control. It has user specific information in it, and shouldn't be shared, e.g. bookmarks, unit test session windows.

>The .dotSettings file should be included and contains project or solution specific settings that can be shared with the team, e.g. naming standards or braces layout.

I believe that not all members in a team use Resharper.

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -103,6 +103,7 @@ $tf/
 # ReSharper is a .NET coding add-in
 _ReSharper*/
 *.[Rr]e[Ss]harper
+*.DotSettings
 *.DotSettings.user
 
 # JustCode is a .NET coding add-in


jingkecn:
Reference: http://stackoverflow.com/questions/22234617/do-i-need-to-maintain-resharpers-dotsettings-user-file-in-source-control-tfs

shiftkey:
I think the linked solution suggests that the `.dotSettings` file should be shared:

> The .dotSettings file **should be included** and contains project or solution specific settings **that can be shared with the team**, e.g. naming standards or braces layout.

jingkecn:
But what if some of team members use Resharper but some don't, is *.DotSettings still necessary for team source control?
But you're right, this still needs some discussion.

shiftkey:
@jingkecn the purpose of the file is to share team-specific settings, and the existing rule about excluding `*.dotSettings.user` files is there for user-specific settings (which should not be shared). The `*.user` naming convention is common in Visual Studio project files, so I think this is all working as intended.

If you don't have R# installed then there's no effect by having this file in your repository.

jingkecn:
@shiftkey Thx for your explanation, I see for now, sorry I misunderstood for this PR, it can be closed now.

PR #2239, opened 2017-01-10
Add ignoring PyCharm IDE project settings directory to Python.gitigno…

mglijer:
Add ignoring PyCharm IDE project settings directory to Python.gitignore file.

PyCharm IDE creates own project settings directory. Ignoring it is a good practice


shiftkey:

PR #2240, opened 2017-01-10
Add debugging files to the gitignore

rogersachan:

`.pdb` files are created by Visual Studio in the root directory solely for Windows debugging. These files cannot be used in other supported IDEs and are therefore not needed.

https://docs.unity3d.com/Manual/WindowsDebugging.html


shiftkey:
@rogersachan thanks!

PR #2241, opened 2017-01-10
Corrected ignore settings for Cake

gep13:
- It is recommended to include the packages.config file when using Cake
- This ensures that the Cake version that is being used is *pinned* and updated only when the user decides


shiftkey:
@gep13 thanks!

gep13:
@shiftkey thank you for merging so quickly!

/cc @damianh

PR #2242, opened 2017-01-11
cmake: add Testing/ directory created by ctest

mcandre:
Ignore logs and other junk generated by cmake's `ctest` command in `Testing/`


shiftkey:
@mcandre thanks!

PR #2244, opened 2017-01-13
Add Google Services

FionaHagen:

Many [Android](https://www.android.com/) apps use Google Services like the Google APIs or [Firebase](https://firebase.google.com/). The file google-services.json contains unique and personal information which should be kept secret. 

https://developers.google.com/android/guides/google-services-plugin
https://github.com/googlesamples/google-services/blob/master/.gitignore


shiftkey:
@FionaHagen thanks!

PR #2246, opened 2017-01-14
[Global/JetBrains] add support for Rider and nested projects

Nihlus:
In newer versions of the IntelliJ platform (such as the one used for Rider), the .idea files are now placed in a subdirectory of the same, named after the current project (such as ".idea/.idea.Everlook/.idea/workspace.xml"). Adding a double-star pattern to these folder rules will also ignore these new files and maintain backwards compatibility with previous revisions of this file.

### DIFF
--- a/Global/JetBrains.gitignore
+++ b/Global/JetBrains.gitignore
@@ -2,24 +2,24 @@
 # Reference: https://intellij-support.jetbrains.com/hc/en-us/articles/206544839
 
 # User-specific stuff:
-.idea/workspace.xml
-.idea/tasks.xml
+.idea/**/workspace.xml
+.idea/**/tasks.xml
 
 # Sensitive or high-churn files:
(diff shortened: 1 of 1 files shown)


shiftkey:
cc @citizenmatt for :eyes: and :thumbsup:

citizenmatt:
Sorry for the delay folks, chasing this down internally.

It's a change specifically for Rider, and doesn't affect other IntelliJ based products. It means that multiple solution files in a single folder will get separate settings folders. This doesn't affect other IntelliJ based products as they are essentially folder based, rather than based on a single project.

The double star pattern is good in that it will continue working not only with current IntelliJ products, as well as Rider.

shiftkey:
@citizenmatt thanks for the :thumbsup:, and thanks for the contribution @Nihlus!

PR #2250, opened 2017-01-21
Ignore DS_Store on OSX

n7down:
Ignores the .DS_Store for OSX machines


shiftkey:

PR #2251, opened 2017-01-23
Updating VisualStudio.gitignore to ignore thumbs.db

jasongaylord:
The thumbs.db file is added by Windows when adding images. This file is used to quickly show thumbnails when browsing in Windows and is not needed by source control.

[https://en.wikipedia.org/wiki/Windows_thumbnail_cache](https://en.wikipedia.org/wiki/Windows_thumbnail_cache)

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -201,6 +201,7 @@ ClientBin/
 *.publishsettings
 node_modules/
 orleans.codegen.cs
+thumbs.db
 
 # Since there are multiple workflows, uncomment next line to ignore bower_components
 # (https://github.com/github/gitignore/pull/1529#issuecomment-104372622)


shiftkey:
Thanks, but see [`Global\Windows.gitignore`](https://github.com/github/gitignore/blob/master/Global/Windows.gitignore)

jasongaylord:
@shiftkey that makes sense and apologies for missing that rule. But then why is there a VisualStudio.gitignore in the root? Why aren't language specific rules at the top level and the VisualStudio.gitignore in the Global?

shiftkey:
@jasongaylord it's a historical thing. Previously we had `CSharp.gitignore` and `VisualBasic.gitignore` which were essentially the same, and it turns out most of the rules were related to Visual Studio rather than each language - hence the unification...

PR #2252, opened 2017-01-23
Add Map Built Data

calben:
Add built data to the gitignore.
Includes data about a map for UE4.14 that is generated when the user builds it (surprise! :D).
Often gigabytes in size.
You definitely don't want it in your repository.


shiftkey:
@calben is there a link to some documentation about this feature, so that others can understand the context around this change?

calben:
It's a pretty new feature that was finished early December, but as users move to UE4.14 it will become relevant.
Right now this file is used for only lighting data, but it could be expanded to include other data.
You can see the feature being discussed in this Youtube video under "Static Lighting Scenarios" https://www.youtube.com/watch?v=6kfMVxNSowM&feature=youtu.be&t=1230.
Unfortunately, I don't think there is any other official documentation on it yet!
The feature slipped under my radar until a Github push failed because of a built data file.

shiftkey:
I asked around on Twitter and got no feedback. Oh well, this seems specific enough.

@calben thanks!

PR #2256, opened 2017-01-24
[Python] add .venv as file

DSIW:

I use virtualenvwrapper with auto loading of python environment after changing directory containing `.venv`.

https://virtualenvwrapper.readthedocs.io/en/latest/tips.html?highlight=venv#automatically-run-workon-when-entering-a-directory

[Oh-My-ZSH's plugin for virtualenvwrapper](https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/virtualenvwrapper/virtualenvwrapper.plugin.zsh#L33-L35) uses this file, too.

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -80,10 +80,9 @@ celerybeat-schedule
 .env
 
 # virtualenv
-.venv/
+.venv
 venv/
 ENV/
-.venv
 
 # Spyder project settings
 .spyderproject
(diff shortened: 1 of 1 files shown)


Python.gitignore
```diff
@@ -83,6 +83,7 @@ celerybeat-schedule
 .venv/
 venv/
 ENV/
+.venv
```
shiftkey: Does this make the `.venv/` rule above obsolete?

DSIW:
Yes, it does. I removed that /.

shiftkey:
@DSIW thanks!

PR #2258, opened 2017-01-25
Update CMake.gitignore

Veemon:

added due to JetBrains .ignore focusing on .idea, and the possibility of developers using both CLion and Terminal to build projects.

### DIFF
--- a/CMake.gitignore
+++ b/CMake.gitignore
@@ -7,3 +7,4 @@ cmake_install.cmake
 install_manifest.txt
 compile_commands.json
 CTestTestfile.cmake
+cmake-build-debug


shiftkey:
@Veemon it's not clear to me what the link is referring to. Is there another resource explaining the purpose of this file, any why it can be safely ignored?

Veemon:
@shiftkey it is the default build folder for some editors and IDE's 
> https://blog.jetbrains.com/clion/2016/10/clion-2016-3-eap-cmake-overload-resolution/

retracting pull request however as I failed to see the redundancy in this change.

PR #2268, opened 2017-01-31
exclude .vscode folder from VisualStuido.ignore

yangzhyo:

.vscode folder should only be committed when user want to share setting with team or others.

[https://code.visualstudio.com/Docs/customization/userandworkspace](https://code.visualstudio.com/Docs/customization/userandworkspace)

[http://stackoverflow.com/questions/32964920/should-i-commit-the-vscode-folder-to-source-control](http://stackoverflow.com/questions/32964920/should-i-commit-the-vscode-folder-to-source-control)

If this is a new template:  No.

[Visual Studio Code](https://code.visualstudio.com/)

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -29,6 +29,9 @@ bld/
 # Uncomment if you have tasks that create the project's static files in wwwroot
 #wwwroot/
 
+# Visual Studio Code setting folder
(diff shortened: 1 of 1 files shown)


shiftkey:

yangzhyo:
got it.

PR #2270, opened 2017-02-02
Ignoring more generated file for Java project

tcthien:

- Adding generated Eclipse project to ignore
- Adding log file, zip, tar.gz, rar to ignore
- Adding generated vagrant file to ignore

### DIFF
--- a/Java.gitignore
+++ b/Java.gitignore
@@ -18,6 +18,9 @@
 *.jar
 *.war
 *.ear
+*.zip
+*.tar.gz
+*.rar
 
 # Vagrant
 .vagrant
--- a/Java.gitignore
+++ b/Java.gitignore
@@ -1,5 +1,13 @@
 *.class
 
+# Eclipse editor
+.classpath
+.project
+.settings/
+
+# Log file
+.log
+
 # BlueJ files
 *.ctxt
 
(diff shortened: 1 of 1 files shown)


Java.gitignore
```diff
@@ -1,5 +1,13 @@
 *.class
 
+# Eclipse editor
```
shiftkey: We have an existing [`Global\Eclipse.gitignore`](https://github.com/github/gitignore/blob/master/Global/Eclipse.gitignore). Please drop these rules.

Java.gitignore
```diff
@@ -1,5 +1,13 @@
...
+.classpath
+.project
+.settings/
+
+# Log file
+.log
```
shiftkey: Should the be `*.log`?

Java.gitignore
```diff
@@ -10,6 +18,12 @@
 *.jar
 *.war
 *.ear
+*.zip
+*.tar.gz
+*.rar
+
+# Vagrant
```
shiftkey: We have an existing [`Global\Vagrant.gitignore`](https://github.com/github/gitignore/blob/master/Global/Vagrant.gitignore). Please drop these rules.

shiftkey:
@tcthien thanks for these changes. I've called out the ones that are stored elsewhere, but the rest seem fine.

tcthien:
Got the idea, I wasn't aware about global ignore. I'm going to remove the duplicate and open another PR

PR #2271, opened 2017-02-02
Update scala.gitignore add intellij

stephennancekivell:
Add intellij. It seems to be more popular than scala-ide.

Add a usefull .gitignore entry.

As in https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore

### DIFF
--- a/Scala.gitignore
+++ b/Scala.gitignore
@@ -21,3 +21,6 @@ project/plugins/project/
 # ENSIME specific
 .ensime_cache/
 .ensime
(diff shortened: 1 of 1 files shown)


shiftkey:
The [`Global\JetBrains.gitignore`](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore) rule is a bit more detailed than this, and I'd rather use their rules than hide everything.

PR #2273, opened 2017-02-03
Ignoring more generated file for Java project

tcthien:
- Adding log file, zip, tar.gz, rar to ignore


shiftkey:
@tcthien thanks!

PR #2276, opened 2017-02-08
Adding a rule for ignoring the target directory from the build

anthonyrichir:

I think adding an ignore for the target directory produced by the build would be very useful for others, just like I find it useful myself.

### DIFF
--- a/Java.gitignore
+++ b/Java.gitignore
@@ -19,3 +19,6 @@
 
 # virtual machine crash logs, see http://www.java.com/en/download/help/error_hotspot.xml
 hs_err_pid*
+
+# Build directory
+target/


sepatus:
i am tired.working 2 days and no sleep.ok.i want rest

drothmaler:
`target/` is the (default) output directory, when building Java-Sources with Maven. Plain Java (javac) has no `target` directory; unless you explicitly chose to specify it, for your project.
--> This should not be in here!

If you are using Maven, please just use `Java.gitignore` in combination with `Maven.gitignore` (which already contains `target/`).

shiftkey:
Seconding what @drothmaler said - this is covered by `Maven.gitignore`.

Closing this out to avoid unnecessary duplication.

anthonyrichir:
I didn't see there was a Maven.gitignore. Sorry for you time.

PR #2285, opened 2017-02-16
Update TeX.gitignore

koppor:
When running latex and using [minitoc](https://www.ctan.org/pkg/minitoc), it also generates .stc, .slf, and .slt files.

**Links to documentation supporting these rule changes:** See manual page 29: http://texdoc.net/texmf-dist/doc/latex/minitoc/minitoc.pdf

grafik


shiftkey:
@koppor thanks!

PR #2286, opened 2017-02-16
python pycharm

vasion:
adding the .idea folder to python ignore for pycharm and intellij


shiftkey:

vasion:
Ah, I don't understand how I did not see that. Thanks

PR #2287, opened 2017-02-16
[SBT] Add some excludes kept in Scala.gitignore .

sherzodv:
Scala.gitignore currently has more entries for SBT than SBT.gitignore itself.
Copy missing entries to SBT.gitignore.

Cleanup Scala.gitignore: https://github.com/github/gitignore/pull/2289


shiftkey:
@sherzodv thanks!

PR #2290, opened 2017-02-17
Ignore dictionaries to reflect docs

aleixosk:
The official docs have been updated regarding this.

Dictionaries are saved under `.\dictionaries\username.xml` and should thus be ignored.


shiftkey:
@aleixosk thanks!

PR #2291, opened 2017-02-20
[Python Gitignore] Ignore Sage's parsed files

pmsosa:
Ignore Sage's parsed files (*.sage.py) that are left after running a .Sage program.

- [Sage](https://www.sagemath.org/) (or SageMath) is a popular mathematics software built atop Python. 
- Sage uses .sage files that it translates into .sage.py files to run.  Essentially these .sage.py are rather irrelevant as one wouldn't want to save them in most cases.
- I suggest this edit to the Python .gitignore, seeing as it would be an overkill to add a whole new template for just for Sage (seeing as Sage is essentially an extension Python).

**Links**
[Sage Official Page](https://www.sagemath.org/)
[Wikipedia on Sage](https://en.wikipedia.org/wiki/SageMath)
[Explanation of Sage extensions](https://doc.sagemath.org/html/en/tutorial/programming.html)

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -76,7 +76,7 @@ target/
 # celery beat schedule file
 celerybeat-schedule
 
-#Sage parsed files
+# SageMath parsed files
 *.sage.py
 
 # dotenv
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -76,6 +76,9 @@ target/
 # celery beat schedule file
 celerybeat-schedule
 
+#Sage parsed files
+*.sage.py
+
 # dotenv
 .env
 


Python.gitignore
```diff
@@ -76,6 +76,9 @@ target/
 # celery beat schedule file
 celerybeat-schedule
 
+#Sage parsed files
```
shiftkey: Add a space after the `#`

shiftkey: SageMath seems to be the latest name here (as per Wikipedia at least). Would that make this a bit clearer?

pmsosa: Indeed; Just updated both suggestions.

shiftkey:
@pmsosa thanks!

PR #2295, opened 2017-02-23
update Objective C .gitignore 

irshadpc:

In every mac there is a auto generated hidden file called .DS_Store it usually kind a behave of a file, so most of the time i fix conflict for this particular file and it is hard to fix for me. 

_TODO_

_TODO_

If this is a new template: 


shiftkey:
Thanks, but see [`Global\JetBrains.gitignore`](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore) and [`Global\macOS.gitignore`](https://github.com/github/gitignore/blob/master/Global/macOS.gitignore).

irshadpc:
@shiftkey i usually take care of objective/swift .gitignore file that because i added that file.

PR #2296, opened 2017-02-23
update VS CAKE path more clear

GhostTW:
tools folder is too common use, ignore tools folder makes more problems.

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -274,5 +274,6 @@ __pycache__/
 *.pyc
 
 # Cake - Uncomment if you are using it
-# tools/**
+# tools/Cake/
+# tools/Addins/
 # !tools/packages.config


shiftkey:
@gep13 are you cool with this?

gep13:
@shiftkey sorry for not getting back to you sooner about this.  We have just had our second daughter 😄 

I am not sure if this is a good idea.  The typical Cake tools folder looks this this:

image

Making the suggested change would require multiple separate additions for the various tools that have been included in the project.

shiftkey:
This is more of a NuGet issue, right? Or could we use a subfolder to namespace better to isolate these things from other tools the user might have installed?

I don't mind helping to put together some sort of PR for this, just curious if you've already tried and dismissed the idea...

gep13:
@shiftkey sorry about length of time to reply.  Just getting back to things after arrival of new baby...

By default we call this folder `tools`, but we are not tied to this.  It is possible through configuration (http://cakebuild.net/docs/fundamentals/default-configuration-values#tools-path) to change the name/location of both the `tools` and the `addins` folder.

So I am not sure if there is anything that needs to be done here, is there?

gep13:
>@GhostTW said...
ignore tools folder makes more problems.

What sort of problems are you running into?

shiftkey:
@GhostTW waiting on clarification about this:

> > ignore tools folder makes more problems.

> What sort of problems are you running into?

gep13:
@GhostTW please let us know if you have any questions about Cake functionality in this area.  Happy to help if we can.

shiftkey:
Closing due to inactivity. @GhostTW if you're able to answer the questions @gep13 was asking earlier we can revisit whether this change is still required.

PR #2297, opened 2017-02-24
Prevent unintended folders from being ignored

maban:

I'm not sure of the best way to fix this (would rather flag it as an issue than suggest this solution get merged), but keeping this in causes any folder in a any project with the name "Icon" to be ignored. I use GitHub for Mac and couldn't figure out why any folder I named "icon" (for front-end assets on a web app) kept disappearing from my code editor (Atom). It's because of this line.

Looking at the history, I understand it was added to prevent custom Finder icons from being included, but is there a way to make this exclusion less general?

Sorry if this is the wrong place – please point me towards where I should raise this as an issue.

### DIFF
--- a/Global/macOS.gitignore
+++ b/Global/macOS.gitignore
@@ -2,10 +2,6 @@
 .AppleDouble
 .LSOverride
 
-# Icon must end with two \r
-# Icon
-
-
 # Thumbnails
 ._*
 
--- a/Global/macOS.gitignore
+++ b/Global/macOS.gitignore
@@ -3,7 +3,7 @@
 .LSOverride
 
 # Icon must end with two \r
-Icon
+# Icon
 
 
 # Thumbnails


shiftkey:
@maban this is the right place - this repository gets used in a few places, including GitHub for Mac.

Unfortunately this rule is an annoying artefact of history - the `\r\r` suffix is how the file is formatted on disk, and unfortunately editors do get confused about this and eagerly change the line ending to be incorrect :metal:. 

Unfortunately it seems to catch a lot of people out, and others have proposed solutions in the past that don't quite work, but I'm not sure I'm ready to drop this rule.

Global/macOS.gitignore
```diff
@@ -3,7 +3,7 @@
 .LSOverride
 
 # Icon must end with two \r
-Icon
+# Icon
```
shiftkey: Rather than comment this rule out, you should remove this whole section (including the comment above).

maban:
Thanks for the clarification! I was stumped for a solution (hence commenting the line out, as I couldn't raise an issue about it).

Given this is bundled with tools and not installed explicitly, I'd like to see this removed. It did catch me out and if I didn't have a more persistent colleague who wanted to figure it out (thanks, @joshje), I wouldn't have known why any folder I named _icon_ disappeared from Atom. And I think the type of people who might use GitHub for Mac over the terminal (it's popular with designers) are more likely to be naming folders that.

Or, would it be better to raise an issue with Atom so that it formats line endings correctly?

I'll amend this PR with the change you suggest, but I'll understand if you want to close it rather than merge it – you know the implications of this change much better than I do.

PR #2300, opened 2017-02-26
D: Ignore generated test files

wilzbach:

DUB (D's default package manager) generated test executables should be ignored by default.

https://github.com/dlang/dub/pull/1029
https://github.com/dlang/dub/blob/0e45525f85838a53c47f824edbd06af43ef8c2f2/source/dub/dub.d#L513

CC @s-ludwig @MartinNowak

### DIFF
--- a/D.gitignore
+++ b/D.gitignore
@@ -20,5 +20,8 @@ docs.json
 __dummy.html
 docs/
 
+# Ignored test executable created by DUB
+/*-test-*
+
 # Code coverage
 *.lst


D.gitignore
```diff
@@ -19,6 +19,7 @@
 docs.json
 __dummy.html
 docs/
+*-test-*
```
shiftkey: This doesn't really have much context - could you add a comment above to indicate what this rule covers?

s-ludwig:
I'm asking myself if this might be a bit too broad, as it could accidentally ignore unrelated files. What about letting `dub init` add `<targetname>-test-*` (with https://github.com/dlang/dub/issues/758 implemented)?

wilzbach:
> This doesn't really have much context - could you add a comment above to indicate what this rule covers?

Thanks a lot for the quick feedback :)
I added an explanatory comment. If I should provide more details, please let me know.

> I'm asking myself if this might be a bit too broad, as it could accidentally ignore unrelated files.

I changed it to `/*-test-*`, so that it only targets files at the root level, which I doubt will have (m)any false positives.

> What about letting dub init add <targetname>-test-* (with dlang/dub#758 implemented)?

That's a good idea as well, but as said imho `/*-test-*` is unique enough that we can avoid the need for DUB to do it's own hacks. Additionally, the fact that the issue is open for more than a year speaks for itself (i.e. this here is an easier & quicker way to "get this to the DUB users")

wilzbach:
Ping @s-ludwig @MartinNowak and other awesome D people: @ZombineDev @CyberShadow @John-Colvin

What's your take on ignoring `/*-test-*` in the root folder of D projects? 
Since https://github.com/dlang/dub/pull/1029 `<package>-test-<config>` is the default for the test binary created by DUB.

s-ludwig:
I don't have a strong opinion on this one. The DUB generated .gitignore already ignores these and there is the potential for blocking files that are not generated by DUB. On the other hand in reality this potential should be so small that it probably doesn't matter in practice. But apart from *-test*, one thing that could definitely be added is `.dub`.

shiftkey:
I'm going to close this out because it looked like the discussion trailed off without any strong feelings, and there are some outstanding questions about whether it affects other files.

> one thing that could definitely be added is `.dub`.

This seems like something worth submitting in another PR, if it makes sense.

PR #2301, opened 2017-02-28
Added Telerik's JustMock configruation files

MichaelKogan:
When Telerik's JustMock add-on is used in Visual Studio it generates a .jmconfig file to maintain the JustMock upgrade information.

http://stackoverflow.com/questions/32696176/what-is-creating-these-files-with-a-jmconfig-extension


shiftkey:
@MichaelKogan thanks!

PR #2305, opened 2017-03-01
Added BizTalk build output files to ignore list

robinhultman:

When compiling BizTalk projects a compiled version of every BizTalk artifact is generated. These should not be version controlled.

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -279,3 +279,9 @@ __pycache__/
 
 # Telerik's JustMock configuration file
 *.jmconfig
+
+# BizTalk build output
+*.btp.cs
+*.btm.cs
+*.odx.cs
+*.xsd.cs


niik:
Hey @robinhultman. Do you happen to have any links to documentation that document these file extensions and why they shouldn't be in version control? Searching through existing projects there seems to be a few [that include btm.cs](https://github.com/search?utf8=%E2%9C%93&q=filename%3A*.btp.cs&type=Code&ref=searchresults) for example.

robinhultman:
Hey. Sorry no offical sources but here is a discussion on the xsd.cs files om msdn forums

https://social.msdn.microsoft.com/Forums/en-US/c9617fa4-2f34-44b5-8cde-a8021cdd7826/issues-with-schemas-in-source-control?forum=biztalkgeneral

And
http://stackoverflow.com/questions/34667660/how-are-the-xsd-cs-files-used/34668185

It is the same principal for all <biztalkartifact>.cs.

niik:
@robinhultman That first link is exactly what I was looking for, thanks. I'll let this sit over night too see if anyone has objections and if not I'll merge it tomorrow.

robinhultman:
@niik Great! Thanks!

PR #2307, opened 2017-03-02
[VisualStudio] ASP.NET Core - bower - default config

marcrocny:
In VS15 (2017) the default .bowerrc repoints the default location to wwwroot/lib, and bower package restore is ON.

VS default flow is bower package restore to wwwroot/lib.

None that I could find, TBH.

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -205,6 +205,9 @@ orleans.codegen.cs
 # (https://github.com/github/gitignore/pull/1529#issuecomment-104372622)
 #bower_components/
 
+# VS15: by default bower location is moved, bower restore is true
+**/wwwroot/lib/
+
 # RIA/Silverlight projects
 Generated_Code/
 
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -204,8 +204,7 @@ orleans.codegen.cs
 # Since there are multiple workflows, uncomment next line to ignore bower_components
 # (https://github.com/github/gitignore/pull/1529#issuecomment-104372622)
 #bower_components/
-
-# VS15: by default bower location is moved, bower restore is true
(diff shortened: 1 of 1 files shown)


VisualStudio.gitignore
```diff
@@ -205,6 +205,9 @@ orleans.codegen.cs
 # (https://github.com/github/gitignore/pull/1529#issuecomment-104372622)
 #bower_components/
 
+# VS15: by default bower location is moved, bower restore is true
```
shiftkey: VS15 is the codename - can we name this to what it'll be publicly launched as?

VisualStudio.gitignore
```diff
@@ -205,6 +205,9 @@ orleans.codegen.cs
...
+**/wwwroot/lib/
```
shiftkey: On line 29 we have this rule:

```
# Uncomment if you have tasks that create the project's static files in wwwroot
#wwwroot/
```

This was added in https://github.com/github/gitignore/pull/1590 and feels familiar to this. Could you:

 - clarify what's in `wwwroot/lib` on your machine so we can document here what we expect (unless there are some external resources to point to) 
 - see if we can update that rule to reflect whatever changes are coming in VS2017

marcrocny:
Changed reference to "ASP.NET Core" from "VS15", since it applies to prior iterations of IDE as well.

VisualStudio.gitignore
marcrocny: The default project setup in ASP.NET Core is that the bower directory is set to wwwroot/lib.  These are external resources that are restored regularly, so there should be no need to add these to VCS.  

The default intended use for other directories in wwwroot is to hold user code and static resources, which should be in VCS.

shiftkey:
@marcrocny thanks!

PR #2308, opened 2017-03-03
.gitignore for OpenFrameworks.cc

C-F-4:
**Enhancement to the current gitignore repository**

Adds .gitignore for [OpenFrameworks](http://openframeworks.cc/)

Github Repository Link: [Github-OpenFrameworks](https://github.com/openframeworks/openFrameworks)

Here's a link to the official HomePage: **http://openframeworks.cc/**

### DIFF
--- a/OpenFrameworks.gitignore
+++ b/OpenFrameworks.gitignore
@@ -1,26 +1,22 @@
 # openFrameworks
-
-[Bb]in/*  
-![Bb]in/data/  
-[Bb]uild/  
-[Oo]bj/  
-*.o  
-[Dd]ebug*/  
-[Rr]elease*/  
-*.mode*  
-*.app/  
-*.pyc  
-*.log  
+[Bb]in/*
+![Bb]in/data/
+[Bb]uild/
+[Oo]bj/
+*.o
+[Dd]ebug*/
+[Rr]elease*/
+*.mode*
+*.app/
+*.pyc
+*.log
 
 # Code::Blocks
-
 *.depend  
 *.layout  
 
-# Windows
-
-# Windows image file caches  
-Thumbs.db  
+# (Windows-Specific) Windows image file caches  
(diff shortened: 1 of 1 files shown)


joeblau:
@C-F-4 - This tempalte is live on [gitignore.io](https://www.gitignore.io) - https://www.gitignore.io/api/openframeworks

C-F-4:
@joeblau Is there a difference between the .gitignore files available on www.gitignore.io and https://github.com/github/gitignore?

joeblau:
gitignore.io references github/gitignore though a submodule so there shouldn't be.  The main difference is that gitignore.io doesn't have strict requirements for what is considered a valid template therefore there are about 100 more templates.

This is still the master resource.

OpenFrameworks.gitignore
```diff
@@ -0,0 +1,73 @@
...
+*.o
+[Dd]ebug*/
+[Rr]elease*/
+*.mode*
+*.app/
+*.pyc
+.svn/
+*.log
+
+# IDE files which should
```
shiftkey: We have existing templates for various IDEs that are up-to-date, and I don't want to take in changes which duplicate these rules.

Please isolate just the rules specific to OpenFrameworks, and add a comment if there are IDE templates that you think complement this template.

OpenFrameworks.gitignore
```diff
@@ -0,0 +1,73 @@
...
+*.ilk
+*.aps
+ipch/
+
+# Eclipse
+.metadata
+local.properties
+.externalToolBuilders
+
+# operating system
```
shiftkey: We have existing templates for operating systems that are up-to-date, and I don't want to take in changes which duplicate these rules.

Please isolate just the rules specific to OpenFrameworks, and add a comment if there are OS templates under `Global` that you think complement this template.

PR #2310, opened 2017-03-06
Create VSCode.gitignore

LestaD:
If this is a new template: 

 - **Link to application or project’s homepage**:  https://code.visualstudio.com


shiftkey:

PR #2312, opened 2017-03-06
Remove Composer's file and directory from Symfony.

Yousha:

We have already a Composer gitignore. 

https://github.com/github/gitignore/blob/master/Composer.gitignore


shiftkey:
@Yousha thanks!

PR #2313, opened 2017-03-08
[Swift] adding new rules

portellaa:

* **Test generated** - `xcodebuild` generates `.gcno` and `.gcda` files that contains information for code coverage. This kind of information is relative to that test, not to project and all the team members. Usually we don't want those files under revision control or we will have changes every time we run tests.
* **Finder** - Finder generates a `.DS_Store` file with information important only to that user, not all the team members in the team. They can have different finder configurations.
* **Git** - if not disabled`git mergetool` generates temporary `.orig` file which we may not want under revision control. I think this happens more frequently with FileMerge from Apple, ,usually this files should be cleaned after a successful merge.

**Note**: Those links explains the points not the rule to the changes

https://gcc.gnu.org/onlinedocs/gcc/Gcov-Data-Files.html

### DIFF
--- a/Swift.gitignore
+++ b/Swift.gitignore
@@ -6,6 +6,10 @@
 build/
 DerivedData/
 
+## Test generated
+*.gcno
+*.gcda
+
 ## Various settings
 *.pbxuser
 !default.pbxuser
@@ -31,6 +35,12 @@ xcuserdata/
 ## Playgrounds
 timeline.xctimeline
 playground.xcworkspace
+
+## Finder
+.DS_Store
+
+## Git
+*.orig # Files created by mergetool
 
 # Swift Package Manager
 #


grzegorzleszek:
https://github.com/github/gitignore/pull/1638

Swift.gitignore
```diff
@@ -32,6 +36,12 @@ xcuserdata/
 timeline.xctimeline
 playground.xcworkspace
 
+## Finder
+.DS_Store
```
shiftkey: This is an OS-specific rule, and covered by `Global/macOS.gitignore`. Please exclude it from this PR.

Swift.gitignore
```diff
@@ -32,6 +36,12 @@ xcuserdata/
...
+
+## Git
+*.orig # Files created by mergetool
```
shiftkey: These rules are not specific to Swift, and are covered by `Global/Patch.gitignore`. Please exclude it from this PR.

shiftkey:
Closing this out due to inactivity and rules duplicating existing templates. Please reopen a fresh PR that only includes Swift-specific changes.

PR #2314, opened 2017-03-08
Improve Objc file

portellaa:

* **Test generated** - `xcodebuild` generates `.gcno` and `.gcda` files that contains information for code coverage. This kind of information is relative to that test, not to project and all the team members. Usually we don't want those files under revision control or we will have changes every time we run tests.
* **Finder** - Finder generates a `.DS_Store` file with information important only to that user, not all the team members in the team. They can have different finder configurations.
* **Git** - if not disabled`git mergetool` generates temporary `.orig` file which we may not want under revision control. I think this happens more frequently with FileMerge from Apple, ,usually this files should be cleaned after a successful merge.

**Note**: Those links explains the points not the rule to the changes

https://gcc.gnu.org/onlinedocs/gcc/Gcov-Data-Files.html

### DIFF
--- a/Objective-C.gitignore
+++ b/Objective-C.gitignore
@@ -6,6 +6,10 @@
 build/
 DerivedData/
 
+## Test generated
+*.gcno
+*.gcda
+
 ## Various settings
 *.pbxuser
 !default.pbxuser
@@ -27,6 +31,12 @@ xcuserdata/
 *.ipa
 *.dSYM.zip
 *.dSYM
+
+## Finder
+.DS_Store
+
+## Git
+*.orig # Files created by mergetool
 
 # CocoaPods
 #
--- a/Swift.gitignore
+++ b/Swift.gitignore
@@ -6,6 +6,10 @@
 build/
 DerivedData/
 
+## Test generated
+*.gcno
+*.gcda
+
 ## Various settings
 *.pbxuser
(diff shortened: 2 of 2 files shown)


Objective-C.gitignore
```diff
@@ -28,6 +32,12 @@ xcuserdata/
 *.dSYM.zip
 *.dSYM
 
+## Finder
+.DS_Store
```
shiftkey: This rule is already managed by `Global/macOS.gitignore`. Please drop it from this PR.

Objective-C.gitignore
```diff
@@ -28,6 +32,12 @@ xcuserdata/
...
+
+## Git
+*.orig # Files created by mergetool
```
shiftkey: This is not related to ObjC or Swift development. Please drop it from this PR.

Objective-C.gitignore
```diff
@@ -6,6 +6,10 @@
 build/
 DerivedData/
 
+## Test generated
```
shiftkey: Can we make this comment more descriptive? What's generating these test files?

shiftkey:
This PR overlaps with #2313, and has similar issues. Please open a fresh PR that focuses only on Objective-C-specific changes.

PR #2315, opened 2017-03-09
Fixes issue excluding sub directories

ehannes:
This .gitignore is not correct. At least was not working for me.
Ignoring /media/* and then excluding !/media/customer/.htaccess is not
working. According to the Git manual, "It is not possible to re-include
a file if a parent directory of that file is excluded."
(https://git-scm.com/docs/gitignore).

Found a solution here:
http://stackoverflow.com/questions/987142/make-gitignore-ignore-everything-except-a-few-files/16318111#16318111

* https://git-scm.com/docs/gitignore
* http://stackoverflow.com/questions/987142/make-gitignore-ignore-everything-except-a-few-files/16318111#16318111


shiftkey:
@ehannes thanks!

PR #2316, opened 2017-03-09
Allow PakBlacklist-*.txt files inside Build dir in UnrealEngine.gitignore

gagallo7:

The PakBlacklist-<BuildConfiguration>.txt is used to disallow some files to be packaged in the pak file.
This can be very helpful to reduce pak file size for mobile platforms.

To avoid the user-defined blueprint class Debug.uasset located in Content/Blueprints folder to be packaged in shipping build, one has to do the following steps:
1. Create the PakBlacklist-Shipping.txt file in Build/Android directory.
2. Insert the content "../../../MyGameName/Blueprints/Debug.uasset
3. Package the game

https://docs.unrealengine.com/latest/INT/Platforms/Android/ReducingAPKSize/#packageblacklist

### DIFF
--- a/UnrealEngine.gitignore
+++ b/UnrealEngine.gitignore
@@ -54,6 +54,11 @@ Binaries/*
 # Builds
 Build/*
 
+# Whitelist PakBlacklist-<BuildConfiguration>.txt files
+!Build/*/
+Build/*/**
+!Build/*/PakBlacklist*.txt
+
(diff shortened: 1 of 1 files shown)


shiftkey:
@gagallo7 this seems reasonable, even if the rules look a bit intimidating. Thanks for contributing!

PR #2319, opened 2017-03-11
Drastically simplify .gitignore for Dart

kevmoo:
Our typical workflow has JS and related build outputs in the `build` directory.

Most users are on new SDKs, so we don't have to worry about these ignores.
Most users don't generate docs locally. If they do, it's easy to add the entry to their .gitignore later.

Most of the entries are legacy – relating to releases that are over 6 months old.

Just starting the cleanup. See this issue for our doc site
https://github.com/dart-lang/site-www/issues/250

### DIFF
--- a/Dart.gitignore
+++ b/Dart.gitignore
@@ -4,9 +4,9 @@
 .packages
 .pub/
 build/
-# If you're building an application, you may want to check-in your pubspec.lock.
+# If you're building an application, you may want to check-in your pubspec.lock
 pubspec.lock
 
-# Directory created by dartdoc.
+# Directory created by dartdoc
 # If you don't generate documentation locally you can remove this line.
 doc/api/
--- a/Dart.gitignore
+++ b/Dart.gitignore
@@ -6,3 +6,7 @@
 build/
 # If you're building an application, you may want to check-in your pubspec.lock.
 pubspec.lock
+
+# Directory created by dartdoc.
+# If you don't generate documentation locally you can remove this line.
+doc/api/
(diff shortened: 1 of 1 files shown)


kevmoo:
@ArgentiApparatus @kwalrath FYI

shiftkey:
This all seems fair. Happy to let this sit for a day or two if others want to :thumbsup: it before I merge.

ArgentiApparatus:
As a previous contributor to .gitignore for Dart with the intent of keeping it up to date I like the idea of simplifying it

shiftkey:
@kevmoo @ArgentiApparatus @kwalrath thanks for the context - let's get this in!

ArgentiApparatus:
Additional thought - there seems to be  non trivial number of people posting dart questions on stack overflow that use dart2js directly. Would a separate dart2js .gitignore be a good idea?

kevmoo:
@ArgentiApparatus not sure how we'd do that cleanly. For folks starting new apps, I think pushing folks towards pub/etc is the right path.

If folks are getting clever w/ dart2js to manage their output, I wonder if these are the folks with a more sophisticated build setup – and likely know how to handle their outputs.

ArgentiApparatus:
@kevmoo - most certainly agree guiding towards pub is the right thing to do.

To be clear, I'm talking about a _separate_ dart2js.githignore, that folks wanting to create a sophisticated build setup could use ... but then again, there is documentation on dartlang.com for that.

kevmoo:
> To be clear, I'm talking about a separate dart2js.githignore, that folks wanting to create a sophisticated build setup could use ... but then again, there is documentation on dartlang.com for that.

I figured. For folks that are getting fancy, I hope they can figure out their own hacks for their `.gitignore`. 😄

PR #2320, opened 2017-03-12
Add Cursive's REPL history file

Cantido:

The file `replstate.xml` contains the history of the Clojure REPL that [Cursive] adds to IntelliJ. Obviously that's user-specific, and not relevant to other users.

[Cursive]: https://cursive-ide.com/

This file is not well-documented, but in cursive-ide/cursive#1325, the Cursive developers state that this is the REPL history file, and that deleting it is acceptable troubleshooting if it's causing trouble.

### DIFF
--- a/Global/JetBrains.gitignore
+++ b/Global/JetBrains.gitignore
@@ -36,6 +36,9 @@
 # JIRA plugin
 atlassian-ide-plugin.xml
 
+# Cursive Clojure plugin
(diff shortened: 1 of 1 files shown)


shiftkey:
@Cantido thanks!

cursive-ide:
Yes, I agree this is a good change - thanks!

PR #2321, opened 2017-03-13
Add ignore for Secondary data files

matt40k:
Secondary data files .ndf (https://msdn.microsoft.com/en-us/library/ms189563.aspx)


shiftkey:
@matt40k thanks!

PR #2322, opened 2017-03-13
Add Vagrant Logs to Exclude List

utkonos:


shiftkey:
@utkonos any links to StackOverflow discussions or other documentation about this change?

utkonos:
@shiftkey Definitely! Here is the official vagrant documentation for debugging and troubleshooting that mentions this log file:
https://www.vagrantup.com/docs/other/debugging.html

shiftkey:
@utkonos this seems to be a user-specified file rather than something generated by vagrant itself, so I'm not sure how useful this will be. Many other templates will just ignore `*.log` files as well, which might be more useful here (along with a comment).

utkonos:
I will submit a new pull request with the suggested adjustments.

PR #2324, opened 2017-03-15
Update Node.gitignore

Rj-ku:
Yarn lock is autogenerated, I see no reason to keep track of it.


drothmaler:
So maybe you should [read a bit more about it... ](https://yarnpkg.com/en/docs/yarn-lock#toc-check-into-source-control)

Rj-ku:
Fair point.

PR #2325, opened 2017-03-16
Add *.pyc to Python.gitignore

avnes:
- *.pyc is the binary output from py_compile in Python 3

- Ref: https://docs.python.org/3/library/py_compile.html

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -2,6 +2,7 @@
 __pycache__/
 *.py[cod]
 *$py.class
+*.pyc
 
 # C extensions
 *.so


shiftkey:
Shouldn't this be handled by the `__pycache__/` rule we added a couple of lines earlier?

avnes:
I am not 100% why it happens, but often I notice *.pyc files get created in the same directory as the *.py file. It could be due to PEP8 checks as indicated at https://www.python.org/dev/peps/pep-3147/#abstract

jamesroutley:
I think `.pyc`s are already handled by the `*.py[cod]` rule on line 3.

avnes:
Omg, I missed that line. You are correct.

PR #2330, opened 2017-03-21
Added .spyproject to ignored paths

random82:

Looks like Spyder 3.1 uses .spyproject folder to store workspace config

[Spyder Config](https://github.com/spyder-ide/spyder/blob/46c9dbd9713d6095e634f4ade430e2dd235acfec/spyder/widgets/projects/config.py)


shiftkey:
@random82 thanks!

PR #2334, opened 2017-03-26
ignore .DS_Store file in MacOS platform.

light-bo:

.DS_Store is a Mac-specific hidden file. It's also not something that you generally want to add to source control, as its contents may change without you necessarily interacting with it.

[Wikipedia introduce info](https://en.wikipedia.org/wiki/.DS_Store)
[stackoverflow data](https://stackoverflow.com/questions/30246832/is-ds-store-file-important)


shiftkey:

PR #2337, opened 2017-03-28
Fix macOS Icon

spencermathews:

About those those pesky `Icon?` files...

The renaming of `OSX.gitignore` to `macOS.gitignore` (#2085) appears to have inadvertently wiped out both the change to UNIX line endings (#482) and, more importantly, the addition of proper characters to exclude `Icon?` files  (#804).

This pull request returns to appropriate UNIX line endings and adds the correct hidden characters necessary to uniquely catch the `Icon\r\r` system files. 

It's a déjà vu moment since prior to the fixing of line endings the Icon rule was properly added by #334 but accidentally reverted without fanfare by the next commit in #419. Since this ignore rule may be vulnerable to editors that automatically standardize line endings, it seems imperative that updates to this file be done with care or else this rollercoaster seems likely to continue.

closes #2297 

* [How to ignore Icon? in git](http://stackoverflow.com/q/17556250/2762964)


shiftkey:
@spencermathews thanks for picking this up - I really wish I'd noticed it at the time.

PR #2341, opened 2017-03-31
Update JetBrains.gitignore

itigoppo:

add .idea directory

### DIFF
--- a/Global/JetBrains.gitignore
+++ b/Global/JetBrains.gitignore
@@ -29,6 +29,7 @@
 
 # IntelliJ
 /out/
+.idea/
 
 # mpeltonen/sbt-idea plugin
 .idea_modules/


Global/JetBrains.gitignore
```diff
@@ -29,6 +29,7 @@
 
 # IntelliJ
 /out/
+.idea/
```
shiftkey: The link mentioned outlines the things that should be shared, and they should be already defined.

I'm 👎 on ignoring everything under `.idea`.

itigoppo: I did it.
```
$  git config --global core.excludesfile ~/gitignore/Global/JetBrains.gitignore
```
but 

```
$ git status
On branch master
Your branch is up-to-date with 'upstream/master'.
Untracked files:

	.idea/

```

So I added .idea .

shiftkey: That's because there's some files _under_ `.idea` that we should be tracking. If you do: 

```
git add .
git status
```

You should see a list of the files it wants to commit. Contrast that with what the JetBrains documentation recommends and if there's any that stand out, let's update the rules for those specific files...

shiftkey:
Closing out due to lack of feedback.

PR #2343, opened 2017-03-31
Ignore mkdocs site directory

ghisvail:
[mkdocs](http://www.mkdocs.org/) is rising as an alternative to Sphinx for project's documentation. The default build command puts the generated documentation in a `site` directory at the root of the project which should be ignored.

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -92,3 +92,6 @@ ENV/
 
 # Rope project settings
 .ropeproject
+
+# mkdocs documentation
+site/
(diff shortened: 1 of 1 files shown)


Python.gitignore
```diff
@@ -92,3 +92,6 @@ ENV/
 
 # Rope project settings
 .ropeproject
+
+# mkdocs documentation
+site/
```
shiftkey: If this is always at the root of the repository it can be written as `/site` - this will help prevent clashes with other folders named the same...

ghisvail: It should be the case.

shiftkey:
@ghisvail thanks!

PR #2346, opened 2017-04-06
Adding iOS git ignore file 

DeepakSinghSA:

No git ignore for iOS


shiftkey:
Thanks, but please use or enhance [`Global\XCode.gitignore`](https://github.com/github/gitignore/blob/master/Global/XCode.gitignore).

PR #2347, opened 2017-04-06
Update macOS.gitignore

DKroot:
Clarified purpose of ._* files. See https://lists.apple.com/archives/applescript-users/2006/Jun/msg00180.html

Clarified purpose of ._* files

https://lists.apple.com/archives/applescript-users/2006/Jun/msg00180.html

### DIFF
--- a/Global/macOS.gitignore
+++ b/Global/macOS.gitignore
@@ -3,9 +3,9 @@
 .LSOverride
 
 # Icon must end with two \r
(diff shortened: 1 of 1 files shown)


shiftkey:
@DKroot I can't take this in due to the line ending changes - I managed to accidentally break this a while ago and someone else noticed it. Apologies!

PR #2349, opened 2017-04-09
Upgrade gradle gitignore

rogersachan:

This PR brings proper support for gradle projects (especially in subfolders) and adds IDE files and package files to the gitignore since they can be generated via gradle. This also properly ignores system-generated files for example `.DS_Store` and the like.

* https://docs.gradle.org/current/userguide/idea_plugin.html
* https://docs.gradle.org/current/userguide/eclipse_plugin.html
* https://docs.gradle.org/current/userguide/ear_plugin.html

### DIFF
--- a/Gradle.gitignore
+++ b/Gradle.gitignore
@@ -7,9 +7,10 @@ libs/
 # Mobile Tools for Java (J2ME)
 .mtj.tmp/
 
-# Package Files #
+# Package Files
 *.war
 *.ear
+*.jar
 
 # virtual machine crash logs, see http://www.java.com/en/download/help/error_hotspot.xml
 hs_err_pid*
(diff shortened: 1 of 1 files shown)


Gradle.gitignore
```diff
@@ -1,14 +1,76 @@
...
+*.bat
+*.setting
 
-# Ignore Gradle GUI config
-gradle-app.setting
+libs/
 
-# Avoid ignoring Gradle wrapper jar file (.jar files are usually ignored)
-!gradle-wrapper.jar
+# Mobile Tools for Java (J2ME)
```
drothmaler: And most of those are already part of `Java.gitignore` - that is typically used together with `Gradle.gitignore`.

rogersachan:
@shiftkey @drothmaler I wouldn't have any issues with this if `.gitignore` files were stackable on Github.com, but this PR was mostly for convenience.

PR #2351, opened 2017-04-11
add editor temp files

dheera:

Add editor temp files


shiftkey:
Please only change one template at a time, and provide more information about why these rules need to be added.

PR #2353, opened 2017-04-14
Create DBeaver.gitignore

peter-janssen:

DBeaver can be installed on any eclipse based "ide" 

[documentation](https://github.com/serge-rider/dbeaver/wiki/Admin-Manage-Connections)

[homepage](http://dbeaver.jkiss.org/)
[GitHub](https://github.com/serge-rider/dbeaver)

### DIFF
--- a/DBeaver.gitignore
+++ b/DBeaver.gitignore
@@ -1,0 +1,2 @@
+# ide config file
+.dbeaver-data-sources*.xml


shiftkey:
@peter-janssen do you think this would be better under `Global/Eclipse.gitignore`?

peter-janssen:
@shiftkey I was debating that myself. But I think the intention is that I can pick and choice which components I have in my project. DBeaver is not installed in Eclipse by default. Furthermore a lot of people will run DBeaver separate of there main ide (Eclipse, Intellij or otherwise). I chose to make a new template to honor the modularity of the project.
On the other hand I see that `Global/Eclipse.gitignore` has more plugins included. These are a bit more main stream though but that in itself is also an argument to give them separate templates.
I'm 51:49 on this...

shiftkey:
For discoverability, I'd suggest having it in `Global/Eclipse.gitignore` as an optional rule - documented by commented-out by default. But I'm not across how this is being used, so I'm mostly guessing here.

Separate templates have a couple of limitations with how they're currently organized:

 - for templates which are relatively small, I'd probably decline to take it in and favour some better documentation as that's more likely to be noticed by users
 - this repository is separate to making things available on GitHub the Website when creating a new repository - that's a whole other thread that I need to clarify in general

peter-janssen:
Seems like this is fixed, closing.

PR #2356, opened 2017-04-19
Edit ignore for Microsoft Visio autosave files to include Office <= 2003

orzechow:

The rule for Microsoft Visio autosave files is specialized for Office 2003 and higher, where the default Visio file extention is `.vsdx`
But older versions of Visio used the file extention `.vsd` thus generating `*.~vsd` autosave files.
The new rule matches both autosave file types.

https://fileinfo.com/extension/vsd


shiftkey:
@orzechow thanks!

PR #2359, opened 2017-04-23
Add moc_*.h in Qt.gitignore

Glordim:

Add moc_\*.h in Qt.gitignore.
(example: moc_predefs.h)
(moc_\*.cpp is already present)

Note: moc_* are auto-generated files (see Qt moc)

### DIFF
--- a/Qt.gitignore
+++ b/Qt.gitignore
@@ -20,6 +20,7 @@
 *.qbs.user.*
 *.moc
 moc_*.cpp
+moc_*.h
 qrc_*.cpp
 ui_*.h
 Makefile*


shiftkey:
Any links to StackOverflow discussions or other documentation about this change?

Glordim:
Very hard to find informations about that...
Maybe here:

https://gyazo.com/76655be87dd256060e30afe39dd6ef14

Source page : https://bugreports.qt.io/browse/QTBUG-57796

That confirm than moc_predef.h is a new feature of qmake. (since Qt 5.8)

shiftkey:
@Glordim thanks for the context. Lots of mentions of "generate" so I'm happy to take this in. For example:

> Then closing this as "Not our bug". CMake needs to generate moc_predefs.h like qmake does.

PR #2367, opened 2017-04-30
Add macOS related entries to Xcode.gitignore

vahan3x:
* Add macOS related entries to Swift.gitignore
* Add macOS related entries to Objective-C.gitignore
* Update reminder comments in all 4 gitignore files

Swift/Obj-C developers (mostly) using macOS as their development platform, Xcode is currently available only for macOS. So it is logical to have macOS gitignore file in same group as Swift, Obj-C and Xcode.

[Xcode](https://en.wikipedia.org/wiki/Xcode)
[Swift](https://en.wikipedia.org/wiki/Swift_(programming_language))
[Objective-C](https://en.wikipedia.org/wiki/Objective-C)

### DIFF
--- a/Global/Xcode.gitignore
+++ b/Global/Xcode.gitignore
@@ -1,7 +1,36 @@
+# macOS
+#
+# gitignore contributors: remember to update Global/macOS.gitignore, Global/Xcode.gitignore, Objective-C.gitignore & Swift.gitignore
+
+*.DS_Store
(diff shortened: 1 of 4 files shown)


shiftkey:
@vahan3x thanks for trying to simplify this, but these are already intertwined enough that I want to avoid any more intertwining to keep my sanity in check!

PR #2370, opened 2017-05-04
Create PowerShell.gitignore

pcgeek86:

Adding a new PowerShell template for .gitignore, to exclude .NET assemblies (DLLs) and ZIP artifacts.

https://msdn.microsoft.com/en-us/library/hk5f40ct(v=vs.90).aspx

If this is a new template: 

https://msdn.microsoft.com/en-us/powershell/mt173057.aspx

### DIFF
--- a/PowerShell.gitignore
+++ b/PowerShell.gitignore
@@ -1,0 +1,5 @@
+# Exclude packaged modules
+*.zip
+
+# Exclude .NET assemblies from source
+*.dll


shiftkey:
I'm going to close this out because the value of maintaining a template like this in this repository isn't clear to me.

@pcgeek86 if you feel like this should still be a template listed here, please read the section about [specialized templates](https://github.com/github/gitignore#specialized-templates) and open a new pull request that follows the new pattern.

PR #2371, opened 2017-05-06
Ignore mypy's cache directory

brettcannon:

 The latest release of mypy caches results in a directory.

https://mypy.readthedocs.io/en/latest/command_line.html (search for `.mypy_cache`)


shiftkey:
@brettcannon thanks!

PR #2374, opened 2017-05-09
add .appx to VisualStudio.gitignore

julienshepherd:

Visual Studio creates .appx files in the UWP project root directory. These files are not needed when writing code.

These are packages created for the Windows Store, you don't need them when you write code. I asked a question on [StackOverflow](http://stackoverflow.com/questions/43862861/is-it-safe-to-ignore-appx-files-in-git-repository-of-a-uwp-app/43865908#43865908) on this topic.

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -183,6 +183,7 @@ AppPackages/
 BundleArtifacts/
 Package.StoreAssociation.xml
 _pkginfo.txt
+.appx
 
 # Visual Studio cache files
 # files ending in .cache can be ignored
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -183,7 +183,7 @@ AppPackages/
 BundleArtifacts/
 Package.StoreAssociation.xml
 _pkginfo.txt
-.appx
+*.appx
 
 # Visual Studio cache files
 # files ending in .cache can be ignored


VisualStudio.gitignore
```diff
@@ -183,6 +183,7 @@ AppPackages/
 BundleArtifacts/
 Package.StoreAssociation.xml
 _pkginfo.txt
+.appx
```
shiftkey: This should already be covered by the earlier `AppPackages/` rule, as per the StackOverflow answer.

If not, could you elaborate on why that's not the case?

julienshepherd: The reason is because there are .appx files created outside the AppPackages/ folder, in the project root folder. And this rule is for this scenario. 

I attached a screenshot from GitHub Desktop.

shiftkey: @julienshepherd cool, so what's different about your setup that means it's not creating it under `AppPackages/`? If the earlier rule isn't something we can rely on then perhaps the other rule needs to go away...

julienshepherd: Indeed. When I create packages for the Windows Store, I choose a different folder than the `AppPackages/` folder (step 6 [here](https://docs.microsoft.com/en-us/windows/uwp/packaging/packaging-uwp-apps#create-an-app-package)
). Anyway, some .appx files are still in the project root folder, but I don't know if this is caused by me choosing a different folder for the Store packages.

shiftkey: Cool. Looks like we need `*.appx` then.

Given we can't rely on `AppPackages/` here we need to figure out what to do about that rule - are there any other files that are output as part of this process that need to be excluded?

julienshepherd: Besides `AppPackages/` folder which contains .appxsym, .appxbundle, .cer, .ps1 and many other file types, and the project root folder .appx files, I didn't see any other output files to be excluded in my Git repo. I would keep the `AppPackages/` folder just in case someone creates the Store packages there.

shiftkey:
@julienshepherd thanks!

PR #2376, opened 2017-05-10
Create PuTTY.gitignore

vuolter:

PuTTYgen is a widely used alternative to OpenSSH under Windows.
This rule allows private keys ignoring.

If this is a new template: 

 - **Link to application or project’s homepage**: https://www.chiark.greenend.org.uk/~sgtatham/putty/


vuolter:
As they say "better late than never" 😝

PR #2378, opened 2017-05-11
Update ExtJS.gitignore

devgaspa:

Ignore folders _.sencha_ and _sass/example_


shiftkey:
@devgaspa any links to StackOverflow discussions or other documentation about this change?

devgaspa:
In fact these files should remain on git. thanks!

PR #2384, opened 2017-05-15
fix Prestashop.gitignore error

King66:
the last commit has no effect:
https://github.com/github/gitignore/commit/af4b2ae6f488f06e08944831e890a4cf291492e0

I think his purpose is to track these php files and ignore other files under /cache/
So I change this file to achieve the goal.


King66:
@shiftkey I have submitted a PR to fix the error but it's ignored!

shiftkey:
@King66 thanks!

PR #2385, opened 2017-05-17
Ignoring OS meta files to Unity projects

douglasgimli:
Gitignore not exclusing OS meta files to Unity Projects


shiftkey:
Thanks, but see [`Global\macOS.gitignore`](https://github.com/github/gitignore/blob/master/Global/macOS.gitignore).

I recommend tailoring your gitignore file at the repository level by combining the framework, tool and OS templates you need.

douglasgimli:
Good call!

PR #2388, opened 2017-05-20
add *_plugin_import.cpp to Qt.gitignore

KoynovStas:
This file is autogenerated by qmake. It imports static plugin classes for
static plugins specified using QTPLUGIN and QT_PLUGIN_CLASS.<plugin> variables.

This file is autogenerated for static build.

### DIFF
--- a/Qt.gitignore
+++ b/Qt.gitignore
@@ -12,6 +12,7 @@
 
 # Qt-es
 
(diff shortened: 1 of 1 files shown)


shiftkey:
@KoynovStas thanks!

PR #2392, opened 2017-05-22
Add .cpp

Dobby233Liu:
**It's Just a Update**

### DIFF
--- a/C++.gitignore
+++ b/C++.gitignore
@@ -7,6 +7,9 @@
 *.o
 *.obj
 
+# Visual Studio
+*.cpp
+
 # Precompiled Headers
 *.gch
 *.pch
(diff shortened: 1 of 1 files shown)


shiftkey:
@Dobby233Liu `*.cpp` seem like they are important for tracking source code. Without more context on _why_ this is a good idea, I can't accept this PR.

C++.gitignore
```diff
@@ -7,6 +7,9 @@
 *.o
 *.obj
 
+# Visual Studio
+*.cpp
```
Yousha: Ignoring C++ files in C++ gitignore? srsly?

Dobby233Liu:
OK, I close that.

PR #2393, opened 2017-05-24
add 7z

fanvil:

_TODO_

_TODO_

If this is a new template: 

### DIFF
--- a/Java.gitignore
+++ b/Java.gitignore
@@ -17,6 +17,6 @@
 *.zip
 *.tar.gz
 *.rar
-
+*.7z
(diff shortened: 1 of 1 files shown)


joeblau:
@fanvil this is already included in https://github.com/dvcs/gitignore

https://www.gitignore.io/api/compressed

drothmaler:
Hmm... IMHO this isn't really Java specific. 
I don't know any place where Java or any Java IDE (that I know) would produce 7z files by default. 
It seems more project specific, to me.

shiftkey:
Yeah, this isn't specific to Java, and is covered by [`Global/Archives.gitignore`](https://github.com/github/gitignore/blob/master/Global/Archives.gitignore).

PR #2394, opened 2017-05-24
Ignoring <module>/release/output.json

leesah:
Android Studio 3 Canary 1 generates a `<module>/release/output.json` file.

Android Studio 3 Canary 1 generates a `<module>/release/output.json` file when selecting from menu: `Build -> Generate Signed APK...`, which looks like:

    [{
      "outputType": {
        "type": "APK"
      },
      "apkInfo": {
        "type": "MAIN",
        "splits": [],
        "versionCode": 32
      },
      "outputFile": {
        "path": "/path/to/the/generated/release/filename.apk"
      },
      "properties": {
        "packageId": "com.example.android",
        "split": ""
      }
    }]

I failed to find any documentation on it.

If this is a new template: 

### DIFF
--- a/Android.gitignore
+++ b/Android.gitignore
@@ -12,6 +12,7 @@
 bin/
 gen/
 out/
+*/release/output.json
 
 # Gradle files
 .gradle/


shiftkey:
@leesah any links to StackOverflow discussions or threads about this change?

leesah:
I didn't specifically search for that in StackOverflow. Maybe I should ask about it if not found, at least to prove this is a change in Android Studio instead of some local configuration in my laptop. Is that what you mean?

shiftkey:
Closing this out in favour of #2684 which seems similar but also doesn't really have any additional context.

PR #2397, opened 2017-05-26
add MANIFEST to Python.gitignore

weinihou:
add MANIFEST which is GENERATED by distutils

If you use distutils.core to setup, the file MANIFEST is generated by distutils when run 
'python setup.py sdist'

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -23,6 +23,7 @@ wheels/
 *.egg-info/
 .installed.cfg
 *.egg
+MANIFEST
 
 # PyInstaller
 #  Usually these files are written by a python script from a template


shiftkey:
@weinihou any links to StackOverflow discussions or other documentation about this change?

dougalsutherland:
Here's a short SO discussion about it: https://stackoverflow.com/questions/29836718/python-package-contains-both-manifest-and-manifest-in-do-i-gitignore-manifest

It also includes a comment at the top
```
# file GENERATED by distutils, do NOT edit
```
which sure sounds like something that should be gitignored.

shiftkey:
@dougalsutherland thanks!

PR #2400, opened 2017-05-31
exclude Eclipse and IntelliJ config files

JimmyLincole:


shiftkey:
Thanks, but these are tracked in `Global/JetBrains.gitignore` and `Global/Eclipse.gitignore`.

PR #2401, opened 2017-05-31
Fix capitalization and spacing of comments

utkonos:
This is a cleanup of the comments so they follow the style of comments in some of the more heavily trafficked gitignote files.


shiftkey:
@utkonos thanks!

PR #2402, opened 2017-06-01
[Android] Cleanup and remove legacy stuff

friederbluemle:

More than four years ago, Google [officially ended support for Eclipse](https://android-developers.googleblog.com/2016/11/support-ended-for-eclipse-android.html) for Android development. Even at that time, most developers had already moved on to the official IDE, Android Studio. This PR removes some of the legacy leftover from the Ant/Eclipse/Maven days:

- Folders like `bin/`, `gen/`, `out/` _do not exist_ with Gradle/Android Studio.
- All build output files `.apk`, `.dex`, `.class` are contained in the `build/` folders, which *are already ignored*. No need to special case and redundantly ignore these specific file types.
- Also consolidated some of the folders that are automatically generated by Android Studio under `# Android Studio generated folders`
- Ignoring `google-services.json` by default for security reasons
- Ignoring `.idea/` since the Android Gradle plugin derives **all** information necessary for the build system from `build.gradle` files. In rare cases where someone wants to add _specific_ files, it should be a conscious decision by the committer, rather than an accidental one.

https://android-developers.googleblog.com/2016/11/support-ended-for-eclipse-android.html

https://stackoverflow.com/questions/16736856/what-should-be-in-my-gitignore-for-an-android-studio-project/38122646

https://stackoverflow.com/questions/37358340/should-i-add-the-google-services-json-from-firebase-to-my-repository

**EDIT (2017-10-16)**
This also reverts #2496 which should not have been merged. Now keystore files are *ignored* again.

Addresses and closes #2394 
**EDIT (2018-02-20)**
Also reverts #2925 which added a non standard/user specific `lint` folder.

**EDIT (2019-12-06)**
Reverted #3233 - In some cases, `.aar` files need to be committed to source control. _Build output_ is _already_ ignored through `build/`.


friederbluemle:
Branch rebase on latest `master`

@shiftkey Have you had a chance to look at this PR? Any comments? Thanks.

friederbluemle:
Rebased onto latest `master`. @shiftkey I understand you're probably busy, could you assign someone else to look into this PR? Thanks!

friederbluemle:
Branch rebased on latest `master`

friederbluemle:
Rebased onto latest `master`

friederbluemle:
Rebased onto latest `master`, and included an additional commit to revert 77145f787f311efa66c570c5ffadea10a429bf21.

Android.gitignore
```diff
@@ -41,11 +22,7 @@ captures/
 .idea/libraries
 
 # Keystore files
-# Uncomment the following line if you do not want to check your keystore files in.
-#*.jks
-
-# External native build folder generated in Android Studio 2.2 and later
-.externalNativeBuild
+*.jks
```
matpag: When you create a new keystore, Android Studio point you to create a *.jks file, however at the end of the process it creates a file without extension. Actually it's a developer job to rename the file adding the extension which is not needed to let the file works correctly.

That said, developers can add .jks or .keystore extension to the file, and it will work in both cases. When Android first came out Eclipse created .keystore files and if we want cover both the cases i think you should add *.keystore too to this line

matpag:
Why we have only one gitignore for Android projects? The gitignore created by Android Studio are 2, one for the project and one inside the module.

With only this gitignore we are still committing a lot of useless things.

The correct way should be

```
Project
    |-> app (module)
    |       |-> gitignore
    |-> gitignore
```

friederbluemle:
@matpag `build/` in the root `.gitignore` will also match the build folder inside of module folders, i.e. the additional `.gitignore` files are *not* needed. What is the "release" folder you mentioned?

Android.gitignore
friederbluemle: Good point. I'm not sure about the keystore generation on Windows, but the official guidelines to [generate the file on command line](https://developer.android.com/studio/publish/app-signing.html#signing-manually) use `my-release-key.jks` as the sample filename, so `*.jks` would match it.
I'm open to also add `*.keystore` if others think it would be useful. I have not seen such a file in a long time, but do remember that the file used to have this extension.

matpag:
Ok for the build/ pattern, i'm not get used to work with only one .gitignore.

Regarding the release folder, when you generate a signed APK for production release AS 3.0+ will create a folder at the module level called _release_. Inside the folder there are 2 files, the compiled APK and a file called output.json, someone else already created a pull request on this topic https://github.com/github/gitignore/pull/2394

Android.gitignore
matpag: Yeah there is a problem about this, using the CLI you can specify the extension while using the GUI if you dont manually add ".jks" at the end of the name it is not added automatically (even if the suggestion say it should be).

There is the potential to open an issue in the Android tracker about this "strange" behaviour

friederbluemle: Okay sure, I'll add `*.keystore` as well.

friederbluemle:
@matpag That's an interesting observation. Thanks! Maybe we can revisit this again once the final version of Android Studio 3.0 is released? At that point we can discuss and implement the necessary changes. Meanwhile, I added the line to exclude files that match `*.keystore` as well.

friederbluemle:
Okay, I just tested using the first release candidate (RC 1) of Android Studio 3.0 (latest pre-release as of Oct 21st 2017). A few observations:

- It seems like this version does *not* generate the mentioned `output.json` file anymore (using default settings in **Generate Signed APK**).
- The release apk output destination defaults to the module directory (_not_ a `release` folder, or the existing `build` folder). For this reason, I re-added the `*.apk` rule back into `.gitignore`
- As you mentioned, when generating a new keystore using the GUI, the file will _not_ have a file extension by default. However, the default destination directory is *not* the current project or module folder, but the user's home folder. This makes sense, as these keystore is not supposed to be committed to source control anyway.

matpag:
I've tried again with my project and Generate Signed APK is still providing the output.json file in the release folder, AS 3.0 RC2

friederbluemle:
Interesting. It seems to be related to the new version of the **Android Gradle plugin**, not (just) Android Studio 3.0 - Building a project that uses an older version of the plugin will not generate the output.json file, but when using `3.0.0-rc2` it will generate it (I just tested and can confirm your findings).
Anyway - Do you strongly feel the issue with the release output should be addressed now and as part of this PR, or can it be done in a follow-up PR, once the final version of Android Studio 3.0 is released and the details are confirmed?
Thanks again for your comments and testing on this!

matpag:
I've performed some more tests about the output.json and i was right when i typed

> Perhaps this is related to the name of the buildType compiled, i should test if this is the case with multi-flavors builds, because if they are related we need a more generic rule

in a previous comment because it seems to be the right logic.

I've created a different buildType called "anotherBuildType" and then created the apk with _Generate Signed APK_, this time the folder was named "anotherBuildType" and not release.
So the https://github.com/github/gitignore/pull/2394 PR is not correct and we need a more general rule like `*/*/output.json`. I don't know if you want to wait for the addition when AS 3.0 will be officially released but i'm pretty confident this behaviour will not change with the final release. They can't make regression like this in RC state and this new file should be helpfull for CI builds.

Yousha:
@friederbluemle May I ask why did you remove `proguard/` directory?

matpag:
@Yousha `proguard/` directory was something generated by Eclipse. Starting with Android Studio that folder should never exist because it is included inside `appModule/build/outputs/mapping/buildType/` and because Eclipse is not officially supported anymore that line is completely useless.

friederbluemle:
@Yousha Yes, as @matpag mentioned, this file only applied to Eclipse based projects. It was also mentioned in the (now removed) comment: `# Proguard folder generated by Eclipse`. *New* Android project can no longer be based on Eclipse, and the very few remaining Eclipse-based projects likely already have a `.gitignore`.

@matpag I did a similar test regarding the build types, and again can confirm your findings. The output folder will have the name of the build/flavor type. Instead of using `*/*/output.json`, we can simply use `output.json` - this is very similar to an existing rule for the file `google-services.json`, which may also exist in variably named sub-directories.

I added it and amended the commit. How does it look now?

matpag:
@friederbluemle I would have used `*/*/output.json` because someone could have named a test or asset file `output.json` somewhere in the test or assets folder, and this could prevent it to correctly commit it. Rules should be the most specifics possible.

Regarding `google-services.json`, while this line its there? `google-services.json` should be committed in version control because it only contains metadata. What you see in the google-services.json is the same thing is compiled in your APK, an attacker just needs to decompile the APK to access those informations, that's nothing secret there...

Read this discussion: https://stackoverflow.com/a/42750187/2910520

friederbluemle:
Well, `*/*/output.json` seems overly specific. What if the module hierarchy has more than one level? Do you also want to add `*/*/*/output.json`, `*/*/*/*/output.json`, etc.? At the same same time, you could ask "What if someone has a `local.properties` in their test folder". I think it's OK to make an explicit manual exception in those 0.1% of repos for which this would apply.

Regarding `google-services.json`: The line is there for a reason. The API keys in that file should definitely *not* be committed to source control for a regular open source project. There might be exceptions for a controlled, closed environment (e.g. enterprise), in which case the line can be commented out. For normal open source repos, you don't want the whole world to use the API keys and resources of your personal backend system (Google API quota). Note that this has nothing to do with an "attacker" - of course there is little that can be done to keep a malicious person from extracting the secrets from an apk file.

While this is certainly a great discussion, I would appreciate it if the PR could be merged soon, so that future projects can start using the consolidated and simplified `.gitignore`.
If there are any further points worth discussing, I'm happy to extract the sections in question to a separate PR.

@shiftkey Please help with the review if you have a chance. The branch is fully rebased once again.

matpag:
I've a different opinion on the first part of your answer, hierarchical sub modules it's something i've never seen in a project, (even a big app like Telegram doen't have them https://github.com/DrKLO/Telegram/tree/master/TMessagesProj) and having another `local.properties` or any configuration file randomly positioned in the project folder it's practically impossible. Moreover, i think it's more likely to have the default app structure with `output.json` in the `module/release` folder to ignore.

About `google-services.json` you are right, in open source projects it should be omitted while it should be included in private ones.

friederbluemle:
It's 2019 now, and this PR from July 2017 **is still pending**.

In the meantime, a lot more stuff got merged into `master`. Examples are: More cruft that only applies to individual users (e.g. a `lint` folder in #2925), or #2496 which potentially introduces a real security risk by not ignoring keystore files by default.
Some of these changes got merged with little to no discussion or references.

I rebased this PR (again) onto the latest `master` branch after resolving conflicts.

@shiftkey - I understand this PR has more line changes than most other PRs in this repo. **Please** let me know if there is there anything I can do to make the review easier? Do you want me to split it up into multiple smaller PRs? The entire discussion is captures and resolved here. @matpag approved these changes on Nov 10, 2017.

matpag:
I will review the changes again when I have time, to see if they still apply to the current situation

friederbluemle:
Thanks @matpag 👍

ikesyo:
Looks like you removed some entries which were added recently (after you submitted this PR initially). Could you revisit this not to break recent changes?

Android.gitignore
```diff
@@ -1,80 +1,27 @@
...
-.idea/tasks.xml
-.idea/gradle.xml
-.idea/assetWizardSettings.xml
-.idea/dictionaries
-.idea/libraries
-.idea/caches
-# Android Studio 3 in .gitignore file.
-.idea/caches/build_file_checksums.ser
-.idea/modules.xml
+.idea/
```
ikesyo: This is not correct. Rationale: https://github.com/github/gitignore/pull/2527#issuecomment-347049128

Android.gitignore
```diff
@@ -1,80 +1,27 @@
...
-# Version control
-vcs.xml
-
-# lint
-lint/intermediates/
-lint/generated/
-lint/outputs/
-lint/tmp/
-# lint/reports/
+google-services.json
```
ikesyo: This was made unignored intentionally in https://github.com/github/gitignore/commit/584dd34b9d96f2fb847103ea8c355169212b6442 so should be reverted.

Android.gitignore
friederbluemle: None of the files inside of `.idea/` are (or should be) required to clean build a fresh clone. Android Studio does not require any of these either. They are auto-generated once opened in Android Studio, which will take information from `build.gradle` (which is in source control).

Android.gitignore
friederbluemle: This change was reverted intentionally. `google-services.json` contains passwords and API secrets that **should not be shared** with the world. In specific scenarios (for example when closely working with other people inside of a company), teams _may elect_ to comment this line and commit `google-services.json` to source control. It is a security risk to not ignore it by default.

friederbluemle:
@shiftkey - 2.5 years, countless rebases, conflict resolutions, and resolved conversations later - is there still any chance this will ever get merged? 🙏

Android.gitignore
ikesyo: See https://groups.google.com/forum/m/#!msg/firebase-talk/bamCgTDajkw/uVEJXjtiBwAJ

> **The general answer is yes, the google-services.json is safe to check in to your repo and is something that should be shared among engineers on your team**.  The JSON file does not contain any super-sensitive information (like a server API key).  It does contain some information like your database URL, Android API key, and storage bucket.  These are not secrets, but if your security rules are not set up correctly an attacker could use them against you.  However since these values are also compiled into your APK as resources, the theoretical attacker would not need your JSON file to get these values anyway.
> 
> For a library or open-source sample we do not include the JSON file because the intention is that users insert their own to point the code to their own backend.  That's why you won't see JSON files in most of our firebase repos on github.

Android.gitignore
ikesyo: Some configuration files in `.idea` dir are meant to be shared with team members as project-wide IDE configurations. Those configs are not in `build.gradle` so ignoring entire `.idea` is not ideal.

friederbluemle: I don't disagree with you. If someone explicitly _wants_ to add certain files to source control, they always have the option to do that. In almost all other cases, none of the files are needed, and are auto-generated by the IDE.   This is true for most Maven/Gradle based projects (which is also mentioned in the links you posted), and especially true for Android Studio.
That's why 9 out of 10 Android projects don't have these files/folders committed to source control. And for those that do, it was often done without explicit intent of the original committer.

Android.gitignore
friederbluemle: We can continue to go around in circles on that one. 😄 There is no "100% correct" answer on that one.

See the accepted and highest rated answer to [this StackOverflow question](https://stackoverflow.com/questions/37358340/should-i-add-the-google-services-json-from-firebase-to-my-repository) here (it's also the first result on Google):

> Should I add the google-services.json (from Firebase) to my repository?
> It seems to make sense to add it to a .gitignore and not include it in a public repo.

Basically the question is:

Do we want this .gitignore file to be more "optimized" for case 1 or case 2:

1. Closed source repo accessible to engineers on a team (e.g. in a company)
2. Public open source repos shared with the world

For _public_ repos, the file should **not** be included (as stated above).
Worst case for ignoring the file by default: The project will not build and you have to add your own file.

For _private_ repos, the file _can_ be added to source control.
Worst case for NOT ignoring the file by default: Someone that is not aware of a default setting in `.gitignore` accidentally commits (and therefore permanently shares) his credentials with the world.

The second scenario sounds a lot more severe to me, therefore the default state should be to _not_ include the file in the repo.

Either way, explicitly _opting out_ of a default security measure makes more sense to me, what do you think?

friederbluemle:
Rebased again onto the latest `master` branch, with #3233 reverted too.

On a side note, the default Android `.gitignore` is now up to 1.4KB, with *many* redundant, invalid, and outdated entries! :(

@shiftkey I know this has become an absurdly long thread, and I really don't want to sound pushy, but if there's any chance you have a few moments to go through this PR again, so we can come to a conclusion and possibly get this merged in this decade, that would be great! 🙏

friederbluemle:
Rebased again onto the latest `master` branch.

Android.gitignore
```diff
@@ -1,85 +1,29 @@
...
 # Local configuration file (sdk path, etc)
 local.properties
 
-# Proguard folder generated by Eclipse
-proguard/
-
-# Log Files
+# Log/OS Files
 *.log
+.DS_Store
```
shiftkey: OS-specific files like this should already exist in a different template - in this case `Global/macOS.gitignore` - to avoid duplication. Can you drop this rule from the file?

Android.gitignore
shiftkey: I don't have any strong feelings on what sort of things under `.idea` should be used, but in situations like this I have some guidance:

 - "What things should be included in version control?" - if a rule is unnecessarily broad, it might interfere with some files that users can commit to the repository and share with others. As `Global/JetBrains.gitignore` shows, there are lots of files in IntelliJ that can be committed, and I've pushed back on PRs in the past where language-specific templates ignore this entire directory.
 - "What things would a typical user expect to be included in version control?" - for more opinionated tooling (which I feel is what Android Studio is doing that @friederbluemle is trying to explain) it may be that these files are unnecessary noise for most users.

If it were me, I'd lean towards being more explicit about what can be committed, but I'll leave it to others to settle on what should be in this template.

friederbluemle: Yes, that's correct, for other IntelliJ/JetBrains products, some information might be required from the `.idea/` folder. For projects that use the Android Gradle pluin, all information about the build system is contained in `build.gradle`, as explained above. I also added it to the PR description.

Android.gitignore
friederbluemle: I completely agree with you that we should avoid unnecessary duplication. For this specific case, can I ask you to make an exception?

Removing this one line here would unavoidably result in **unintentionally committed** `.DS_Store` files in many repositories. A simple example would be: Someone not knowing about the `macOS.gitignore` file, just looking for something to get started on Android. Or someone else setting up a project on Linux, then people start contributing from macOS. `Global/macOS.gitignore` currently contains 26 lines (of which 25 would potentially be not applicable for Android).
Also, `.DS_Store` has been part of the default `.gitignore` generated by Android Studio for a long time, since many mobile developers use macOS.

If you really do think this line has to be removed for this PR to be merged, just leave a quick comment here, and I'll remove it immediately.

friederbluemle:
Thanks @shiftkey for getting back to us! 🙇 👍 

I rebased the PR again and replied to your comment above! Please have another look!

friederbluemle:
Rebased again. 2021 might be the year! :crossed_fingers:

Android.gitignore
Hipska: Still existing?

friederbluemle: That's right, `.DS_Store` continues to be part of the default `.gitignore` as generated by Android Studio.

Hipska: Yeah, but here we are creating templates GH will generate from, so this still goes into global os specific gitignore or won't get merged..

friederbluemle: What are your thoughts on my comment above?

Android.gitignore
```diff
@@ -1,89 +1,31 @@
...
```
bdougie, line 10:
```suggestion
```
I am going to remove this change since it is still unresolved. My recommendation is still to use the Global Macos file for this.

bdougie:
Removed the .DS_store in order to get this merged. Feel free to open up another PR or fork the file as a template.

friederbluemle:
Thank you @bdougie 👍

PR #2403, opened 2017-06-01
Update python gitignore - ignore common archived and compressed file formats

sr-murthy:

_TODO_

_TODO_

If this is a new template: 


shiftkey:
Thanks, but this is covered by [`Global/Archives.gitignore`](https://github.com/github/gitignore/blob/master/Global/Archives.gitignore).

PR #2404, opened 2017-06-05
Add .gitignore file for Hugo.

felicianotech:

Hugo is missing and has a directory, where files are generated, that should be ignored.

https://gohugo.io/tutorials/automated-deployments#setting-up-version-control

If this is a new template: 

- **Link to application or project’s homepage**: https://gohugo.io/
- Hugo has grown greatly in the last few years. It's now [considered the # 2 most popular static site generator](https://www.netlify.com/blog/2017/05/25/top-ten-static-site-generators-of-2017/?utm_content=buffer4bc8a&utm_medium=social&utm_source=twitter.com&utm_campaign=buffer
) right behind Jekyll.


Hugo.gitignore
```diff
@@ -0,0 +1 @@
+/public/
```
bep: I suspect the more correct would be "public/"

felicianotech: The tutorial link leads with a forward slash. *shrug*. Updated.

joeblau:
@felicianotech this is included on https://github.com/dvcs/gitignore

https://www.gitignore.io/api/hugo

shiftkey:
@felicianotech apologies for the delay in replying to this.

I had a lot of these templates requested, but I couldn't quite figure out the right place for them. I ended up settling on #2857, and you'll see the Hugo template can now be found at `community/Golang/Hugo.gitignore`.

Hugo.gitignore
TamaMcGlinn: leading with a forward slash is the better approach. It ignores only the one directory called public in the root of the git repository. Without a leading slash, 'public/' ignores all directories that are called public, no matter where they are.

PR #2405, opened 2017-06-05
adding a template for OSX .DS_Store files

majormoses:

I have to do this every time I create a new repo and do not want OSX metadata files in source control.

If this is a new template: 

 - **Link to application or project’s homepage**: https://en.wikipedia.org/wiki/.DS_Store

### DIFF
--- a/Osx-DS.gitignore
+++ b/Osx-DS.gitignore
@@ -1,0 +1,2 @@
+# files that OSX creates to track metadata that is of no use to a remote system
+.DS_Store


shiftkey:
@majormoses thanks, but see `Global/macOS.gitignore`.

majormoses:
@shiftkey maybe I am missing something but I do not see any option to select these global .gitignore "profiles" from the UI:

image

majormoses:
I also verified that if you choose any profile it does not include all gobals so I am confused as to what is special about it. More information in the readme would be helpful. This does not really make it clear to me:

> The files in the root directory are for `.gitignore` templates that are project specific, such as language or framework specific templates. Global (operating system or editor specific) templates should go into the Global/ directory.

majormoses:
@shiftkey I'd love an explanation if you can spare it, I am still confused.

drothmaler:
In git there are the `.gitignore` files, that contain the [local gitignore-configuration](https://help.github.com/articles/ignoring-files/#create-a-local-gitignore) that is **specific to the _project_**.
And there is a [global gitignore-config](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore), that is **specific to _your_ (operating) system** and to the **editors and tools _you_ use**.
 
Those global setting should not be part of the project's `.gitignore` file - because they would just flood the file with tons of generic gitignore-rules - and it would probably still not match everybody's development environment. That's why github doesn't display those global gitignore-templates to add them to you project.

majormoses:
@drothmaler Thanks that helped, I understood the difference from a git perspective but did not understand how that played into not having those exposed as features which I guess seems pretty obvious when you make the connection. As a maintainer who deals with multiple people across multiple platforms (both in my org and out of it) I'd rather not rely on users setting things up properly (they never will) which is why I tend to include these "global" options into the projects. It's a shame that these global profiles are not exposed via github as it means that I will have to continue keeping quite large `.gitconfig` templates to just be copied as initial commits to each repo rather than doing that at creation. Was hoping to be able to reduce this effort.

PR #2406, opened 2017-06-06
Remove Backup*/ entry

masaeedu:
There are many valid use cases for having a folder with a name that **starts** with "Backup". Having `Backup*/` gitignored is too aggressive. 

The user will simply delete the `Backup` folder if they don't want it anyway, since it isn't a regularly generated artifact, so it doesn't make any sense to have it in `.gitignore`.

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -212,7 +212,6 @@ Generated_Code/
 # to a newer Visual Studio version. Backup files are not needed,
 # because we have git ;-)
 _UpgradeReport_Files/
-Backup*/
 UpgradeLog*.XML
 UpgradeLog*.htm
 


shiftkey:
This change was introduced due to how VS migrates solutions between versions, as you can have multiple folders each starting with `Backup`: https://github.com/github/gitignore/commit/e6117c95dc86aa0d8935862a4d70414be588c48f

As this is part of the default behaviour of VS (I still believe this is the case for VS2017 but I haven't tested), and this area is more complex than just the `Backup*` folders, I can't take in this change.

For reference, you can bypass the gitignore rules by including a `-f` when you want to track a folder in Git:

```
$ git add -f Backup/
```

masaeedu:
@shiftkey Yeah, I know about this, but it results in actual content with the word Backup in it being mysteriously gitignored, and you realize when you're doing `git clean -xdf` several commits later. In my case I realized the problem when some C# code related to backup functionality disappeared.

As I've explained in the PR description, the Backup folder is something the user will simply delete if they don't want it hanging around. It is a one time operation, not an artifact that gets generated regularly. There is no sense in gitignoring it, especially at the expense of introducing subtle problems in users' repos.

I'd really appreciate if you could reconsider this.

masaeedu:
Worst case you should anchor the rule to the root, i.e. `/Backup*/`.

shiftkey:
> Worst case you should anchor the rule to the root

I don't think we can do that - from memory the upgrade and backup folders are created alongside the solution file, which can be anywhere in the repository.

masaeedu:
As can the `.gitignore` file. In general the folder containing the `sln` file is parent to any folders containing `csproj` files, so as long as the `.gitignore` is placed adjacent to the `.sln` the rules will apply to all your Visual Studio related code.

PR #2409, opened 2017-06-07
Add entry for PyCharm

emiranda04:
When using PyCharm, '.idea/' is added by this IDE. If the developer is using Atom or VS Code (or any other editor for that matter), he/she might not want to have it included as part of the project.


shiftkey:
@emiranda04 thanks, but see `Global/JetBrains.gitignore` which is more detailed up-to-date.

PR #2413, opened 2017-06-10
Add Tabs Studio files to VisualStudio.gitignore

yannduran:

Tabs Studio is a Visual Studio extension that produces several *.tss files. Anyone who has this extension will have .tss file in their solution. These files are generated and therefore do not need to be added to source control.

[Tabs Studio](https://tabsstudio.com/)


shiftkey:
@yannduran thanks!

PR #2414, opened 2017-06-12
handle vscode files in unity .gitignore

omid3098:

Using vscode with Unity, .vscode directory should be ignored.
I've only added single line to handle this directory.

based on this repository:
https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore

If this is a new template: 


shiftkey:
Thanks, but see `Global/VisualStudioCode.gitignore` for the editor-specific template.

PR #2417, opened 2017-06-23
Added SuperCharger

gabrielizalo:
Supercharger is an extension designed to significantly improve the Visual Studio development experience. It builds upon and enhances previous VS10x products like CodeMAP, Editor View Enhancer, Comments Extender while also adding new high-quality tools.
Info: http://supercharger.tools/

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -293,3 +293,6 @@ __pycache__/
 *.btm.cs
 *.odx.cs
 *.xsd.cs
+
+# Supercharger
+.vs/Supercharger/


VisualStudio.gitignore
```diff
@@ -293,3 +293,6 @@ __pycache__/
 *.btm.cs
 *.odx.cs
 *.xsd.cs
+
+# Supercharger
+.vs/Supercharger/
```
shiftkey: Line 28 has `.vs/` defined, which should cover this rule.

gabrielizalo: Ah sure... Sorry.. I'll check why all Supercharger files were added to my solution.
;)

PR #2418, opened 2017-06-24
Add *.log and *.iml to JetBrains.gitignore

chenyufeng1991:
When we use mac and IntelliJ IDEA to develop JavaWeb  project with Maven, maybe we will generate many log files with suffix *.log . Consider about version control and release, we have to ignore log file .  
Same as *.iml files, these files will auto generate when we create a project in IntelliJ IDEA, but *.iml files are useless between developers.

https://github.com/chenyufeng1991/StartSpringMVC
This project is supporting these rule changes, and in my team, we follow these rules.

If this is a new template: 

 This is not a new template.

Thanks!


shiftkey:
@chenyufeng1991

> When we use mac and IntelliJ IDEA to develop JavaWeb project with Maven, maybe we will generate many log files with suffix *.log . 

What tool is generating these log files? I think there's probably a more suitable place to add this rule, as there are some other templates like `Maven.gitignore` and `Java.gitignore` but it's hard to say without more context.

> Same as *.iml files, these files will auto generate when we create a project in IntelliJ IDEA, but *.iml files are useless between developers.

[The JetBrains documentation](https://intellij-support.jetbrains.com/hc/en-us/articles/206544839) mentions that .iml files are useful to share between developers:

>

robmaceachern:
There are some situations where .iml files probably shouldn't be checked in. From the same [JetBrains post](https://intellij-support.jetbrains.com/hc/en-us/articles/206544839):
> You may consider not to share the following:
> - .iml files for the Gradle or Maven based projects, since these files will be generated on import
> - ...
> - XML files under .idea/libraries in case they are generated from Gradle or Maven project

Perhaps it'd be worth having a separate JetBrains-Gradle.gitignore? 🤷‍♂️

shiftkey:
> .iml files for the Gradle or Maven based projects, since these files will be generated on import

Are these found under a specific directory? Maybe we can make this rule more explicit than `*.iml`

We already have a Gradle-specific section:

```
 # Gradle:
 .idea/**/gradle.xml
 .idea/**/libraries
```

Let's see if we can get this specific enough to satisfy both what

shiftkey:
Closing due to inactivity. @chenyufeng1991 if you're happy to follow up with my feedback we can revisit this.

PR #2419, opened 2017-06-26
Added link to source

soreng:

To include a link to the source of the file

none

If this is a new template: 

none


shiftkey:
@soreng thanks!

PR #2420, opened 2017-06-26
Add extended DVI (XDV) to TeX.gitignore

niklasholm:

* XDV is the XeTeX-equivalent of DVI, an intermediary format that has to be converted before reading.

* https://www.ctan.org/pkg/xetex


shiftkey:
@niklasholm thanks!

PR #2422, opened 2017-07-06
.DS_Store is a file name, not an extension

danilcha:
I changed `*.DS_Store` to `.DS_Store`, because it is a filename: https://en.wikipedia.org/wiki/.DS_Store


shiftkey:
@danilcha great point - not sure why it's taken this long for someone to notice!

PR #2423, opened 2017-07-07
Add vendor to Golang git ignore list

hamidfzm:
Golang Dependency Management == Vendoring

Vendoring is the only method for golang dependency management. In this, you save a local copy of the dependent libraries that your application shall use. Each library is for a vender which its source codes should not be added to main project and only name of those dependencies should be tracked by tools like glide or ... .

So I think we should ignore vendor directory.


shiftkey:
As talked about in #2266 and #2452, we believe this should be opt-in rather than enabled by default.

PR #2426, opened 2017-07-11
request add /target/  in java.ignore

Think-007:

to ignore the target files in eclipse or idea

null

If this is a new template: 

not a new template.


shiftkey:
If this is specific to the tools then it should live in [`Global\JetBrains.gitignore`](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore) or [`Global\Eclipse.gitignore`](https://github.com/github/gitignore/blob/master/Global/Eclipse.gitignore)

PR #2427, opened 2017-07-12
Add React gitignore file

kristinbaumann:
This PR adds the gitignore file for React, a UI Library that gained popularity in the last couple of years. 

For the coding of a React project you need several node modules, that are generated upon installation and easily occupy a lot of space, and therefore should be ignored. The necessary package management  system can create error logs, which should be ignored as well.

To run a React project you need to bundle your files, e.g. with Webpack. This generates a build directory that is generated automatically and should be ignored for versioning with git.

[https://facebook.github.io/react/](https://facebook.github.io/react/)


joeblau:
@kristinbaumann This is included in https://github.com/dvcs/gitignore

https://www.gitignore.io/api/reactnative

shiftkey:
I'm closing this our because `Node.gitignore` covers most of these rules, and I don't believe `/build/` matches the conventions from webpack (they use `dist` in the example on their home page, but this can be changed by the user easily, and other tools exist for bundling your React app).

PR #2428, opened 2017-07-13
Visual Studio: ignore 'ncrunchproject' and 'ncrunchsolution' files

NRKirby:

When using the gitignore template prior to this commit and when NCrunch is added to a solution, 'ncrunchproject' and 'ncrunchsolution' files are tracked - I don't believe these files should be in source.

This commit fixes the issue by using the following pattern:

> `*ncrunch*` 

**Note:** Using the pattern above means that `_NCrunch_*` and `nCrunchTemp_*` were no longer required, hence me removing them in this commit.

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -121,9 +121,8 @@ _TeamCity*
 *.coveragexml
 
 # NCrunch
-_NCrunch_*
+*ncrunch*
 .*crunch*.local.xml
-nCrunchTemp_*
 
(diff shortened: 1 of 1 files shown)


shiftkey:

> It's recommended that the last two files listed in the above table (`.v3.ncrunchsolution` and `.v3.ncrunchproject`) be stored inside a source control system to be shared with team members. 

See https://github.com/github/gitignore/pull/2056 has some other details

PR #2432, opened 2017-07-18
use .DS_Store instead of *.DS_Store

yorkieindie:
the .DS_Store is the name of a file generated by macOS, not a file extension

https://git-scm.com/docs/gitignore
http://man7.org/linux/man-pages/man3/fnmatch.3.html
https://stackoverflow.com/questions/107701/how-can-i-remove-ds-store-files-from-a-git-repository
http://www.filetypehelp.com/dsstore-file-extension/

### DIFF
--- a/Global/macOS.gitignore
+++ b/Global/macOS.gitignore
@@ -1,5 +1,5 @@
 # General
-*.DS_Store
+.DS_Store
 .AppleDouble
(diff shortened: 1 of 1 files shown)


shiftkey:
@yorkieindie oops, I already merged #2422. Thanks anyway!

PR #2435, opened 2017-07-24
Create Red.gitignore

southwolf:

To ignore files generated during tests for Red (using quick-test framework).

https://github.com/red/red/issues/2921

https://github.com/red/red/blob/master/docs/red-system-quick-test.txt
http://static.red-lang.org/red-system-quick-test.html

If this is a new template: 

**Red language websites:**

http://www.red-lang.org/
http://red.github.io/

### DIFF
--- a/Red.gitignore
+++ b/Red.gitignore
@@ -1,0 +1,9 @@
+# Libraries
+crush.dll
+crush.dylib
+
+# Files generated during test
+quick-test/quick-test.log
+quick-test/runnable/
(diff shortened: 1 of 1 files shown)


shiftkey:
@southwolf apologies for the delay in replying to this.

I had a lot of these templates requested, but I couldn't quite figure out the right place for them. I ended up settling on #2857, and you'll see the Red template can now be found at `community/Red.gitignore`.

PR #2436, opened 2017-07-25
Update JetBrains.gitignore

knight7:
Adds .idea/**/preferred-vcs.xml

The preferred-vcs.xml file isn't ignored by the pre-existing JetBrains.gitignore, has recently begun showing up after updating to 2017.1.5 / 2017.2 versions of JetBrains IDEs, and seems irrelevant/useless in a repository already under Git.
The file lists 'ApexVCS' as the preferred VCS despite the IDE already recognizing the project is under Git, and the file is re-created after deletion each time the IDE is run.

### DIFF
--- a/Global/JetBrains.gitignore
+++ b/Global/JetBrains.gitignore
@@ -4,6 +4,7 @@
 # User-specific stuff:
 .idea/**/workspace.xml
 .idea/**/tasks.xml
+.idea/**/preferred-vcs.xml
 .idea/dictionaries
 
 # Sensitive or high-churn files:


shiftkey:
@knight7 i don't see any mention of this in [the official docs](https://intellij-support.jetbrains.com/hc/en-us/articles/206544839) which seem up-to-date. Do you have any other links to discussions about this new file handy?

knight7:
@shiftkey I'm sorry; I don't at this time.

shiftkey:
Hey @hhariri & @maartenba, who would know if this file is important to version or not?

maartenba:
Let me check...

maartenba:
Seems to come from this plugin - https://plugins.jetbrains.com/plugin/9238--jedide-plugin-for-handy-force-com-development-with-jetbrains-ides

If intended to share these settings with tea members it should not be ignored, otherwise can safely be ignored.

citizenmatt:
My vote is to not add to the global gitignore.

shiftkey:
@maartenba @citizenmatt thanks for the info!

@knight7 if we're going to include this, it should be in it's own section, with comments to explain where it's coming from, and **commented out by default**.

shiftkey:
Closing due to inactivity and the JetBrains crew suggesting this can be committed safely.

PR #2437, opened 2017-07-26
JetBrains: out/ ignored recursively

koppor:

`.gitignore` may be used to in a projects containing multiple Java projects. Therefore, in other settings, directories are ignored recursively (e.g., `.idea_modules`).

`/out/` was already in, so I think it is agreed that IntelliJ creates the `out` folder.


shiftkey:
@koppor thanks!

PR #2439, opened 2017-07-28
Update JetBrains.gitignore.

meyerlasse:
- Ignore all cmake build directories, not just debug.

If one uses a different cmake configuration, like an additional one for a release build, these build directories should also be ignored (e.g. cmake-build-release).


shiftkey:
@meyerlasse any links to StackOverflow discussions or other documentation about this change?

Yousha:
@shiftkey I think these links helps:
https://stackoverflow.com/a/34898501/1429432

And this one by JetBrains staff:
https://intellij-support.jetbrains.com/hc/en-us/community/posts/206020944/comments/204888850

shiftkey:
@Yousha thanks!

PR #2440, opened 2017-07-30
Remove TODO from VS.gitignore

jamesqo:

I search my GitHub repo regularly for the string "TODO" and this always shows up. Please consider removing the todo; I'm also open to changing it to something else like `Note: `.

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -151,7 +151,7 @@ publish/
 # Publish Web Output
 *.[Pp]ublish.xml
 *.azurePubxml
-# TODO: Comment the next line if you want to checkin your web deploy settings
+# Comment the next line if you want to checkin your web deploy settings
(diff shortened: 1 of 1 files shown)


shiftkey:
@jamesqo good point!

VisualStudio.gitignore
```diff
@@ -151,7 +151,7 @@ publish/
 # Publish Web Output
 *.[Pp]ublish.xml
 *.azurePubxml
-# TODO: Comment the next line if you want to checkin your web deploy settings
```
shiftkey: :thumbsup: to some other prefix here. I'm not really sure if there's a better way here to indicate that you _can_ introduce an optional rule here, but `Note:` is a fine substitute.

jamesqo: OK, I went with `Note: `.

shiftkey:
@jamesqo thanks!

PR #2448, opened 2017-08-07
Update VisualStudio.gitignore

FelipeVG83:

_TODO_

_TODO_

If this is a new template: 


VisualStudio.gitignore
```diff
@@ -75,6 +75,7 @@ artifacts/
 *.pidb
 *.svclog
 *.scc
+*.txt
```
shiftkey: Text files seem important to version control. Could you tell me a bit about why you think they should be excluded?

PR #2451, opened 2017-08-11
Fix comments on same line causing ignore to break

Xeio:

The current ActionScript ignore file doesn't work. bin and obj directories are not ignored because of the comment on the same line following the pattern.

Also removed redundant declaration of "bin/"


shiftkey:
@Xeio thanks!

PR #2452, opened 2017-08-11
vendor/ folder added

MarcusMann:

Added vendor folder to be ignored by git, in order to work well with glide, godep and etc ...

https://github.com/golang/dep,
http://glide.sh/

### DIFF
--- a/Go.gitignore
+++ b/Go.gitignore
@@ -12,3 +12,6 @@
 
 # Project-local glide cache, RE: https://github.com/Masterminds/glide/issues/736
 .glide/
+
+# Ignores the vendor directory provided by some applications such as: Glide, Godep, etc...
+vendor/


shiftkey:
@MarcusMann this was removed in https://github.com/github/gitignore/pull/2266 as part of cleaning things up. Without more information around why it's better/safer to now ignore this folder, especially from those who know more about the Go ecosystem than myself, I'm going to stick with the discussion there.

cc @AlekSi

AlekSi:
Please see #2266 for some discussion about it.

AlekSi:
See also https://github.com/golang/dep/blob/master/docs/FAQ.md#should-i-commit-my-vendor-directory.
I think a decision to ignore `vendor/` should explicit opt-in – just add it to `.gitignore` yourself.

PR #2458, opened 2017-08-20
Add default file, README in libraries and translations in Drupal

wirgen:
- sites/example.sites.php
Configuration file for Drupal's multi-site directory aliasing feature. But its only example. To use this file, copy and rename it to "sites.php"
- sites/*/translations
This folder stores temporarily downloaded translations
- sites/all/libraries/README.txt
README file for libraries


shiftkey:
@wirgen thanks!

PR #2459, opened 2017-08-20
Small fix - Meta files

KTGLeiden:
.Meta files were actually included in the commit, which is probably not what you want ;-)

**Reasons**
META files are included now, which is in unwanted, I would say.

### DIFF
--- a/Unity.gitignore
+++ b/Unity.gitignore
@@ -24,7 +24,7 @@ ExportedObj/
 *.pdb
 
 # Unity3D generated meta files
-*.pidb.meta
+*.meta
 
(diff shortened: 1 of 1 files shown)


shiftkey:
@KTGLeiden @garyritchie I'm not sure which way to go on this. Any external resources/discussions about the _types_ of files that are caught by this would help, as I'm not familiar with the context.

KTGLeiden:
@garyritchie 
Ah, then this is probably a misunderstanding. I saw meta files being included multiple times online. Sorry guys!

PR #2460, opened 2017-08-22
Use desktop.ini instead of Desktop.ini

NeverMine17:

Some Windowses uses desktop.ini instead of Desktop.ini .

https://ru.wikipedia.org/wiki/Desktop.ini

Translation:
> **d**esktop.ini is a configuration file that contains data on external systems of the system folder in Microsoft Windows operating system: icon, text color, wallpaper, etc.

If you visit it on English wiki you will be redirected to .INI page.

### DIFF
--- a/Global/Windows.gitignore
+++ b/Global/Windows.gitignore
@@ -8,6 +8,7 @@ ehthumbs_vista.db
 
 # Folder config file
 Desktop.ini
+desktop.ini
 
 # Recycle Bin used on file shares
 $RECYCLE.BIN/


Global/Windows.gitignore
```diff
@@ -8,6 +8,7 @@ ehthumbs_vista.db
 
 # Folder config file
 Desktop.ini
+desktop.ini
```
shiftkey: This can actually be folded up into the previous rule:

`[Dd]esktop.ini`

NeverMine17: Wow, I didn't know that. Thanks!

Yousha:
Modified version of this commit/PR is here now: https://github.com/github/gitignore/pull/2504

So I think you can close this PR.

PR #2462, opened 2017-08-23
update java.gitignore

hudongyang123:
Add the support for maven.

_Because Java web projects often use maven tools._

### DIFF
--- a/java.gitignore
+++ b/java.gitignore
@@ -1,0 +1,16 @@
+
+*.class
+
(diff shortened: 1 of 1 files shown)


shiftkey:
Thanks, but see [Java.gitignore](https://github.com/github/gitignore/blob/master/Java.gitignore) and [Maven.gitignore](https://github.com/github/gitignore/blob/master/Maven.gitignore).

PR #2464, opened 2017-08-29
updated static and media files in django section

Cadmus:
the media, and static_storage files is not needed while pushed the items in the git, it will increase the memory usage.

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -51,6 +51,8 @@ coverage.xml
 
 # Django stuff:
 *.log
+.static_storage/
+.media/
 local_settings.py
 
 # Flask stuff:


shiftkey:
@Cadmus any links to StackOverflow discussions or other documentation about this change?

Cadmus:
In the django project application files are uploaded into media folder,it's included into project root and once we make the collectstatic for the project it will create static_storage folder with all the static files (js,css,etc) but the same things is there in "static" folder . so both folders have "media" and "static_storage" raise the file size.

shiftkey:
@Cadmus 👍

cyliang:
Although I agree files in static folder and media folder shall be excluded from git repository, this pull request is not quite reasonable.

The paths of [static folder](https://docs.djangoproject.com/en/2.0/ref/settings/#std:setting-STATIC_ROOT) and [media folder](https://docs.djangoproject.com/en/2.0/ref/settings/#std:setting-MEDIA_ROOT) are actually personal preferences and the ones added in this pull request are neither Django's default nor something suggested/illustrated in Django's documentation/[tutorial](https://docs.djangoproject.com/en/2.0/howto/static-files/). In fact, Django does not provide default values for these two paths.

There are other reasons these two paths shall not be included in the gitignore:
1. Both folders are not necessary to be included in the project folder since files in both folders are not served by Django itself but instead a web server.
2. Depending on deployment environment, Django may have permission problems in automatically creating these two folders. In case these folders are decided to be located in the project folder, include both folders in git repository, with ideal permission modes set, is often a simple way to prevent the folder creation problem.

PR #2465, opened 2017-08-29
Nanoc: Ignore tmp/nanoc/, not tmp/

ddfreyne:

Nanoc stores its temporary data in `tmp/nanoc/` now, rather than `tmp/` as before.

https://github.com/nanoc/nanoc/pull/1071:

> Temporary data is now stored in tmp/nanoc/ rather than tmp/, so that Nanoc can ± safely assume that anything in tmp/nanoc/ is 100% under control of Nanoc.

### DIFF
--- a/Nanoc.gitignore
+++ b/Nanoc.gitignore
@@ -4,7 +4,7 @@
 output/
 
 # Temporary file directory
-tmp/
+tmp/nanoc/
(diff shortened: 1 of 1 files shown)


shiftkey:
@ddfreyne thanks!

PR #2467, opened 2017-09-01
Node - update Compiled binary addons link to https

rdela:

Save a redirect in Node.gitignore build/Release Compiled binary addons: http to https

https://nodejs.org/api/addons.html


shiftkey:
@rdela thanks!

PR #2469, opened 2017-09-03
Add ignoring package-lock.json

Jqmey:

* `package-lock.json` is automatically generated and contains full tree of dependencies. It should not be published, `package.json` should be used intead.

* [https://docs.npmjs.com/files/package-lock.json](https://docs.npmjs.com/files/package-lock.json)

### DIFF
--- a/Node.gitignore
+++ b/Node.gitignore
@@ -1,4 +1,3 @@
-
 # Logs
 logs
 *.log


drothmaler:
Nope... Please check the documentation link you provided... It explicitly states: 

> This file is **intended to be committed into source repositories**, and serves various purposes:

PR #2471, opened 2017-09-08
Added MATLAB/Simulink gitignore

IndefiniteBen:

MATLAB/Simulink are two highly used packages that many people use to develop things, and also use GitHub for source control, it would be a benefit to have a gitignore that has the temporary/compile/autosave MathWorks files in it and can be chosen up repo creation.

http://nl.mathworks.com/help/simulink/ug/set-up-git-source-control.html

If this is a new template: 

mathworks.com

### DIFF
--- a/MathWorks.gitignore
+++ b/MathWorks.gitignore
@@ -1,0 +1,18 @@
+## Autosave files
+# Windows default autosave extension
+*.asv
+
+# OSX / *nix default autosave extension
+*.m~
+
+# MATLAB Version Model Backup
+*.slx.r20*
+*.mdl.r20*
+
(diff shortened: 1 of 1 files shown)


shiftkey:
@IndefiniteBen thanks for this, but we already have a [`Global/Matlab.gitignore`](https://github.com/github/gitignore/blob/master/Global/Matlab.gitignore).

Could you add anything you need to that file instead? Thanks!

IndefiniteBen:
Ah I see. Is there any way to be able to select that gitignore from the list when creating a new repo, or is that somewhere Global gitignores will never be accessible? I shall modify that one instead.

PR #2474, opened 2017-09-12
Remove double-spaced lines

Ebola16:
I know this doesn't follow the contributing guidelines but it's a really simple change for consistency. I tested some regex and might as well make it into a PR.


shiftkey:
@Ebola16 thanks!

PR #2475, opened 2017-09-12
Fix failure to ignore Simulink autosave files

Ayase-252:

Name of autosave file generated by Simulink is formatted as `model_file_name.autosave`.  Therefore, an asterisk is needed to ignore such files.  `.autosave` is not enough.

[Autosave Options](https://cn.mathworks.com/help/simulink/gui/simulink-preferences-model-file-pane.html#brtjhn8)


shiftkey:
@Ayase-252 thanks!

PR #2476, opened 2017-09-14
Add VSCode in python gitignore

shimulch:
IDE related files and folders are not part of a software repository

Visual Studio Code doc states that it makes a .vscode file on the workspace 


shiftkey:

PR #2479, opened 2017-09-19
Add cabal.project.local~ to Haskell.gitignore

RyanGlScott:

When using the `cabal` build tool, if one calls `cabal new-configure` and an existing `cabal.project.local` file exists in the current directory, it will be moved to `cabal.project.local~` as a backup before creating a fresh `cabal.project.local` to take its place. Just as we currently ignore `cabal.project.local`, we ought to ignore `cabal.project.local~`.

[Documentation about `cabal.project.local`](https://github.com/haskell/cabal/blob/282eb213e6ce1ce8740b8120a422be4f3228c7bc/Cabal/doc/nix-local-build.rst#developing-multiple-packages)

### DIFF
--- a/Haskell.gitignore
+++ b/Haskell.gitignore
@@ -17,5 +17,6 @@ cabal.sandbox.config
 *.eventlog
 .stack-work/
 cabal.project.local
+cabal.project.local~
 .HTF/
 .ghc.environment.*


mcandre:
This would technically be covered by the Emacs formula as well, but I think it's a good idea to include this in Haskell.gitignore anyway, for non-Emacs users.

PR #2480, opened 2017-09-20
Update Autotools.gitignore

Ckins:

add files generated by the autotool m4. It is common in open source projects, like google protobuf.

http://www.gnu.org/software/m4/m4.html

If this is a new template: 


shiftkey:
@Ckins thanks!

PR #2482, opened 2017-09-20
Add exclusion of Visual Studio trace files (*.e2e)

jimbob7198:
Reason for change:
Exclude Visual Studio Trace files (*.e2e) as these are runtime generated files and not candidates for source control inclusion

https://docs.microsoft.com/en-us/dotnet/framework/wcf/service-trace-viewer-tool-svctraceviewer-exe


shiftkey:
@jimbob7198 thanks!

PR #2483, opened 2017-09-21
Add *.snk strong name key files

wolf99:
Strong name key files shouldn't be included in a repository AFAIK. they are intended to sign build output to verify that it comes from the correct publisher. having an *.snk in a repository would allow anyone to produce build as if they were the original publisher.

If *.snk files are needed for CI build then they should be encrypted with a password that is an environment variable.

I guess some OSS projects might like to have *.snk files in their repos but that would be an exception to the rule.

https://stackoverflow.com/questions/10474852/any-security-issues-adding-a-strong-name-key-to-source-control-for-an-open-sourc

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -210,6 +210,7 @@ ClientBin/
 *.jfm
 *.pfx
 *.publishsettings
+*.snk
 orleans.codegen.cs
 
 # Since there are multiple workflows, uncomment next line to ignore bower_components
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -210,8 +210,22 @@ ClientBin/
 *.jfm
 *.pfx
 *.publishsettings
-*.snk
 orleans.codegen.cs
+
+# Uncomment following rule to ignore strong name key files.
+# "Strong-naming gives an application or component a unique identity that other 
+# software can use to refer explicitly to it. For example, strong-naming enables 
+# application authors and administrators to specify a precise servicing version 
+# to be used for a shared component. This enables different applications to specify 
+# different versions without affecting other applications. In addition, you can 
+# use the strong name of a component as security evidence to establish a trust 
+# relationship between two components."
+# Having an *.snk in a repository would allow anyone to produce build as if they 
+# were the original publisher. 
+# If *.snk files are needed for CI build then they should be encrypted with a 
+# password that is an environment variable. Use of a password protected PFX may
+# be prefereable in this case (see `*.pfx` line above).
+#*.snk
 
 # Since there are multiple workflows, uncomment next line to ignore bower_components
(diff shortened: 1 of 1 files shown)


shiftkey (commented):
@wolf99 I worry about the inverse scenario here - someone thinking they've committed a key that's important (and set up correctly), but it wasn't tracked and is removed when a `git clean` happens.

By default these templates should be conservative with what they can **safely** exclude.

What about having a rule commented out with a comment above it explaining why you can enable this, for people to opt-in? Otherwise I'd just leave it as-is.

wolf99:
@shiftkey thanks for the info - I've adjusted as requested.

shiftkey:
@wolf99 thanks!

akrone-hach:
Hello from the future! I appreciate the link in the VisualStudio.gitignore for more info on this.

Ashokvyas786:
[anil_kumar_cv.pdf](https://github.com/github/gitignore/files/9054230/anil_kumar_cv.pdf)

sliekens:
Hello from the even more distant future! I also found this PR by following the link in my gitignore.

I am here to spread the word that Microsoft wants you to CONSIDER adding the snk to source control, as this allows other developers to modify your libraries and recompile them with the same key.

https://learn.microsoft.com/en-us/dotnet/standard/library-guidance/strong-naming#:~:text=%E2%9C%94%EF%B8%8F%20CONSIDER%20adding%20the%20strong%20naming%20key%20to%20your%20source%20control%20system.

I believe excluding the snk by default is still a good default, but you might want to consider this.

image

Youssef1313:
While the `*.snk` line is commented out, I think it's not correct to say that adding it to source control is a security risk. But I'm going to refer to @blowdart on this to advise if this PR should be reverted.

blowdart:
Oh goodness, this is an old issue :) And it's been merged already, but as @sliekens says for open-source or modern projects checking in a .snk is, in fact, [encouraged](https://learn.microsoft.com/en-us/dotnet/standard/library-guidance/strong-naming). 

The salient part

> ✔️ CONSIDER adding the strong naming key pair (public + private) to your source control system.
> 
> A publicly available key pair lets developers modify and recompile your library source code with the same key.
> 
> You shouldn't make the strong naming key pair public if it has been used in the past to give special permissions in [partial-trust scenarios](https://learn.microsoft.com/en-us/previous-versions/dotnet/framework/code-access-security/using-libraries-from-partially-trusted-code). Otherwise, you might compromise existing environments.
> 
> If you can't check in the public + private key pair, then check in the public key and use [public signing](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/security#publicsign) for regular builds. Public signing still allows developers to recompile and use your library in most scenarios.

My opinion here is to **_not_** add .snk to .gitignore, for modern development it's

PR #2491, opened 2017-10-02
TeX.gitignore modification for Emacs users

cmower:
For Emacs users: currently the `auto` directory and Emacs-Lisp files `.el` are still considered for commits.

Two minor modifications have been made.

1. `/auto/*` changed to `./auto/*`: the latter ignores the `auto` directory.
2. additional line `*.el` that ensures all Emacs-Lisp files are ignored.


shiftkey:
@cmower thanks!

PR #2493, opened 2017-10-05
Add sccprj/ folder to global MATLAB.gitignore

caioproiete:

From the [MathWorks docs](http://www.mathworks.com/help/simulink/ug/work-with-derived-files-in-projects.html):
> Best practice is to exclude derived files, such as .mex*, the contents of the slprj folder, **sccprj folder**, or other code generation folders from source control, because they can cause problems

- http://www.mathworks.com/help/simulink/ug/work-with-derived-files-in-projects.html


shiftkey:
@caioproiete thanks!

PR #2494, opened 2017-10-07
Create Emacs.gitignore

braegel:
This is my suggestion for ignoring annoying Emacs files.

### DIFF
--- a/Emacs.gitignore
+++ b/Emacs.gitignore
@@ -1,0 +1,3 @@
+# Ignoring Emacs backup- and autosave-files
+*#
+*~


cmower:
@braegel you don't need a `.gitignore` file for this. You can modify your `.emacs` file to handle these "annoying" files, see [here](https://www.emacswiki.org/emacs/AutoSave).

braegel:
I love emacs and its auto files, but as it happens some people (including me) accidentally add these to a git repository, where they do not belong.

shiftkey:
@braegel thanks, but see [Global/Emacs.gitignore](https://github.com/github/gitignore/blob/master/Global/Emacs.gitignore).

PR #2495, opened 2017-10-07
[Python gitignore] add .idea to gitignore;

opalczynski:

PyCharm IDE stores the project settings in set  of the xml files stored under the .idea directory;
As I am working - and probably not only me :) with PyCharm (which is one of the most popular IDE for python), adding this each time to the gitgnore file created from template is little bit irritating. 

* [Official PyCharm IDE documentation](https://www.jetbrains.com/help/pycharm/project-and-ide-settings.html)

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -102,3 +102,6 @@ venv.bak/
 
 # mypy
 .mypy_cache/
+
+# PyCharm IDE settings
+.idea


xlui:
same to me, i'm also tired of adding `.idea` to `.gitignore` again and again!

joeblau:
If you use https://www.gitignore.io, it already generates a template that ignores `.idea`: https://www.gitignore.io/api/pycharm%2Biml

mcandre:
@opalczynski As joeblau mentions, there is already a gitignore formula for PyCharm. Note that both gitignore.io and the bash curl alias accept multiple, comma-separated formulae to generate an aggregate .gitignore file.

drothmaler:
The PyCharm template on gitignore.io is just an alias for `Global\Jetbrains.gitignore`... This is the gitignore file to use for all IDEA bases IDEs from Jetbrains.

joeblau:
@drothmaler It's like that right now, but I've created design patterns to extend the PyCharm template if it needs more modifications.

PR #2496, opened 2017-10-08
Commenting out Android keystore files

malhobayyeb:

Android keystore files are critical and losing them is destructive.

They should be commented out by default, the user should explicitly uncomment their line if they intend not to include them into their repository.

### DIFF
--- a/Android.gitignore
+++ b/Android.gitignore
@@ -41,7 +41,8 @@ captures/
 .idea/libraries
 
 # Keystore files
-*.jks
+# Uncomment the following line if you do not want to check your keystore files in.
+#*.jks
 
 # External native build folder generated in Android Studio 2.2 and later
 .externalNativeBuild


shiftkey:
@malhobayyeb thanks!

matpag:
I'm against this change, keystores should never be uploaded with the project source files. Exposing them from source control will greatly increase security flaws. If an attacker takes this file and break the password (which its easy) he potentially could publish app updates in your place.

Yousha:
I'm against this change too, this increases security risk...

friederbluemle:
Sorry, but I strongly disagree with this change as well.

Keystore files should be kept on an *offline* encrypted USB drive (or otherwise secure medium). If you do choose to commit a jks file to source control, it should be an *explicit decision* - meaning you would _manually have to comment out this line_ (instead of the other way around).

@shiftkey I reverted this commit and also put it onto my PR that has been open for a while:
https://github.com/github/gitignore/pull/2402
@matpag @Yousha It would be great if you could also take a quick look at it. Thanks!

PR #2498, opened 2017-10-13
[Visual Studio] Add ignore for Azure Stream Analytics local run output folder

t-my:

When using visual studio with stream analytics facet, output folder was not ignored. 

https://docs.microsoft.com/en-us/azure/stream-analytics/stream-analytics-tools-for-visual-studio


shiftkey:
@t-my this seems fine, but I'm going to reach out to my contacts on the VS team and see if I can discourage them from doing this sort of thing in the future.

PR #2500, opened 2017-10-15
add Idea File and Eclipse File

TinChiWay:

can this files and folders would be include?

### DIFF
--- a/Java.gitignore
+++ b/Java.gitignore
@@ -18,5 +18,15 @@
 *.tar.gz
 *.rar
 
+# Idea Files and Folders #
+*.iml
+/.idea/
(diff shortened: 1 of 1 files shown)


Yousha:
You can simply use https://github.com/github/gitignore/blob/master/Global/Eclipse.gitignore and https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore for developing Java applications

shiftkey:
Yep, these templates are intended to be references - so we split out editors into their own templates to make things more maintainable.

TinChiWay:
thanks，i ignored the Global folder

PR #2501, opened 2017-10-15
Fixed typos in README.md

flowonyx:
There were just a couple of typos in the README that I fixed.

I noticed two missing words that were not a big deal but should be fixed anyway. The first one could be OK without the change but the second one needs the added "is".


shiftkey:
@flowonyx thanks!

PR #2503, opened 2017-10-18
Ignore Nuget package directory name casing

synepis:
It looks like VS2017 changed it's nuget package directory name to be capitalized.

If you are using Ubuntu for Windows as your git shell, directory names are case sensitive, and will be seen as changes by git.

Since it doesn't matter for Windows users, perhaps we can add this case ignoring to .gitignore.


shiftkey:
@synepis thanks!

PR #2504, opened 2017-10-23
Windows.gitignore - Change desktop.ini to [Dd]esktop.ini

Yousha:

Some Windowses uses **desktop.ini** instead of **Desktop.ini**.
This commit can close PR #2460 by @NeverMine17

https://ru.wikipedia.org/wiki/Desktop.ini


shiftkey:
@Yousha thanks!

PR #2506, opened 2017-10-27
Fix link to go to current anchor

cliffordp:

Old link pointed to a non-existent anchor.


shiftkey:
@cliffordp thanks!

PR #2509, opened 2017-10-29
Update Terraform.gitignore

hantuzun:
Improvements: 

 * `.terraform` directories could be created in any subdirectory when user runs `terraform init`.
 * Quite an interesting variations of`.tfstate` files are created by Terraform such as `.*.tfstate.d` or  `.*.tfstate.lock.info`. It's best not to name any file with `*.tfstate.*` pattern.
 * `*.tfvars` files often contain private data and not supposed to be committed.

_TODO_

_TODO_

If this is a new template: 

### DIFF
--- a/Terraform.gitignore
+++ b/Terraform.gitignore
@@ -1,10 +1,9 @@
-# Compiled files
+#  Local .terraform directories
+**/.terraform/*
+
+# .tfstate files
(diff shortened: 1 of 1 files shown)


shiftkey:
@hantuzun thanks!

PR #2510, opened 2017-10-30
Update VisualStudio.gitignore

denis-peshkov:
added additional files to be ignored


drothmaler:
Ignoring StyleCop reports might be a good Ideas, but for `Thumbs.db` there already is `Global/Windows.gitignore`

VisualStudio.gitignore
```diff
@@ -51,6 +51,13 @@ project.fragment.lock.json
 artifacts/
 **/Properties/launchSettings.json
 
+# StyleCop
+StyleCopReport.xml
+
+# Thumbnails created by windows
+Thumbs.db
```
shiftkey: This is covered by `Global/Windows.gitignore` - please remove it.

VisualStudio.gitignore
```diff
@@ -51,6 +51,13 @@ project.fragment.lock.json
 artifacts/
 **/Properties/launchSettings.json
 
+# StyleCop
+StyleCopReport.xml
```
shiftkey: Any documentations or discussions to explain the context behind this change?

VisualStudio.gitignore
denis-peshkov: Removed

VisualStudio.gitignore
denis-peshkov: Unfortunately I can't found out any documents described the output files...
But this file appears very often in our projects, so I propose add it for now and add/change in future, when will have more info.

PR #2515, opened 2017-11-13
Add .vscode/

Kamys:
My vsсode stores tasks in .vscode/ so I think this file also need to add


neico:
https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore

shiftkey:

PR #2524, opened 2017-11-23
UnrealEngine - Include Plugins folder

halbich:
When working with plugins, we should also ignore their Binaries and Intermediate folders.

### DIFF
--- a/UnrealEngine.gitignore
+++ b/UnrealEngine.gitignore
@@ -50,6 +50,7 @@ SourceArt/**/*.tga
 
 # Binary Files
 Binaries/*
+Plugins/*/Binaries/*
 
 # Builds
 Build/*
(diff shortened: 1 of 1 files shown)


shiftkey:
@halbich any external links to documentation or discussions about this, for context?

halbich:
@shiftkey Documentation about UE plugins can be found here: [https://docs.unrealengine.com/latest/INT/Programming/Plugins/](https://docs.unrealengine.com/latest/INT/Programming/Plugins/). But it is pretty simple.

The main reason why I proposed this addition is the fact that I added new plugin to my game and folders generated by UE compiler was not automatically ignored by current template. If we are ignoring build folders created during game compile time, I think that it would be coherent approach that we treat plugins the same as we treat game itself. Because compiler does the same - creates Binaries and Intermediate folder for each plugin and also for a game itself. And doesn't matter if it is custom plugin or plugin by another developer - as long as there is C++ code in it, the compiler will generate these folders.

shiftkey:
@halbich thanks! This seems safe enough to take in.

PR #2526, opened 2017-11-26
Add gradle properties

mitohato:

Information of proxy is automatically added in gradle.properties.
It is url, port number, username and pass.

Nothing.

If this is a new template: 

 - AndroidStudio: https://developer.android.com/studio/index.html
--


ideaman924:
I second this. This file is not needed.

drothmaler:
Nope... Typically `gradle.properties` should be committed, as it contains project specific configuration properties (e.g. for multi-module builds) (see the first paragraph in this gradle documentation section: https://docs.gradle.org/current/userguide/build_environment.html#sec:gradle_configuration_properties  as well as this discussion: https://discuss.gradle.org/t/best-practice-for-gradle-properties-and-scm/5809/4 ). 
Machine specific properties can be placed into `local.properties` (which already is ignored) or into `gradle.properties` in Gradle user home or the could be passed as system property. 
If Android Studio actually stores machine-specific settings into the project `gradle.properties` file instead of `local.properties` than you should better file a AndroidStudio-Bug

shiftkey:
Seconding what @drothmaler said. This file seems important to track.

PR #2527, opened 2017-11-26
Remove all .idea files

ideaman924:
I've looked through the folder. All of them are generated when the IDE first launches.

When moving my repository across different machines, I've seen very bad behavior from the IDE, such as modules failing to load unless I remove them and add them manually, or some Java tests running bad when the code is actually fine.

If there is a reason the .idea folder removal is unhelpful, I would very much like to hear it.

_TODO_

_TODO_

If this is a new template: 

### DIFF
--- a/Android.gitignore
+++ b/Android.gitignore
@@ -34,11 +34,7 @@ captures/
 
 # IntelliJ
 *.iml
-.idea/workspace.xml
-.idea/tasks.xml
-.idea/gradle.xml
-.idea/dictionaries
-.idea/libraries
+.idea/
 
 # Keystore files
 # Uncomment the following line if you do not want to check your keystore files in.


drothmaler:
Well, moving (instead of cloning) a repository across different machines doesn't seem to be a valid git usage scenario according to my understanding. Also I don't know how a gitignore file would help in this case.
But maybe I misunderstood you. Do you really mean physically moving the repo? Also what are the differences between the machines (OS, IntelliJ-/Android-Studio-Versions)?
Maybe your actual problem are EOL-Differences or Version incompatibilities...

ideaman924:
Ah, sorry. I should’ve said I worked on the same repository on multiple machines. They all have the same Android Studio version (3.0.1 for now) so there shouldn’t be any version differences. 

When I finish fleshing out my code on, say, a laptop, I push it to GitHub, move to my desktop, pull it, and open it there. That’s when Android Studio complains about “missing modules” and “an IDE error occurred.” Eventually with enough Gradle refreshes and invalidating caches the IDE behaves once more, but this is pretty cumbersome because I frequently move around devices.

I have checked that removing the files from the repository do not cause any adverse effects - as far as I know. I did build correctly yesterday with all repos freshly synced from GitHub with the files excluded from tracking.

shiftkey:
> If there is a reason the .idea folder removal is unhelpful, I would very much like to hear it.

The [JetBrains template](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore) points to official documentation around what can be ignored in version control, which is what we're currently basing our decisions on: https://intellij-support.jetbrains.com/hc/en-us/articles/206544839

They also point to that template, so it'd be nice to ensure we're all on the same page rather than just ignoring `.idea/` completely.

ideaman924:
I stand corrected. After a few hours the module error returned. I guess this is a problem with my environment and not exactly the gitignore problem.

Thread closed.

PR #2528, opened 2017-11-27
Add @std/esm ESM module cache

DenisCarriere:

`@std/esm` is starting to be widely adopted to support ESM Modules, it creates a `.esm-cache` folder with the compiled Javscript files. 

https://github.com/standard-things/esm

CC: @jdalton Author of `@std/esm`

### DIFF
--- a/Node.gitignore
+++ b/Node.gitignore
@@ -57,3 +57,5 @@ typings/
 # dotenv environment variables file
 .env
 
+# Standard Things ESM modules cache (https://github.com/standard-things/esm)
+.esm-cache


jdalton:
Hi @DenisCarriere!

Thanks for the ping! The `@std/esm` loader has changed to using the same cache location has `@babel/register, `nyc`, and friends *(the `node_modules/.cache` folder) so no longer needs its own ignore entry.

DenisCarriere:
@jdalton Even better! 👍

PR #2530, opened 2017-11-29
Ignore Visual Studio Code cache/options directory

marska:
Add .vscode cache/options directory to ignored.


drothmaler:
There is an extra ignore file for VSC: `Global/VisualStudioCode.gitignore`

marska:
ok

PR #2531, opened 2017-11-29
Fixing .gitignore

matzar:

Fixing paths for ignored directories 

https://github.com/github/gitignore/commit/e0c2baea55e9537dbc50e834a090da34c4229841


shiftkey:
@matt360 thanks!

matt360:
@shiftkey In the thick (when I was testing the previous changes) I got rid of '/' before the folder names and because it was working for my particular face as well as other simple cases. I and my team-mates use the template available here with our projects and it's actually the only one that's working. But this is not the version that is available for everyone when creating a new repository. Is there a way we can maybe have this new updated version there?

PR #2532, opened 2017-11-30
Update Android.gitignore

roh-7:
Ignoring all files in the .idea folder generated by intellij. Ignoring the release and debug folders generated at the time of release

### DIFF
--- a/Android.gitignore
+++ b/Android.gitignore
@@ -16,6 +16,8 @@ out/
 # Gradle files
 .gradle/
 build/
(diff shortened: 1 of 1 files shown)


Android.gitignore
```diff
@@ -16,6 +16,8 @@ out/
 # Gradle files
 .gradle/
 build/
+app/debug
```
shiftkey: Any external links to documentation or discussions about this change?

PR #2534, opened 2017-12-01
Add .vscode

AlexLomm:

Excluding [VSCode ](https://code.visualstudio.com/)(editor) specific files


shiftkey:

PR #2537, opened 2017-12-04
add pycharm .idea directory to python gitignore

lucgiffon:
add pycharm environment files (.idea) to gitignore

I prefer ignoring the files that could be not ignored but maybe i'm wrong.

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -102,3 +102,6 @@ venv.bak/
 
 # mypy
 .mypy_cache/
+
+# pycharm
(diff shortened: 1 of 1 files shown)


shiftkey:
Thanks, but see [`Global\JetBrains.gitignore`](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore) as that has more detailed rules.

lucgiffon:
Ok, thank you.

PR #2538, opened 2017-12-04
Ignore node_modules

stfnhrrs:
Rails 5 uses npm to handle javascript dependancies, node_modules should not be checked in to git.

http://guides.rubyonrails.org/5_1_release_notes.html#yarn-support


shiftkey:
@stfnhrrs thanks!

PR #2540, opened 2017-12-05
Update Jekyll.gitignore

tasken:

Add .jekyll-cache/ as per https://github.com/envygeeks/jekyll-assets


tasken:
If https://github.com/github/pages-gem/issues/189 is applied will be neccesary

shiftkey:
@tasken this seems like a pretty safe change irrespective of the linked issue. Thanks!

PR #2543, opened 2017-12-06
Ignoring the models builder out of date flag

richarth:
The out of date models flag is automatically generated by the system, we don't want this in Git


shiftkey:
@richarth thanks!

PR #2546, opened 2017-12-11
Added Godot gitignore (https://godotengine.org/)

henriiquecampos:

Godot Engine is becoming a major game engine in game development industry, and in preparation for the next major release, Godot Engine 3.0, I've prepared this .gitignore in order to make it easy for new projects to start with a template that will ignore automatic generated files, such as `.import` ones which will be generated as soon as a compatible file, such as a `.png`, is added to the project. 

Also since Godot now supports C# by using Mono, I added the `.mono` directory.

You can check how Godot Engine handles the import process in the [official documentation](http://docs.godotengine.org/en/latest/learning/workflow/assets/import_process.html?highlight=import).

If this is a new template: 

 Godot has a [github repository](https://github.com/godotengine/godot) with more than 10.000 stars and over 2.500 forks and an [official homepage](https://godotengine.org/) with contents being updated frequently.

### DIFF
--- a/Godot.gitignore
+++ b/Godot.gitignore
@@ -1,0 +1,3 @@
+.import/
+.mono/
+export_presets.cfg


MrJustreborn:
You shouldn't ignore *.import, since in this file are all the settings how to import the asset (eg Filter/Mipmap, etc)

Godot.gitignore
```diff
@@ -0,0 +1,8 @@
+temp/
+.import/
+.mono/
+*.import
```
akien-mga: This should not be ignored, it's what defines the import settings for each resource. If you ignore those, your project will always have the default import settings (so say farewell to pixel art).

Godot.gitignore
```diff
@@ -0,0 +1,8 @@
...
+*.DS_Store
+*.directory
+*.fscache
```
akien-mga: No supported version of Godot outputs the `.fscache` file anymore.

Godot.gitignore
```diff
@@ -0,0 +1,8 @@
+temp/
```
akien-mga: I don't know this being output by Godot?

Godot.gitignore
```diff
@@ -0,0 +1,8 @@
...
+.import/
+.mono/
+*.import
+*.DS_Store
+*.directory
```
akien-mga: `.DS_Store` and `.directory` are files, not extensions, so no need to put a leading `*`.

They're also not Godot-related, they come respectively from macOS's Finder and KDE's Dolphin, so I don't know if they're meant to be included in the Godot template (haven't checked how this gitignore repo is used exactly).

Godot.gitignore
henriiquecampos: I see some duplicates of `*.import`, should the folder be ignored or the file extension or none? Or just the `*.stex` and `*.oggstr`? I'm bit of confused with these imports extensions.

Also what about the **.mono** folder?

akien-mga: - `.import/`, the folder at the root of your project repo (where `project.godot` is), is where Godot stores all the files it imports automatically based on your source assets and their `assetname.import` flags. This should be git-ignored.
- `*.import` files are text files with the flags to tell Godot how to import each asset. They need to be kept, unless you want to reconfigure all the time.

willnationsdev:
You forgot the `logs/` folder.

akien-mga:
> You forgot the logs/ folder. Unless that should be included for some reason?

That was during unstable development, it should not longer be generated in the project folder.

shiftkey:
Closing this out in favour of #2613 which has some more detailed rules (and refers to the example from a demo project)

PR #2547, opened 2017-12-13
added kirby.gitignore

wumble:
gitignore file for kirby cms (https://github.com/getkirby/getkirby.com)

it would be nice to have kirby in this list

### DIFF
--- a/kirby.gitignore
+++ b/kirby.gitignore
@@ -1,0 +1,10 @@
+.DS_Store
+/scripts
+/site/accounts
+/site/cache
+/thumbs
+/site/config/config.dev.getkirby.com.php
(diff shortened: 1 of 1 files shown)


joeblau:
Is this the same thing as Kirby 2? 

https://github.com/dvcs/gitignore/blob/master/templates/Kirby2.gitignore

shiftkey:

@wumble if you still feel this should be a template listed in this repository, please read the section about [specialized templates](https://github.com/github/gitignore#specialized-templates) and open a new pull request that follows the new pattern.

I'm also curious about the question @joeblau asked earlier too - is this related to Kirby 2?

PR #2549, opened 2017-12-13
ignore MSBuild Binary and Structured Log

enricosada:
msbuild 15 added a `/bl` argument to generate a binary log.
these can be viewed with http://msbuildlog.com/ or inside visual studio 2017


shiftkey:
@enricosada thanks!

PR #2551, opened 2017-12-17
Adding helm gitignore to ignore compressed packages

sairam546:

Added helm gitignore to ignore compressed packages after dependency build or dependency update

https://docs.helm.sh/developing_charts/#syncing-your-chart-repository contains working of charts. When we do depency build or dependency update it will download charts as packages in format .tgz which are not needed to be included in source code.

If this is a new template: 

 - https://helm.sh/

### DIFF
--- a/Helm.gitignore
+++ b/Helm.gitignore
@@ -1,0 +1,2 @@
+#ignore packages
+**/.tgz


joeblau:
Could this be a compressed archive template?  I'm not sure if this is Helm specific.

https://www.gitignore.io/api/compressedarchive

shiftkey:
Seconding @joeblau - this doesn't seem specialized enough to require it's own gitignore template.

PR #2555, opened 2017-12-25
Added a few common CAM file extensions

PTS93:

These are common gerber file extensions that should be covered by this gitignore as it already ignores other gerber extensions.

https://github.com/OSHPark/OSHPark-Eagle-Tools/blob/master/cam/oshpark-2layer.cam

### DIFF
--- a/Eagle.gitignore
+++ b/Eagle.gitignore
@@ -4,6 +4,9 @@
 *.s#?
 *.b#?
 *.l#?
+*.b$?
+*.s$?
+*.l$?
 
 # Eagle project file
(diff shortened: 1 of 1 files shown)


shiftkey:
@PTS93 thanks!

Eagle.gitignore
```diff
@@ -31,14 +34,19 @@ eagle.epf
 *.drl
 *.gpi
 *.pls
+*.ger
+*.gpi
```
jayvdb: Note this was a duplicate entry. c.f https://github.com/github/gitignore/pull/2741

PR #2558, opened 2017-12-27
ignore vscode filefolder

vimiix:
VSCODE is becoming a major editor in python development.

[https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore](https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore)

If this is a new template: 

 - **Link to application or project’s homepage**: [https://code.visualstudio.com/](https://code.visualstudio.com/)

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -97,3 +97,6 @@ ENV/
 
 # mypy
 .mypy_cache/
+
+# vscode
+.vscode/


shiftkey:
Thanks, but duplicating existing templates adds additional maintenance work that I'm not able to support. The recommended advice is compose together templates in your repository, based on the languages and tools used.

PR #2562, opened 2018-01-01
Update Node.gitignore

rgbkrk:

`.next` directories contain the build artifacts for a next.js deployment and should not be placed in version control

[next.js `.next` build for production](https://github.com/zeit/next.js#production-deployment)


shiftkey:
@rgbkrk thanks!

PR #2563, opened 2018-01-04
Added Xcode Patch to the template file.

justinhartman:

While already thorough, the template didn't cater for certain Xcode project files. I have included a patch so that project files can be excluded. By retaining the `xcodeproj` files it causes conflicts between other users working on the same source code projects.


justinhartman:
I'm trying to follow the guideline of only modifying one template per pull request but having issues. After I created the pull request I worked on an additional new template and committed the code to my fork yet, however, I see that commit is now appearing on my pull request. I'm going to close the pull request (can't seem to find a way to delete the request) and try again. 

Sorry guys.

PR #2564, opened 2018-01-04
Update Xcode.gitignore

justinhartman:

While already thorough, the template didn't cater for certain Xcode project files. I have included a patch so that project files can be excluded. By retaining the xcodeproj files it causes conflicts between other users working on the same source code projects.


justinhartman:
Closing this.

PR #2566, opened 2018-01-04
[Xcode] add some new rules

justinhartman:

While already thorough, the template didn't cater for certain Xcode project files. I have included a patch so that project files can be excluded. By retaining the xcodeproj files it causes conflicts between other users working on the same source code projects.

### DIFF
--- a/Global/Xcode.gitignore
+++ b/Global/Xcode.gitignore
@@ -21,3 +21,10 @@ DerivedData/
 !default.mode2v3
 *.perspectivev3
 !default.perspectivev3
+
+## Xcode Patch
+*.xcodeproj/*
+!*.xcodeproj/project.pbxproj
+!*.xcodeproj/xcshareddata/
+!*.xcworkspace/contents.xcworkspacedata
+/*.gcno


Global/Xcode.gitignore
```diff
@@ -21,3 +21,10 @@ DerivedData/
 !default.mode2v3
 *.perspectivev3
 !default.perspectivev3
+
+## Xcode Patch
+*.xcodeproj/*
```
shiftkey: The cited StackOverflow thread only mentions these two rules:

```
*.xcodeproj/
DerivedData/
```

How did you come to these specific rules?

justinhartman: Apologies for taking so long to respond to your question. It's been a long time since I made these changes so I'll try my best to explain it all :smile:.

#### 1. `*.xcodeproj/`

There is a good [thread over here][1] that gets into some more specifics but the issue with excluding the `.xcodeproj/` in its entirety is that you are excluding files that shouldn't be excluded and may need to be shared across development team members. The additional rules I added address this problem of simply excluding the entire folder.

##### `.xcodeproj/` structure

In order to address the additional rules I added in [patch-1][patch] I have included the follow `tree` output of an `Xcode` project; in particular, the `.xcodeproj/` folder. I will use this to reference rules `#2`, `#3` and `#4` below.

```bash
LicenseGenerator-iOS/LicensesViewController.xcodeproj/
├── project.pbxproj
├── project.xcworkspace
│   └── contents.xcworkspacedata
└── xcshareddata
    └── xcschemes
        └── LicensesViewController.xcscheme
```

#### 2. `!*.xcodeproj/project.pbxproj`

In the above cited thread [this comment][2] discusses the efficacy of ignoring everything inside a `*.xcodeproj` except the `project.pbxpro` file itself.

A developer would want to include this file as it contains project data such as the list of source code and resource files along with build properties. It also saves group data which organises the files in the project ([ref][6], [ref][8]).

While the up-votes and logic is sound, rules `#3` and `#4` also need consideration and shouldn't just be ignored as rules `#1` and `#2` would achieve doing this.

#### 3. `!*.xcodeproj/xcshareddata/`

The `/xcshareddata/` folder contains important `xml` data contained within a `*.xcscheme` file. In my example above the `.xcscheme` file is called `LicensesViewController.xcscheme` and this contains important information covering the:

- Build,
- Test,
- Launch,
- Profile,
- Analyse, and
- Archive configuration for the project.

View [a gist file][9] containing the source code of the `LicensesViewController.xcscheme` file.

On [this page][5] it says: 

>The XCSCHEME file can be shared with other users or protected for a single user, which is useful for multi-developer projects.

While this is not conclusive enough to include the rule based on this, the page goes on to say that:

> If the file is checked "Shared" in the "Manage Schemes" window, the file is stored in the `xcshareddata/xschemes` directory.

This is an important statement because a user in `Xcode` has to check the option to share the scheme which in turn shows intent to want to share the file with other developers. If the file is excluded by rules `#1` and `#2` then the objective of sharing the scheme with others via SCM is lost.

#### 4. `!*.xcworkspace/contents.xcworkspacedata`

The `*.xcworkspace` folder (`project.xcworkspace` in the example) contains the `contents.xcworkspacedata` file which contains information describing the workspace or projects and lists the projects that are included as top-level entities in the project. 

In an environment where a user doesn't share workspaces this file can be ignored, however if there is related projects in the same workspace which share configuration then this file should not be ignored ([ref][7]).

This rule was added because I am working on an _assumption_ that when applying a `.gitignore` file to an `Xcode` project the intention is to share that project with other developers and if not, to keep your project as clean as possible yet retaining core configuration across project life-cycles.

This can be done either as part of a team or generally via public SCM (e.g. Github) so the need to include `contents.xcworkspacedata` is important in order to share workspace and project information with other members, as well as yourself, especially if there are related projects in the same workspace.

#### 5. `/*.gcno`

[This thread][3] discusses how `.gcno` files are used to manage code coverage information and the [Apple docs][4] give further information. Not only do these files contain coverage information but they also contain profile data stored in a platform-independent format. 

The `.gcno` files are placed in the same directory as a build object file and are used with the `gcc` compiler. They should therefore not be included in a source code repository as they are very build specific and can mess with other members build configuration settings if committed to a repository.

[1]: https://stackoverflow.com/questions/49478/git-ignore-file-for-xcode-projects
[2]: https://stackoverflow.com/questions/49478/git-ignore-file-for-xcode-projects#comment461987_49488
[3]: https://stackoverflow.com/questions/32738645/what-is-objectivec-gcno-should-it-be-included-in-source-code-repository
[4]: https://developer.apple.com/library/prerelease/ios/qa/qa1514/_index.html
[patch]: https://github.com/justinhartman/gitignore/commit/e1f0409ddcb9d3f3e08114640e685109e272d9f5
[5]: https://fileinfo.com/extension/xcscheme
[6]: https://fileinfo.com/extension/pbxproj
[7]: https://stackoverflow.com/a/10960309/1322242
[8]: https://stackoverflow.com/questions/10956312/is-the-project-xcworkspace-file-important
[9]: https://gist.github.com/justinhartman/44753ca01e765af146dc7a7336a15cd4

justinhartman: I hope this helps and clarifies matters?

brntbeer:
@shiftkey bump on @justinhartman's clarification. 😄 

I just started an xcode project and i didn't know if i should ignore this file or not. thanks @justinhartman for this PR!

brntbeer:
Yeah, as i explored in my own repo, `!*.xcodeproj/project.pbxproj` needs to be kept in this for sure.

shiftkey:
Thanks @justinhartman!

Coeur:
I'm not happy with this change.
The linked Stack Overflow thread is from 10 years ago, with mistakes and outdated workarounds: we shouldn't have the rule `*.xcodeproj/` at all.

PR #2569, opened 2018-01-04
New template file for Images

justinhartman:
There is no template to exclude images from a project. This file addresses that by excluding all
popular image formats for web, desktop and 3D.

N/A


shiftkey:
@justinhartman thanks!

PR #2571, opened 2018-01-04
New Template file for Iconic3

justinhartman:
There is currently no template file for Iconic3 and this file aims to change that.

Ionic have a pre-defined `.gitignore` file which formed the basis for this template. You can view the file [over here][2]. I have added extra items to the default Ionic3 files.

If this is a new template:  
- **Link to application or project's homepage**: [Iconic3 Framework][1]

[1]: https://ionicframework.com
[2]: https://github.com/ionic-team/ionic/blob/master/.gitignore

### DIFF
--- a/Ionic3.gitignore
+++ b/Ionic3.gitignore
@@ -1,0 +1,58 @@
+# Log and tmp data
+*~
+*.sw[mnpcod]
+*.log
+*.lock
+*.tmp
+*.tmp.*
+log.txt
+
+# NPM (Node Package Manager) files
+npm-debug.log*
+node_modules/
+tmp/
+temp/
+.package.tmp.json
+
+# Preprocessors cache files
+.idea/
+.sourcemaps/
+.vscode/
+.sass-cache/
+.versions/
+coverage/
+dist/
+.tmp/
+hooks/
+
+# Ionic 3 plugins files
+platforms/
+plugins/
+plugins/android.json
+plugins/ios.json
+www/
+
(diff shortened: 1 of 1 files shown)


Ionic3.gitignore
```diff
@@ -0,0 +1,58 @@
...
+hooks/
+
+# Ionic 3 plugins files
+platforms/
+plugins/
+plugins/android.json
+plugins/ios.json
+www/
+
+# Trash data
```
shiftkey: These rules are covered by existing OS-specific templates under the `Global` folder, and I want to avoid the maintenance work that comes with duplicating

Ionic3.gitignore
```diff
@@ -0,0 +1,58 @@
...
+
+# NPM (Node Package Manager) files
+npm-debug.log*
+node_modules/
+tmp/
+temp/
+.package.tmp.json
+
+# Preprocessors cache files
+.idea/
```
shiftkey: Many of these rules are actually IDE-specific rules, which we have defined under the `Global` subfolder. Please focus the template on what is important for using Ionic.

shiftkey:
I'm going to close this out because there are a lot of rules covered by existing templates in this repository, and I want to avoid duplication wherever possible.

@justinhartman if you feel like this should still be a listed template here, please read the section about [specialized templates](https://github.com/github/gitignore#specialized-templates) and open a new pull request that focuses on the rules specific to using Ionic.

PR #2572, opened 2018-01-04
Fixed the unignore of folder 'packages'

mbukejlovic:
Considering the VisualStudio gitignore file contains a line "**/[Pp]ackages/*", these 2 lines from Umbraco gitignore never get applied properly to the desired files, even though Umbraco lines were added after the VS lines. I guess the VS's line is more specific, so it gets a priority of some sort.


shiftkey:
@mbukejlovic thanks!

ImDavaloper:
Maybe it is not always the best idea to put the yarn.lock file in repository.

PR #2573, opened 2018-01-04
Add to *.lock

DavaEngineer:

_TODO_
Maybe it is not always the best idea to put the yarn.lock file in your repository.

_TODO_

If this is a new template: 

### DIFF
--- a/Node.gitignore
+++ b/Node.gitignore
@@ -1,6 +1,7 @@
 # Logs
 logs
 *.log
+*.lock
 npm-debug.log*
 yarn-debug.log*
 yarn-error.log*


luftywiranda13:
But, having lock files listed in `.gitignore` is harmful

https://github.com/luftywiranda13/remove-lockfiles#why
https://github.com/facebookincubator/create-react-app/pull/2014#issuecomment-300811661

shiftkey:
@ImDavaloper please provide more context for this change. As @luftywiranda13 mentioned, some lock files that should be committed will be caught by this rule...

shiftkey:
[From the docs](https://yarnpkg.com/lang/en/docs/yarn-lock/):

> All `yarn.lock` files should be checked into source control (e.g. git or mercurial). This allows Yarn to install the same exact dependency tree across all machines, whether it be your coworker’s laptop or a CI server.

PR #2575, opened 2018-01-06
Visual Studio 2017 puts the name of the project after Release and Debug

BenMcLean:
Ran into this while building the Darkplaces Quake engine in Visual Studio 2017 Community.

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -13,16 +13,18 @@
 *.userprefs
 
 # Build results
-[Dd]ebug/
-[Dd]ebugPublic/
-[Rr]elease/
-[Rr]eleases/
+[Dd]ebug*/
+[Dd]ebugPublic*/
+[Rr]elease*/
+[Rr]eleases*/
 x64/
 x86/
 bld/
 [Bb]in/
 [Oo]bj/
 [Ll]og/
+*.lib
+*.exp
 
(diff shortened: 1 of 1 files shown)


ButchDean:
Isn't this just something that you change in the project settings rather than VS2017 default behavior...

Project -> Properties -> Configuration Properties -> General -> Output Directory

?

shiftkey:
You can set anything to be the output paths by changing your project settings.

I don't want to accept PRs for every different permutation out there in the wild.

BenMcLean:
My understanding is that this is something VS 2017 does to all projects **by default**, and is different from previous versions of VS.

shiftkey:
@BenMcLean it's something that they've been doing for a while - see [the VS2015 project](https://github.com/xonotic/darkplaces/blob/master/darkplaces-wgl-vs2015.vcxproj#L73-L92) as well as [the VS2017 project](https://github.com/xonotic/darkplaces/blob/master/darkplaces-wgl-vs2017.vcxproj#L73-L92).

BenMcLean:
Oh so its not by default, you're probably right. But shouldn't we at least exclude exp and lib files?

shiftkey:
@BenMcLean I'm not clear on where those are coming from - if you can figure it out, please open a new PR.

PR #2576, opened 2018-01-11
Add .nvuser NVidia Nsight configuration file

jerzywie:
Nsight is a GPU and graphics debugging suite from NVidia. 

There is an edition that integrates with Visual Studio. User specific settings are saved in files with the .nvuser extension.


shiftkey:
@jerzywie thanks!

PR #2579, opened 2018-01-16
Update TeX.gitignore

meneguzzi:
Added a temporary file used in MacTeX

This is a temp file that always gets created when compiling tex files using in my installation of MacTeX (for a while now). I have checked and the same temp file does not appear in Windows or Linux.

I have not found any documentation describing this file, but it does get created in my installation.

If this is a new template: 

 - **Link to application or project’s homepage**:  https://www.tug.org/mactex/

### DIFF
--- a/TeX.gitignore
+++ b/TeX.gitignore
@@ -10,6 +10,7 @@
 *.fot
 *.cb
 *.cb2
+.*.lb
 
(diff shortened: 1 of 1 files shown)


shiftkey:
@meneguzzi thanks!

PR #2580, opened 2018-01-19
Add rptproj.rptuser

raterus:

The VS 2017 version of the Business Intelligence project contains a user file (*.rptproj.rptuser) that should be ignored as well.  This file is generated by the "Microsoft Reporting Services Projects" extension for visual studio.  https://marketplace.visualstudio.com/items?itemName=ProBITools.MicrosoftReportProjectsforVisualStudio

This was discussed on Stack Overflow (I wasn't involved), but this change was never requested.  https://stackoverflow.com/questions/47076716/which-reporting-services-file-types-are-safe-to-ignore-on-git

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -247,6 +247,7 @@ UpgradeLog*.htm
 *.rdl.data
 *.bim.layout
 *.bim_*.settings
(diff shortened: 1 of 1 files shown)


shiftkey:
@raterus thanks!

PR #2581, opened 2018-01-21
Update VisualStudio.gitignore

AngelosP:

The actual name of the folder which contains publish profiles is "PublishProfiles"

I don't have any documentation to link to, but I've confirmed this is the case with Visual Studio 2017

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -173,7 +173,7 @@ publish/
 # Microsoft Azure Web App publish settings. Comment the next line if you want to
 # checkin your Azure Web App publish settings, but sensitive information contained
 # in these scripts will be unencrypted
-PublishScripts/
+PublishProfiles/
 
 # NuGet Packages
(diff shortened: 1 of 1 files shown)


shiftkey:
> I don't have any documentation to link to, but I've confirmed this is the case with Visual Studio 2017

Could this have been something that's changed between major versions of VS? If so, I'd recommend making this an additive change rather than removing the old value.

shiftkey:
For example, this is the PR where the original rule came from: https://github.com/github/gitignore/pull/1919

frotaadriano:
I want to mute this thread and stop to receive e-mails. I was in link "mute
the thread" and fellow instructions and nothing.

2018-01-21 7:18 GMT-02:00 Angelos Petropoulos <notifications@github.com>:

shiftkey:
Closing due to inactivity. @AngelosP if you're able to provide more context I'm happy to revisit this.

PR #2582, opened 2018-01-22
Solidity/Truffle .gitignore

FelipeArgento:
There was no suggested file for either Truffle or Solidity. The added file is a standart .gitignore that helps a lot.

There is no official documentation supporting these gitignore files, they are what I've been using, based on official projects that use the same technology:

https://github.com/ethereum/go-ethereum/blob/master/.gitignore
https://github.com/ethereum/node-ethereum/blob/master/.gitignore

If this is a new template: 

https://github.com/trufflesuite/truffle
https://github.com/ethereum

### DIFF
--- a/SolidityTruffle.gitignore
+++ b/SolidityTruffle.gitignore
@@ -1,0 +1,24 @@
+# depedencies
+node_modules
+
+# testing
+coverage
+
+# production
+build
+build_webpack
+
+# misc
+.DS_Store
+.env
+npm-debug.log
+.truffle-solidity-loader
+.vagrant/**
+blockchain/geth/**
+blockchain/keystore/**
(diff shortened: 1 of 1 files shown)


shiftkey:

@kfinlay if you feel like this should still be a template listed here, please read the section about [specialized templates](https://github.com/github/gitignore#specialized-templates) and open a new pull request that follows the new pattern (it could live under `community/JavaScript/` as it looks like it shares a lot of Node-specific rules).

StEvUgnIn:
I agree with @shiftkey. You should edit your file and exclude most rules present in `node.gitignore` and put it in community repository instead.

Good idea, but I'm pretty sure Truffle generate `.gitignore` for you.

PR #2587, opened 2018-01-26
Added ignore for files created by `catkin_make_isolated`

felixdivo:

Simply execute `catkin_make_isolated` and `build` and `devel` will have different names, but should be ignored as well.

See [REP No. 134](http://www.ros.org/reps/rep-0134.html#catkin-make-isolated-options).


shiftkey:
@felixdivo thanks!

PR #2588, opened 2018-01-29
Add cache directories for different editors to the Unity gitignore

brendon-ai:

I use Unity with Visual Studio Code, and it's nice to not have the .vscode directory pushed to git. Same goes for people using Rider. Both of these are mentioned as supported code editors on the Unity blog.

https://blogs.unity3d.com/2018/01/05/discontinuing-support-for-monodevelop-unity-starting-in-unity-2018-1/


shiftkey:
@brendon-ai thanks, but I want to avoid duplicating rules which can easily fall out of sync.

For reference, [`Global\VisualStudioCode.gitignore`](https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore) and [`Global\JetBrains.gitignore`](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore) are more granular than these changes, and I'd recommend composing your own gitignore based on the technologies and tools you need.

PR #2591, opened 2018-01-31
Update Python.gitignore

buzzjmd:
pytest generates a .pytest_cache/ directory in the working directory where it is run. It was generated by a test failure when trying to import a module that didn't exist.

**Links to documentation supporting these rule changes:** 

I couldn't find any documentation about this.
 
 - **Link to application or project’s homepage**:  [pytest home page](https://docs.pytest.org)

PR #2594, opened 2018-02-04
Add ignore '.pytest_cache' directory in Unit test section

raimon49:

It added support to the specifications of the modified cache directory in the unit test tool pytest 3.4.0.

And I also saw #2591 discussion.

See also: [Pytest Changelog history](https://docs.pytest.org/en/latest/changelog.html)

> The default cache directory has been renamed from `.cache` to `.pytest_cache` after community feedback that the name `.cache` did not make it clear that it was used by pytest.

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -45,6 +45,7 @@ nosetests.xml
 coverage.xml
 *.cover
 .hypothesis/
+.pytest_cache/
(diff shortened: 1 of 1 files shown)


shiftkey:
@raimon49 thanks!

PR #2597, opened 2018-02-07
.idea/**/datasources.xml is now safe to store

DKroot:

Since IDEA 14 (2014), IDEA and other contemporaneous JebBrains IDEs don't store sensitive info in `datasources.xml` and it makes sense to version control `datasources.xml` for sharing project data sources among a team. `dataSources.local.xml` is user-specific: contain usernames and could be sensitive from a security perspective, albeit passwords could be externalized to e.g. Mac Keychain.

References: https://intellij-support.jetbrains.com/hc/en-us/articles/206544839, https://youtrack.jetbrains.com/issue/IDEA-127105

### DIFF
--- a/Global/JetBrains.gitignore
+++ b/Global/JetBrains.gitignore
@@ -9,7 +9,6 @@
 # Sensitive or high-churn files:
 .idea/**/dataSources/
 .idea/**/dataSources.ids
-.idea/**/dataSources.xml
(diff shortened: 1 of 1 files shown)


shiftkey:
> Since IDEA 14 (2014), ...

I think this is long enough to wait for holdouts. Thanks @DKroot!

PR #2599, opened 2018-02-09
*.pyc files are ignored

dishantrathi:
By default, the .pyc files are ignored and are not visible in the Project Window (IDE).

 "*.pyc" files are ignored by Python interpreter

https://www.jetbrains.com/help/pycharm/cleaning-pyc-files.html

If this is a new template: 

 - **Link to application or project’s homepage**: No Support

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -2,6 +2,7 @@
 __pycache__/
 *.py[cod]
 *$py.class
+*.pyc
 
 # C extensions
 *.so


shiftkey:
This shouldn't be necessary, as per https://github.com/github/gitignore/pull/2325#issuecomment-288318637 - can you confirm that rule is working for your situation?

dishantrathi:
@shiftkey You are right #2325 , Haven't checked that. ".py[cod]"

PR #2600, opened 2018-02-09
Added missing entries included by the dotnet/core project as well as …

CoreyScogin:
….DS_Store for Mac development.

dotnet/core includes a few extra entries. Include this as well.

Visual Studio development on a Mac is becoming more popular. Ignore .DS_Store.

https://github.com/dotnet/core/blob/master/.gitignore

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -53,6 +53,13 @@ project.lock.json
 project.fragment.lock.json
 artifacts/
 **/Properties/launchSettings.json
+.DS_Store
+*.swp
+*.*~
+
+
+# Visual Studio Code
+.vscode
 
 # StyleCop
 StyleCopReport.xml


srasku:
Some of these rules are already in [`macOS.gitignore`](https://github.com/github/gitignore/blob/master/Global/macOS.gitignore#L1-L2) and [`Vim.gitignore`](https://github.com/github/gitignore/blob/master/Global/Vim.gitignore#L1-L5).  Your tilde rule also already seems to be in the current [`VisualStudio.gitignore`](https://github.com/github/gitignore/blob/master/VisualStudio.gitignore#L214).  There also seems to be `.vscode` rules in [`VisualStudioCode.gitignore`](https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore).  I'm not familar with Visual Studio so I don't know if they're equivalent.

CoreyScogin:
I didn't realize there was a Global folder for OS entries.

Doesn't this mean that the VisualStudio.gitignore should be in the Global folder?
> The files in the root directory are for .gitignore templates that are project specific, such as language or framework specific templates. Global (operating system or editor specific) templates should go into the Global/ directory.


I didn't catch the `*~` would also match `*.*~` . I'll remove that.

Since there are rules for JetBrains Rider and other IDEs in [VisualStudio.gitignore](https://github.com/github/gitignore/blob/master/VisualStudio.gitignore#L214), it makes sense for the [VisualStudioCode.gitignore](VisualStudioCode.gitignore) entries to also be included.

shiftkey:
> Doesn't this mean that the VisualStudio.gitignore should be in the Global folder?

No, `VisualStudio.gitignore` covers a lot of different project and language options, so it's being handled uniquely until we can figure out a better way.

shiftkey:
> Since there are rules for JetBrains Rider and other IDEs in VisualStudio.gitignore, it makes sense for the VisualStudioCode.gitignore entries to also be included.

I'm going to push back on this for a couple of reasons:

 - the `JetBrains.gitignore` rules are more nuanced than what was merged (if the same PR were opened today, I wouldn't accept it)
 - Same thing for the `VisualStudioCode.gitignore` changes - the recommended `gitignore` rules are more specific than what's proposed here

Also, the more duplication of rules like this that occurs, the more easier it is for things to get out of sync.

PR #2602, opened 2018-02-12
Added .gitignore for 1C external data processors on unmanaged formes

darthunix:
This is a gitignore file for 1C external data processors on unmanaged formes versioned with [precommit1c](https://github.com/xDrivenDevelopment/precommit1c) git hook

### DIFF
--- a/1C.gitignore
+++ b/1C.gitignore
@@ -1,0 +1,6 @@
+# 1C external data processors on unmanaged formes
+
+**/Form/**/form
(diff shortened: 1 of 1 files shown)


shiftkey:

@darthunix if you feel like this should still be a template listed here, please read the section about [specialized templates](https://github.com/github/gitignore#specialized-templates) and open a new pull request that follows the new pattern.

PR #2604, opened 2018-02-13
FIRST Robotics Competition .gitignore

JamieSinn:

Support FIRST Robotics Competition Teams that use GitHub for version control. (Java and CPP - LabVIEW already has a supported GitIgnore - https://github.com/github/gitignore/blob/master/LabVIEW.gitignore)

Teams need a fast way to create a gitignore that is project agnostic and a quick reference during project creation.

https://github.com/github/gitignore/blob/master/Java.gitignore - Includes the contents of the /bin folder

If this is a new template: 

 - **Link to application or project’s homepage**: https://www.firstinspires.org/robotics/frc/

### DIFF
--- a/FRC.gitignore
+++ b/FRC.gitignore
@@ -1,0 +1,40 @@
+# WPILib
+bin/
+build/
+dist/
+sysProps.xml
+workbench.xmi
+
+# IDE
+*.ipr
+*.iml
+*.iws
+out/
+tmp/
+.settings/
+.loadpath
+.project
+/.idea
+.classpath
+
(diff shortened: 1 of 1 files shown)


shiftkey:
I'm going to close this out because the value of maintaining a template like this isn't clear to me. It looks like this duplicates a lot of existing templates we have in the repository.

@kfinlay if you feel like this should still be a template listed here, please read the section about [specialized templates](https://github.com/github/gitignore#specialized-templates) and open a new pull request that follows the new pattern.

PR #2605, opened 2018-02-13
Added *.ncrunchsolution

Pvlerick:
NCrunch solution specific file.

This file pattern is missing.


shiftkey:
NCrunch [recommends](https://www.ncrunch.net/documentation/considerations-and-constraints_using-ncrunch-with-source-control) the `.ncrunchsolution` and `.ncrunchproject` be versioned.

Pvlerick:
I would not have thought of versioning these, but if they say so... :-)

PR #2606, opened 2018-02-14
add yarn.lock

hailincai:

yarn.lock file is auto generated file, we don't need to put it in the git repository

The comment in the yarn.lock file said it is autogenerated

If this is a new template: 

### DIFF
--- a/Node.gitignore
+++ b/Node.gitignore
@@ -59,3 +59,7 @@ typings/
 
 # next.js build output
 .next
+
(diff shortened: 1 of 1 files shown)


shiftkey:
The `yarn.lock` file should be committed to version control, according to [the documentation](https://yarnpkg.com/en/docs/configuration#toc-use-yarn-lock-to-pin-your-dependencies).

PR #2608, opened 2018-02-22
Remove typings ignore

CKGrafico:
Nobody is using v1 of typings anymore on new projects and 'typings' is the recommended name for the folders of your custom types.
On the other hand the official Visual Studio ignore is not ignoring this folder.


shiftkey:
@CKGrafico thanks!

PR #2610, opened 2018-02-23
Add Pycharm settings

treussart:

For include settings files of Pycharm


shiftkey:

PR #2612, opened 2018-02-25
Ignored default sqlite database of Django

jtogrul:
Ignore default SQLite database created by Django framework for development and test purposes.

* Supporting that file name is correct: https://docs.djangoproject.com/en/2.0/intro/tutorial02/#database-setup
* Supporting that this file should be ignored: https://devcenter.heroku.com/articles/gitignore

If this is a new template: 

 - **Link to application or project’s homepage**: _Not a new template_

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -56,6 +56,7 @@ coverage.xml
 .static_storage/
 .media/
 local_settings.py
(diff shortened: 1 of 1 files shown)


shiftkey:
@jtogrul thanks!

PR #2613, opened 2018-02-25
Create Godot.gitignore

fahall:
A simple .gitignore for the Godot game engine.

 - **Project’s homepage**: [The Godot Engine](www.godotengine.org)
 - **Template**: [.gitignore example from godot demo project.](https://github.com/godotengine/godot-demo-projects/blob/master/.gitignore)

### DIFF
--- a/Godot.gitignore
+++ b/Godot.gitignore
@@ -1,0 +1,12 @@
+
+# Godot-specific ignores
+.import/
+export.cfg
+export_presets.cfg
+
+# Mono-specific ignores
+.mono/
+
+# System/tool-specific ignores
+.directory
+*~
--- a/Godot.gitignore
+++ b/Godot.gitignore
@@ -6,7 +6,3 @@ export_presets.cfg
 
 # Mono-specific ignores
 .mono/
-
-# System/tool-specific ignores
-.directory
-*~


aBARICHELLO:
Why ignore the .import files? They are important while cloning a project to prevent things like all PNG textures having filter on.

shiftkey:
@BastiaanOlij @akien-mga (who wrote the linked gitignore) - anything to add here?

akien-mga:
Seems fine, though should `# System/tool-specific ignores` be part of the Godot .gitignore? Those are not Godot-related, just standard files produced by users' file managers and IDEs.

shiftkey:
@akien-mga good point, and yes, this should be Godot-specific rules.

Godot.gitignore
```diff
@@ -0,0 +1,12 @@
...
+# Godot-specific ignores
+.import/
+export.cfg
+export_presets.cfg
+
+# Mono-specific ignores
+.mono/
+
+# System/tool-specific ignores
+.directory
```
shiftkey: This isn't specific to Godot - could you please remove it?

shiftkey:
@fahall thanks!

PR #2615, opened 2018-02-27
Adding Python bytecode file overlook

776948741:
**Adding Python bytecode file overlook**

_TODO_

**https://stackoverflow.com/questions/2998215/if-python-is-interpreted-what-are-pyc-files**

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -103,3 +103,6 @@ venv.bak/
 
 # mypy
 .mypy_cache/
+
+# py Bytecode
+*.pyc


lemcheyWhite:
Hi, looks like it was already added - https://github.com/776948741/gitignore/blob/1b120f876352da00200c2daff1f3f3173f603abc/Python.gitignore#L3

776948741:
Yes, I added this ignored item in the warehouse of my fork. There is no ignore item in the version of fork source. When PyC executes Python program, it will automatically generate for accelerating operation. It should be managed without pushing to the version library

shiftkey:
Closed as Line 3 should already cover this rule.

PR #2616, opened 2018-03-02
Added htlatex for TeX.gitignore

BlueNalgene:

I needed these temporary files ignored on one of my personal repos, and others might need it as well.  htlatex (the command from the program [TeX4ht](https://en.wikipedia.org/wiki/TeX4ht) is rather common.

These temporary files are undocumented by TeX4ht and do not serve any purpose once complete.  An [issue from 2011](http://tug.org/pipermail/tex4ht/2011q1/000274.html) was raised with the maintainers, but not implemented.  These files can be safely removed after the work is done.

If this is a new template: 

n/a - not a new template

### DIFF
--- a/TeX.gitignore
+++ b/TeX.gitignore
@@ -110,6 +110,14 @@ acs-*.bib
 *.gaux
 *.gtex
 
+# htlatex
+*.4ct
+*.4tc
+*.idv
+*.lg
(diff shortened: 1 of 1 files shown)


shiftkey:
@BlueNalgene thanks!

PR #2617, opened 2018-03-02
Dart.gitignore: Fix URL in comments

kevmoo:
Points to valid URI

The URL has been updated – pointing to latest

See that the old URI redirects to the new one


shiftkey:
@kevmoo thanks!

PR #2618, opened 2018-03-04
Keep `.vscode` from appearing in untracked files

icylace:

Whenever I ran `git status` in my project directory the `.vscode` subdirectory would erroneously show up in the untracked files list.

### DIFF
--- a/Global/VisualStudioCode.gitignore
+++ b/Global/VisualStudioCode.gitignore
@@ -1,3 +1,4 @@
+.vscode
 .vscode/*
 !.vscode/settings.json
 !.vscode/tasks.json


shiftkey:
@icylace the rules in this gitignore should cover the specific files that VS Code generates that shouldn't be tracked - I'm 👎 to disabling this globally.

icylace:
Yes, this gitignore does cover specific files but why not also cover the folder name itself as I'm suggesting ?

shiftkey:
@icylace because the existing rules should cover this - line 2 will ignore all files, and the lines below that will then track specific files that should be in version control

chrismcgahee:
Thank you for your information, Ron. I am an amateur to say the least but
can follow instructions, in most cases, quite well. Though I am rather tech
savvy to an extent, I feel as though I will be staying away from any
developments and code, etc. None the less, I thank you very much for your
emails and my apologies for not responding before now.

Regards,

Chris McGahee

chrismcgahee:
I've come to realize that my technical skills are not to the level of
riding toad, development, etc. I think all of you for your emails but will
be stepping back from this. Have a great day!

Regards,

Chris McGahee

shiftkey:
@icylace I believe the current template should cover what you need. Please try it out and open a new PR if there are any changes that need to be included.

PR #2619, opened 2018-03-06
ReactNative PR

Arshiamidos:
The OS that is not OS is not OS !!

### DIFF
--- a/ReactNative.gitignore
+++ b/ReactNative.gitignore
@@ -1,0 +1,54 @@
+# OSX
+#
+.DS_Store
+
+# Xcode
+#
+build/
+*.pbxuser
+!default.pbxuser
+*.mode1v3
+!default.mode1v3
+*.mode2v3
+!default.mode2v3
+*.perspectivev3
+!default.perspectivev3
+xcuserdata
+*.xccheckout
+*.moved-aside
+DerivedData
+*.hmap
+*.ipa
+*.xcuserstate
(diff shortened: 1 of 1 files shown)


shiftkey:
I can't take this in in it's current form, because it seems to duplicate a number of existing templates:

 - [Global/macOS.gitngore](https://github.com/github/gitignore/blob/master/Global/macOS.gitignore)
 - [Global/XCode.gitngore](https://github.com/github/gitignore/blob/master/Global/Xcode.gitignore)
 - [Node.gitngore](https://github.com/github/gitignore/blob/master/NodeJS.gitignore)
 - [Android.gitignore](https://github.com/github/gitignore/blob/master/Android.gitignore)

Buck is the only new rules that aren't captured, and might deserve it's own template. What I'd love to see in this template is the rules specific to ReactNative, that we can document as a reference for others to consume.

Arshiamidos:
can be merged now?

joeblau:
@Arshiamidos - We have this merged in at [@dvcs/gitignore](https://github.com/dvcs/gitignore).  We have the concept of a stack which supports template composition removing redundancies.


If you want to see what the template looks like you can check it out here:
- https://www.gitignore.io/api/reactnative.

You can also generate your template locally:
```sh
$ git clone git@github.com:dvcs/gitignore.git
$ cd gitignore/templates
$ cat ReactNative.*
```

PR #2622, opened 2018-03-07
Added an item in TeX.gitignore

bnuzhoub:
To ignore backup files generated by vim <https://www.vim.org>.

_TODO_

_TODO_

If this is a new template: 

### DIFF
--- a/TeX.gitignore
+++ b/TeX.gitignore
@@ -210,6 +210,9 @@ pythontex-files-*/
 TSWLatexianTemp*
 
 ## Editors:
+# vim
+*~
+
 # WinEdt
 *.bak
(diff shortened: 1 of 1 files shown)


vonzcy:
Do we really need to consider the temp file "~"? This is a quite general one.

bnuzhoub:
They are left by vim, existing and detected by git even if vim has been closed.  This is why I'd like to ignore them.

shiftkey:
Thanks, but given [`Global\Vim.gitignore`](https://github.com/github/gitignore/blob/master/Global/Vim.gitignore) already exists I want to avoid duplicating these rules in other templates.

PR #2623, opened 2018-03-09
Add .DS_Store to Node .gitignore

kareniel:
This file often finds itself in git repositories by mistake. Since the file is user-specific, it is safe to assume it won't be needed there.


shiftkey:

PR #2624, opened 2018-03-11
Kicad: Add more formats and documentation link

patrickelectric:

KiCad have more formats

[Link to documentation that support my insane changes.](http://kicad-pcb.org/help/file-formats/)

### DIFF
--- a/KiCad.gitignore
+++ b/KiCad.gitignore
@@ -1,4 +1,5 @@
 # For PCBs designed using KiCad: http://www.kicad-pcb.org/
+# Format documentation: http://kicad-pcb.org/help/file-formats/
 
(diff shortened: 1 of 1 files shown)


shiftkey:
@patrickelectric thanks!

adamheinrich:
Hi @patrickelectric, is it really a good idea to ignore cache lib? If someone opens the project on a machine with different libraries, it will be a disaster.

patrickelectric:
Theoretically, the components that aren't default in kicad, should exist as a submodule or a subdirectory of the projects, and you can include this libraries/folder using the kicad variable path relative to the .pro file.

PR #2625, opened 2018-03-13
Update documentation link cargo.toml vs cargo.lock

schnoddelbotz:

The old/outdated link is redirected, but not directly to the desired location.


shiftkey:
@schnoddelbotz thanks!

PR #2626, opened 2018-03-14
don't include the Go dependency management tool vendor folder.

skandix:
don't want commit the vendor folder as it contains the packages for the project when using the [`dep`](https://github.com/golang/dep) tool to lock the projects dependencies.

### DIFF
--- a/Go.gitignore
+++ b/Go.gitignore
@@ -10,3 +10,6 @@
 
 # Output of the go coverage tool, specifically when used with LiteIDE
 *.out
+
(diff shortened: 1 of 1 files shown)


shiftkey:
@skandix thanks for this, but it was removed in #2266 - please read that context and see if you agree with it.

shiftkey:
Closing this out as we had an earlier discussion about vendoring files in golang projects, and unless the situation has changed I'm happy to stick with that rule for now.

PR #2628, opened 2018-03-16
Update Swift.gitignore

doingself:
ignore .DS_Store in Mac OS

Ignore hidden files on Mac OS


doingself:
Thank you very much

PR #2629, opened 2018-03-18
Update .gitignore with .cpcache

tirkarthi:

The current clojure deps tool creates .cpcache directory in the project directory while using .deps.edn

https://clojure.org/reference/deps_and_cli

> Project directory
>     The current directory
>     Contents:
>           deps.edn - optional project deps
>           .cpcache - project cache directory, same as the user-level cache directory, created if there is a deps.edn

If this is a new template: 

* None


shiftkey:
@tirkarthi thanks!

PR #2631, opened 2018-03-19
allow .ssh directory

areyer:

My .ssh directory missed in my git repository as the vi-swapfile exclude was to greedy


shiftkey:
@areyer thanks!

PR #2632, opened 2018-03-20
Add IntelliJIDEA IDE gitignore

aderm:

Exclude IntelliJIDEA IDE project index from git on Mac pro machine.


shiftkey:
@aderm thanks, but see [`Global\JetBrains.gitignore`](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore)

PR #2637, opened 2018-03-21
Feat: Ignore eslint cache file

douglasgimli:

The JS linter tool `Eslint` creates a cache file that should be ignored when the `--cache`is added to the command.

https://eslint.org/docs/user-guide/command-line-interface#caching

### DIFF
--- a/Eslint.gitignore
+++ b/Eslint.gitignore
@@ -1,0 +1,1 @@
+.eslintcache


shiftkey:
I'm going to close this out because it's a trivial template for one specific tool, and ESLint already has some great documentation already around this to help users be aware of it.

PR #2638, opened 2018-03-21
Feat: Ignore stylelint cache file

douglasgimli:

The CSS linter tool `Stylelint` creates a cache file that should be ignored when the `--cache`is added to the command.

https://stylelint.io/user-guide/cli/

### DIFF
--- a/Stylelint.gitignore
+++ b/Stylelint.gitignore
@@ -1,0 +1,1 @@
+.stylelintcache


shiftkey:
I'm closing this out because it's a trivial template for one specific tool, which makes it too narrow for tracking and managing here.

PR #2642, opened 2018-03-26
added: *.iobj, *.ipdb

linlux:
"These files are produced when Incremental Link-Time Code Generation (LTCG) is enabled."
https://stackoverflow.com/questions/31554559/possible-to-stop-generating-ipdb-iobj-files-by-visual-studio-2015


shiftkey:
@linlux thanks!

PR #2643, opened 2018-03-28
Add exclude credential master key of Rails 5.2

chimame:

From Rails 5.2, the function to handle confidentiality is added.
A master key for restoring confidentiality is created, but it is better not to manage git.

- https://github.com/rails/rails/commit/69f976b859cae7f9d050152103da018b7f5dda6d
- https://github.com/rails/rails/commit/f27319a72a4ccfbffc575b752e4d91136f23725e#diff-24ae9dbe3383b08f36a08bfac6cf5fc7


shiftkey:
@chimame thanks!

PR #2644, opened 2018-03-28
Add the idea caches to gitignore

Terik23:

After updating to a new android studio, projects began to store some cache in the folder .idea/caches/
For example:

```
~/ExampleAndroidProject/.idea/caches
$ ls
build_file_checksums.ser
```

### DIFF
--- a/Android.gitignore
+++ b/Android.gitignore
@@ -39,6 +39,7 @@ captures/
 .idea/gradle.xml
 .idea/dictionaries
 .idea/libraries
+.idea/caches
 
 # Keystore files
 # Uncomment the following line if you do not want to check your keystore files in.


shiftkey:
@Terik23 any links to StackOverflow discussions or other documentation about this change?

5teveroy:
The only thing I find is here and it seems to support the idea of ignoring `build_file_checksums.ser`:
https://github.com/wh0/wh0.github.io/blob/master/_posts/2018-02-17-build-file-checksums-ser.md

The name "caches" suggests the directory should be ignored altogether, but I don't have any hard info on that.

shiftkey:
@Terik23 @5teveroy thanks for the details - that's good enough for me!

PR #2645, opened 2018-03-29
Update VisualStudio.gitignore

Duracell1989:
If you have a project named for example 'MyService.User'; the *.user will ignore this folder.
What you want; is to exclude all the *.csproj.user and *vb.proj.user files.


Skippern:
VisualStudio.gitignore probably should be moved to the Global subfolder, together with other OS and editor files?

Duracell1989:
Moved the VisualStudio.gitignore to the global folder.

shiftkey:
> VisualStudio.gitignore probably should be moved to the Global subfolder, together with other OS and editor files?

This change was not necessary, and will likely add more confusion. Some context: `VisualStudio.gitignore` was put in place to consolidate `CSharp.gitignore` and `VisualBasic.gitignore` into one template (the rules ended up being less language-specific and more tooling specific, and were common to both), and it's been kept at the root to help ensure people would find it and didn't creating new templates again.

It's the only exception I'm aware of to our existing conventions about `Global`, because of it's unique purpose of being a tool that services a broad ecosystem of tools and languages.

ArtureMsM:
A wonderful platform for innovation, creativity, progress, unite to achieve ones stupendous aspirations and goals in this age of e-digital-marketing, be it for business or for professional services; sky-is-the-limit!
00917900132499, 00912228103445

Duracell1989:
Reverted the VisualStudio.gitignore back to the main folder.
Added the vbproj.user files to the ignored files.

Duracell1989:
@shiftkey:  I think this PR can be merged?

PR #2646, opened 2018-03-30
JetBrains: add .idea/httpRequests

ikemo3:

This directory used by HTTP Client in IntelliJ IDEA (and other product), contains only requests history.

https://www.jetbrains.com/help/idea/http-client-in-product-code-editor.html#viewingResponse


shiftkey:
@ikemo3 thanks for linking to the docs for this feature!

PR #2647, opened 2018-03-30
Update Android.gitignore

leesah:
Ignore .idea/caches/build_file_checksums.ser

I noticed this today after upgrading to Android Studio 3.1. And the file name suggests it's checksums of some cached files and should not be checked in to git.

https://stackoverflow.com/questions/49557737/should-i-add-idea-caches-build-file-checksums-ser-to-gitignore


shiftkey:
@leesah thanks for this, but I just merged #2644 which was the same change

PR #2648, opened 2018-03-31
Update Android.gitignore

leesah:
Ignore .idea/assetWizardSettings.xml for Android Studio 3.1

This file was created when I create Image Asset or Vector Asset. It contains the full path to where my Android Studio is installed.

See [here](https://github.com/leesah/nirvana/blob/3c1204ab20a058b524f7e1bf81b7bc9e3046a865/.idea/assetWizardSettings.xml#L38).

I haven't found any.

### DIFF
--- a/Android.gitignore
+++ b/Android.gitignore
@@ -37,6 +37,7 @@ captures/
 .idea/workspace.xml
 .idea/tasks.xml
 .idea/gradle.xml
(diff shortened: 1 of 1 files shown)


kevinc123:
Why not ignore complete .idea folder??

shiftkey:
@leesah thanks!

PR #2651, opened 2018-04-05
add .history folder to gitignore

samuelstein:
Ignore local history changes from .history folder

### DIFF
--- a/Global/VisualStudioCode.gitignore
+++ b/Global/VisualStudioCode.gitignore
@@ -3,3 +3,4 @@
 !.vscode/tasks.json
 !.vscode/launch.json
 !.vscode/extensions.json
+.history


samuelstein:
sure: https://dailydotnettips.com/2015/05/04/how-to-navigate-through-history-of-files-in-visual-studio-code/. It's a relatively new feature of vs code.

Global/VisualStudioCode.gitignore
```diff
@@ -3,3 +3,4 @@
 !.vscode/tasks.json
 !.vscode/launch.json
 !.vscode/extensions.json
+.history
```
shiftkey: Shouldn't this be inside `.vscode/`? If so, it should be covered by the rule on line 1

samuelstein: no the folder is outside of .vscode

shiftkey: I'd like to see some documentation that explains this file better - describing the feature is one thing, but it'd be nice to confirm we're looking at the right file in the right location before merging this in.

It also seems suboptimal for this to not be using the `.vscode` folder, especially if it's an in-the-box feature of VSCode, for what looks like a generated file.

sara0871:
👋

samuelstein:
i could not find better official documentation. @shiftkey try it for yourself. got to user settings and activate it via `"local-history.enabled": 1,`

shiftkey:
On VSCode 1.22.2 I don't seem to be able to enable this feature:

shiftkey:
Closing this out as I'm not able to understand the requested change and verify it's correct

PR #2652, opened 2018-04-05
Add shelf to JetBrains.gitignore

twhitbeck:

.idea/shelf is a directory of user-specific stuff, should not be committed.

Can't find any, but these files are obviously personal files


shiftkey:
> . I couldn't seem to find the commit that reverted it.

You'll want to look at #1475 which was aligning the template with the JetBrains docs

twhitbeck:
Ok yes I agree that the docs don't explicitly mention `.idea/shelf`, but I still think it should be included. There are lots of things in this file which are not listed in the docs.

PR #2654, opened 2018-04-06
Added system files in gitignore

aamir-nazir:
Most of the times we don't need to track system files in git repository.


shiftkey:

PR #2655, opened 2018-04-06
Create StandardML.git

k4rtik:

Add a gitignore for Standard ML

https://www.smlnj.org/doc/CM/index.html

The documentation doesn't explicitly mention this as most SML/NJ developers have been using Subversion. But this can be found in most SML projects on GitHub, e.g. https://github.com/MLton/mlton/blob/master/lib/.gitignore

If this is a new template: 

 - **Link to application or project’s homepage**: http://smlnj.org/

### DIFF
--- a/StandardML.git
+++ b/StandardML.git
@@ -1,0 +1,2 @@
+## Ignore all SML/NJ Compiler Manager build directories
+.cm/


shiftkey:

@k4rtik if you feel like this should still be a template listed here, please read the section about [specialized templates](https://github.com/github/gitignore#specialized-templates) and open a new pull request that follows the new format.

PR #2657, opened 2018-04-07
Python.gitignore improvements

akelsch:

All commits one by one:
1. Reorder rules in groups by rule count and alphabetically within groups
2. Capitalization and trimming of unnecessary long comments
3. I don't get a `wheels` directory with Python 3.6.5, It's `python-wheels` instead
4. `lib64` is a symlink (Linux) or junction (Windows), not a directory

I can provide some terminal output:

```bash
user@host:~$ python --version
Python 3.6.5
user@host:~$ python -m venv gitignore
user@host:~$ ls -l gitignore/share/
total 0
drwxrwxrwx 0 user user 512 Apr  7 22:05 python-wheels
```

```bash
user@host:~$ ls -l gitignore/
total 0
drwxrwxrwx 0 user user 512 Apr  7 22:05 bin
drwxrwxrwx 0 user user 512 Apr  7 22:05 include
drwxrwxrwx 0 user user 512 Apr  7 22:05 lib
lrwxrwxrwx 1 user user   3 Apr  7 22:05 lib64 -> lib
-rw-rw-rw- 1 user user  69 Apr  7 22:05 pyvenv.cfg
drwxrwxrwx 0 user user 512 Apr  7 22:05 share
```

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -10,7 +10,7 @@ dist/
 downloads/
 eggs/
 lib/
-lib64/
+lib64
 MANIFEST
 parts/
 python-wheels/


akelsch:
Does the sorting/reordering hold you back from merging @shiftkey? Or is it something else?

I really would love to see this merged because I regularly get my gitignores from this repo.

shiftkey:
> Does the sorting/reordering hold you back from merging @shiftkey? Or is it something else?

Lack of available time to look at it is my current constraint

akelsch:
Alright, I thought there is something wrong with my PR.

I'll be patient then :)

shiftkey:
@akelsch this seems fine. Some extra questions:

> 3. I don't get a `wheels` directory with Python 3.6.5, It's `python-wheels` instead

Do you know how far back this change was? If it's a recent one, we should make this an addition to the existing rule, for users who aren't yet on the version with this change.

> 4. `lib64` is a symlink (Linux) or junction (Windows), not a directory

We're talking about this in #2703 but I think your work here to just change that rule is what I prefer

akelsch:
> Do you know how far back this change was? If it's a recent one, we should make this an addition to the existing rule, for users who aren't yet on the version with this change.

I wasn't able to find something about `python-wheels` in the cpython, pip and wheel repositories. There are also no release notes on this.

But apparently this rule affects Debian (and thus Ubuntu) only:
> When these binary packages are installed, `*.whl` files must be placed in the `/usr/share/python-wheels` directory. **The location inside a virtual environment will be rooted in the virtual environment, instead of `/usr`.**

Source: https://www.debian.org/doc/packaging-manuals/python-policy/ch-module_packages.html

akelsch:
Anything new on this @shiftkey?

akelsch:
I guess this PR is dead, especially since there are merge conflicts now 😕

PR #2658, opened 2018-04-08
maven-wrapper.jar can be removed from the source

gzm55:

remove the exceptions for maven-wrapper.jar

https://github.com/takari/maven-wrapper#usage-without-binary-jar

### DIFF
--- a/Maven.gitignore
+++ b/Maven.gitignore
@@ -7,6 +7,3 @@ release.properties
 dependency-reduced-pom.xml
 buildNumber.properties
 .mvn/timing.properties
-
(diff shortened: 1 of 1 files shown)


shiftkey:
@gzm55 I'm not quite clear on this change:

 - the original rule was about ensuring the file is committed
 - If that's no longer needed then the `!` can be dropped and the file will be ignored
 - Removing the rule means the file may still be tracked by Git

Can you clarify what should be happening here?

gzm55:
@shiftkey another commit to ignore the wrapper jar, the Git should no longer track this file.

shiftkey:
@gzm55 thanks!

PR #2662, opened 2018-04-13
[Fastlane] Specify screenshot files only, rather than whole folder (in case the user has frameit config)

juliancadi:

The whole screenshots folder shouldn't be ignore since there could be frameit configuration in it (Framefile.json, fonts, background.jpg, *.strings, etc...). It'd be better to just ignore the generated PNG screenshot files.

https://docs.fastlane.tools/actions/frameit/#usage

### DIFF
--- a/Swift.gitignore
+++ b/Swift.gitignore
@@ -64,5 +64,5 @@ Carthage/Build
 
 fastlane/report.xml
 fastlane/Preview.html
-fastlane/screenshots
+fastlane/screenshots/*/*.png
 fastlane/test_output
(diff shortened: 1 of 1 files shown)


Swift.gitignore
```diff
@@ -64,5 +64,5 @@ Carthage/Build
 
 fastlane/report.xml
 fastlane/Preview.html
-fastlane/screenshots
+fastlane/screenshots/*/*.png
```
shiftkey: Should this be `**/*.png`?

juliancadi: That's right!

shiftkey:
@juliancadi thanks!

PR #2663, opened 2018-04-13
[TeX] Ignore generated *.xmpi files

koppor:

Package [xmpincl](https://www.ctan.org/pkg/xmpincl) generates `*.xmpi` files out of `*.xmp` files during the inclusion process.

https://texdoc.net/pkg/xmpincl -> page 3:

> basically create a new file .xmpi


shiftkey:
@koppor thanks!

PR #2664, opened 2018-04-16
Visual studio cache issue

kilian442:
/.vs/ do not catch the visual studio cache folder but .vs/ do

_TODO_

_TODO_

If this is a new template: 


shiftkey:
> `/.vs/` do not catch the visual studio cache folder but `.vs/` do

This is because a leading `/` forces the check to only ignore from the root folder. I suspect you have a nested solution and the `.vs` directory being generated in a nested directory.

Unity.gitignore
```diff
@@ -6,7 +6,7 @@
 Assets/AssetStoreTools*
 
 # Visual Studio cache directory
-/.vs/
+.vs/
```
shiftkey: This is now consistent with `VisualStudio.gitignore`

PR #2665, opened 2018-04-16
Added gitignore for django

null-none:


null-none:
Added gitignore for django

shiftkey:
@null-none our current `Python.gitignore`  already contains Django-specific rules:

https://github.com/github/gitignore/blob/4b37ac26ffaee3e6566723ea187202d1e44ebb52/Python.gitignore#L54-L57

Could anything missing from that file be added there, rather than creating a new file?

shiftkey:
Closing this out as I believe the Django-specific rules should be added to `Python.gitignore`

PR #2666, opened 2018-04-16
[Fastlane] Specify screenshot files only, rather than whole folder (i…

juliancadi:

Already done and merged for Swift.gitignore (#2662). It also applies to Objective-C projects, not to Android projects though. Therefore it wasn't changed on Android.gitignore file.

https://docs.fastlane.tools/actions/frameit/#usage


shiftkey:
@juliancadi thanks!

PR #2674, opened 2018-04-24
Rust - Recursively ignore target directory

thomaseizinger:

Repositories with Rust code often have multiple modules in them. Those modules have a `target` folder themselves and can be arbitrarily nested.

For example:

```
.
├── jsonrpc
│   ├── src
│   ├── target
│   ├── v1
│   │   ├── src
│   │   └── target
│   └── v2
│       ├── src
│       └── target
├── src
│   └── bin
└── target
```

This rules makes sure that all of these `target` directories are ignored.

This is just the way `cargo` works.


thomaseizinger:
This was partly a misunderstanding on my part. If the workspace is correctly configured, `cargo` only uses 1 output directory.

PR #2675, opened 2018-04-25
Update Android.gitignore

rogerhu:
I don't think the .idea should ever be checked in.

Noticed it was adding other .idea files.  Since Gradle has been used, it is unnecessary.

https://stackoverflow.com/a/26290130

### DIFF
--- a/Android.gitignore
+++ b/Android.gitignore
@@ -34,6 +34,7 @@ captures/
 
 # IntelliJ
 *.iml
+.idea
 .idea/workspace.xml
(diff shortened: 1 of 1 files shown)


drothmaler:
Yes it should, because it also contains stuff like code formatter settings, that you **do** want to share, with your team...

shiftkey:
Please refer to the JetBrains documentation about this: https://intellij-support.jetbrains.com/hc/en-us/articles/206544839

PR #2676, opened 2018-04-25
Create SilverStripe.gitignore

zanderwar:
Adding SilverStripe default .gitignore

https://silverstripe.org/download
https://github.com/silverstripe


shiftkey:

@zanderwar if you feel like this should still be a template listed here, please read the section about [specialized templates](https://github.com/github/gitignore#specialized-templates) and open a new pull request that follows the new pattern.

PR #2679, opened 2018-04-25
Add Python venv and Django settings

kraigb:
For improving experience with PTVS.
Prevents default venvs from being added to source control with project creation.

Entries are copied directly from python.gitignore, https://github.com/github/gitignore/blob/master/Python.gitignore.

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -297,6 +297,19 @@ paket-files/
 __pycache__/
 *.pyc
 
+# Python virtual environments
+.env
+.venv
+env/
+venv/
+ENV/
+env.bak/
+venv.bak/
+
+# Python/Django
+local_settings.py
+db.sqlite3
+
 # Cake - Uncomment if you are using it
 # tools/**
 # !tools/packages.config


kraigb:
@qubitron for review

qubitron:
LGTM!

VisualStudio.gitignore
```diff
@@ -297,6 +297,19 @@ paket-files/
 __pycache__/
 *.pyc
 
+# Python virtual environments
```
jayvdb: Why not use `Python.gitignore` which has these rules?

kraigb: When a new project is created in Visual Studio with an option selected for creating a git repository, VS specifically uses VisualStudio.gitignore to initialize the .gitignore for that repo, regardless of the language used in the project. (More specifically, each release of VS copies VisualStudio.gitignore during its build process, and that file is what's used for .gitignore during new project creation.) Because VS does not provide a user with the option to select a .gitignore base or to customize it, and because we want to exclude venvs from new Python projects by default, we need to have this change in VisualStudio.gitignore. Does that make sense? 

As a further background, we're trying to address a behavioral issue with VS and Python when creating a git repository when creating a project. By default, VS is set up to do an auto-commit of the project after it's created. However, creating a project from a Python template in VS also typically creates a venv at the same time, which ends up being committed unnecessarily to source control. The steps to remove it from source control requires one to use CLI git commands, rather than being able to remain in VS. 

There are cumbersome workarounds for this, but the simplest answer is this small modification to VisualStudio.gitignore so that we get the right behavior by default.

jayvdb: Interesting.  And IMO, that situation makes it worth the duplication.  Perhaps document the reason for it as a comment in the source file.

kraigb:
@jayvdb Added a comment per your request. Are you able to merge the change? We're hoping to have it in place before the next Visual Studio release (15.8).

kraigb:
Who has the ability to merge this change?

shiftkey:
@kraigb I do, but I'm currently occupied with moving countries at the moment so I haven't been able to keep up with reviews. As I wrote over in https://github.com/github/gitignore/pull/2720#issuecomment-392967907 I have some concerns about duplicating rules in `VisualStudio.gitignore`.

PR #2685, opened 2018-05-01
Update VisualStudio.gitignore

jfversluis:

Visual Studio Code is part of the family now and creates its own working folder, for supporting multiple root spaces since version 1.1.8. It would be nice to keep these files out of Git.

https://code.visualstudio.com/docs/editor/multi-root-workspaces#_settings

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -327,3 +327,6 @@ ASALocalRun/
 
 # MFractors (Xamarin productivity tool) working folder 
 .mfractor/
+
+# Visual Studio Code working folder
+.vscode/** 


drothmaler:
Thanks, but please see `Global/VisualStudioCode.gitignore` for VSCode...

jfversluis:
Ah, I found it was odd there was nothing for it yet. Thanks!

But how does this work? I mean, it isn't very unlikely that Visual Studio (full) and Code are used in the same project right? So, wouldn't it make sense to incorporate it in the `VisualStudio.gitignore` as well?

shiftkey:
> So, wouldn't it make sense to incorporate it in the `VisualStudio.gitignore` as well?

`VisualStudio.gitignore` has accrued a rich history of rules due to the myriad of technologies and extensions that it has supported over the years. I would prefer to keep `VisualStudioCode.gitigore` focused on just what VSCode needs and supports.

Also please note that there are a number of files under the `.vscode` directory that can be safely versioned.

jfversluis:
Fair enough! Thanks!

PR #2686, opened 2018-05-01
Matlab - Exclude coder temporary folders

14roiron:

When using the matlab coder to generate C/C++/Mex files, matlab generate a lot of files that should not be committed, including documentation.
For the C sources codes, for me they should not be commited by defaut, as it depends on the architectures/toolbox versions... 

https://www.mathworks.com/examples/matlab-coder/mw/coder_product-coderdemo_hello_world-hello-world

### DIFF
--- a/Global/Matlab.gitignore
+++ b/Global/Matlab.gitignore
@@ -18,6 +18,9 @@ helpsearch*/
 slprj/
 sccprj/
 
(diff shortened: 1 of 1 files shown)


shiftkey:
@14roiron thanks!

PR #2688, opened 2018-05-02
Ignore Reporting Project Upgrade backup files

nschonni:

I took a look at #1887 but settled on a more directed fix for what I'm hitting right now.
The conversion process from the old custom rptproj to the MSBuild based projects creates a MYProject.rptproj.bak backup file

This doesn't call out the backup directly, but is the process that is creating these files https://blogs.msdn.microsoft.com/sqlrsteamblog/2017/09/25/msbuild-support-for-reporting-services-projects-now-available/


shiftkey:
@nschonni thanks!

PR #2689, opened 2018-05-03
Create CRBasic.gitignore

patricktokeeffe:
Add generic `.gitignore` for Campbell Scientific's CRBasic language. There is no central homepage for the language, but documentation is publicly available ([.pdf](https://s.campbellsci.com/documents/us/manuals/loggernet.pdf)) and there is a website (sales page) for [the IDE](https://www.campbellsci.com/crbasiceditor).

### DIFF
--- a/CRBasic.gitignore
+++ b/CRBasic.gitignore
@@ -1,0 +1,5 @@
+# optional table definintions created at compile time
+*.tdf
+
+# any encrypted files
+*_Enc.*


shiftkey:

@patricktokeeffe if you feel like this should still be a template listed here, please read the section about [specialized templates](https://github.com/github/gitignore#specialized-templates) and open a new pull request that follows the new pattern.

PR #2692, opened 2018-05-05
Create Flutter.gitignore

saschpe:
Ignore rules for the Flutter SDK (see flutter.io). Based of the default
.gitignore rule generated for new projects. Does not include
rules already covered by Dart.gitignore.

Google's Flutter SDK needs custom ignore rules on top of Dart.gitignore

https://github.com/flutter/flutter/blob/master/.gitignore

If this is a new template: 

 - **Link to application or project’s homepage**: https://flutter.io/

### DIFF
--- a/Flutter.gitignore
+++ b/Flutter.gitignore
@@ -1,0 +1,5 @@
+.flutter-plugins
+
+# Firebase / Google Cloud Services
+android/*/google-services.json
+ios/*/GoogleService-Info.plist


shiftkey:
> Does not include rules already covered by Dart.gitignore.

@saschpe correct me if I'm wrong here, but given how Flutter is coupled to Dart I think it's worth keeping those rules together - if you were to create a new Flutter project using just these rules you might find yourself committing the wrong files from the beginning.

Two potential solutions come to mind:

 - add a section for Flutter to `Dart.gitignore` - this simplifies maintenance, might affect discoverability, but reflects that Flutter is part of the Dart ecosystem
 - duplicate the rules from `Dart.gitignore` into `Flutter.gitignore` - this is preferred if you anticipate significantly more distinct rules for Flutter, and want to avoid churning the file significantly

saschpe:
@shiftkey There are other popular Dart projects around, most notably [AngularDart](https://webdev-dartlang-org-dev.firebaseapp.com/), that would yield another section in Dart.gitignore soon as well. I don't have a strong opinion in either direction...

@RinaFave01 huh?

shiftkey:
I'm going to close this out due to inactivity, but based on #2857 I'm interested in:

 - an update to `Dart.gitignore` to add Flutter-specific rules (opt-in or enabled by default)
 - a `community/Dart/Flutter.gitignore` template that refers to the main `Dart.gitignore` template

PR #2695, opened 2018-05-10
Added jigsaw template

RohanSakhale:

Jigsaw is a Static Site Generator build on top of Laravel5 PHP Framework

https://github.com/tightenco/jigsaw/blob/master/stubs/site/.gitignore

If this is a new template: 

https://jigsaw.tighten.co/

### DIFF
--- a/Jigsaw.gitignore
+++ b/Jigsaw.gitignore
@@ -1,8 +1,5 @@
 # Jigsaw Static Site Generator
 
-# Jigsaw specific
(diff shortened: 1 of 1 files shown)


shiftkey:

@RohanSakhale if you feel like this should still be a template listed here, please read the section about [specialized templates](https://github.com/github/gitignore#specialized-templates) and open a new pull request that follows the new pattern.

PR #2697, opened 2018-05-11
Cases for Jetbrains Rider's with Unity3d

PalashBansal96:

Cases for Jetbrains Rider's user specific stuff
For people using JetBrains Rider as a replacement for Visual Studio in Unity3D


shiftkey:

PR #2698, opened 2018-05-12
Add latexrun default directory for auxiliary files

eelregit:

Add the default directory where latexrun outputs all intermediate files.
Also add a new section for possible future latex build tools using similar mechanisms.

The author haven't documented this, or at least I couldn't find it.
Below is the line that specifies this default build path
https://github.com/aclements/latexrun/blob/38ff6ec2815654513c91f64bdf2a5760c85da26e/latexrun#L90

### DIFF
--- a/TeX.gitignore
+++ b/TeX.gitignore
@@ -40,6 +40,10 @@
 *.synctex.gz(busy)
 *.pdfsync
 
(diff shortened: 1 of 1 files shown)


shiftkey:
@eelregit thanks!

PR #2700, opened 2018-05-15
Update VisualStudio.gitignore - Add visual studio code

Noctis-:

Seems like Visual Studio Code creates a folder when you open the folder with it.


shiftkey:

PR #2701, opened 2018-05-16
add  .DS_Store  .xceserstate  in Objective-C.gitignore

jinkulishi:
Those two files just like two flies when I use git to commit. so I try to  add it into the lastest Objective-C.gitignore,then everything is calm.

### DIFF
--- a/Objective-C.gitignore
+++ b/Objective-C.gitignore
@@ -28,6 +28,11 @@ xcuserdata/
 *.dSYM.zip
 *.dSYM
 
+##File .DS_Store
+*/.DS_Store
+/.DS_Store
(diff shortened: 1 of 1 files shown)


mys721tx:
Wouldn't it be better to use global gitignore for [Xcode](https://github.com/github/gitignore/blob/master/Global/Xcode.gitignore) and [macOS](https://github.com/github/gitignore/blob/master/Global/macOS.gitignore) instead?

shiftkey:
Please review the current rules in `Global/XCode.gitignore` and `Global/macOS.gitignore` because I think those templates are what you need to be adding to your projects.

PR #2702, opened 2018-05-16
Added ignore for Jetbrains Rider

GoodPie:

Rider is becoming more and more popular with Unity developers, especially with the latest changes to the Unity plugin. Adding the idea cache folder to the ignore provides convenience in the same way ignoring the visual studio cache does.

Not really much to support but with the [continued Unity integration](https://www.jetbrains.com/rider/whatsnew/) improvement, it is worth adding.


shiftkey:
This is a bit different to the VS temporary directory. 
We have a detailed set of rules for all the JetBrains IDES over in [`Global\JetBrains.gitignore`](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore), which I recommend adding to your repository .gitignore.

PR #2703, opened 2018-05-16
Remove directory slashes.

monokrome:
These are unnecessary and can lead to issues. For instance, a user may use Python 3 to create a virtualenv such as:

```sh
python3 -m venv .
```

The problem is that, in this case, Python 3 creates a symlink (lib64 -> lib) which isn't ignored based on these rules. Removing directory slashes solves the problem.

Another potential solution would be to specifically ignored files within that directory (since git doesn't recognize directories, just files) - which I'm happy to change this to.

Hopefully others find this helpful! :smiley_cat:

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -8,19 +8,19 @@ __pycache__/
 
 # Distribution / packaging
 .Python
-build/
-develop-eggs/
-dist/
-downloads/
-eggs/
-.eggs/
-lib/
-lib64/
-parts/
-sdist/
-var/
-wheels/
-*.egg-info/
+build
+develop-eggs
+dist
+downloads
+eggs
+.eggs
+lib
+lib64
+parts
+sdist
+var
+wheels
+*.egg-info
 .installed.cfg
 *.egg
 MANIFEST


jayvdb:
I think it is much clearer to have the slashes, and they are used extensively.

```console
$ git grep '^[^#].*\/$' | wc -l
417
```

If there are specific problems with some of these, then it would be good to find another way to achieve the same result, or only remove the `/` where necessary and include a comment explaining why.

shiftkey:
This feels like a big hammer to address the issue, and I think it's important to clarify rules that should target directories. From the [Git SCM docs](https://git-scm.com/docs/gitignore#_pattern_format) (emphasis mine):

> If the pattern ends with a slash, it is removed for the purpose of the following description, but it would only find a match with a directory. In other words, `foo/` will match a directory `foo` and paths underneath it, but **will not match a regular file or a symbolic link `foo` (this is consistent with the way how pathspec works in general in Git)**.

I'm fine with addressing the issue with symlinks not being handled correctly by this template, if we want to simplify the changes down to that. @monokrome can you confirm which of these rules actually need to be applied to symlinks and directories?

PR #2705, opened 2018-05-19
Do not ignore launchsettings.json for VS

giggio:

Ignoring `launchSettings.json` does not make much sense. Now .NET CLI even considers this file when running with `dotnet run`, as you can read [here](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-run?tabs=netcore2x).

This settings will be useful if shared among project members, so it should be commited to the repo.

Also, on the default `.gitignore` file generated by Visual Studio it is not ignored, so this causes confusion, as depending on how `.gitignore` was created it could be commited or not.

https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-run?tabs=netcore2x

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -52,7 +52,6 @@ BenchmarkDotNet.Artifacts/
 project.lock.json
 project.fragment.lock.json
 artifacts/
-**/Properties/launchSettings.json
 
 # StyleCop
 StyleCopReport.xml


shiftkey:
@giggio thanks for the feedback.

This was added in https://github.com/github/gitignore/pull/2141 but there was some discussion at the time that this might be useful. cc @NTaylorMullen to confirm this is A Good Thing™ to do now.

shiftkey:
Thanks all!

sven5:
Just a question: How do you advice to have a shared launchsettings.json with different Environment for each developer?

shiftkey:
@sven5 I'd recommend asking a question on StackOverflow: https://stackoverflow.com/questions/tagged/visual-studio-code

monokrome:
Seems like a disappointing non-answer from @shiftkey on this one for @sven5. This is a major oversight of not ignoring this. Assuming that everyone's environment is exactly the same is surely going to cause problems in the future.

PR #2708, opened 2018-05-21
add .idea/

cloisonne:

add .idea/ which was PyCharm configuration files.

https://stackoverflow.com/questions/9237345/how-do-i-ignore-pycharm-configuration-files-in-a-git-repository

If this is a new template: 


shiftkey:

PR #2710, opened 2018-05-23
Ignore */.DS_Store

SConaway:

Ignore */.DS_Store

_TODO_

If this is a new template: 


shiftkey:

PR #2711, opened 2018-05-24
Add all Sass sourcemap file formats

rootbeersoup:

This ignores the other Sass/SCSS sourcemap formats `sass` creates. If `*.css.map` is ignored by default, it stands to reason a user would probably want to exclude `*.sass.map` and `*.scss.map` as well.


shiftkey:
@rootbeersoup thanks!

PR #2714, opened 2018-05-25
Ignore Terraform crash log files

mastertinner:

Terraform creates crash logs when it crashes. These should be ignored.

https://github.com/hashicorp/terraform/blob/master/.gitignore


shiftkey:
The change introducing this rule https://github.com/hashicorp/terraform/commit/9650542e3d1fe87011688f046061f05cdbb282ed doesn't really add much context.

Got an example of what the error files look like when they're generated so we can track _why_ we're ignoring `*.log` here?

shiftkey:
@mastertinner can we be explicit then and name the new rule `crash.log`?

mastertinner:
Sure. Makes sense. I changed it.

PR #2715, opened 2018-05-25
Add Rider IDE-specific information

julia-vaseva:

Cover JetBrains Rider IDE

Here is a reference to Rider knowledge base article: https://rider-support.jetbrains.com/hc/en-us/articles/207097529

### DIFF
--- a/Global/JetBrains.gitignore
+++ b/Global/JetBrains.gitignore
@@ -1,4 +1,4 @@
-# Covers JetBrains IDEs: IntelliJ, RubyMine, PhpStorm, AppCode, PyCharm, CLion, Android Studio and WebStorm
+# Covers JetBrains IDEs: IntelliJ, RubyMine, PhpStorm, AppCode, PyCharm, CLion, Android Studio, WebStorm and Rider
 # Reference: https://intellij-support.jetbrains.com/hc/en-us/articles/206544839
 
 # User-specific stuff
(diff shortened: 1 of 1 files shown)


Global/JetBrains.gitignore
```diff
@@ -7,6 +7,13 @@
 .idea/**/dictionaries
 .idea/**/shelf
 
+# Rider IDE-specific stuff
+# Reference: https://rider-support.jetbrains.com/hc/en-us/articles/207097529
+.idea/**/modules.xml
+.idea/**/contentModel.xml
+*.iml
```
mihaipopescu: What about `.idea/**/*.iml` instead? Is there any iml file generated outside of .idea folder?

shiftkey: I'm 👎 to this rule because `*.iml` files are considered share-able in the reference link on line 2:

> This format is used by all the recent IDE versions by default. Here is what you need to share:
>
> - All the files under .idea directory in the project root except the **workspace.xml**, **usage.statistics.xml**, and **tasks.xml** files which store user specific settings

Global/JetBrains.gitignore
```diff
@@ -7,6 +7,13 @@
...
+*.DotSettings.user
```
shiftkey: This is a ReSharper file and can live in it's own section

Global/JetBrains.gitignore
```diff
@@ -7,6 +7,13 @@
 .idea/**/dictionaries
 .idea/**/shelf
 
+# Rider IDE-specific stuff
+# Reference: https://rider-support.jetbrains.com/hc/en-us/articles/207097529
+.idea/**/modules.xml
+.idea/**/contentModel.xml
```
shiftkey: Got a source for this rule?

shiftkey:
Closing this out due to inactivity (and conflicts). @julia-vaseva if you're interested in seeing this land please respond to the feedback and we can revisit it.

PR #2717, opened 2018-05-27
Add .vscode/ directory to Unity.gitignore

glowurms:

As more people start using VS Code for C# with Unity, this ignore will be more convenient to have as a default. The `.vscode/` directory is created and used by the VS Code editor and not necessary for a Unity Project or simple scripts to be used with Unity.


shiftkey:

PR #2718, opened 2018-05-27
Update Dart.gitignore

bcko:

updated according to https://www.dartlang.org/guides/libraries/private-files

https://www.dartlang.org/guides/libraries/private-files


Dart.gitignore
```diff
@@ -3,11 +3,27 @@
...
+*.dart.js
+*.info.json      # Produced by the --dump-info flag.
+*.js             # When generated by dart2js. Don't specify *.js if your
+                 # project includes source files written in JavaScript.
+*.js_
+*.js.deps
+*.js.map
+
+
+# files and directories created by development environments
```
shiftkey: These are not specific to Dart, and are maintained in `Global/JetBrains.gitignore` or `Global/macOS.gitignore`.

Please tidy this section up.

PR #2719, opened 2018-05-28
Add ".serverless" to gitignore

tejashah88:
This is for when new projects are initialize for use with the [Serverless framework](https://serverless.com/). Whenever you initialize a new Serverless project, it generates a hidden .serverless folder (see the gitignore below).

Serverless node.js template [gitignore](https://github.com/serverless/serverless/blob/master/lib/plugins/create/templates/aws-nodejs/gitignore).

### DIFF
--- a/Node.gitignore
+++ b/Node.gitignore
@@ -62,3 +62,6 @@ typings/
 
 # vuepress build output
 .vuepress/dist
+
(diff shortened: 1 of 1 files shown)


shiftkey:
@tejashah88 thanks!

PR #2720, opened 2018-05-29
Support other editors in VisualStudio.gitignore

markrendle:
Add the settings directories for:
- VS Code: `.vscode`
- JetBrains Rider: `.idea`

I keep having to add these entries to the `.gitignore` files generated by GitHub using the Visual Studio option. People work on .NET projects using lots of different editors these days, so they should all be included.

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -32,6 +32,12 @@ bld/
 # Visual Studio 2017 auto generated files
 Generated\ Files/
 
+# VS Code cache/options directory
+.vscode/
+
+# JetBrains Rider cache/options directory
+.idea/
+
 # MSTest test Results
 [Tt]est[Rr]esult*/
 [Bb]uild[Ll]og.*


shiftkey:
@markrendle just a heads up, I've been declining this sort of PRs for a while for a few reasons:

 - we have existing templates for these editors - `Global\VisualStudioCode.gitignore` and `Global\JetBrains.gitignore` which are more detailed
 - the duplication of rules across multiple templates becomes an additional maintenance burden, and things fall out of sync because only one place gets updated
 - `VisualStudio.gitignore` is it's own special layer of maintenance hell because of all the various sorts of projects you can use it for

I'm open to suggestions about how to improve this story, but for the moment I don't really have the bandwidth to do anything radical with how we organize these templates.

drothmaler:
Well personally I'm a fan of the modularity of this repository and voted agains such changes numerous times. Also as far as I understand, all the Global/* templates are intended for the use in my personal _global_ gitignore-file, **not** the _project_ gitignore. I use [gitignore.io](http://gitignore.io) to create/update my _global_ gitignore and it is a combination of more than 15 gitignore's. This keeps the project gitignores small an clean.

On the other hand, I know that many people aren't aware of the _global_ gitignore... So in practice many projects put global stuff into there project gitignore, because telling each and every new committer about the global gitignore and how to use it, can be annoying.

But even if I decide to put global stuff in a project gitignore, I still prefer to do this based on small modules instead of a giant one fits all gitignore!

Maybe it would be the best option to extend the GitHub Create Repository dialog to be able to select multiple gitignores (optional with an extra checkbox to include global ones) instead of just one. But I have no idea where to suggest such a feature... Maybe @shiftkey can help...

shiftkey:
@drothmaler thanks for the context. I'm not a user or fan of "use Global for your global gitignore", and happily compose together multiple templates when crafting a project-specific gitignore, so I sit somewhere in between these use cases.

> Maybe it would be the best option to extend the GitHub Create Repository dialog to be able to select multiple gitignores (optional with an extra checkbox to include global ones) instead of just one. But I have no idea where to suggest such a feature... Maybe @shiftkey can help...

The Create Repository dialog here is just consuming and filtering this repository, and I still want to maintain this repository as a reference set of templates that are refined over time, but this doesn't address the ability to compose templates at all.

I've opened https://github.com/github/gitignore/pull/2742 to help tackle the underlying issue about needing metadata for these templates, and I'd love feedback from people who are consuming this repository.

markrendle:
@shiftkey Forgot about this. I see your point about the VisualStudio profile being maintenance hell. Would it not make more sense to have more specific files for different project types? A .NET-specific profile, rather than VS-specific, might be easier to maintain?

shiftkey:
I had another discussion about #2816 and we settled on splitting these out into two templates:

 - `DotNet.gitignore` - a new template fresh with whatever current .NET requires (leaner)
 - `Global/VisualStudio.gitignore` - the current template, moved alongside the other IDEs

I was hoping someone would step up to contribute the beginnings of a `DotNet.gitignore` template but that seems to have stalled out, so maybe someone reading this will notice.

I think that approach here will help balance the current needs of `VisualStudio.gitignore` (which is very editor-centric) and supporting the modern .NET ecosystem (where the editor matters less these days).

PR #2725, opened 2018-06-04
Remove Packages folder from .gitignore

mne1991:
It is very common to have `Packages` folder in domain business logic (it happened to me and to other colleagues), and because of this entry, it has been excluded from our repository. Therefore I would remove this folder from the ignore list.

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -178,8 +178,6 @@ PublishScripts/
 
 # NuGet Packages
 *.nupkg
-# The packages folder can be ignored because of Package Restore
-**/[Pp]ackages/*
 # except build/, which is used as an MSBuild target.
 !**/[Pp]ackages/build/
 # Uncomment if necessary however generally it will be regenerated when needed


drothmaler:
But it is much more common to [have a NuGet-`Packages` folder in your solution](https://docs.microsoft.com/en-us/nuget/consume-packages/packages-and-source-control), so I would vote for keeping it, in the gitignore.
In many cases this `Packages` folder is probably in the root of the repository, so it might be an option to change the rule, so that it only matches the folder there. 
But honestly I don't think a collision with a "Packages" namespace in the BL is very likely... If It actually happens in a particular project, you can always whitelist the specific `Packages` folder in the affected sub project.

mne1991:
Actually I removed it because I have never had in any project that nuget Package folder (can't it be something legacy?). However, if you mean it is really a common situation, then keeping it ignored does make sense. I was Just wondering how can I generate this folder (by nuget) in my project?

drothmaler:
Well I think all .NET project, that I have worked on, or contributed to, did use Nuget for 3rd part dependencies (sometimes even for internal ones). But there certenly is no rule to use it :-)

paulcbetts:
+1 to changing Packages to `Packages/` to only match the root, -1 to removing it entirely

shiftkey:
@paulcbetts that's what we had before https://github.com/github/gitignore/pull/1131

paulcbetts:
@shiftkey Blargh, just leave it alone then I guess, Domain Business Logic'ers will just have to change their names, or customize their .gitignore

shiftkey:
@mne1991 I'm 👎 on changing these defaults - lots of people don't want to store their package outputs in Git, which is what this change would introduce. 

You can add a additional rule like this if you happen to have other folders which are caught by this convention which you want to track:

```
!src/some/path/to/[Pp]ackages/
```

PR #2726, opened 2018-06-04
Added IntelliJ File-Ignoring

PascalKu:

I added the IntelliJ files.


shiftkey:

PR #2727, opened 2018-06-04
Added Visual Studio Code cache directory to Unity gitignore

esnho:

Support to Visual Studio Code

https://code.visualstudio.com/docs/getstarted/settings#_settings-file-locations


shiftkey:
Thanks, but see [`Global\VisualStudioCode.gitignore`](https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore) as that template has guidance for files that can be shared in version control.

PR #2732, opened 2018-06-08
remove not needed blank line from macOS global gitignore

casaper:

* There was a unnecessary blank line

_Too small a change, to link documentary_


drothmaler:
Did you read the comment, right above your change??

shiftkey:
> There was a unnecessary blank line

The line is actually very necessary, because `Icon\r\r` (with two `\r`) is the whole rule.

PR #2733, opened 2018-06-08
Rider IDE (Intellij Idea) cache directory added

skibitsky:

Rider IDE creates a cache folder which is not ignored by GitHub default Unity gitignore


shiftkey:

skibitsky:
Rider IDE is the only way to work with Unity scripts on macOS and Linux (MonoDevelop and VS for Mac sucks). Rider is also very popular on Windows, it has various integrations with Unity. I suggest to add this ignore to the default Unity gitignore because there is an ignore for VS cache ([click!](https://github.com/github/gitignore/blob/master/Unity.gitignore#L8)). Why not to have an ignore for Rider too? 
@shiftkey

shiftkey:
>  I suggest to add this ignore to the default Unity gitignore because there is an ignore for VS cache

That cache is temporary files generated by VS that shouldn't be committed. Rider and the JetBrains suite of editors actually have files that can be committed to version control, which are already documented in this repository. That's why I'm handling them differently here.

I'm also thinking about how to manage these sorts of "related templates" over in #2742, but for the moment (and for my sanity) I need to keep these as two distinct templates.

PR #2734, opened 2018-06-11
Ignore usage statistics file

bamapookie:

.idea/misc.xml used to contain usage statistics.  This was broken out to usage.statistics.xml in the latest EAP (2018.2 EAP).  It should be excluded from git.

See https://youtrack.jetbrains.com/issue/IDEA-192913


shiftkey:
@bamapookie thanks!

PR #2736, opened 2018-06-14
Updated for Codekit 3

codethirteen:
Updating for Codekit 3's new config file extension: https://codekitapp.com/help/

### DIFF
--- a/Global/CodeKit.gitignore
+++ b/Global/CodeKit.gitignore
@@ -1,3 +1,4 @@
 # General CodeKit files to ignore
 config.codekit
+config.codekit3
 /min


shiftkey:
@codethirteen any links to StackOverflow discussions or other documentation about this change?

shiftkey:
@codethirteen it looks like the `.codekit3` file isn't supposed to be shared? The documentation mentions that this does update settings when you get it from somewhere else, but doesn't quite indicate whether this file should be versioned or not.

shiftkey:
> Finally, gitignore.io does list the file.

The source of that change doesn't provide any other context: https://github.com/dvcs/gitignore/pull/61

shiftkey:
I'm going to take this in because it's additive to what was used for versions 1 and 2 and there's been no pushback about that. Thanks for the additional context @codethirteen!

PR #2728, opened 2018-06-05
Trim trailing whitespace

jayvdb:
**Reasons for making this change:**

Trailing whitespace is ugly.

**Links to documentation supporting these [rule] changes:** 

n/a

PR #2737, opened 2018-06-14
Remove trailing whitespace

jonahweissman:

Trailing whitespace should be avoided whenever possible.

https://softwareengineering.stackexchange.com/questions/121555/why-is-trailing-whitespace-a-big-deal

(I used `find . -not -path "./.git/*" -type f -exec sed -i 's/ $//g' {} \;` to get rid of the trailing whitespace)


shiftkey:
Closing in favour of #2728

PR #2738, opened 2018-06-18
Patch artifacts

jayvdb:

`patch` generates .rej and .orig files if all hunks were not applied.

https://linux.die.net/man/1/patch

If this is a new template:

 - **Link to application or project’s homepage**: https://en.wikipedia.org/wiki/Patch_(Unix) / http://savannah.gnu.org/projects/patch/

There are cases were patch uses `#` for rejects, and use `~` instead of .orig , but I am not confident about how what pattern could catch those without false positives.


shiftkey:
@jayvdb thanks!

PR #2740, opened 2018-06-18
Autotools: De-duplicate autom4te entry

jayvdb:

Duplication added in https://github.com/github/gitignore/pull/2480

leaving the original `/autom4te.cache` is probably also ok.  Someone putting their autotools files in a subdirectory is off on their own path and many of the other rules will still not work for them, so it isnt particularly useful to only change one.

But I've made the change to avoid any possible regression.  Perhaps @Ckins might be able to shed some light on why the leading `/` causes problems.

### DIFF
--- a/Autotools.gitignore
+++ b/Autotools.gitignore
@@ -9,7 +9,7 @@ Makefile.in
 
 # http://www.gnu.org/software/autoconf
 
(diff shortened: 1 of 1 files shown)


shiftkey:
@jayvdb thanks!

PR #2741, opened 2018-06-18
Remove exact duplicates in the same file

jayvdb:

Duplicates add nothing.  Two were added quite recently, but most date back more than a year.

Can this repo please have some CI to catch these very basic errors.
c.f. https://github.com/github/gitignore/pull/2728#issuecomment-397936514


shiftkey:
@jayvdb thanks!

PR #2742, opened 2018-06-18
[RFC] add metadata alongside templates

shiftkey:
This PR is an experiment in introducing machine-parseable data alongside the current set of gitignore templates, to address some of the current limitations.

## The Problem

There's a significant amount of implicit context that's accrued over time with this repository, and it's not communicated clearly to consumers of this repository.

Some examples:

 - we receive lots of PRs to add editor support to language templates (VSCode is a recent example of this), which adds rules that don't match what's captured in the editor template. This also adds a maintenance burden that isn't sustainable at scale.
 - we use comments to link templates together like XCode/ObjectiveC/Swift, but we fall back to manual work to keep these up to date
 - modelling the ecosystem of tools with a single file can be hard to follow, and while the templates are intended to be composed together based on your environment we leave consumers to their own devices

## Prior Art

This idea comes from [choosealicense.com](https://choosealicense.com), which is a static site for learning more about common software licenses. The source for this site is [a Jekyll project](https://github.com/github/choosealicense.com) which attaches metadata to each license file in the form of YAML front matter.

An example of this is the [the MIT License file](https://github.com/github/choosealicense.com/blob/gh-pages/_licenses/mit.txt). In this situation the front matter is used to provide content related to each license, but we could use a similar pattern for gitignore files as a way to store related data for tools to build upon.

## A Potential Solution

Having a unified file containing metadata and the contents would have been neat, but as I suspect a lot of consumers are using these files directly I don't want to break them - instead I'm proposing additive metadata alongside the raw gitignore templates.

Examples:
 - `VisualStudio.gitignore` is complemented by `VisualStudio.yml`
 - `Global/JetBrains.gitignore` is complemented by `Global/JetBrains.yml`

These files are only necessary when metadata needs to be tracked. If no file exists, the the template is considered standalone.

The current `VisualStudio.yml` file is probably the best example of what metadata can give us:

```yml
aliases:
  - CSharp
  - C#
  - FSharp
  - F#
  - VisualBasic
  - VB

editors:
  - Global/JetBrains.gitignore
  - Global/VisualStudioCode.gitignore
```

**This helps guide consumers of this repository that this template is associated with some other popular languages, and that there are other editors that could potentially be added when generating this template.**

The contents of these files are described in more detail in [this document](https://github.com/github/gitignore/blob/add-metadata/docs/metadata.md) and I've used this PR to show what moving to this metadata file looks like for some common headaches I've been dealing with:

 - Visual Studio supports a number of different languages - C#, F#, Visual Basic - how can users discover that?
 - The unified XCode/Objective-C/Swift templates - how can we make this more maintainable?
 - JetBrains supports a number of different languages with their suite of editors - how can we make this clearer?

## Bringing The World Along

This doesn't completely address scenarios that people raise like "How can we improve the GitHub New Repository experience?" but the current process of "throw all the rules into every template" is a recipe for maintenance hell, so I wanted to help clarify where I see the separation of responsibilities:

 - **this repository** - used to track reference templates and metadata about how they relate
 - **contributors to this repository** - keep templates and metadata up-to-date
 - **consumers of this repository** - take these reference templates and metadata and bake them into their apps, build tooling to leverage the new metadata

Having a sane way to manage these templates and automate things will be a great thing for the long-term, and getting stuff out of my head and into metadata means it becomes a shared set of knowledge. And being able to work with structured data means

I'd like to hear from people who work with this repository their initial reactions to this proposal:

 - Does this interest you at all?
 - What other metadata about these templates should we be capturing?
 - Are there things in this problem space I haven't thought about yet that I should be aware of?

I'd also love to hear from @joeblau who has been doing some amazing work with `gitignore.io` over the years - does this help make your work easier?

I'll also see if I can get some feedback from GitHub about how these changes might affect the New Repository setup experience.


Android.yml
```diff
@@ -0,0 +1,2 @@
+editors:
+  - Global/JetBrains.gitignore
```
jayvdb: EOL at EOF ... everywhere ;-)

jayvdb:
Appreciate the relationships being described so far.
There are a few which replicate parts of other templates.
e.g. `Global/VirtualEnv.gitignore` should be a subset of `Python.gitignore`.

The following gives some idea of the amount of duplication found throughout, and could help find possible new ways to compose larger templates from smaller ones.

```console
$ find . -type l | xargs rm
$ find . -name '*.gitignore' | xargs  grep '^[^ #]' | cut -d ':' -f 2 | sort | uniq -c | sort -rn | grep -v '^ *[12] '
     17 *.o
     15 *.log
     13 build/
     12 *.so
     12 *.exe
     11 *.dll
     11 *.a
     10 *.dylib
      8 *.bak
      8 *~
      7 bin/
      6 *.tmp
      6 *.out
      6 *.obj
      6 *.lo
      6 *.lib
      6 *.la
      6 *.jar
      5 *.zip
      5 *.pch
      5 node_modules/
      5 dist/
      5 *.class
      4 *.ipa
      4 /.htaccess
      4 .htaccess
      4 *.gch
      4 .env
      4 *.egg
      4 .cache
      4 *.app
      4 *.apk
      3 xcuserdata/
      3 *.xcscmblueprint
      3 *.xccheckout
      3 .vs/
      3 /vendor/
      3 *.userprefs
      3 *.user
      3 *.tgz
      3 /target/
      3 target/
      3 *.suo
      3 *.so.*
      3 *.slo
      3 .settings/
      3 .sass-cache/
      3 *.rbc
      3 *.pyc
      3 .project
      3 *.pidb
      3 *.perspectivev3
      3 *.pdf
      3 *.pdb
      3 *.pbxuser
      3 [Oo]bj/
      3 *.msi
      3 *.moved-aside
      3 *.mode2v3
      3 *.mode1v3
      3 *.mo
      3 *.map
      3 lib/
      3 *.lai
      3 fastlane/test_output
      3 fastlane/report.xml
      3 fastlane/Preview.html
      3 .DS_Store
      3 DerivedData/
      3 !default.perspectivev3
      3 !default.pbxuser
      3 !default.mode2v3
      3 !default.mode1v3
      3 *.d
      3 /cache/*
```

VisualStudio.yml
```diff
@@ -0,0 +1,14 @@
...
+  - F#
+  - VisualBasic
+  - VB
+
+editors:
+  - Global/JetBrains.gitignore
+  - Global/VisualStudioCode.gitignore
+
+reviewers:
+  - Brendan Forster (@shiftkey)
```
jayvdb: maybe add an .editorconfig ? ;-)

shiftkey: LOL perhaps

shiftkey:
@jayvdb thanks for the feedback!

> There are a few which replicate parts of other templates.
> e.g. `Global/VirtualEnv.gitignore` should be a subset of `Python.gitignore`.

Yeah, identifying how current templates relate as part of their ecosystem helps. I'm not really across Python tooling but could a hint like this to include a related template (because it's more important than an editor to the setup) help?

```yaml
# Python.yml
include:
  - Global/VirtualEnv.gitignore
```

> The following gives some idea of the amount of duplication found throughout, and could help find possible new ways to compose larger templates from smaller ones.

For the moment I don't want to go too overboard with this syntax and fragment things too much. Let's focus on finding the accidental and deliberate relationships between templates.

jayvdb:
Do you want to use `include` to describe relationship which are not fully realised at the moment, such as `Plone.yml` & `TurboGears2.yml` 'should' have
```
include:
  - Python.gitignore
```

They contain some of the entries, but all entries apply as they are a layer above. p.s. I am not proposing modifying the actual file, but describing the links between them only in the yaml.

There are quite a few like this, such as Jekyll is Ruby
etc.

shiftkey:
@jayvdb there's a couple of interesting relationships to clarify here - maybe something will fall out of these thoughts:

 - choosing a `Python.gitignore` project, enhancing it with `VirtualEnv.gitignore`, and maybe some editors that support Python
 - choosing `Plone.gitignore` because that's a specialized template for the project you are working on, which should then pull in `Python.gitignore` because it's a dependency

That means potentially building up a graph of templates `Plone` -> `Python` -> `Global/VirtualEnv.gitignore`, and these would be splatted onto disk in the order of traversal.

Does that make sense? Does `include` as metadata make sense for both cases?

jacobbednarz:
@shiftkey ❤️ the intent here! 

I wonder if we could leverage [GitHub's code owners](https://help.github.com/articles/about-codeowners/) instead of the `reviewer` key to streamline the review/change process? To combat the two files needing to stay in sync issue, perhaps we could look to convert everything to YAML and then on master merge have the `gitignore` compatible files written out to disk with all the logic you've described above? I see you've mentioned it as a potential longer term thing but I _think_ that might help the maintenance and consumer sides equally since there will be a sane management of the files but also keep the expected output that people are used to at the moment.

joeblau:
> That means potentially building up a graph of templates Plone.gitignore -> Python.gitignore -> Global/VirtualEnv.gitignore, and these would be splatted onto disk in the order of traversal.

That is basically what I've done with the [**stack** concept in the dvcs.gitignore](https://github.com/dvcs/gitignore/) repo.  You can have:

```
Python.gitignore
Python.VirtualEnv.stack -> VirtualEnv.gitignore
```

then
```
Plone.gitignore
Plone.Python.stack -> Python.gitignore
Plone.VirtualEnv.stack -> VirtualEnv.gitignore
```

This structure allows for composition of templates removing duplication so you can reference VirtualEnv in multiple templates.  Since the stacks are just symlinks to actual templates, you can use your terminal to execute.

```sh
cat Python.* > .gitignore
```

joeblau:
@shiftkey To address the main challenges, I think I've got solutions to all of those challenges using just a flat file system.  You can use the [@dvcs/gitignore](https://github.com/dvcs/gitignore) repo as an example for where I implement these solutions.

> we receive lots of PRs to add editor support to language templates (VSCode is a recent example of this), which adds rules that don't match what's captured in the editor template. This also adds a maintenance burden that isn't sustainable at scale.

This can be solved with the `.patch` file extension concept.  When @github/gitignore was a main upstream, I needed a way to allow "unapproved" additions to templates.  If you create a `.patch` extension, you can combine that with with @jacobbednarz mentioned of a *CODEOWNERS* file and allow the public or specific owners to modify the `.patch` files.  This will separate the "official" part of the template (`.gitignore`) from the "community" part of the template (`.patch`).

With the `.patch` files you can specify owners or just let anyone in the community approve the changes.

```sh
-rw-r--r--  1 josephblau  staff   1.0K Jun 19 06:59 JetBrains.gitignore
-rw-r--r--  1 josephblau  staff   176B Dec 31 11:08 JetBrains.patch
```

Create Official Template
```sh
cat JetBrains.gitignore > .gitignore
```

Create Community Template
```sh
cat JetBrains.* > .gitignore
```

> we use comments to link templates together like XCode/ObjectiveC/Swift, but we fall back to manual work to keep these up to date

[I mentioned how I solved these above with the stack concept.](https://github.com/github/gitignore/pull/2742#issuecomment-398360987)


Another issue you may run into is having aliases of a template.

What I've done for linking templates is to use a simple symlink.  For Example, Apple renamed OSX to macOS.  What I've done is represent OSX as symlink of macOS that way when one template is changed, the other is also changed

```sh
lrwxr-xr-x  1 josephblau  staff    15B Mar 26  2017 OSX.gitignore@ -> macOS.gitignore
-rw-r--r--  1 josephblau  staff   402B Jun 19 06:59 macOS.gitignore
```

This allows for tons of aliases, such as all of the JetBrains projects (Webstorm, AppCode, CLion, IntelliJ, PHPStorm, PyCharm).

```sh
comm OSX.gitignore macOS.gitignore
# returns nothing because files are the same
```

> modelling the ecosystem of tools with a single file can be hard to follow, and while the templates are intended to be composed together based on your environment we leave consumers to their own devices

*My opinionated design goal*

One of the key takeaways with the design of having file extensions and symlinks is that you don't want any extra magic metadata in the files.  The goal is to use standard Linux command line tools to be able to recreate your templates.  If templates contain reference information to other  templates, that means users will need special tools outside of things like `cat`, `grep`, `ls`, `|` and `>` to create a template.

A second added benefit is that you can keep everything working exactly how it works now and just add functionality to make the repo easier to maintain.

jayvdb:
Re CODEOWNERS, see https://github.com/jusx/mergeable/issues/24 for workaround and potential dropin solution in near future.

joeblau:
The one challenge that I can see with *CODEOWNERS* file is that the approvers would have to be part of the GitHub organization with write access to the repository.

jayvdb:
Write access not necessary with mergeable probot.
But that is technical. Policy is a different matter.

VisualStudio.yml
whoisj: 👍

VisualStudio.gitignore
```diff
@@ -286,10 +286,6 @@ paket-files/
 # FAKE - F# Make
 .fake/
 
-# JetBrains Rider
```
whoisj: I'm good with this. :+1:

PatMyron:
I didn't know [gitignore.io](https://www.gitignore.io) existed, so this is how I solved some of these problems myself: https://stackoverflow.com/a/53165122/4122849

I've since updated it to run in the background, swallow any output, and append to previous results in case of connectivity problems:
https://github.com/PatMyron/dotfiles/blob/46875b963c3c6c26c1d993a80698c95cd43d946a/.bashrc#L185-L190

It's definitely still a work in progress and has some quirks (anything removed from templates won't be automatically removed, but I find that acceptable personally), but I'd love to hear thoughts about this approach

gistofj:
@shiftkey I'm no longer a good person for this. Better to include @dscho or @jeffhostetler.

westside333 (changes requested):
add-metadata

stonehead8111:
1

Gatito1010:
> Este PR es un experimento en la introducción de datos analizables por máquina junto con el conjunto actual de plantillas de gitignore, para abordar algunas de las limitaciones actuales.
> 
> ## El problema
> Hay una cantidad significativa de contexto implícito que se acumula con el tiempo con este repositorio, y no se comunica claramente a los consumidores de este repositorio.
> 
> Algunos ejemplos:
> 
> * recibimos muchos PR para agregar compatibilidad con el editor a las plantillas de idioma (VSCode es un ejemplo reciente de esto), que agrega reglas que no coinciden con lo que se captura en la plantilla del editor. Esto también agrega una carga de mantenimiento que no es sostenible a escala.
> * usamos comentarios para vincular plantillas como XCode/ObjectiveC/Swift, pero recurrimos al trabajo manual para mantenerlos actualizados
> * modelar el ecosistema de herramientas con un solo archivo puede ser difícil de seguir, y aunque las plantillas están diseñadas para componerse juntas en función de su entorno, dejamos a los consumidores con sus propios dispositivos.
> 
> ## Estado de la técnica
> This idea comes from [choosealicense.com](https://choosealicense.com), which is a static site for learning more about common software licenses. The source for this site is [a Jekyll project](https://github.com/github/choosealicense.com) which attaches metadata to each license file in the form of YAML front matter.
> 
> An example of this is the [the MIT License file](https://github.com/github/choosealicense.com/blob/gh-pages/_licenses/mit.txt). In this situation the front matter is used to provide content related to each license, but we could use a similar pattern for gitignore files as a way to store related data for tools to build upon.
> 
> ## A Potential Solution
> Having a unified file containing metadata and the contents would have been neat, but as I suspect a lot of consumers are using these files directly I don't want to break them - instead I'm proposing additive metadata alongside the raw gitignore templates.
> 
> Examples:
> 
> * `VisualStudio.gitignore` is complemented by `VisualStudio.yml`
> * `Global/JetBrains.gitignore` is complemented by `Global/JetBrains.yml`
> 
> These files are only necessary when metadata needs to be tracked. If no file exists, the the template is considered standalone.
> 
> The current `VisualStudio.yml` file is probably the best example of what metadata can give us:
> 
> ```yaml
> aliases:
>   - CSharp
>   - C#
>   - FSharp
>   - F#
>   - VisualBasic
>   - VB
> 
> editors:
>   - Global/JetBrains.gitignore
>   - Global/VisualStudioCode.gitignore
> ```
> 
> **This helps guide consumers of this repository that this template is associated with some other popular languages, and that there are other editors that could potentially be added when generating this template.**
> 
> The contents of these files are described in more detail in [this document](https://github.com/github/gitignore/blob/add-metadata/docs/metadata.md) and I've used this PR to show what moving to this metadata file looks like for some common headaches I've been dealing with:
> 
> * Visual Studio supports a number of different languages - C#, F#, Visual Basic - how can users discover that?
> * The unified XCode/Objective-C/Swift templates - how can we make this more maintainable?
> * JetBrains supports a number of different languages with their suite of editors - how can we make this clearer?
> 
> ## Bringing The World Along
> This doesn't completely address scenarios that people raise like "How can we improve the GitHub New Repository experience?" but the current process of "throw all the rules into every template" is a recipe for maintenance hell, so I wanted to help clarify where I see the separation of responsibilities:
> 
> * **this repository** - used to track reference templates and metadata about how they relate
> * **contributors to this repository** - keep templates and metadata up-to-date
> * **consumers of this repository** - take these reference templates and metadata and bake them into their apps, build tooling to leverage the new metadata
> 
> Having a sane way to manage these templates and automate things will be a great thing for the long-term, and getting stuff out of my head and into metadata means it becomes a shared set of knowledge. And being able to work with structured data means
> 
> I'd like to hear from people who work with this repository their initial reactions to this proposal:
> 
> * Does this interest you at all?
> * What other metadata about these templates should we be capturing?
> * Are there things in this problem space I haven't thought about yet that I should be aware of?
> 
> I'd also love to hear from @joeblau who has been doing some amazing work with `gitignore.io` over the years - does this help make your work easier?
> 
> I'll also see if I can get some feedback from GitHub about how these changes might affect the New Repository setup experience.

Gatito1010:
> agregar metadatos

PR #2745, opened 2018-06-25
Update Node.gitignore

obie3:
added the history dir for VS Code IDE

_TODO_

_TODO_

If this is a new template:

### DIFF
--- a/Node.gitignore
+++ b/Node.gitignore
@@ -71,3 +71,6 @@ typings/
 
 # Serverless directories
 .serverless
+
+# VS Code history directory
+.history


shiftkey:
@obie3 any links to StackOverflow discussions or other documentation about this change? I asked about this over in #2651 but couldn't get any clues...

obie3:
@shiftkey I just finished going through your convo with @samuelstein, My bad I didnt know a PR was made for the patch already, as @samuelstein rightly pointed out there is no official doc for it and secondly the dir is not include .vscode folder and will be always be committed along other dirs in the project if not included in gitignore explicitly, well I'm also a mac user and installed mine as a third party app from the vscode market, you can look it up there, thanks and cheers

shiftkey:
@obie3 I'd love to understand this change because I couldn't even find the right extension to install, but even after that I think this should live under `Global/VisualStudioCode.gitignore` so I'm going to close this out.

Feel free to submit a PR there, this does feel like a good candidate for the template but it currently lacks the appropriate context.

PR #2746, opened 2018-06-28
updated SFDC gitignore

paustint:

The force.com gitignore was lacking many IDE project files

https://github.com/mailtoharshit/gitignore/blob/master/.gitignore
https://github.com/github/gitignore/blob/master/ForceDotCom.gitignore
https://github.com/mailtoharshit/gitignore/blob/master/.gitignore

### DIFF
--- a/ForceDotCom.gitignore
+++ b/ForceDotCom.gitignore
@@ -1,4 +1,68 @@
+# Credit:
+# https://github.com/mailtoharshit/gitignore/blob/master/.gitignore
+# https://github.com/github/gitignore/blob/master/ForceDotCom.gitignore
+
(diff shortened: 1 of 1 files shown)


shiftkey:
> The force.com gitignore was lacking many IDE project files

We track many of the rules you included in distinct templates under the `Global` namespace. This helps with maintainability on our end and makes it easier for users to compose templates together.

Are you able to focus these changes on just what updates the `ForceDotCom` template needs?

shiftkey:
Closing this out due to inactivity. @paustint if you're interested in landing this PR can we focus this PR on the non-IDE-specific changes needed for this template, and then identify which of the templates under `Global` need further work?

PR #2748, opened 2018-06-29
Update Swift.gitignore

fidmor89:
Adding a comment in case one wants to ignore .xcworkspace

Most people using cocoa pods will likely run pod install after getting the repository so the xcworkspace will be recreated. 

[https://stackoverflow.com/questions/22345575/should-xcworkspace-be-in-gitignore](https://stackoverflow.com/questions/22345575/should-xcworkspace-be-in-gitignore)

### DIFF
--- a/Swift.gitignore
+++ b/Swift.gitignore
@@ -47,6 +47,9 @@ playground.xcworkspace
 # https://guides.cocoapods.org/using/using-cocoapods.html#should-i-check-the-pods-directory-into-source-control
 #
 # Pods/
+#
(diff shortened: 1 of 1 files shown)


shiftkey:
@fidmor89 thanks!

Could you also update [`Global/Xcode.gitignore`](https://github.com/github/gitignore/blob/master/Global/Xcode.gitignore) so we can keep these in sync?

fidmor89:
Hey @shiftkey I just sent a PR for [Global/Xcode.gitignore](https://github.com/github/gitignore/blob/master/Global/Xcode.gitignore)

PR #2752, opened 2018-07-01
Update Xcode.gitignore

fidmor89:
Updating file to match the one for swift and objective c.

To keep this file in sync with its swift and objective-c counterparts 

Note: on the diff tool it might look like some lines where removed but in reality not one of those is missing (just a different line)

### DIFF
--- a/Global/Xcode.gitignore
+++ b/Global/Xcode.gitignore
@@ -2,17 +2,11 @@
 #
 # gitignore contributors: remember to update Global/Xcode.gitignore, Objective-C.gitignore & Swift.gitignore
 
-## User settings
-xcuserdata/
-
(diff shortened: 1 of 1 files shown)


shiftkey:
@fidmor89 thanks!

Coeur:
@shiftkey ah! I didn't notice this pull request before, but it's not a good one: we lose the comments and it introduces some junk.

shiftkey:
@Coeur not sure I follow:

> we lose the comments

This looks reasonably well-documented to me - can you clarify in the PR where you think things could be better?

> it introduces some junk

I believe this PR was opened to ensure `XCode.gitignore` also has new rules previously defined with `ObjectiveC.gitignore` and `Swift.gitignore` - can you indicate which specific rules shouldn't be here?

PR #2753, opened 2018-07-01
Added .pytest_cache/ into Python.gitignore

sanjaypra555:

Generally developer adds tests into repository during development phase, pytest adds .pytest_cache/ directory while testing.


shiftkey:
@sanjaypra555 this rule is already captured on Line 48:

https://github.com/github/gitignore/blob/65cdade87324774700104eee79dec1e7b91d56d7/Python.gitignore#L48

PR #2755, opened 2018-07-03
Add Intermediate folder for exclusion

NN---:
Some projects use Intermediate instead of obj, e.g.: https://github.com/github/gitignore/blob/master/UnrealEngine.gitignore#L70

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -23,6 +23,7 @@ bld/
 [Bb]in/
 [Oo]bj/
 [Ll]og/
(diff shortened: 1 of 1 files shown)


shiftkey:
@NN given this is already covered by `UnrealEngine.gitignore`, why does this need to be duplicated here?

NN---:
There are C++ projects that are not using Unreal Engine and intermediate folder is named “Intermediate”.

NN---:
@shiftkey I can use unrealengine template too.

PR #2760, opened 2018-07-05
Add IDEs and containers ignore

KenanBek:

Ignore idea and containerization tool generated files and folders.

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -1,3 +1,9 @@
+# IDEs / Containers
+.idea
+.vscode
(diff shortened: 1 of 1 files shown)


shiftkey:
These are already covered in more details by existing templates:

 - `Global/VisualStudioCode.gitignore`
 - `Global/Vagrant.gitignore`

It is recommended to tailor your repository `.gitignore` to suit the tools you are using, using these reference templates, as duplicating rules everywhere becomes a maintenance burden over time.

PR #2761, opened 2018-07-07
Possible security issue with .DS_Store

parnissolle:
There is potential security issues with data stored in .DS_Store files.


mys721tx:
It would be better to setup a global gitignore with `macOS.gitignore` to ignore all macOS related files all together.

shiftkey:
Seconding what @mys721tx said, we have `Global/macOS.gitignore` which covers this and other OS-specific rules.

PR #2763, opened 2018-07-11
Updated Unity.gitignore with visual studio code settings dir

paulhayes:
Unity has support for visual studio code. Given that there's already an entry to ignore Visual Studio's settings it seems prudent to ignore this IDE's settings file too.


shiftkey:
> Given that there's already an entry to ignore Visual Studio's settings it seems prudent to ignore this IDE's settings file too.

This setting is for a temporary cache for VS 2015 and later, but  [`Global/VisualStudioCode.gitignore`](https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore) shows this `.vscode` folder contains things that can be committed and shared.

paulhayes:
Thanks, didn't realise.

PR #2764, opened 2018-07-18
Added  files to C.gitignore as when coding on different OS and setup

GabrielPora:
swap file used by various operating systems
Property List is specific to Mac
nfs files are specific to working over network

### DIFF
--- a/C.gitignore
+++ b/C.gitignore
@@ -50,3 +50,8 @@ modules.order
 Module.symvers
 Mkfile.old
 dkms.conf
+
+# Unix and Linux Generated files
+*.swp
+*.nfs.*
+*.plist


shiftkey:
I'd prefer for these rules to be ported to their OS-specific templates, like [`Global/Linux.gitignore`](https://github.com/github/gitignore/blob/master/Global/Linux.gitignore) and [`Global/macOS.gitignore`](https://github.com/github/gitignore/blob/master/Global/macOS.gitignore), as these will then benefit everyone.

Please also add more context about _why_ these files can be safely excluded from version control.

GabrielPora:
Thank you for the feedback, I will do this in future commits.

PR #2765, opened 2018-07-24
alkathirikhalid

alkathirikhalid:

1. Modified ignoring individual files contained in .idea directory/folder to ignoring the whole .idea directory/folder, all these files are generate we could simply ignore all.
2. Added ignore Mac generated temp image file .DS_Store that are found in Android image resources, that stores custom attributes of its containing folder, such as the position of icons when developing on Mac.

1. https://git-scm.com/docs/gitignore
2. https://en.wikipedia.org/wiki/.DS_Store


fabiofdsantos:
@alkathirikhalid,
I think you should change the title of this issue.

Android.gitignore
```diff
@@ -63,3 +57,6 @@ fastlane/Preview.html
 fastlane/screenshots
 fastlane/test_output
 fastlane/readme.md
+
+# Mac temp image file
+.DS_Store
```
shiftkey: This is an OS-specific rule, and is already covered in `Global/macOS.gitignore`. Please remove it from this template.

Android.gitignore
```diff
@@ -34,13 +34,7 @@ captures/
...
 # IntelliJ
 *.iml
-.idea/workspace.xml
-.idea/tasks.xml
-.idea/gradle.xml
-.idea/assetWizardSettings.xml
-.idea/dictionaries
-.idea/libraries
-.idea/caches
+.idea/
```
shiftkey: I don't agree with this bit

>  all these files are generate we could simply ignore all.

We are tracking a separate `Global/JetBrains.gitignore` which refers to [this documentation](https://intellij-support.jetbrains.com/hc/en-us/articles/206544839) from JetBrains that indicates there are many files under `.idea` that can be committed to version control.

PR #2766, opened 2018-07-24
Ignore *nix swap files

bdeanindy:

ViM developers may appreciate not having *nix swap files in their commits accidentally, this prevents that from happening.

[Description of what swap files are for non-ViM reviewers](https://stackoverflow.com/questions/11313639/what-is-a-swap-file)

### DIFF
--- a/Node.gitignore
+++ b/Node.gitignore
@@ -71,3 +71,7 @@ typings/
 
 # Serverless directories
 .serverless
+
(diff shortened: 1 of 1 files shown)


shiftkey:
Thanks, but see [`Global/Vim.gitignore`](https://github.com/github/gitignore/blob/master/Global/Vim.gitignore).

PR #2767, opened 2018-07-30
Removed project files from gitignore for UE

apap04:

So project files won't be ignored in games with code. These files shouldn't be ignored, even if the game has no code.


shiftkey:
cc @samsheff @robipresotto for feedback as they added these rules

shiftkey:
@apap04 any links to StackOverflow discussions or other documentation about how this change is beneficial?

robipresotto:
The files should be ignored because it's auto-generated by the engine and could contain personal configuration. I don't see any benefit in that changes.

shiftkey:
> The files should be ignored because it's auto-generated by the engine and could contain personal configuration.

@apap04 thoughts on ☝️ ?

PR #2770, opened 2018-08-04
Fix issue with out/

trenrandrews:
If you have a package called something like com.businessname.out.domain with out/ those will get skipped also.

_TODO_

_TODO_

If this is a new template:

### DIFF
--- a/Global/JetBrains.gitignore
+++ b/Global/JetBrains.gitignore
@@ -39,7 +39,7 @@ cmake-build-*/
 *.iws
 
 # IntelliJ
-out/
+/out/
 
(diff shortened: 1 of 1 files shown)


shiftkey:
@trenrandrews any links to StackOverflow discussions or other documentation about how this change is beneficial?

shiftkey:
Closing this out due to inactivity. @trenrandrews if you're able to provide more context about why the nested `out` directory should be committed (I suspect this will catch everyone's `out` folders and cause other unexpected issues) then I'm happy to revisit the PR.

PR #2771, opened 2018-08-05
add pycharm settings folder to Python.gitignore

megagru:

Since **PyCharm** is a common IDE for python based projects, which stores **project specific settings** in the folder **.idea/** right beside the projects code, the settings folder should be added to .gitignore.

https://www.jetbrains.com/help/pycharm/project-and-ide-settings.html#location_of_directories


shiftkey:

BrianMSheldon:

PR #2773, opened 2018-08-07
Specific changes requires fro our Github Enterprise instance

gunjvm:
Its a compilation of all file extensions we want to restrict for our Enterprise Instance. 
We want the template to be present as CM i.e for Capital Market users of our enterprise Instance.
The users can use this template in each of their repositories.

We want to impose restrictions on users to not upload binary or any files except pure source code to our github Instance. 

Specific changes requires for our Github Enterprise instance

Specific changes requires fro our Github Enterprise instance

https://github.com/gunjvm/gitignore


shiftkey:
@gunjvm I can't take this in because it's a company-specific template which isn't the purpose of this repository.

I'd recommend reaching out to GitHub support (if you haven't already) about being able to customize your GitHub Enterprise instance to support this workflow.

PR #2775, opened 2018-08-08
Ignore logs for Symfony4

TheGamingBook:
Ignore /var/log/* for Symfony

Folder structure for logs has been changed for Symfony 4, from /var/logs/* in v3.x to /var/log/* in v4.x

https://symfony.com/doc/master/logging.html
Under Where Logs are Stored


shiftkey:
@TheGamingBook thanks!

PR #2780, opened 2018-08-11
Adding Django.gitignore

amiteshbhartiya:

To add git-ignore for Django

**Reference:** 

https://github.com/django/django/blob/master/.gitignore


shiftkey:
@amiteshbhartiya thanks for the submission.

As this overlaps with lots of rules already defined in `Python.gitignore`, we have a section for additional Django rules here:

https://github.com/github/gitignore/blob/ea28c14da0faf75047165c10223635ba95566ad7/Python.gitignore#L54-L57

Could you add any extra rules that are Django-specific in there?

amiteshbhartiya:
Hey @shiftkey,  as django being the popular python framework don't you think having separate file would help developer in convenient way ? The way have separate file for laravel, zend and symfony php frameworks. ?

shiftkey:
> as django being the popular python framework don't you think having separate file would help developer in convenient way ? 

Not really. I've had no issues with receiving and merging Django contributions to the existing Python template in the past.

> As framework is most popular way to develop code, keeping gitignore at framework level would improve usability, just a food for thought

This repository is intended to be a set of reference templates, and when rules are duplicated across templates like this it becomes a maintenance burden.

PR #2784, opened 2018-08-18
add .dmypy.json to Python.gitignore.

GabrielC101:

MyPy includes a daemon server that creates a file named .dmypy.json to
manage socket connection data. 

MyPy includes a server, dmypy, to check typing in Python code. This server creates a file .dmypy.json to handle socket information. 
https://github.com/python/mypy/blob/v0.620/mypy/dmypy_util.py#L11

The MyPy project has included .dmypy.json in it's own .gitignore file:
https://github.com/python/mypy/blob/master/.gitignore#L16

It's very unlikely anyone would want to reuse this file. As MyPy is used almost exclusively to aide developers, it is unlikely to be wanted in a production environment.

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -103,3 +103,5 @@ venv.bak/
 # mypy
 .mypy_cache/
 .dmypy.json
+dmypy.json
+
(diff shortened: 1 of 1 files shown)


shiftkey:
@GabrielC101 thanks!

PR #2785, opened 2018-08-19
Update Terraform.gitignore

obounaim:

Ignore override files as they are usually used locally
to override resources or variables without changing the code

https://www.terraform.io/docs/configuration/override.html


shiftkey:
@obounaim thanks!

PR #2788, opened 2018-08-22
ignore ipython profile

amigcamel:

When invoking ipython shell if a virtual environment is activated, ipython will read profiles under `profile_default` or read `ipython_config.py`, if exist.
I'd suggest we ignore these since they're often not related to the project per se.

https://ipython.readthedocs.io/en/stable/development/config.html

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -73,7 +73,7 @@ target/
 .ipynb_checkpoints
 
 # IPython
-default_profile/
+profile_default/
 ipython_config.py
 
 # pyenv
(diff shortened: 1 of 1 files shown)


Python.gitignore
```diff
@@ -72,6 +72,10 @@ target/
 # Jupyter Notebook
 .ipynb_checkpoints
 
+# IPython
+default_profile/
```
shiftkey: I'm not clear on where this rule is being used in the documentation linked to - is this supposed to be `profile_default`?

amigcamel: Sorry for the typo. It should be `profile_default`, and I've updated my branch.

shiftkey:
@amigcamel thanks!

PR #2789, opened 2018-08-23
Update Node.gitignore

piyush97:

Package-lock ignore

### DIFF
--- a/Node.gitignore
+++ b/Node.gitignore
@@ -39,6 +39,9 @@ jspm_packages/
 # TypeScript v1 declaration files
 typings/
 
+# Package-lock.json file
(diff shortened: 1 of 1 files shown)


shiftkey:
This should be in version control according to [NPM docs](https://docs.npmjs.com/files/package-locks#using-locked-packages):

> It is highly recommended you commit the generated package lock to source control: this will allow anyone else on your team, your deployments, your CI/continuous integration, and anyone else who runs `npm install` in your package source to get the exact same dependency tree that you were developing on.

piyush97:
Oh, Thanks! @shiftkey

PR #2790, opened 2018-08-25
Processing: Add ignore rules for Linux ARM exports

Niels-NTG:
Since Processing version 3.0.1 you can compile Linux ARM versions of your Processing sketch, distinct from the regular 32 and 64 bit Linux exports.

- https://github.com/processing/processing/releases/tag/processing-0247-3.0.1


shiftkey:
@Niels-NTG thanks!

PR #2791, opened 2018-08-25
add .idea folder

nerdoc:

_TODO_

_TODO_

If this is a new template:

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -94,6 +94,9 @@ venv.bak/
 .spyderproject
 .spyproject
 
+# PyCharm
+.idea/
+
 # Rope project settings
 .ropeproject
 


Python.gitignore
```diff
@@ -94,6 +94,9 @@ venv.bak/
 .spyderproject
 .spyproject
 
+# PyCharm
```
kimbaudi: .idea folder is not only applicable to PyCharm, but also Rider and possibly other IntelliJ IDEs. I think it would be better to replace the comment `#PyCharm` with `#IntelliJ project specific settings` in the description.

nerdoc:
ah yes, I'll change that.

shiftkey:
@nerdoc I recommend tailoring your `.gitignore` based on the tools you are using.

BrianMSheldon:

PR #2793, opened 2018-08-26
Added Hugo.gitignore

cansurmeli:

Hugo is a rising star in static site development because:
- it's easy to develop with,
- it provides nearly instantaneous export of the static site(usually measured in milliseconds or seconds at most)
- of the performance it yields once the site is deployed.

Therefore a `gitignore` file is long overdue.

 - **Link to application or project’s homepage**: [http://gohugo.io](http://gohugo.io)

### DIFF
--- a/Hugo.gitignore
+++ b/Hugo.gitignore
@@ -1,0 +1,44 @@
+# Compiled source #
+###################
+*.com
+*.class
+*.dll
+*.exe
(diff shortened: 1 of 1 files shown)


shiftkey:
Closing this in favour of #2404 - please add your feedback to that PR

PR #2794, opened 2018-08-28
Update Perl6.gitignore

p6steve:
Add some macOS and vi hidden files to gitignore

Some basic changes that make this gitignore file work for my (commonplace) perl6 env

### DIFF
--- a/Perl6.gitignore
+++ b/Perl6.gitignore
@@ -4,4 +4,7 @@
 # precompiled files
 .precomp
 lib/.precomp
-
+# macOS
+.DS_Store
+# vi
+.swp


p6steve:
Hi Brian - I am new to GitHub and felt that this small pull request would be helpful to others. Feel free to reject - I will not be offended. Feedback and comments always welcome!

shiftkey:
@p6steve thanks for the suggestion.

I'm going to close this out because we have existing OS- and editor-specific templates under the `Global` folder - check out [Global/macOS.gitignore](https://github.com/github/gitignore/blob/master/Global/macOS.gitignore) and [Global/Vim.gitignore](https://github.com/github/gitignore/blob/master/Global/Vim.gitignore) for additional templates you can use locally when working with Perl.

PR #2795, opened 2018-08-31
adding idea and impl for go gitignore

nanofaroque:

Currently, it does not have the capability to remove .idea and *.impl

_TODO_

If this is a new template:

### DIFF
--- a/Go.gitignore
+++ b/Go.gitignore
@@ -10,3 +10,5 @@
 
 # Output of the go coverage tool, specifically when used with LiteIDE
 *.out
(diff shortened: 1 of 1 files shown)


shiftkey:
@nanofaroque I'm not 100% sure where `.impl` is coming from, but we have [`Global\JetBrains.gitignore`](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore) which I believe covers the IntelliJ suite of IDEs. Does that help with your use case?

PR #2802, opened 2018-09-09
Add .nox directory for the nox testing tool

SethMichaelLarson:

This is a new Python testing tool, similar to tox and is being used in a lot of Google Cloud projects.

Couldn't find specific documentation about the `.nox/` directory but the tool uses this directory to house it's virtualenvs.

 - **Link to application or project’s homepage**: https://nox.readthedocs.io/en/stable/

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -38,6 +38,7 @@ pip-delete-this-directory.txt
 # Unit test / coverage reports
 htmlcov/
 .tox/
+.nox/
 .coverage
 .coverage.*
 .cache


shiftkey:
I couldn't find docs, but I saw a couple of projects also using this rule and I found [this code](https://github.com/theacodes/nox/blob/f8462b812444e1feb31b02a942d52a6e52a74d9d/nox/__main__.py#L63-L65):

```python
    parser.add_argument(
        "--envdir", default=".nox", help="Directory where nox will store virtualenvs."
    )
```

Feel free to add a comment here if you see any other details about this, but for now I'm good to merge this in!

SethMichaelLarson:
Thanks! cc: @theacodes

PR #2806, opened 2018-09-12
Added wildcard ignore for .keystore files

Orbyt:
Keystore files with an extension of `.keystore` are also valid and commonly used.

### DIFF
--- a/Android.gitignore
+++ b/Android.gitignore
@@ -43,8 +43,9 @@ captures/
 .idea/caches
 
 # Keystore files
-# Uncomment the following line if you do not want to check your keystore files in.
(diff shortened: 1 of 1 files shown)


shiftkey:
@Orbyt any links to StackOverflow discussions or other documentation about this change?

Orbyt:
@shiftkey `JKS` is generally used to denote "Java Key Store", but the extension used is largely redundant. When generating keystores, some developers opt for the `.jks` extension, but it is also very common to simply use a `.keystore` extension. Additionally, the [default debug keystore](https://developer.android.com/studio/publish/app-signing#debug-mode) generated by the Android SDK tools is named `debug.keystore`.

PR #2810, opened 2018-09-14
Ignore build cache file for Android Studio 3.1+

stuckj:

This build cache file is generated and contains full paths (specific to the user's checkout directory).

See this SO post: https://stackoverflow.com/questions/49557737/should-i-add-idea-caches-build-file-checksums-ser-to-gitignore

and this Google issue tracker ticket: https://issuetracker.google.com/issues/77544553

for justification.


shiftkey:
@stuckj thanks!

PR #2814, opened 2018-09-18
Adding .DS_Store to VisualStudio.gitignore

sayedihashimi:

`.DS_Store` is a file that gets generated by macOS. It contains metadata about the containing folder. The file is not meant to be shared. 


mys721tx:
Please consider setting up a [global .gitignore](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore) using appropriate files under the Global directory.

PR #2815, opened 2018-09-18
No need to fully exclude CodeRush settings

Droni:
Currently, CodeRush provides the capability to store team settings and images used in Rich Comments and they should be shared among all team members. I have corrected the gitignore file to exclude only personal settings.

https://community.devexpress.com/blogs/rorybecker/archive/2018/07/02/coderush-new-options.aspx


shiftkey:
@Droni thanks!

PR #2816, opened 2018-09-18
[WIP] Rename VisualStudio.gitignore to DotNet.gitignore

matkoch:
Unfortunately, you don't have issues enabled, so please let me use this first to discuss the change I would like to make:

Some time ago, the change was made to put everything CSharp related into the VisualStudio.gitignore file (f06c0504ac984ee44ddc820c609a4e9490ec7d19). While this might have been a good idea back then, it is very misleading nowadays. Not only is this a discoverability issue - many people probably look for _CSharp_ - but also the ecosystem of IDEs has changed a lot, and people are using not only VisualStudio, but also Rider, VSCode, Sublime, Atom, etc..

So my suggestion is to move everything that is purely CSharp related to its own `.gitignore` file.

What do you think?


shiftkey:
@matkoch please read the history behind this change in #492 and this discussion #2720 and maybe #2742 too.

We originally had this headache of people only adding rules to one template, and once we identified that most of the rules are related to the VS project system (and the ecosystem of things supported) the initial language you choose doesn't really matter.

This is intended to be a reference set of templates, so please tell me more about your use case and how you're consuming this file.

matkoch:
_Thanks for your quick response and for the link to the original PR. It wasn't referenced in the commit, unfortunately._

I do understand the reasoning back then, but as mentioned, this is no longer true for today's ecosystem. Phil wrote that:

> Visual Studio is the de-facto IDE for developers doing .NET, C++, or Windows RT development on Windows

Calling it `VisualStudio.gitignore` is not very inclusive, having in mind that other IDEs exist and are widely used. The discoverability issue was raised back then as well (https://github.com/github/gitignore/pull/492#issuecomment-11603289), but no one actually addressed it. The background to this issue is actually a plugin for IDEA/Rider (https://plugins.jetbrains.com/plugin/7495--ignore), where you can select from multiple `.gitignore` files, which then finally get merged.

Me personally, I would be happy to have it renamed to `DotNet.gitignore`.

shiftkey:
> Me personally, I would be happy to have it renamed to `DotNet.gitignore`.

Again, this is not quite accurate either. It doesn't cover Visual C++, SQL Server, Biztalk, client-side technologies, Node.js and Python tools for VS, Installers, and others that have accrued in that file over time.

What if we went a slightly different approach:

 - `DotNet.gitignore` - a new template fresh with whatever current .NET requires (leaner)
 - `Global/VisualStudio.gitignore` - the current template, moved alongside the other IDEs

I suspect there's work I need to do on my end to address downstream users, but maybe that's a way to break this impasse?

matkoch:
SGTM, do you want me to get started with the new template?

shiftkey:
@matkoch please do - it'll be great to get eyes on it from across the ecosystem while I figure out the rest

shiftkey:
@matkoch please open a fresh PR with just the new template changes, so we can separate that discussion from this.

PR #2817, opened 2018-09-20
added *.orig for backups kept by Team Explorer's Diff tool on edit/save

birbilis:


birbilis:
update: seems I was wrong and most probably the *.orig is not saved by VS tools, but by the default Diff tool used by TortoiseGit (or else one used by GitExtensions addon for VS)

in any case, one wouldn't want those .orig files to end up in their repo

shiftkey:
For reference, we have [`Global/Patch.gitignore`](https://github.com/github/gitignore/blob/cd872a98f0faf94f3472645e9f8cacfbcd463b71/Global/Patch.gitignore) available which covers this rule.

PR #2819, opened 2018-09-25
Remove patch files from repo

AurelienLavorel:

Patch files should be hidden

### DIFF
--- a/Magento.gitignore
+++ b/Magento.gitignore
@@ -2,6 +2,8 @@
 # Magento Default Files    #
 #--------------------------#
 
+/PATCH_*.sh
+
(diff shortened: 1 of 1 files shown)


shiftkey:
@AurelienLavorel any links to StackOverflow discussions or other documentation about this change?

AurelienLavorel:
If you look at [Magento documentation.](https://devdocs.magento.com/guides/m1x/other/ht_install-patches.html#id), Magento installed patches are available on *app/etc/applied.patches.list*, not in root folder.

shiftkey:
@AurelienLavorel thanks!

PR #2821, opened 2018-09-27
Update VisualStudio.gitignore

KevinRaney:
This will ignore SQL Server Reporting Services user files

https://stackoverflow.com/questions/47076716/which-reporting-services-file-types-are-safe-to-ignore-on-git


shiftkey:
@KevinRaney thanks!

PR #2825, opened 2018-10-02
New dependency manager delphi for delphi

snakeice:

[boss](https://github.com/HashLoad/boss) use the module folder for the download dependencies and it is not necessary commit 

If this is a new template:

 - **Link to application or project’s homepage**: [boss](https://github.com/HashLoad/boss)

### DIFF
--- a/Delphi.gitignore
+++ b/Delphi.gitignore
@@ -64,3 +64,6 @@ __recovery/
 
 # Castalia statistics file (since XE7 Castalia is distributed with Delphi)
 *.stat
+
+#Boss dependency manager vendor folder
+modules/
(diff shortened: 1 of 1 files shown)


shiftkey:
@snakeice thanks!

jpluimers:
Very few people use the boss dependency manager in Delphi, so I think the `modules/` should not be ignored by default.

I bumped into this when converting projects from a different version control to git, using this .gitignore template, all modules in all projects (each project was in a different repository) were ignored.

shiftkey:
> Very few people use the boss dependency manager in Delphi, so I think the `modules/` should not be ignored by default.

@jpluimers if you feel strongly about this, a PR could be opened to make this opt-in by default (the rule is commented out, but have a comment to explain when you should enable the rule). That's been our practice when a rule is recommended but adoption is varied.

jpluimers:
@shiftkey if you have time for that, please do.

For the past and upcoming months it's over my ears into supporting someone with Alzheimer.

PR #2826, opened 2018-10-03
Ignore backup of the report in VisualStudio

CoperNick8:
Ignore backups created during upgrading to newer version of VS

Example of the files created during upgrade from VS2005 to VS2013:
Alarms - Backup.rdl
Alarms - Backup (2).rdl

Those files are in the same directory as original Alarms.rdl, so it is not possible to filter by folder name.

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -242,6 +242,8 @@ UpgradeLog*.XML
 UpgradeLog*.htm
 ServiceFabricBackup/
 *.rptproj.bak
+* - Backup.rdl
+* - Backup (*).rdl
 
 # SQL Server files
 *.mdf


shiftkey:
@CoperNick8 thanks for the contribution, but I'm not going to take in this change in in it's current form for three main reasons:

 - the `VisualStudio.gitignore` template is already sprinkled with obscure rules, and I'd like to avoid accruing more template debt if possible
 - having to define two rules makes me feel this is a brittle change - if it were a more predictable folder name I'd have no issues
 - the setup to trigger this - upgrading from VS2005 to VS2013 - feels rather uncommon for being in 2018. It's also a one-time operation (hopefully) so I'm not sure its valuable to add to every new repository

PR #2827, opened 2018-10-03
[Godot] Don't ignore export presets 

Calinou:

Export presets are useful to keep around, especially for use in CI/CD.

### DIFF
--- a/Godot.gitignore
+++ b/Godot.gitignore
@@ -1,8 +1,5 @@
-
 # Godot-specific ignores
 .import/
-export.cfg
-export_presets.cfg
 
 # Mono-specific ignores
 .mono/


KellyThomas:
@akien-mga You were part of the conversation at https://github.com/github/gitignore/pull/2613

Are there any benifit to keeping these files out of the repo?

akien-mga:
It's kind of project specific whether you want to commit the presets to your repo, but yeah, I guess it makes sense to default with them not being ignored.

They can contain things like `keystore/release_password` and `apk_expansion/public_key`, so users should just be aware of that before pushing their game source on a public repo.

shiftkey:
> They can contain things like `keystore/release_password` and `apk_expansion/public_key`, so users 
should just be aware of that before pushing their game source on a public repo.

In this case I'd like to keep the rules around (whether commented out or not, I'll leave up to y'all to agree on) and properly documented (so people are aware of what this rule represents).

shiftkey:
I'm not sure where we ended up with this - can we close this out unless we've come to a consensus about what changes we should make?

akien-mga:
I think it can be merged as is, having the export presets is necessary to keep game-specific config. For open source games, it also makes it easier to package them for Linux distros by generating the data.pck file from the command line based on the "Linux/X11" preset (or any other name they'd set for it).

Calinou:
Maybe we could leave the rule commented out, making it obvious it might contain secrets depending on the configured export presets.

There's no way to revoke Android keystore credentials at all, so people need to be extra careful :slightly_smiling_face:

shiftkey:
> As far as I know, there's no way to revoke Android keystore credentials at all, so people need to be extra careful 🙂

If this is the case I'd prefer the rules remain enabled, but with a comment about sharing if the files have been vetted for sensitive details.

Something like this:

> # These files are excluded by default because they can contain credentials and other sensitive information, but otherwise may be useful in *[scenarios/reasons here]
> # You can remove these rules and commit these files if you wish to share them, but be careful!
> export.cfg
> export_presets.cfg

andyfreer:
I think from a security perspective anything containing sensitive credentials **must** be in .gitignore and users warned of the dangers of committing e.g. keystore credentials.

The main issue here is there are useful settings bundled along with those settings that are useful stored in git

Therefore I think the best solution would be to split the credentials parts out into a separate cfg that is ignored and users warned not to commit, then keep this config and remove the .gitignore

Calinou:
@andyfreer We'll do this in a future Godot version, but since this isn't implemented yet, I'll close this PR for now.

kraybit:
What's the current state of this?

Looks like in Godot 4 beta-16 (the year is 2023 now), `export_presets.cfg` is no longer ignored in the default `.gitignore` when creating a new project.

And it looks like credentials are stored in there?

Excerpt from `export_presets.cfg`, after making a MacOS-preset:
```
...
notarization/apple_id_name=""
notarization/apple_id_password=""
...
```

Just looking for insight/clarity here.

- If I worry about storing credentials in a repository, should I add `export_presets.cfg` to git ignore?
- Any other files?

Cheers!

geekley:
> We'll do this in a future Godot version, but since this isn't implemented yet, I'll close this PR for now.

@Calinou since https://github.com/godotengine/godot/pull/76165 was merged and is out (since 4.1.0 I think?), shouldn't this be reopened? Or would it cause issues with people on earlier versions? Still, if so, it should probably be like:
`# export_presets.cfg # Uncomment if not using Godot <version> or later`

Calinou:
> shouldn't this be reopened?

Indeed, this can be considered now but I can't reopen this pull request as I previously removed the branch.

That said, remember that Godot 3.x still has this issue and it's still actively supported. As a result, we may want to start shipping separate `.gitignore` files for Godot 3 and Godot 4.

micycle8778:
can that change be backported to godot 3? seems pretty important

PR #2828, opened 2018-10-06
Added .DS_Store for Mac operating system!

RDIL:
.DS_Store files are useless and annoying so I am preventing them from being committed!


mys721tx:
It would be better to setup a global gitignore with `macOS.gitignore` to ignore all macOS related files all together.

PR #2831, opened 2018-10-09
Add source created for VSCODE

mizamelo:

_TODO_

_TODO_

If this is a new template:


shiftkey:

PR #2833, opened 2018-10-10
Update Julia.gitignore

yakir12:
added the build folder for the documentation.

Documenter.jl creates all the needed files for the documentation in a doc/build folder. Best to ignore that.

https://juliadocs.github.io/Documenter.jl/stable/man/hosting/#.gitignore-1


yakir12:
After some discussion, I think this might not be general enough to be included in the base-gitignore. Closing this now.

PR #2834, opened 2018-10-10
Update JetBrains.gitignore

mmanciop:
Adding *.iml to the ignore list for JetBrains/IntelliJ

_TODO_

_TODO_

If this is a new template:

### DIFF
--- a/Global/JetBrains.gitignore
+++ b/Global/JetBrains.gitignore
@@ -40,6 +40,7 @@ cmake-build-*/
 
 # File-based project format
 *.iws
+*.iml
 
 # IntelliJ
 out/


mmanciop:
Uhm, well, the *.iml is pretty much the default file format of IntelliJ community edition. Does https://intellij-support.jetbrains.com/hc/en-us/community/posts/206878055-Why-is-the-iml-file-stored-in-project-root- provide enough info?

shiftkey:
@mmanciop not really, because `.iml` gets a shout out in [this doc](https://intellij-support.jetbrains.com/hc/en-us/articles/206544839) which we cite at the top of the template:

mmanciop:
I guess I am having some philosophical misunderstanding here. I always interpreted these .gitignore templates as “stuff not to version if you use tech X”.

In all teams I worked, we always excluded IDE setup files from versioning, except those like auto-formatting rules that become have to be imported. *.iml in particular are auto-generated by IntelliJ and seem to me to be equivalent more or less to .project in Eclipse, that is also something I would never version.

What am I missing here?

develar:
I am responsible for configuration store. `*.iml` files should be committed. Since IntelliJ Platform 2017.2 even gradle/maven projects doesn't produce a lot of changed `*.iml` files (auto-generated data stored under system caches now).

> In all teams I worked, we always excluded IDE setup files from versioning

If you follow this rule, then this `.gitignore` is not for you — you can simply ignore the whole `/.idea/` and `*.iml` files. As far I see this file for those, who commit IntelliJ Platform project files into version control but still want to ignore known workspace-only files (`workspace.xml` and other such files).

mmanciop:
Thanks for the explanations.

PR #2835, opened 2018-10-11
Ignore the new Android Bundle generated files

dureyingenieria:
A couple of weeks ago Google added a new format for Android generated files apart from the typical APKs: the [Android Bundle Files](https://developer.android.com/guide/app-bundle/#aab_format). They give more flexibility when uploading apps to the Google Play Console, and they shouldn't be tracked in Git projects.

### DIFF
--- a/Android.gitignore
+++ b/Android.gitignore
@@ -1,6 +1,7 @@
 # Built application files
 *.apk
 *.ap_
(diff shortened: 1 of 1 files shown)


shiftkey:
@dureyingenieria thanks!

PR #2836, opened 2018-10-12
add IntelliJ Rider config file ignore

xiaomoinfo:
add IntelliJ Rider config file ignore


shiftkey:
Thanks, but see [`Global\JetBrains.gitignore`](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore) for the Rider config.

PR #2837, opened 2018-10-13
Ignoring multi-directory a path without ** is not enough

chhh:
Ignoring multi-directory a path without ** is not enough for nested projects

The ignored path with multiple subdirs is not ignored when there are several of them in nested projects.

_TODO_

If this is a new template:


shiftkey:
@chhh thanks!

PR #2838, opened 2018-10-14
[Python] set .python-version as Opt-In

aeroastro:
This PR set `.python-version` as opt-in, and add detail comment like `.ruby-version`.

One of the major objectives of `pyenv` as stated [here](https://github.com/pyenv/pyenv/tree/75ca7613ada2533fb5a37b38eaff8c3b529dfcc9#pyenv-does) is:

* Provide support for per-project Python versions.

To achieve this objective, `.python-version` is used to set a local application-specific Python version. When this file is created in a project, it usually follows that developers lock Python version **by intention, not by accident**. Therefore, IMHO, it should be usually included in the target project.

If we want to support multiple versions, we can omit calling `pyenv local`, or opt-in `.python-version` as an ignored file.

Also, opt-in feature, by which `.python-version` is initially visible with `git`, improves developers' awareness on versions and avoids unnecessary cross version problems. In other words, opt-in helps maintain the situation where `.python-version` is intentionally added to ignores by developers who know well on version issues.

Feedback is welcome.

https://github.com/pyenv/pyenv
https://github.com/pyenv/pyenv/blob/75ca7613ada2533fb5a37b38eaff8c3b529dfcc9/COMMANDS.md#pyenv-local

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -78,7 +78,9 @@ profile_default/
 ipython_config.py
 
 # pyenv
-.python-version
+# for a library or a package, you might want to ignore this file since the code is
+# intended to run in multiple environments; otherwise, check them in:
+# .python-version
 
 # celery beat schedule file
 celerybeat-schedule


shiftkey:
I don't have any strong feelings on this, so I'd love to hear from recent contributors to the template: @SethMichaelLarson @amigcamel @GabrielC101 @jtogrul @raimon49 @cyliang

Also cc @vltr who added this rule initially - I'd love your input about this change!

amigcamel:
I always use `pyenv` with [`pyenv-virtualenv`](https://github.com/pyenv/pyenv-virtualenv). Supposed I have a python 3.7.0 project, the workflow would be:

    pyenv virtualenv 3.7.0 test_env
    pyenv local test_env

Then, a `.python-version` file with `test_env` inside will be generated.  
If I include the file in `git` and someone who happens to have `test_env` which is built on, say, 3.6.5, clones it, the virtualenv will be activated but with the wrong python version. As a result, I always add `.python-version` into `git` to avoid the potential confusion.

vltr:
Even though I added the rule initially, I always check my `.gitignore` to see if `.python-version` is indeed in there; if not, I add it, always. There are three main reasons behind this:

1. `.python-version` is not a visible file, so there's no guarantee that every developer may be aware of it (or use `pyenv` nor even know what this file is all about);
2. If you need to set a Python version to your script, add them to your classifiers inside `setup.py` or, at least, check the current version using `sys.version_info[:2]` or even add that information to your README. Don't rely on the existence of `pyenv` inside a machine since it's not the only tool used (we have `virtualenv` in the stdlib nowadays);
3. I strongly believe that `pyenv` should not be used to "lock" a Python version - see number 2. I use multiple versions of Python inside the same `.python-version` file so `tox` can work nicely, example:
    ```
    $ cat .python-version 
    sanic-boom-3.7.0
    3.5.6
    3.6.6
    3.7.0
    ```
    I can even use Python 2.7.x if needed.

Nightmare mode: picture a `.python-version` file lost inside, let's say, a Node dependency (that uses Python to build if required, using `node-gyp`) and the merely existence of `pyenv` in your machine. Now, go figure what went wrong with that build (let's say, Atom). "Happy troubleshooting!"

In other words, it is good to use a tool like `pyenv` in a daily basis to have all sorts of Python environments for your projects, but your toilet paper will be, in the end, your responsibility.

aeroastro:
Thank you for providing detailed comments and ideas on this pull request, and sorry for the delayed response.

I've read your comments and I've further studied python ecosystem after submitting this PR. Then, I've found out that the idea of this PR is not good.

PR #2839, opened 2018-10-14
vim ignore

hasantayyar:

Broken vim sessions keeps the *.swp files. Those files usually ignored.

http://vimdoc.sourceforge.net/htmldoc/message.html

Project page: https://www.vim.org/


shiftkey:
Thanks, but see [`Global\Vim.gitignore`](https://github.com/github/gitignore/blob/master/Global/Vim.gitignore).

PR #2841, opened 2018-10-17
[Rust] target any directory (not just root)

j-campbell:
Allow `target` to be ignored when the Cargo dir is not the root of the repo

The current version only works if the rust folder is the root of a repository


shiftkey:
> The current version only works if the rust folder is the root of a repository

@j-campbell should we just update that then to drop the first `/` and make it work for nested projects?

cc @schnoddelbotz @waynenilsen @deepak @cnd who have contributed to the template before for their thoughts

waynenilsen:
It would probably be best to try to be consistent with whatever `cargo new` is doing. I do think this makes sense for workspaces as well. The only reason that `cargo new` wouldn't just use `target/` is because of the footgun of some directory like `**/src/**/target/` getting ignored accidentally. This is entirely possible. Perhaps add a negation rule as well to include these sources. This could end up being a PR to cargo as well I would think.

cnd:
I'm agreed with the change.

shiftkey:
@j-campbell does the feedback from @waynenilsen make sense here?

j-campbell:
Not allowing the target folder to be tracked by default would solve this problem? As @waynenilson said there are a few situations where cargo not being the root folder comes up.

shiftkey:
I'm not sure where we ended up with this - can we close this out unless we've come to a consensus about what changes we should make?

shiftkey:
Closing this out as we don't seem to have reached a consensus here.

@j-campbell feel free to open a fresh PR if this is something you still care about.

PR #2842, opened 2018-10-17
vscode directory not ignored

JonasWeismueller:

.vscode/ folder is not being ignored when asterisk is used, so use just . .vscode/ instead
$ git --version
git version 2.17.1

### DIFF
--- a/Global/VisualStudioCode.gitignore
+++ b/Global/VisualStudioCode.gitignore
@@ -1,4 +1,4 @@
-.vscode/*
+.vscode/
 !.vscode/settings.json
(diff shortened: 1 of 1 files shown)


shiftkey:
@JonasWeismueller can you elaborate on _what_ is not being ignored? There are rules below with the `!` to indicate that these files can be committed - are those what you encountered?

JonasWeismueller:
Ah, I guess you are right. ".vscode/settings.json" is in the directory and that was confusing me, thanks for pointing.

PR #2843, opened 2018-10-17
Added Sublime3 project file ignore

hfsf:
Included Sublime3 project extensions in Python.gitignore


shiftkey:
Thanks, but see [`Global/SublimeText.gitignore`](https://github.com/github/gitignore/blob/master/Global/SublimeText.gitignore)

PR #2847, opened 2018-10-18
Added *.VC.db to Unity.gitignore

FutureClemens:

*.VC.db is an auto generated IntelliSense file from VS2015 Update 2 (replacing the .sdf file).

https://stackoverflow.com/questions/36407386/what-is-the-vc-db-file-in-visual-studio-projects


shiftkey:
@FutureClemens thanks!

PR #2502, opened 2017-10-16
Update Unity.gitignore - expanded functionality.

matt360:
**Reasons for making this change:**

Prevents directories: 
`Library`
`Temp`
`Obj`
`Build`
`Builds`
`Assets/AssetStoreTools`
from being included in the remote if the project sits in the `Assets` folder.
The change doesn't change the current functionality.

**Links to documentation supporting these rule changes:** 
The issue came up in my recent professional project. https://github.com/matt360/TrashKittens

PR #2850, opened 2018-10-23
[Unity] Adding VS Code and flexibility

pctroll:

Added support to VS Code cache directory. A lot of developers are starting to use VS Code as their editor of choice because of its lightweight and seamless support on both Windows and Mac (which is the usual mix in Unity development teams nowadays).

The second commit reflects on different development practices for Unity projects in which the project directory is not the root one, but a subdirectory inside it along with other directory for doc and raw files.

> The workspace setting file is located under the `.vscode` folder in your root folder.


Unity.gitignore
```diff
@@ -1,36 +1,37 @@
...
+**/[Ll]ibrary/
+**/[Tt]emp/
+**/[Oo]bj/
+**/[Bb]uild/
+**/[Bb]uilds/
+**/Assets/AssetStoreTools*
 
-# Visual Studio cache directory
-.vs/
+# Visual Studio / Visual Studio Code cache directory
```
shiftkey: This isn't a cache directory, and we have an existing template to show what can be committed in  [`Global\VisualStudioCode.gitignore`](https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore)

Unity.gitignore
```diff
@@ -1,36 +1,37 @@
-[Ll]ibrary/
-[Tt]emp/
-[Oo]bj/
-[Bb]uild/
-[Bb]uilds/
-Assets/AssetStoreTools*
+**/[Ll]ibrary/
```
shiftkey: We've gone through multiple changes to get this pattern right so I'm hesitant to do this again:

 - for the longest time we had a4661a2d39f9cd733ff616dbcc2918f959828d76 which was `/[Ll]ibrary/` to ignore just the root folder
 - then e0c2baea55e9537dbc50e834a090da34c4229841 in #2502 changed it to `*/[Ll]ibrary/` 
 - then dc0ca6d8d5e61dae1bd051f9a99c1414890069f6 in #2532 changed this to `[Ll]ibrary/`  

I'd love to figure out the right pattern here once and for all, rather than endlessly churning these patterns. cc @matzar who has contributed to recent PRs.

Unity.gitignore
```diff
@@ -1,36 +1,37 @@
...
-*.user
-*.userprefs
-*.pidb
-*.booproj
-*.svd
-*.pdb
-*.opendb
+**/ExportedObj/
+**/.consulo/
+**/*.csproj
```
shiftkey: I don't believe we need to prefix `**` to expressions which just map to file extensions. Can you provide more information about this change?

matzar: @shiftkey at the time I've tested the current .gitignore with git BASH v1.9 and v2.* and it worked fine for the ignored folders. The reason for creating a wildcard (`*/`) was that git BASH v1.9 did not seem to work with .gitignore. But it might have been the case that I did not have full admin rights.  

If you were to ask me what's the best practice for defining ignore folders and extension files for .gitignore that would be: 

Folders: 
`[Nn]ameOfIgnoredFolder/` 
`.nameOfIgnoredFolder/` 

Extnesion files: 
`*.extensionOfIgnoredFiles` 

I believe [30d40a](https://github.com/github/gitignore/pull/2850/commits/30d40a225ac122684ce8f8498d923be15e54817a) is unnecessary

Unity.gitignore
matzar: @shiftkey at the time I've tested the current .gitignore (`[Ll]ibrary/`) with git BASH v1.9 and v2.* and it worked fine for the ignored folders. The reason for creating a wildcard (`*/`) was that git BASH v1.9 did not seem to work with `/[Ll]ibrary/` but it might have been the case that I did not have the full admin rights.  But again `[Ll]ibrary/` seems to work with all the vesrions.

If you were to ask me what's the best practice for defining ignore folders and extension files in .gitignore that would be: 

Folders: 
`[Nn]ameOfIgnoredFolder/` 
`.nameOfIgnoredFolder/` 

Extnesion files: 
`*.extensionOfIgnoredFiles` 

I believe [30d40a](https://github.com/github/gitignore/pull/2850/commits/30d40a225ac122684ce8f8498d923be15e54817a) is unnecessary

Unity.gitignore
matzar: I've tested the usage of `**/` and it does not seem to make any difference. For reference I would look at [UE4 .gitignore](https://github.com/github/gitignore/blob/master/UnrealEngine.gitignore) and see how it's used there for whitelisting files

shiftkey:
Closing this out due to inactivity and outstanding review feedback

PR #2851, opened 2018-10-23
Add .pyre directory for Pyre type checker

kitsuyui:
Pyre is a new type-checker for Python developed by facebook in May 2018.
https://pyre-check.org/

`.pyre` directory stores result (including logs) and analysis (cache) for type checking.
This is environment dependent. So it should be ignored in Git.

- https://pyre-check.org/docs/overview.html

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -109,3 +109,6 @@ venv.bak/
 .mypy_cache/
 .dmypy.json
 dmypy.json
+
+# Pyre type checker
+.pyre/


kitsuyui:
In addition, default analysis directory `.pyre` is defined here;
https://github.com/facebook/pyre-check/blob/bf080498fe2b2398b3df72dfaa6c3f945169305e/client/filesystem.py#L47-L55

shiftkey:
@kitsuyui thanks!

PR #2852, opened 2018-10-23
Ignore Vim Swap file (*.swp)

vuonghv:

Ignore Vim Swap file (*.swp)


shiftkey:
Thanks, but see [`Global/Vim.gitignore`](https://github.com/github/gitignore/blob/master/Global/Vim.gitignore)

PR #2853, opened 2018-10-24
(add) Ignore dist folder

doroninds:

_TODO_

_TODO_

If this is a new template:

### DIFF
--- a/Node.gitignore
+++ b/Node.gitignore
@@ -74,3 +74,6 @@ typings/
 
 # FuseBox cache
 .fusebox/
+
+# Distribution folder
+dist/


shiftkey:
@doroninds any links to StackOverflow discussions or other documentation about this change?

doroninds:
@shiftkey [What is the role of src and dist folders?](https://stackoverflow.com/questions/23730882/what-is-the-role-of-src-and-dist-folders)

shiftkey:
@doroninds this feels like a community convention that's not enforced by tooling, and I worry about accidentally causing other people to lose their code because they named the folder wrong and wanted to commit their work anyway. Are there examples of tooling using this directory by default?

In these sorts of situations we'd have the rule commented out, with a header comment indicating the reason for enabling this rule. Does that feel suitable here?

PR #2854, opened 2018-10-26
Ignore the default ipython notebook.

ewosborne:

I often run Jupyter Notebook from within a directory controlled by git. Sometimes I do this just to have a scratch space to test things in. Jupyter creates a notebook with the default name of Untitled.ipynb, which I view as Jupyter's version of a tmp file.  I think this should be ignored by git, as anything I want to keep will end up in a notebook with a real name.

No documentation.

If this is a new template:

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -76,6 +76,7 @@ target/
 # IPython
 profile_default/
 ipython_config.py
+Untitled.ipynb
 
 # pyenv
 .python-version


shiftkey:
@amigcamel as the contributor who added the other IPython rules I'd love your input about this change

amigcamel:
According to the [official documentation](https://jupyter-notebook.readthedocs.io/en/stable/config.html), one can configure the default filename by specifying `ContentsManager.untitled_file`, where the default value is `Untitled`. I personally more prefer not including `Untitled.ipynb` into `.gitignore` as the following reasons:
1. A user can configure the default filename to anything other than `Untitled.ipynb`.
1. *Jupyter notebook* can change the default filename for any reason in the future.
1. `Untitled` might be really meaningful to a user for certain reasons.

I'd say it's more like a developer's responsibility to check if a file should really be versioned by `git` whenever we make changes.

shiftkey:
> I'd say it's more like a developer's responsibility to check if a file should really be versioned by `git` whenever we make changes.

Agreed. I'm going to close this out because I'd rather this file be accidentally committed by the user than omitted from version control.

PR #2855, opened 2018-10-26
Add .idea (JetBrains Rider) directory ignoring Unity gitignore

mirancon:

Jet Brains Rider become popular script editor in Unity


shiftkey:
Thanks, but we have an existing JetBrains template at [`Global/JetBrains.gitignore`](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore) which has more detailed rules about what you can commit to the repository.

PR #2856, opened 2018-10-27
[Java] added *.properties

xxammuxx:
*.properties

_TODO_

_TODO_

If this is a new template:

### DIFF
--- a/Java.gitignore
+++ b/Java.gitignore
@@ -18,6 +18,6 @@
 *.zip
 *.tar.gz
 *.rar
-
+*.properties
 # virtual machine crash logs, see http://www.java.com/en/download/help/error_hotspot.xml
 hs_err_pid*


shiftkey:
@xxammuxx any links to StackOverflow discussions or other documentation about this change?

Tristan971:
The Spring project, Java's most popular web application framework (see https://dzone.com/articles/most-popular-java-web-frameworks for example), relies on those heavily ; they should definitely not be ignored. See : https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-external-config.html

shiftkey:
@Tristan971 thanks for the link. And these paths can be _safely_ ignored from version control? Users won't expect these files to be committed and tracked?

shiftkey:
@xxammuxx given the link from @Tristan971 above, does it still make sense to add this rule?

Tristan971:
No they indeed cannot be ignored and should 100% be version controlled.
Unfortunately I do not believe this PR is good in the long term.

xxammuxx:
@Tristan971 is right about it. This can be ignored.

PR #2809, opened 2018-09-14
Create Vuejs.gitignore

mogya:
**Reasons for making this change:**

To add git-ignore for Vue.js

**Links to documentation supporting these rule changes:**

https://github.com/vuejs-templates/webpack/blob/develop/.gitignore

**Link to application or project’s homepage**

https://vuejs.org/index.html

PR #2712, opened 2018-05-24
Rename Magento.gitignore to Magento1.gitignore

Nolwennig:

PR #2361, opened 2017-04-23
Create Drupal8.gitignore

lpalgarvio:
Create a new Drupal8.gitignore for Drupal 8 with the code from dvcs/gitignore
See the latest [gitignore](https://github.com/dvcs/gitignore/blob/master/templates/Drupal8.gitignore)
And the [pull request](https://github.com/dvcs/gitignore/pull/4)

Drupal 8's directory structure is fundamentally different from past versions.

PR #2360, opened 2017-04-23
Update and rename Drupal.gitignore to Drupal7.gitignore

lpalgarvio:
Replace the current Drupal.gitignore for Drupal 7 with the one patched at dvcs/gitignore and rename the file to Drupal7.gitignore
See the latest [gitignore](https://github.com/dvcs/gitignore/blob/master/templates/Drupal7.gitignore)
And the [pull request explaining the change](https://github.com/dvcs/gitignore/pull/3)

PR #2857, opened 2018-10-27
introduce community templates

shiftkey:
This PR is my attempt to address some limits of how templates are currently organised, so I can potentially land some newer templates and not lose my sanity.

[**Rendered README**](https://github.com/github/gitignore/blob/experiment-ecosystem-groupings/README.md)

### But why?

Here's my current hell:

 - dumping all templates at the root leads to a mess - discoverability is hard, maintenance is hard, etc etc
 - introducing version-specific templates leads to even more mess
 - we have a lot of niche templates which would like to be added, but I've been hesitant to merge these as-is 

This PR introduces an `community` subdirectory, to help better organise templates. The core flows I want to address are:

 - there are major changes between versions of a template
 - specialised templates based on a product or framework

The root of the repository will remain focused on active and popular templates, but organising things means we can support a more diverse range of templates in the future.

### Supporting different versions of a framework

The first example is the current situation with Drupal. Here's two PRs waiting for my review:

 - #2361 - Add Drupal 8
 - #2360 - Update and rename Drupal.gitignore to Drupal7.gitignore

Rather than have two versioned templates at the root for `Drupal8.gitignore` and `Drupal7.gitignore` we should move to this pattern for major frameworks:

 - maintain `Drupal.gitignore` as the evergreen "latest" template
 -  define an `community/Python/Drupal7.gitignore` for those who need to access the older template

At some point in the future, if Drupal 9 has significant changes, we can go through the flow again:

 - `Drupal.gitignore` is updated to support Drupal 9's rules
 - `community/Python/Drupal8.gitignore` is added with the old rules

Another example is this PR:

 - #2712 - Rename Magento.gitignore to Magento1.gitignore 

In this PR I've instead gone with this flow:

 -  define an `community/PHP/Magento1.gitignore` for those who need to access the older template
 - someone later can come in and upgrade `Magento.gitignore` to the v2 format

### Supporting specialized frameworks

A request to add `Vue.js` was added in #2809 but the template itself is just four rules:

```
# this duplicates Node.gitignore
node_modules
# this duplicates Global/macOS.gitignore
.DS_Store
# unclear what these are for
docs/_book
test/
```

This instead could live under `community/JavaScript/Vue.gitignore` and look like this:

```
# gitignore template for Vue.js projects
#
# Recommended template: Node.gitignore

# TODO: where does this rule come from?
docs/_book

# TODO: where does this rule come from?
test/
```

As a first pass expanding on this idea, I've been able to **resolve 30+ outstanding PRs** (some overlap because they propose the same change)

### TODO

 - [x] find and incorporate PRs that fit this pattern
 - [x] gather feedback


foca:
I like having the version specific templates separate, but would go a step forward and put the newest template as well in that directory and make the top level, “unversioned” file just a symlink to the latest template.

shiftkey:
@foca interesting idea. I'd actually prefer to deprecate the use of symlinks because Windows support for it is still suboptimal, but I'll keep that in mind.

foca:
Ah, that’s a good point. Shouldn’t respond to issues before my first
coffee. Disregard, then :)

Lin-Buo-Ren:
I wonder if you've decided the leading letter's case of the `community` subdirectory?

lpalgarvio:
good work =)

lpalgarvio:
@shiftkey just a minor correction, drupal is php but was by mistake placed under python. can you fix it? id provide a patch, but i have my week and weekend overdosed with work and responsabilities

shiftkey:
@lpalgarvio I'm also swamped (and trying to wrap things up before going on holidays tomorrow) so either it'll get done after that or someone can submit a change...

PR #2859, opened 2018-10-28
[Terraform] added plan files

saidelmark:

It is very convenient to save plans to a file, but this file should not be commited: they often contain secrets, credentials, etc.

https://www.terraform.io/docs/commands/plan.html#out-path

### DIFF
--- a/Terraform.gitignore
+++ b/Terraform.gitignore
@@ -24,3 +24,6 @@ override.tf.json
 # Include override files you do wish to add to version control using negated pattern
 #
 # !example_override.tf
+
+# Ignore files created by `terraform plan -out=...`
+*plan


Terraform.gitignore
```diff
@@ -24,3 +24,6 @@ override.tf.json
 # Include override files you do wish to add to version control using negated pattern
 #
 # !example_override.tf
+
+# Ignore files created by `terraform plan -out=...`
+*plan
```
shiftkey: Is this rule too vague? What's an example of the default file names here?

saidelmark: There is no default value (plans are not saved by default).

But when developers do save plans, they usually (well, I noticed such usage and never met anything else) name the file "**plan**" or "blablabla.**plan**".

shiftkey:
@alastairtree @obounaim @mastertinner @jkodroff I'd love your thoughts on whether this change makes sense

rubberduck203:
As a heavy Terraform user, this change makes sense.
All of our terraform projects at work contain a `*.plan` rule.

mastertinner:
Personally, I haven't used this feature yet. What @shiftkey suggests makes sense when reading the documentation and the warnings about these files. I agree that we should have such a rule. However, I'm not sure if it's the right glob. I have seen files like `*.tfplan` being used in other places (e.g. https://github.com/hashicorp/terraform/issues/7976) and I think that would be bore stringent and a better glob. But I'm not sure what's the actual community best practice here.

rubberduck203:
What about 

```
plan
*.plan
*.tfplan
```
?

That’s pretty specific and should cover 99% of use cases. 
I’ve personally never seen `*.tfplan`, but it follows common terraform naming conventions.
Anyway, I just happened to notice this discussion and wanted to confirm that `*.plan` is a common file extension for plans and you definitely don’t want them getting checked in.

mastertinner:
I would stick with `*.tfplan only. It seems to have become the convention to name Terraform related files `*.tf*`.

rubberduck203:
@mastertinner, you have two actual TF users here telling you that `*.plan` is actual convention in use.

jkodroff:
@shiftkey I'm cool with anything that improves the current default.  Just let me know what the consensus is.

FnTm:
For what its worth - my opinion as a recent TF user would be to go with the `.tfplan` option, as I've not noticed docs actually saying that `.plan` should be the suffix used. I have actually seen a couple of Github Issues mentioning the `.tfplan` option, which is why I used it myself afterward.
A disclaimer though - I may not have been using TF for long enough to see the conventions in other places yet.

PR #2863, opened 2018-10-29
ignore .DS_Store from repos using Node.gitignore

williamsthing:

.DS_Store files are currently useless for GitHub repositories.

?


shiftkey:
Thanks, but we're tracking this rule and many other macOS-specific in [`Global/macOS.gitignore`](https://github.com/github/gitignore/blob/master/Global/macOS.gitignore). Please pull these rules into your project-specific template.

PR #2864, opened 2018-10-31
Add Xilinx Vivado Design Suite 

lkxajd:
…ad.html)

Maybe there is somebody that want to use Vivado to develop their FPGAs.

None.
Wrote it because of personal use.

If this is a new template:

 - **Link to application or project’s homepage**: https://www.xilinx.com/support/download.html


shiftkey:
Closed in favour of #2885

PR #2865, opened 2018-10-31
Add VS Code for Python

cseas:

The folder added stores only local settings and is not necessary for a project to be shared with someone.

If this is a new template:

 - **Link to application or project’s homepage**: https://github.com/Microsoft/vscode

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -112,3 +112,6 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+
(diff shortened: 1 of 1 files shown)


shiftkey:
Thanks, but see [`Global\VisualStudioCode.gitignore`](https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore) which shows what files can be stored in version control and shared.

PR #2866, opened 2018-11-01
Add a leading ** to target/

jamesray1:
This is handy for projects with more than one target directory.

_TODO_


waynenilsen:
I actually think that [this comment](https://github.com/rust-lang/cargo/issues/6245#issuecomment-438331437) is the best long term solution.

> Having a submodule `src/target/mod.rs` doesn't seem like an entirely uncommon module name. An alternative would be to have cargo new create a sub-`.gitignore` inside new crates folders when run in an existing git repo (maybe even detecting if you have a workspace and only adding it if not).

In the short term I think it makes the most sense to leave the current behavior and include an option for workspace configuration. 

```
# Generated by Cargo, will have compiled files and executables
# Use /**/target for working with workspaces, otherwise use /target 
# WARNING if you have a submodule named target it will need another gitignore file to include it with this rule 
#   https://doc.rust-lang.org/book/second-edition/ch14-03-cargo-workspaces.html
# /**/target
/target
```

It is a bit verbose but more documentation is better than less I find.

tchajed:
gitignore patterns by default match the basename in any subdirectory; `/**/target` matches the same paths as `target`. In addition, a trailing slash is slightly better since it only matches directories (and expresses that intent better).

In summary: the current line is good (`/target/`) by default, and the recommendation should be `target/` for working with workspaces.

Also, as an aside, isn't `!src/**/target/` plus `target/` also a pretty good solution, albeit only relevant for a small subset of users? I'll admit I don't really understand what directory layouts look like when using workspaces.

waynenilsen:
Here are the different directory structures that we are interested in supporting 

workspace with crates foo and bar, crate foo has a module target that should be included

```
foo
|-src
    |-target (should not be ignored)
|-target (should be ignored)
|-Cargo.toml
bar
|-src
   |-baz
      |-target (should not be ignored)
|-target (should be ignored)
|-Cargo.toml
Cargo.toml
.gitignore
```

standard

```
src
|-target (should not be ignored)
target (should be ignored)
Cargo.toml
.gitignore
```


I do think the @tchajed rules are what we want

```
/target/
!src/**/target/
```

Just checking, do the gitignore rules above ignore and not ignore as we would like?

tchajed:
Awesome, thanks for the example! I missed one subtlety in the negative rule: `**` only works for full path matches, so the correct rule is `/**/src/**/target/`. The correct rules are:

```
target/
!/**/src/**/target/
```

Here's the behavior for your first example @waynenilsen (this is with no added files):
```
$ tree
.
├── Cargo.toml
├── bar
│   ├── src
│   │   └── baz
│   │       └── target
│   │           └── mod.rs
│   └── target
│       └── lib.o
├── foo
│   ├── Cargo.toml
│   ├── src
│   │   └── target
│   │       └── mod.rs
│   └── target
│       └── lib.o
└── src
    └── target
        └── mod.rs

11 directories, 7 files

$ git status -u
...
Untracked files:

	.gitignore
	Cargo.toml
	bar/src/baz/target/mod.rs
	foo/Cargo.toml
	foo/src/target/mod.rs
	src/target/mod.rs
...
```

and for the simpler single-crate example:

```
$ tree
.
├── Cargo.toml
├── src
│   └── target
│       └── mod.rs
└── target
    └── lib.o

3 directories, 3 files

$ git status -u
...
Untracked files:

	.gitignore
	Cargo.toml
	src/target/mod.rs
...
```

Nemo157:
Note that workspaces do _not_ need to ignore `target` in subdirectories, a workspace will use a single `target` directory at the workspace root for all crates. It's only when you have a repository that includes multiple crates that are not in a workspace that you will need to ignore additional `target` directories (which seems like a niche usecase to me).

tchajed:
@Nemo157 that makes more sense and lines up with the Cargo workspace documentation. Perhaps it would be bad to correctly handle nested `target` directories, since a workspace is a better solution.

PR #2867, opened 2018-11-02
Added ignore for basic editor level settings for error in path, env etc.

AngrySoilder:
Added Ignore for Basic Editor Level Changes Which Usually Causes:-
Env Error or Not Found
Uncompatible system level paths
Non existing keys etc.

Added For:-
Sublime-text
vscode

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -112,3 +112,9 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+
+# vscode stuff igonred
+.vscode/*
+
+# sublime stuff ignored
+*.sublime-*


shiftkey:
Thanks, but to avoid duplicating rules we have existing templates for [`Global/VisualStudioCode.gitignore`](https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore) and [`Global/SublimeText.gitignore`](https://github.com/github/gitignore/blob/master/Global/SublimeText.gitignore) which I recommend adding to your project-specific templates.

PR #2871, opened 2018-11-02
Add `ARM` and `ARM64` to Visual Studio's ignore file.

cwize1:

Recent editions of Visual Studio now permit building Windows drivers and Windows desktop apps for the ARM and ARM64 architectures. Visual Studio uses the architecture name for some of its build directories, 

This change follows existing precedent used for the `x64` and `x86` platforms.

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -20,6 +20,8 @@
 [Rr]eleases/
 x64/
 x86/
+[Aa][Rr][Mm]/
+[Aa][Rr][Mm]64/
 bld/
 [Bb]in/
 [Oo]bj/


shiftkey:
@cwize1 I'm a bit nervous about ignoring _every_ folder named some variation of `arm`/`arm64`, but given the precedence with `x86`/`x64` and no complaints about that I guess I'm fine with it.

superyyrrzz:
I renamed my folder storing [ARM templates](https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/overview) from `arm` to `arm-temlpate` for this.

PR #2875, opened 2018-11-08
Add "network-security.data"

watilde:
emacs adds `network-security.data` file automatically and the file should not be committed.

+ http://network-security.data


shiftkey:
> `http://network-security.data` 

@watilde this link doesn't work - any links to StackOverflow discussions or other documentation about this change?

shiftkey:
@watilde thanks!

PR #2877, opened 2018-11-08
Added CMakeLists.txt.user file to a CMake .gitignore

skident:
Added new file which should be present in CMake .gitignore


shiftkey:
@skident any links to StackOverflow discussions or other documentation about this change?

cwfitzgerald:
Both QTCreator and Visual Studio use CMakeLists.txt.user to store user specific data for their respective cmake integrations.

shiftkey:
@skident thanks!

PR #2883, opened 2018-11-22
[JetBrains] starts to ignore all *.iml and modules.xml files

MrVine:

`*.iml` and `.idea/modules.xml` files will be generated on import so should be ignored by default


shiftkey:
@MrVine that link is the same one that we link to in the header of the template - can you elaborate on why this change is needed?

shiftkey:
Closing this out due to inactivity and outstanding questions

PR #2886, opened 2018-11-23
Added .gradle folder to unity.gitignore

BabyBluee:
The newest version of Unity creates a .gradle folder when creating an Android build, that is just unnecessary clutter.

This small addition helps solve that problem by ignoring it.


shiftkey:
Looks like this was already covered by #2906 - thanks for the contribution anyway @BabyBluee!

PR #2892, opened 2018-11-27
Remove trailing whitespace in Node.gitignore

22a:

We don't need this whitespace.

NA


shiftkey:
Looks like this was already covered by #2910 - thanks for the contribution anyway @22a!

PR #2893, opened 2018-11-27
Create Magento2.gitignore

rafaelstz:

There two very different versions of Magento, version 1 and 2, each one has a specific gitignore version.

The original Magento 2 gitignore file is not congruent because it's versioning many auto-generated files and ignoring files which are versioned by default. Also you can see that it's ignoring folder and files which are not related to the project, but related with the machine, environment and IDE (e.g. `.vscode`, `.idea`, `.DS_Store`), following the best practices of gitignore we have to ignore them into some global gitignore and not in a project scope.

> Basically it's a right way gitignore for that project.

https://git-scm.com/docs/gitignore
https://amasty.com/knowledge-base/topic-magento-related-questions.html#5803

https://github.com/magento/magento2

---

I created this PR based of that: https://github.com/github/gitignore/pull/1840


shiftkey:
@rafaelstz thanks!

PR #2894, opened 2018-11-27
remove rsconnect folder from R.gitignore

colearendt:

The `rsconnect` folder does not contain any sensitive information (the deployment token is stored elsewhere). Moreover, it is recommended to be _tracked_ by VCS (and therefore _not_ in gitignore) in order to enable collaboration on published content.

http://docs.rstudio.com/connect/user/publishing.html#collaboration

### DIFF
--- a/R.gitignore
+++ b/R.gitignore
@@ -31,6 +31,3 @@ vignettes/*.pdf
 # Temporary files created by R markdown
 *.utf8.md
 *.knit.md
-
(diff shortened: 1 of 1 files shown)


shiftkey:
@colearendt thanks for the link! I've pasted the key paragraph in for reference:

> This `rsconnect` folder should be added and committed into version control. It does not contain any private or secure information. However, it does contain the URL for the RStudio Connect server and the content URL. This information allows future publications to easily target the same endpoint.

PR #2896, opened 2018-11-29
feat: Add missing executables and generated

erdtsksn:

There are missing sections that should be added to .gitignore

https://gohugo.io/hugo-pipes/introduction/

### DIFF
--- a/community/Golang/Hugo.gitignore
+++ b/community/Golang/Hugo.gitignore
@@ -1,8 +1,12 @@
+### Hugo ###
 # gitginore template for Hugo projects
 # website: https://gohugo.io
 
 # generated files by hugo
 /public/
+/resources/_gen/
 
 # executable may be added to repository
 hugo.exe
+hugo.darwin
+hugo.linux


shiftkey:
@erdtsksn any links to StackOverflow discussions or other documentation about this change?

erdtsksn:
> @erdtsksn any links to StackOverflow discussions or other documentation about this change?

https://discourse.gohugo.io/t/why-does-hugo-create-the-resources-directory/13021/4

https://discourse.gohugo.io/t/why-does-hugo-create-the-resources-directory/13021/3
**bep** who is one of the core contributor said you may add it to git if you like. But also he stated it is a **resource cache**. In my opinion it should be in gitignore if it is cache.

https://github.com/netlify-templates/one-click-hugo-cms/tree/master/bin
Netlify have a one-click template to create hugo websites. As you may see in the link above. They include hugo binaries inside bin folder.

shiftkey:
@erdtsksn great, thanks for the extra context!

PR #2897, opened 2018-11-29
The global .gitignore ignores DLL, this is a problem with plugins in …

allanolivei:
…Unity 3D

_TODO_

_TODO_

If this is a new template:

### DIFF
--- a/Unity.gitignore
+++ b/Unity.gitignore
@@ -39,3 +39,5 @@ sysinfo.txt
 # Crashlytics generated file
 Assets/StreamingAssets/crashlytics-build.properties
 
+#Do not ignore plugins
+!/**/*.dll


Unity.gitignore
```diff
@@ -39,3 +39,5 @@ sysinfo.txt
 # Crashlytics generated file
 Assets/StreamingAssets/crashlytics-build.properties
 
+#Do not ignore plugins
+!/**/*.dll
```
shiftkey: This came up in #1603 and after a bit of discussion I think I was concerned about this being recursively applied to all paths - can this rule be made more specific?

allanolivei:
```suggestion
!/Assets/Plugins/**/*.dll
```

shiftkey: I guess? I'd need some input from people who are more familiar with Unity at this point.

cc @ksmit799 @theDrake @1nfinite9 @owenblacker @FutureClemens @kilian442 @mariuz for feedback (peeps who have made changes to the template over the last year)

FutureClemens: DLLs in a Unity project are allowed to be placed within any "Plugins" directory or "Plugins" sub-directory. Also, the "Plugins" directory itself does not need to be placed directly in the Assets folder and can be placed in any directory. And there might be multiple plugins directories in your project.

Therefore, valid places are:
Assets\[0-n Directories]\Plugins\[0-n Directories]\plugin.dll

Examples from existing plugins, that are working in my projects (Unity 2017 & 2018):
Assets\Plugins\x86_64\plugin.dll
Assets\Plugins\WSA\PhoneSDK81\x86\plugin.dll
Assets\MyTests\Plugins\x86_64\Test.dll

I am not sure if a line like this makes sense at all for a unity gitignore file, but if you do it correctly, it at least does not do any harm.

allanolivei:
```suggestion
!/Assets/**/Plugins/**/*.dll
```

capture

Thanks @FutureClemens.

ksmit799: I'm not sure if this makes sense within the scope of the global Unity gitignore, seeing as the current Unity gitignore doesn't ignore any dll files by itself. The fact that an external gitignore is impacting on Unity's makes this more of an edge case then something that should be applied globally.

shiftkey:
Closing this out due to inactivity and outstanding review feedback.

If someone else thinks these changes would still add value to the template, please open a fresh PR.

PR #2898, opened 2018-11-30
[Sass] Add .scssc to ignore

devinrhode2:


shiftkey:
@devinrhode2 any links to StackOverflow discussions or other documentation about this change?

devinrhode2:
It's Generated from Ruby sass.. You could try searching SO, use Ruby sass,
or maybe search their codebase? Not really interested in finding
evidence/proof it was getting generated..
https://www.google.com/search?q=gitignore+sass+files+scssc+sass-cache&oq=gitignore++sass+files+scssc+sass-cache&aqs=chrome..69i57.48267j1j4&client=ms-android-google&sourceid=chrome-mobile&ie=UTF-8

shiftkey:
The best example I could find was the [`.gitignore`](https://github.com/sass/sass-site/blob/master/.gitignore) from `https://github.com/sass/sass-site` and it's rule is `*.scssc` which is different to what you've included here. I tried to `git blame` back in history but it wasn't clear to me where that came from.

The top 5 google results for `gitignore ruby-sass ".scssc"` all look like this too:



> Not really interested in finding evidence/proof it was getting generated..

I just like to have external resources to cite as a sanity check that the rule is correct and the files can be safely ignored in version control. I've been bitten before by merging in new rules that weren't fully understood and ended up impacting users.

shiftkey:
Closing this out due to inactivity and lack of clarity about the value of the proposed change.

PR #2900, opened 2018-12-02
Improve comment capitalization consistency

theDrake:

One comment in Unity.gitignore has capitalization that differs from all the others. This bothers OCD people like me, so I figured we might as well fix it.


shiftkey:
@theDrake thanks!

PR #2902, opened 2018-12-03
[JetBrains] Add ignores for file-based projects

scadgek:
When IntelliJ project is created as a file-based (i.e. without `.idea` folder, but with `.iws`, `.iml` and `.ipr` files), and this is a Gradle or Maven project with auto-import, should ignore them as well for the same reason we ignore `.idea` folder content in that case.


shiftkey:
@scadgek thanks!

PR #2903, opened 2018-12-03
[VisualStudio] Added line to exclude coverity analysis directory.

hype8912:

Coverity Desktop analysis folder was not being excluded but git

I don't have any outside documentation to link to other than all the developers here have Coverity Desktop Analysis loaded in Visual Studio and it creates a .cov hidden directory in the top level of the source code similar to the .vs cache directory. I tried searching to find some documentation of it online but I didn't find any. This is a line we add to all our git ignore files to exclude the Coverity folder.

If this is a new template:

 - **Link to application or project’s homepage**: Not a new template.

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -336,3 +336,6 @@ ASALocalRun/
 
 # Local History for Visual Studio
 .localhistory/
+
+# Coverity Analysis Folder
+.cov/


shiftkey:
@hype8912 my only concern with this PR as-is is that it's not clear to me that the folder can be safely ignored. The `.vs` cache is used for generated files, but I don't have enough context on this to understand whether the `.cov` folder benefits from being in version control (so files are shared between developers on a project) like some VS extensions.

Can you elaborate on what's being stored in `.cov`?

hype8912:
I reached out to my contact at Coverity for a formal explanation of the directory. I will get with back with you on this.

shiftkey:

@hype8912 if you do hear back and have more context, please open a fresh pull request.

PR #2904, opened 2018-12-06
Only unignore directories that have at least a character ending with .cache

wldevries:

We have an Fmod project in our C# repository that generates a `.cache` directory containing cache files. After updating the `.gitignore` file from this repository it is included in the git changes due to PR #1503.

The `?` makes sure we only include directories that end with `.cache`, but start with at least one other character.


shiftkey:
@wldevries thanks!

PR #2906, opened 2018-12-12
[Unity] Ignore .gradle/ build cache

ksmit799:

Newer version of Unity now use Gradle as the default build system for Android builds, thus creating a .gradle/ cache directory that sticks around after builds. This directory is merely a local cache, and serves no purpose being tracked by Git.

https://docs.gradle.org/current/userguide/directory_layout.html


shiftkey:
Tentatively pulling this in but [`Gradle.gitignore`](https://github.com/github/gitignore/blob/master/Gradle.gitignore) might be the preferred option here (to avoid duplicating rules across templates).

PR #2907, opened 2018-12-13
Ignore health check temp database

HassanHashemi:
Ignore liveness db which belongs to:
https://github.com/Xabaril/BeatPulse

It's not necessary to commit local (development time db) to source control.

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -223,6 +223,7 @@ ClientBin/
 *.pfx
 *.publishsettings
 orleans.codegen.cs
+livenessdb
 
 # Including strong name files can present a security risk
 # (https://github.com/github/gitignore/pull/2483#issue-259490424)


shiftkey:
@HassanHashemi any links to StackOverflow discussions or other documentation about this change?

I see the project's `gitignore` has this rule:

```
*livenessdb
```

And this is mentioned in the README:

> All liveness results are stored into a SqLite database persisted to disk with _livenessdb_ name.

But it'd be nice to confirm this rule reflects that the right files are being targeted.

Also, feel free to extract this to it's own section and add a comment indicating this is specific to an extension.

HassanHashemi:
@shiftkey I applied your review, However i am pretty sure about `livenessdb` rule. 
No idea about `*livenessdb`, So i think it's safe to have `livenessdb` for now.

@unaizorrilla any comments?

unaizorrilla:
Hi @HassanHashemi 
Well, the database name for healthchecks ui  on release version is "healthchecksdb" not livenessdb.
Is safe to ignore it, it is created automatically on each execution.

shiftkey:
> Well, the database name for healthchecks ui on release version is "healthchecksdb" not livenessdb.
Is safe to ignore it, it is created automatically on each execution.

@unaizorrilla just a heads up: this is still referred to as `livenessdb` in the README and a couple of other places



@HassanHashemi are you fine with updating the rule?

shiftkey:
@unaizorrilla thanks for the details about the new repository.

I'd recommend making it clearer that the old repository is no longer active and where users should go - GitHub has the ability to [archive repositories](https://help.github.com/articles/archiving-repositories/) to make this visible to others (and prevent them submitting issues or PRs to the wrong location).

VisualStudio.gitignore
```diff
@@ -336,3 +336,6 @@ ASALocalRun/
 
 # Local History for Visual Studio
 .localhistory/
+
+#BeatPulse healthcheck temp database
```
shiftkey: whitespace after the `#` for readability

PR #2908, opened 2018-12-14
Add ignorer to .env.test variable in NodeJS

lmedson:

while working on my project, I realized that I always needed to add an ignore in the .env variable of tests, where I call .env.test, and I thought that many other people using this pattern go through the same thing, so I decided to try to sub.

I dont have a specific documentation link talking about having to use this pattern, but it's almost a convention, separating .env from tests by .env.test

If this is a new template:

None

### DIFF
--- a/Node.gitignore
+++ b/Node.gitignore
@@ -56,6 +56,7 @@ typings/
 
 # dotenv environment variables file
 .env
+.env.test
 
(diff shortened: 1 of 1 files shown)


shiftkey:
@lmedson thanks!

PR #2910, opened 2018-12-16
[Node] fix wsp

kesoji:

Fixed wsp to match to other lines.

None


shiftkey:
@kesoji thanks!

PR #2911, opened 2018-12-16
Added PyCharm project directory in Python's gitignore.

dynamicmindset:

PyCharm creates a .idea directory for storing project settings, so this line is for developers using PyCharm.

https://www.jetbrains.com/help/pycharm/project-and-ide-settings.html
"Project settings are stored with each specific project as a set of xml files under the .idea folder. If you specify the template project settings, these settings will be automatically used for each newly created project."

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -103,6 +103,9 @@ venv.bak/
 # Rope project settings
 .ropeproject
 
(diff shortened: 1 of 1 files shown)


shiftkey:

BrianMSheldon:

PR #2912, opened 2018-12-18
Unity Logs

FutureClemens:

Unity 2018.3 generates logs after updating their packages. It is good to know that this happens, but this is not part of the project and therefore should not be stored in the github imho.

Cannot find any resources, but if you upgrade a project from 2018.2 to 2018.3, a message from the engine pops up and informs you about it and the directory and logfile is created.


shiftkey:
@FutureClemens thanks!

PR #2915, opened 2018-12-22
[CakePHP] Ignoring user's migration lock file

ahmyi:
Ignoring lock files htat are generated after cakephp migration is done

### DIFF
--- a/CakePHP.gitignore
+++ b/CakePHP.gitignore
@@ -17,9 +17,13 @@
 /logs/*
 !/logs/empty
 
+/config/Migrations/schema-dump-default.lock
+
 # CakePHP 2
 
 /app/tmp/*
 /app/Config/core.php
 /app/Config/database.php
 /vendors/*
+
+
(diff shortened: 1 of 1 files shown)


shiftkey:
@ahmyi any links to StackOverflow discussions or other documentation about this change?

ahmyi:
https://github.com/cakephp/migrations/issues/252

ahmyi:
https://book.cakephp.org/3.0/en/migrations.html#skipping-the-schema-lock-file-generation

ravage84:
For development, the schema.lock file should be committed. If more than one person works on a project, a difference in the database environment (server version, server/client configuration) can result in a diff in the schema.lock file (compared to other developer's). The next time you [bake a migration diff](https://book.cakephp.org/migrations/2.x/en/#generating-a-diff-between-two-database-states), you potentially get a different diff (migration script) than your peer developers get.  A committed ``schema.lock`` can help you detect that.

As for when you deploy your code through git, it probably shouldn't or you should use ``--no-lock`` option when executing the migrations (as @ahmyi already linked to, [link moved since](https://book.cakephp.org/migrations/2.x/en/#skipping-the-schema-lock-file-generation))).

So, I would recommend to not exclude it.

PR #2916, opened 2018-12-25
Ignore vscode settings

hammacktony:

Ignore vscode settings folder


shiftkey:
Thanks, but see [`Global\VisualStudioCode.gitignore`](https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore).

PR #2918, opened 2018-12-31
Unified base folder name format

johnsoncodehk:
I don't think there is a need to adapt to capitalization, just to make it look more uniform.


shiftkey:
@johnsoncodehk thanks!

PR #2919, opened 2019-01-02
Add Flask.gitignore

Tuzgai:

There is currently not a gitignore for Flask.

I'm working through the Flask tutorial and found the gitignore here:
http://flask.pocoo.org/docs/1.0/tutorial/layout/

If this is a new template:

 - **Link to application or project’s homepage**: http://flask.pocoo.org

### DIFF
--- a/Flask.gitignore
+++ b/Flask.gitignore
@@ -1,0 +1,17 @@
+# Reference for Flask gitignore: http://flask.pocoo.org/docs/1.0/tutorial/layout/
+
+.gitignore
+venv/
+
+*.pyc
+__pycache__/
+
+instance/
+
+.pytest_cache/
+.coverage
+htmlcov/
+
+dist/
+build/
+*.egg-info/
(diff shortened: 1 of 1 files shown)


jwflory:
For what it's worth, all these changes and the Flask specific things are already contained in the `Python.gitignore`:

https://github.com/github/gitignore/blob/e931ef7f3e7d8f7aa0e784c14bd291ad4448b1ab/Python.gitignore#L65-L67

wirecat:
👋 @Tuzgai since we have an entry in in the [Python.gitignore](https://github.com/github/gitignore/blob/main/Python.gitignore#L64-L66) for some of these values already (like @jwflory also pointed out), I would prefer we evaluate Pythong.gitignore for any future flask extension.

Thank you for contributing! I'm going to close this one out since it's been open a while and we do have this section already. Feel free to re-open or spin up a new PR for further discussion.

PR #2921, opened 2019-01-03
google-services.json

aslamanver:

google-services.json file should be commented on default if they want they will change otherwise it'll destroy the project

### DIFF
--- a/Android.gitignore
+++ b/Android.gitignore
@@ -52,7 +52,7 @@ captures/
 .externalNativeBuild
 
 # Google Services (e.g. APIs or Firebase)
-google-services.json
+# google-services.json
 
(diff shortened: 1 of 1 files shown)


shiftkey:
@aslamanver any links to StackOverflow discussions or other documentation about this change?

aslamanver:
> @aslamanver any links to StackOverflow discussions or other documentation about this change?

I faced a big issue on using this code, because at the first time when I updated my .gitignore it removed my google-services.json file. So my advice is better we can keep the code in comment if anyone want to un-comment it, so they can do it and they know about it.

shiftkey:
@aslamanver works for me!

PR #2923, opened 2019-01-06
✨ Create iOS.gitignore

erikhuizinga:

By default, the template for Android ignores the Android equivalent of this file: `google-services.json`. Because this file contains API keys, it should be ignored by default.

https://github.com/github/gitignore/blob/18654b47c86bef38add5cd091ec2dac7d6e6e37b/Android.gitignore#L54-L55

If this is a new template:

   - https://developer.apple.com/
   - https://firebase.google.com/docs/ios/setup
   - https://firebase.google.com/docs/flutter/setup#configure_an_ios_app

### DIFF
--- a/iOS.gitignore
+++ b/iOS.gitignore
@@ -1,0 +1,2 @@
+# Google Services (e.g. APIs or Firebase)
+GoogleService-Info.plist


Coeur:
I am not in favor of creating an `iOS.gitignore` for that.
Firebase on iOS is most often installed through `pod 'Firebase'`: https://cocoapods.org/pods/Firebase
And being from that readme or from the official doc, there is never a mention of ignoring such file.

Hiding API keys to developers is a concern that could apply for a thousand of different third-parties, not just Firebase, and not just iOS: a tech architect would eventually put all the sensitive data in a folder of its own.

erikhuizinga:
Just cleaning up my old open PRs.

PR #2924, opened 2019-01-07
Update Hugo.gitignore

BhaveshSGupta:
Correcting Spelling for ignore and adding folder when run with --disableFastRender

### DIFF
--- a/community/Golang/Hugo.gitignore
+++ b/community/Golang/Hugo.gitignore
@@ -1,10 +1,11 @@
 ### Hugo ###
-# gitginore template for Hugo projects
+# gitignore template for Hugo projects
 # website: https://gohugo.io
 
 # generated files by hugo
 /public/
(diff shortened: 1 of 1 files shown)


community/Golang/Hugo.gitignore
```diff
@@ -1,10 +1,11 @@
 ### Hugo ###
-# gitginore template for Hugo projects
+# gitignore template for Hugo projects
 # website: https://gohugo.io
 
 # generated files by hugo
 /public/
 /resources/_gen/
+/isableFastRender/
```
shiftkey: Should this be `disableFastRender`?

shiftkey:
As reported on https://github.com/gohugoio/hugo/issues/5600 I believe this is a problem with how Hugo was invoked, and not something we should address in this template.

PR #2927, opened 2019-01-08
Android.gitignore: Add build_file_checksums.ser and modules.xml files

Yousha:

**Android.gitignore** was missing `.idea/caches/build_file_checksums.ser` and `.idea/modules.xml`

New Android projects in Android Studio +3 created with a **.gitignore** file that contain `build_file_checksums.ser` and `modules.xml`

Also: https://stackoverflow.com/a/17803964/1429432

If this is a new template:

### DIFF
--- a/Android.gitignore
+++ b/Android.gitignore
@@ -42,6 +42,9 @@ captures/
 .idea/dictionaries
 .idea/libraries
 .idea/caches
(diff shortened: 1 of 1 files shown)


shiftkey:
I think https://stackoverflow.com/a/38122646 is the relevant SO answer (for Android Studio 3.2+ at least).

PR #2928, opened 2019-01-09
Added macOS meta-data

androdev700:

Avoiding the addition of unnecessary .DS_Store files from macOS


shiftkey:
Thanks, but see [`Global\macOS.gitignore`](https://github.com/github/gitignore/blob/master/Global/macOS.gitignore).

PR #2929, opened 2019-01-09
add it

ayush1612:

_TODO_

_TODO_

If this is a new template:


shiftkey:
It's not clear to me what this template is for, and without more context I believe `.html` files are probably something that should be in version control. @ayush1612 are you able to elaborate?

PR #2930, opened 2019-01-11
Update JetBrains.gitignore

marcramser:

That way it also works if you have a folder structure like:

/.gitignore
/backend
/frontend
/frontend/.idea/.workspace.xml

_TODO_

If this is a new template:


shiftkey:
I would recommend using multiple `.gitignore` templates in your repository, and placing the `Global/JetBrains.gitignore` contents under `frontend/.gitignore`.

We've taken the same approach with the `Unity.gitignore` template, which seems to have a similar limitation.

PR #2933, opened 2019-01-16
Unity: ignore Rider cache as well as Visual Studio

sapphon:

IDE cache files are not typically committed to source control, and the master branch currently ignores one such IDE's cache already (VS).

https://unity3d.com

https://www.jetbrains.com/help/idea/creating-and-managing-projects.html


shiftkey:

PR #2934, opened 2019-01-17
Update Node.gitignore

byeonggukgong:

It need `coverage.lcov` file to integrate **nyc** and **Codecov**. I recommend to add it because it's created locally when someone will check that the node command that creating `coverage.lcov` file works correctly.

https://github.com/istanbuljs/nyc#integrating-with-codecov

### DIFF
--- a/Node.gitignore
+++ b/Node.gitignore
@@ -16,6 +16,7 @@ lib-cov
 
 # Coverage directory used by tools like istanbul
 coverage
+coverage.lcov
 
 # nyc test coverage
 .nyc_output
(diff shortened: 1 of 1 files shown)


Node.gitignore
```diff
@@ -16,6 +16,7 @@ lib-cov
 
 # Coverage directory used by tools like istanbul
 coverage
+coverage.lcov
```
shiftkey: Could this be generalized to `*.lcov` if it's a file rather than a folder?

byeonggukgong: Okey, I'll do that :)

shiftkey:
@byeonggukgong thanks!

PR #2936, opened 2019-01-21
Add compilation db to Qt gitignore

martonmiklos:

QtCreator above 4.8 version is utilizing a so called "compilation database" which placed in the compile_commands.json and contains generated information.

https://www.kdab.com/clang-tidy-part-2-integrate-qmake-and-other-build-systems-using-bear/


shiftkey:
@martonmiklos thanks!

PR #2939, opened 2019-01-22
Unity: Never ignore asset meta data

Flate:

If I were to use the Unity.gitignore (for example as part of a project) and the VisualStudio.gitignore (for example as my global gitignore) at the same time, I could run into problems, because the VS one ignores *.meta and Unity relies heavily on these.

This change just makes sure that the *.meta files in the Assets directory are always added to the version control. 

I can see that the necessity of this is debatable, but due to the fact that Visual Studio is the de facto default for Unity under Windows, I think this fix is reasonable.

https://docs.unity3d.com/Manual/ExternalVersionControlSystemSupport.html

### DIFF
--- a/Unity.gitignore
+++ b/Unity.gitignore
@@ -5,6 +5,9 @@
 [Bb]uilds/
 [Ll]ogs/
 
+# Never ignore Asset meta data
+![Aa]ssets/**/*.meta
+
 # Uncomment this line if you wish to ignore the asset store tools plugin
 # [Aa]ssets/AssetStoreTools*
 


shiftkey:
@Flate thanks for the details!

> I could run into problems, because the VS one ignores *.meta and Unity relies heavily on these.

I dug into whether we could drop the rule from `VisualStudio.gitignore` but it looks like it was added here: https://github.com/github/gitignore/commit/2f8c09da0f8ebc34514cecde0e58ece0d0a30739 

The PR itself #95 doesn't have any further context, so I'm fine with adding this rule for the Unity-specific case, rather than potentially breaking some unknown Visual C++ usage in the wild.

shiftkey:
@AdamKane if you believe this should be reverted, please open a new PR and provide context for the change

PR #2940, opened 2019-01-23
Add the entire .idea folder

elkshadow5:

This folder is created purely at compile-time and will be re-written every time you compile with a JetBrains IDE. Additionally, all of the files in this directory control user-specific settings that will very often change for every user in the project. Let me know if/why this reasoning is incorrect please.

https://intellij-support.jetbrains.com/hc/en-us/articles/206544839/comments/200744669

### DIFF
--- a/Global/JetBrains.gitignore
+++ b/Global/JetBrains.gitignore
@@ -1,6 +1,9 @@
 # Covers JetBrains IDEs: IntelliJ, RubyMine, PhpStorm, AppCode, PyCharm, CLion, Android Studio and WebStorm
 # Reference: https://intellij-support.jetbrains.com/hc/en-us/articles/206544839
 
+# JetBrains Compile-time Files
+.idea/
+
 # User-specific stuff
 .idea/**/workspace.xml
 .idea/**/tasks.xml


Global/JetBrains.gitignore
```diff
@@ -1,6 +1,9 @@
 # Covers JetBrains IDEs: IntelliJ, RubyMine, PhpStorm, AppCode, PyCharm, CLion, Android Studio and WebStorm
 # Reference: https://intellij-support.jetbrains.com/hc/en-us/articles/206544839
 
+# JetBrains Compile-time Files
+.idea/
```
shiftkey: This makes all the subsequent rules unnecessary. I'd prefer this template reflect what should be in version control - are you able to elaborate on what files are not covered by this current set of rules?

shiftkey: I'm deferring to this part of the JetBrains documentation in the link at the top of the gitignore template:

> - All the files under `.idea` directory in the project root except the `workspace.xml`, `usage.statistics.xml`, and `tasks.xml` files which store user specific settings

It's not clear to me what's in those files that might be per-user (it's been a while since I've used any JetBrains IDEs to see for myself). Care to elaborate on why you think those can be safely ignored?

iamandrewluca:
Added JetBrains `.gitignore`

```bash
⇒  git status .idea/
On branch master
Your branch is up to date with 'origin/master'.

Untracked files:

        .idea/project-name.iml
        .idea/dataSources.xml
        .idea/encodings.xml
        .idea/misc.xml
        .idea/modules.xml
        .idea/vcs.xml

```

shiftkey:
Closing this out because this outstanding question wasn't answered:

> It's not clear to me what's in those files that might be per-user (it's been a while since I've used any JetBrains IDEs to see for myself). Care to elaborate on why you think those can be safely ignored?

I'm still in favour of versioning what can be shared between users of JetBrains IDEs, and ignoring `.idea/` goes against the goal of the template.

BrianMSheldon:

PR #2942, opened 2019-01-23
Ignore CMake FetchContent directory

joaomlneto:

CMake's `FetchContent` module creates a `_deps` directory on the root (configurable by modifying `FETCHCONTENT_BASE_DIR`).

https://cmake.org/cmake/help/latest/module/FetchContent.html


shiftkey:
@joaomlneto thanks!

PR #2944, opened 2019-01-25
Add pip-wheel-metadata to Python.gitignore

pappasam:

PEP-517 has resulted in some updates to the Python build process. As a result, a new directory called pip-wheel-metadata is created on fresh builds. This PR adds this directory into the Python default gitignore. 

See
https://github.com/pypa/pip/blob/e5f4bbb7ddff87f48f2b5815513e4211ccdde83a/src/pip/_internal/req/req_install.py#L568

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -20,6 +20,7 @@ parts/
 sdist/
 var/
 wheels/
+pip-wheel-metadata/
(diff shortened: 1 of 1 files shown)


shiftkey:
@pappasam thanks!

PR #2946, opened 2019-01-26
Add templates for JetBrains IDEs

burdoto:
JetBrains IDEs usually create the `.idea` subdirectory containing workspace informations, some of which should really not be included in a Git repository. These are three examples for JetBrains Rider (CSharp) and JetBrains IntelliJ IDEA (Java and Gradle).

If this is a new template:

- https://www.jetbrains.com/idea/?fromMenu
- https://www.jetbrains.com/rider/?fromMenu

### DIFF
--- a/CSharp-Rider.gitignore
+++ b/CSharp-Rider.gitignore
@@ -1,0 +1,33 @@
+# Common IntelliJ Platform excludes
+
+# User specific
+**/.idea/**/workspace.xml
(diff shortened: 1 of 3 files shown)


shiftkey:
Thanks, but see [`Global/JetBrains.gitignore`](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore).

PR #2948, opened 2019-01-27
[Codesniffer] remove duplicate Composer rule

Yousha:

**/vendor/** directory is already defined in it's own file: **Composer.gitignore**
https://github.com/github/gitignore/blob/master/Composer.gitignore

https://github.com/github/gitignore/blob/master/Composer.gitignore

https://github.com/squizlabs/PHP_CodeSniffer


shiftkey:
@Yousha thanks!

PR #2949, opened 2019-01-28
Updated Unity.gitignore to allow subfolders by default

Smalnardev:
The existing .gitignore also ignores folders subfolders with the given names. For example, a plugin or user code might exist at Assets/Buildings/Library/code.cs which would be ignored. This change makes only the given folders in the base directory get ignored.

### DIFF
--- a/Unity.gitignore
+++ b/Unity.gitignore
@@ -1,9 +1,9 @@
-[Ll]ibrary/
-[Tt]emp/
-[Oo]bj/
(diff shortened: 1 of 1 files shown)


cscharfen:
@shiftkey This is now obsolete because we merged the other PR -> https://github.com/github/gitignore/pull/2980

shiftkey:
Closing this out as it has been superseded by #2980

PR #2954, opened 2019-01-29
[Unity] added preamble

AdmiralSnyder:
... to let the file point to its canonical source
e.g. here is such a preamble, too:
https://github.com/github/gitignore/blob/master/VisualStudio.gitignore


shiftkey:
@AdmiralSnyder thanks!

PR #2956, opened 2019-02-04
added .DS_Store

profjsb:

Mac OSX adds `.DS_Store` to directories which are viewed in Finder. This file
is not needed as part of a normal git repo.


shiftkey:
Thanks, but see [`Global/macOS.gitignore`](https://github.com/github/gitignore/blob/master/Global/macOS.gitignore)

PR #2961, opened 2019-02-12
Add KDiff3 gitignore

dariusz-wozniak:

Add KDiff 3 files that may be ignored.

https://github.com/KDE/kdiff3/blob/master/.gitignore

### DIFF
--- a/KDiff3.gitignore
+++ b/KDiff3.gitignore
@@ -1,0 +1,9 @@
+*.BACKUP.*
+*_BACKUP_*
+*.BASE.*
+*_BASE_*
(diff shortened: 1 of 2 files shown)


shiftkey:
@dariusz-wozniak would you be interested in submitting a standalone `Global/KDiff.gitignore` template with these rules? As KDiff can be used standalone with Git, that feels more discoverable and usable than appending it to the `VisualStudio.gitignore` template...

dariusz-wozniak:
@shiftkey You're right. Please see new commits for new KDiff3 gitignore file.

PR #2962, opened 2019-02-13
Ignore Visual Studio Code cache directory for Unity project

agirault:

Ignore visual studio cache directory for Unity projects

> The workspace setting file is located under the `.vscode` folder in your root folder.
> \- https://code.visualstudio.com/docs/getstarted/settings

### DIFF
--- a/Unity.gitignore
+++ b/Unity.gitignore
@@ -14,6 +14,9 @@
 # Visual Studio cache directory
 .vs/
 
+# Visual Studio Code cache directory
+.vscode/
+
 # Gradle cache directory
 .gradle/
 


LucasSteer:
Is it common to not want to include ones settings within a git repo? For my personal projects I prefer to have `settings.json` move with the repo so that project specific settings don't get lost if I pick up the development on other machines. I'm not sure what the best practice and desired utility is for multi-member teams, however.

shiftkey:
This is not a cache directory, but a place for storing VSCode-specific configurations for a project, so that they can be shared. 

We have an existing template for this [`Global/VisualStudioCode.gitignore`](https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore) that should be used instead of excluding the entire directory.

PR #2963, opened 2019-02-13
Migration backup

OsirisTerje:

Visual Studio 2017 have a built-in tool for converting from Nuget Package config files to Package References.  During this process the package config file is deleted and the csproj files are modified. Before deletion these are copied into a backup folder named MigrationBackup/.
This folder should not go into source control

https://docs.microsoft.com/en-us/nuget/reference/migrate-packages-config-to-package-reference

https://docs.microsoft.com/en-us/nuget/reference/migrate-packages-config-to-package-reference#how-to-roll-back-to-packagesconfig


shiftkey:
@OsirisTerje thanks!

PR #2965, opened 2019-02-15
Update VisualStudio.gitignore

rishabh-husky:

enc_temp_folder/  folder is created after building a project in visual studio.

Take a look at this repository for example.
https://github.com/jackmott/dungeonbuilder/blob/master/.gitignore

If this is a new template:

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -26,6 +26,7 @@ bld/
 [Bb]in/
 [Oo]bj/
 [Ll]og/
+enc_temp_folder/
 
 # Visual Studio 2015/2017 cache/options directory
 .vs/


shiftkey:
@rishabh-husky 

> enc_temp_folder/ folder is created after building a project in visual studio.

What sort of project? I'd love some more context here so we can track _why_ we're introducing a new rule like this.

rishabh-husky:
The project created using empty c++ project template seems to have this folder. This could be due to auto recovery done by visual studio. There is not much I could find on the internet. But, it happened in my two c++ projects.

shiftkey:
@rishabh-husky the only real clue I could uncover about this came from [this obscure thread](https://groups.google.com/forum/#!topic/wx-dev/zGy1KRU0SeQ):

> enc_temp folder is the visual studio autosave/autorecover directory, I think it's used during edit and continue

Is this something you're using in your projects?

samrrr:
What about add this folder to gitignore? Git still try submit this folder.

SpriteOvO:
The folder seems to be created when Visual Studio starts debugging, but for unknown reasons, it is not always created.
When I stop debugging, the folder is automatically deleted.

SpriteOvO:
I've investigated!
If you edit the code after you start debugging, Visual Studio will create the folder and back up the original contents of the code file.

bdougie:
There are merge conflicts present due to merges elsewhere. Closing, but feel free to re-open and for the fixes.

PR #2967, opened 2019-02-19
Updated Unity's gitignore file to include JetBrains Rider's cache folder

Tanshaydar:
JetBrains Rider's full support for Unity is on the run for a long time now and many developers are using it. However, each time .idea folder must be added to .gitignore file. It should be there by default.

An example case can be found here: https://stackoverflow.com/questions/53236846/the-gitignore-setting-for-the-idea-folder-of-jetbrains-riders-does-not-work

### DIFF
--- a/Unity.gitignore
+++ b/Unity.gitignore
@@ -14,6 +14,9 @@
 # Visual Studio cache directory
 .vs/
 
+# JetBrain Rider cache directory
+.idea/
+
 # Gradle cache directory
 .gradle/
 


shiftkey:
Thanks, but see [`Global/JetBrains.gitignore`](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore) - this is not a local cache, and there are various files in that directory that can be committed and shared with others.

Tanshaydar:
JetBrains Rider, when initialized on a Unity project, creates a local folder named **.idea**

This is not JetBrains related. This is Unity game engine related, hence the pull request is made for Unity's gitignore file.

I've been using Rider since its EAP times and all of my Unity related projects has this folder, which I manually exclude with .gitignore file. I'd rather have this on default instead of slapping my forehead each time.

shiftkey:
That gitignore template is relevant to all IDEA-based editors (including Rider), and it's the template that JetBrains themselves link at the bottom of their ["How to manage projects under Version Control Systems"](https://intellij-support.jetbrains.com/hc/en-us/articles/206544839) documentation.

Tanshaydar:
No it is not.
A Unity project is not a C# project.
Unity does not rely Rider's, or any other IDEs' templates. 

What baffles me here is that Github officially maintains a Github plugin for Unity here: https://github.com/github-for-unity/Unity
And the default .gitignore file here has the same line: https://github.com/github-for-unity/Unity/blob/master/src/.gitignore

Another example: https://gist.github.com/ryanzec/730699ce8c1c88764f914329acbcdba1
See the comment: **Unity .gitignore file (Using Rider IDE)**

Another example: https://leavingharbor.com/posts/unity-and-git
See the comment: **I'm actually using JetBrains' Rider so my .gitignore looks something like this**

If you are still going to show me the same JetBrains doc, then we can close the case for once and all. I've not going to re-iterate myself saying that this is all Unity related, not JetBrains related.

Officially you have .vs folder completely ignored in the Unity .gitignore file, but VS Code file is not the same? https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore
Are you going to suggest that Untiy depends on the .vs folder contents and should be added, too?

PR #2968, opened 2019-02-20
Ignore Windows dump files

rdev5:
Ignores several possible dump file extensions where OS information (including username, network shares, etc.) may be recorded during application crash. This issue was only discovered after being pushed to a remote origin from a *non-VisualStudio* repository.

Without further analysis, the suspected origin of the dump file is a SublimeText plugin. Consideration should be given as to whether these ignore lines should be introduced for other project types where software running in the local repository directory may be involved.

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -224,6 +224,11 @@ ClientBin/
 *.publishsettings
 orleans.codegen.cs
 
+# Dump files containing OS information including username, network shares, etc. recorded during crash
+*.dmp
+*.dump
+*.mdmp
+
 # Including strong name files can present a security risk
 # (https://github.com/github/gitignore/pull/2483#issue-259490424)
 #*.snk
--- a/Global/Windows.gitignore
+++ b/Global/Windows.gitignore
@@ -5,6 +5,9 @@ ehthumbs_vista.db
 
 # Dump file
 *.stackdump
+*.dmp
+*.dump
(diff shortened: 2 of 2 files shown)


shiftkey:
@rdev5 do you think the `Global/Windows.gitignore` template is a better fit for it? We already have one ignore rule in there for dump files, and I think tools other than Visual Studio can also generate dump files like this:

https://github.com/github/gitignore/blob/d2c1bb2b9c72ead618c9f6a48280ebc7a8e0dff6/Global/Windows.gitignore#L6-L7

rdev5:
@shiftkey Yes, I think that would be appropriate. I was actually looking for that, albeit, in the root of this repo.

Analyzing the minidump reveals it did in fact come from **sublime_text.exe** (0xC0000005: The thread tried to read from or write to a virtual address for which it does not have the appropriate access.)

VisualStudio.gitignore
```diff
@@ -224,6 +224,11 @@ ClientBin/
 *.publishsettings
 orleans.codegen.cs
 
+# Dump files containing OS information including username, network shares, etc. recorded during crash
```
shiftkey: @rdev5 can we drop these changes from this template, now they have a more appropriate home?

shiftkey:

@rdev5 if you'd still like to see this change land, feel free to open a fresh PR that addresses the feedback above.

PR #2969, opened 2019-02-21
[Global/Eclipse] add ".project" file

jskelin:
ignore project description file

Ignore generated file by eclipse .

https://help.eclipse.org/luna/index.jsp?topic=%2Forg.eclipse.platform.doc.isv%2Freference%2Fmisc%2Fproject_description_file.html

If this is a new template:

### DIFF
--- a/Global/Eclipse.gitignore
+++ b/Global/Eclipse.gitignore
@@ -11,6 +11,9 @@ local.properties
 .loadpath
 .recommenders
 
+# Project description file
+.project
+
 # External tool builders
 .externalToolBuilders/
 


shiftkey:
@jskelin thanks!

shiftkey:
@jskelin actually, this comment from the link makes me curious:

> The purpose of this file is to make the project self-describing, so that a project that is zipped up or released to a server can be correctly recreated in another workspace. 

This to me feels like it has value, and _might_ be useful to have in version control, even if it's generated. Can you elaborate on why it's safe to exclude this from version control?

larry1186:
Looks like @jskelin does not have this project forked anymore. I think it would still be beneficial to at least address the .project file in the template. I have opened my own Pull Request #3305 to make it opt-in to ignore. Could this possibly be extended to .cproject already ignored or perhaps .classpath/.settings?

A few other templates have the .project file listed, but should be consolidated to the Eclipse .gitignore.

shiftkey:
@larry1186 thanks for submitting #3305 - I'm going to close this out as it's now unnecessary

PR #2971, opened 2019-02-23
[Node] Ignore diagnostic report files

Richienb:

To ignore diagnostic reports created by NodeJS, usually only relevant to the current computer.

**More information:**

[NodeJS Reporting Documentation](https://nodejs.org/api/report.html)
[What is NodeJS?](https://nodejs.org)

### DIFF
--- a/Node.gitignore
+++ b/Node.gitignore
@@ -5,7 +5,7 @@ npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
 
-# Diagnostic reports
+# Diagnostic reports (https://nodejs.org/api/report.html)
 report.*.json
 
 # Runtime data
--- a/Node.gitignore
+++ b/Node.gitignore
@@ -5,6 +5,9 @@ npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
 
+# Diagnostic reports
+report.*.json
+
 # Runtime data
 pids
 *.pid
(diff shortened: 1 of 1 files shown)


Richienb:
Any update on this?

Node.gitignore
```diff
@@ -5,6 +5,9 @@ npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
 
+# Diagnostic reports
```
shiftkey: Is it worth including a link to the documentation about this, given it's an experimental feature?

Node.gitignore
```diff
@@ -5,6 +5,9 @@ npm-debug.log*
...
+report.*.json
```
shiftkey: I'm a bit nervous about this rule catching other files that happen match this pattern (by luck or sheer misfortune) - is there a more strict way we could write this expression?

I see from the docs that the file paths look like this: `report.20181221.005011.8974.001.json`. If that's the case, this could be rewritten as `report.[0-9]*.[0-9]*.[0-9]*.[0-9]*.json`.

Thoughts?

Node.gitignore
Richienb: Done.

Node.gitignore
Richienb: Done.

Richienb:
@shiftkey I've applied your recommendations.

shiftkey:
@Richienb thanks!

PR #2972, opened 2019-02-24
[WordPress] updated ignored list

rohanvakharia:

To ignore everything except the plugins and theme

https://core.trac.wordpress.org/browser/trunk/.gitignore

If this is a new template:

https://wordpress.org

### DIFF
--- a/WordPress.gitignore
+++ b/WordPress.gitignore
@@ -1,19 +1,23 @@
-*.log
-wp-config.php
-wp-content/advanced-cache.php
-wp-content/backup-db/
-wp-content/backups/
-wp-content/blogs.dir/
-wp-content/cache/
-wp-content/upgrade/
-wp-content/uploads/
-wp-content/mu-plugins/
-wp-content/wp-cache-config.php
+# ignore everything in the root except the "wp-content" directory.
+!wp-content/
+
+# ignore everything in the "wp-content" directory, except:
+# "mu-plugins", "plugins", "themes" directory
+wp-content/*
+!wp-content/mu-plugins/
+!wp-content/plugins/
+!wp-content/themes/
+
+# ignore these plugins
 wp-content/plugins/hello.php
 
-/.htaccess
-/license.txt
-/readme.html
-/sitemap.xml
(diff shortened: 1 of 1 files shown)


WordPress.gitignore
```diff
@@ -1,19 +1,23 @@
...
-/.htaccess
-/license.txt
-/readme.html
-/sitemap.xml
-/sitemap.xml.gz
+# ignore specific themes
+wp-content/themes/twenty*/
 
+# ignore node dependency directories
+node_modules/
```
malithmcr: Does Wordpress use node_modules in the root?

rohanvakharia: Yes. 
Here you will find the list of packages using by WordPress:
https://core.trac.wordpress.org/browser/trunk/package.json

shiftkey:
@rohanvakharia thanks!

kolorafa:
How does those rules ignore all files in root directory?
https://github.com/github/gitignore/blob/499ae899e7b54e701e878759f73d9092302fd07a/WordPress.gitignore#L1-L2

Isn't there something missing like /*  ?
Current (after merge) gitignore includes wp-config:

```
$ git add . --dry-run|grep wp-config
add 'wp-config-sample.php'
add 'wp-config.php'
```

hdwebpros:
I'm not sure why this PR was merged. It's confusing. 

If I see `# ignore everything in the root except the "wp-content" directory. ` then I shouldn't see `# ignore log files and databases` since those are in the root.

Also, like @kolorafa mentioned, it seems that there is something missing. We've used:

```
# Ignore everything, but all to descend into subdirectories
*
!*/
```
To ignore everything, but allow the including of what you want included. So, you could have

```
# "themes/my-theme" directory
!wp-content/themes/my-theme/**
```
Later in your gitignore file and have that work fine.

pedrosanta:
It's just me or this merge is a utter and complete _trainwreck_? I've got all seed projects broken, basically allowing me to add/stage pretty much of a complete WP install into the commit.

This wasn't nowhere near as tested once by the submitter.

PR #2973, opened 2019-02-24
Add .phpunit.result.cache to the .gitignore

rohanvakharia:

PHPUnit version 7.3 adds a new argument `--cache-result` which allows you to do things like re-run test failures.


shiftkey:
@rohanvakharia thanks!

https://github.com/laravel/laravel/commit/6b40d49efdcb30aad4e4e51ea9c2c19481806ac0 is also some helpful context about this change.

PR #2975, opened 2019-02-25
Update VisualStudio.gitignore file

JFMG:

Added
mono_crash.* // Some executed unit tests in a .NET Core project may lead to a mono crash which then generates this file

Changed:
// Just to be conform with the other folder name conventions
x64/ => [Xx]64/
x86/ => [Xx]86/

// Visual Studio For Mac sometimes creates multiple bin and obj folders with increasing numbers, e.g. "bin 2/" (bin n/ where n = i + 1 and i starts from 0)
[Bb]in/ => [Bb]in*/
[Oo]bj/ => [Oo]bj*/
[Ll]og/ => [Ll]og*/

🤷🏻‍♂️

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -21,14 +21,14 @@ mono_crash.*
 [Dd]ebugPublic/
 [Rr]elease/
 [Rr]eleases/
-[Xx]64/
-[Xx]86/
+x64/
+x86/
 [Aa][Rr][Mm]/
 [Aa][Rr][Mm]64/
 bld/
-[Bb]in*/
-[Oo]bj*/
-[Ll]og*/
+[Bb]in/
+[Oo]bj/
+[Ll]og/
 
 # Visual Studio 2015/2017 cache/options directory
 .vs/
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -13,19 +13,22 @@
 # User-specific files (MonoDevelop/Xamarin Studio)
 *.userprefs
 
+# Mono auto generated files
+mono_crash.*
+
 # Build results
 [Dd]ebug/
 [Dd]ebugPublic/
 [Rr]elease/
 [Rr]eleases/
-x64/
-x86/
+[Xx]64/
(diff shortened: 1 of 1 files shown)


shiftkey:
@JFMG some initial feedback:

> `mono_crash.*`

Good catch!

> // Just to be conform with the other folder name conventions
> `x64/ => [Xx]64/`
> `x86/ => [Xx]86/`

Any idea where this came from? I'm tentatively 👍, just wondering if there's extra context because `x` in these platform names has been a thing for a loooooong time.

> // Visual Studio For Mac sometimes creates multiple bin and obj folders with increasing numbers, e.g. "bin 2/" (bin n/ where n = i + 1 and i starts from 0)

This feels like an uncommon scenario where the tooling might not be able to write to its preferred paths. I'm 👎 on this without more information or context for this behaviour.

JFMG:
@shiftkey 

> Good catch!

Thanks :-)



> Any idea where this came from?

I wish I'd know 😅 

> This feels like an uncommon scenario where the tooling might not be able to write to its preferred paths.

Yep, this might be true. I've seen it a few times when Visual Studio for Mac and for Windows were in use. My guess is that in some cases the obj and bin folders get locked and therefore VS is not able to write into these folders. I didn't find any further information about this issue. It might be a bug in the tooling as you said.

VisualStudio.gitignore
```diff
@@ -13,19 +13,22 @@
 # User-specific files (MonoDevelop/Xamarin Studio)
 *.userprefs
 
+# Mono auto generated files
+mono_crash.*
```
shiftkey: @JFMG given the lack of clarity and documentation about where the other behaviour is coming from, are we able to focus on just this change? I'd be fine with future PRs if you uncover more context about the below rules.

JFMG: Yep, sure. I will update the PR

JFMG: @shiftkey Please review my latest PR

shiftkey:
@JFMG thanks!

PR #2976, opened 2019-02-25
Add .DS_STORE for mac os

jordan1997:
Mac os created .DS_STORE into each folder. We don't need to include this file into the git repository.

The previous version don't be well for MacOs user

### DIFF
--- a/Android.gitignore
+++ b/Android.gitignore
@@ -78,3 +78,6 @@ lint/generated/
 lint/outputs/
 lint/tmp/
 # lint/reports/
+
+# MacOs
+*.DS_STORE


drothmaler:
Please see `Global/macOS.gitignore`

jordan1997:
> Please see `Global/macOS.gitignore`

Ok, I catch a point

shiftkey:
Thanks, but see [`Global/macOS.gitignore`](https://github.com/github/gitignore/blob/master/Global/macOS.gitignore) for the list of macOS-specific rules.

jordan1997:
I check, .DS_STORE included in Global/macOs.gitignore
@shiftkey 
Do we need to add all rules from Global/macOS.gitignore into the Android.gitignore?

shiftkey:
>Do we need to add all rules from Global/macOS.gitignore into the Android.gitignore?

No, these templates are intended to be a reference - duplicating rules across templates for convenience adds to the maintenance burden I cannot currently support.

Please compose them together in your projects based on your setup and tools used.

PR #2977, opened 2019-02-26
Add ignore rules for Pipenv

JarryShaw:
Pipenv uses Pipfile.lock to maintain Python package information
(metadata, hash, etc.) installed as described in Pipfile. Thus,
Pipfile.lock may vary on different operating systems, platforms
when collaborating. This PR adds Pipfile.lock into the Python
default gitignore. See http://pipenv.org

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -82,6 +82,9 @@ ipython_config.py
 # pyenv
 .python-version
 
+# pipenv
+Pipfile.lock
+
 # celery beat schedule file
 celerybeat-schedule
 
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -83,7 +83,9 @@ ipython_config.py
 .python-version
 
 # pipenv
-Pipfile.lock
+#   According to pypa/pipenv#598, it is recommended to include Pipfile.lock in VCS.
+#   However, in case of collaboration, Pipfile.lock may conflict on different OS.
+#Pipfile.lock
 
 # celery beat schedule file
 celerybeat-schedule
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -83,7 +83,7 @@ ipython_config.py
 .python-version
 
 # pipenv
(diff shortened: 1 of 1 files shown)


drothmaler:
According to the [Pipenv docs](https://github.com/pypa/pipenv/blob/master/docs/basics.rst#-general-recommendations--version-control) and https://github.com/pypa/pipenv/issues/598 it is generally recommended to add `Pipfile.lock` to the VCS, unless you have a specific reason not do do so (e.g. if you are targeting multiple Python versions, or if you are dealing with legacy dependencies, that are not platform-agnostic).

So -1 for adding it by default. 
But it might be nice to add it as an commented Opt-In ignore. Something like 

```
# you might want to ignore the pipenv lock-file if ... bla bla bla
# Pipfile.lock
```

JarryShaw:
Noted. Adjusted according to @drothmaler 's comment.

Python.gitignore
```diff
@@ -82,6 +82,11 @@ ipython_config.py
 # pyenv
 .python-version
 
+# pipenv
+#   According to pypa/pipenv#598, it is recommended to include Pipfile.lock in VCS.
```
shiftkey:
```suggestion
#   According to pypa/pipenv#598, it is recommended to include Pipfile.lock in version control.
```

Python.gitignore
```diff
@@ -82,6 +82,11 @@ ipython_config.py
...
+#   However, in case of collaboration, Pipfile.lock may conflict on different OS.
```
shiftkey: This comment from the linked issue is helpful context for problems with the lockfile - @JarryShaw are you able to elaborate in this comment on the problems that users might see with this lockfile? https://github.com/pypa/pipenv/issues/598#issuecomment-361811946

JarryShaw: Just rephrased a bit. Hopefully it is now better :)

PR #2980, opened 2019-03-06
[Unity] Added leading slashes to ignored directories so that valid subdirectories aren't ignored incorrectly

LucasSteer:

The lines at the top of the .gitignore would ignore the folders at the root of the directory (which is correct behaviour), but would also ignore **any subfolder with that name** anywhere else in the project (which is incorrect behaviour).

Example of Incorrect Behaviour: `/Assets/Scripts/Library/Example.cs` would previously be incorrectly ignored.

This change fixes that issue and **only** ignores the folders at the root of the project directory.

StackOverflow post detailing a similar situation: https://stackoverflow.com/questions/24139478/when-to-use-leading-slash-in-gitignore

My own testing replicates the behaviour described in that post.

### DIFF
--- a/Unity.gitignore
+++ b/Unity.gitignore
@@ -1,9 +1,9 @@
-[Ll]ibrary/
-[Tt]emp/
-[Oo]bj/
-[Bb]uild/
-[Bb]uilds/
-[Ll]ogs/
+/[Ll]ibrary/
+/[Tt]emp/
+/[Oo]bj/
+/[Bb]uild/
+/[Bb]uilds/
+/[Ll]ogs/
 
 # Never ignore Asset meta data
 ![Aa]ssets/**/*.meta
--- a/Unity.gitignore
+++ b/Unity.gitignore
@@ -1,3 +1,5 @@
+# This .gitignore file should be placed at the root of your Unity project directory
+
 /[Ll]ibrary/
 /[Tt]emp/
 /[Oo]bj/
@@ -9,7 +11,7 @@
 ![Aa]ssets/**/*.meta
 
 # Uncomment this line if you wish to ignore the asset store tools plugin
-# [Aa]ssets/AssetStoreTools*
+# /[Aa]ssets/AssetStoreTools*
 
 # Visual Studio cache directory
 .vs/
--- a/Unity.gitignore
+++ b/Unity.gitignore
@@ -8,7 +8,7 @@
 /[Ll]ogs/
 
 # Never ignore Asset meta data
-![Aa]ssets/**/*.meta
+!/[Aa]ssets/**/*.meta
 
 # Uncomment this line if you wish to ignore the asset store tools plugin
 # /[Aa]ssets/AssetStoreTools*


shiftkey:
This set of rules has been through a bit of churn over the years:

 - it was configured this way all the way back in #1481 by @Tandoori 
 - then it was updated in to support subdirectories in #2531 by @matzar

I'm not familiar with how Unity projects should be managed within Git repositories in this situation, but if we can reach a consensus here I'd be happy to merge whatever is agreed upon.

cc @Flate @shana @johnsoncodehk @FutureClemens @ksmit799 who have contributed to the Unity template recently for any additional feedback on this change.

johnsoncodehk:
I agree with this PR.

Previously removing the slash seems to be because @matzar Unity project is not in the root directory.

But consider this directory structure:

- Root
  - ServerProject
    - Library
  - UnityProject
    - Library
    - Assets/Scripts/Library

The previous Unity.gitignore couldn't handle it, but we can now place Unity.gitignore in Root/UnityProject to properly ignore the Library directory.

(git has handled nested .gitignore, why should handle nested Unity projects in Unity.gitignore?)

cscharfen:
Definitely, go for this PR ;)

ksmit799:
There's no reason to ignore these folders globally, and the current definition (for many people) would lead to unexpected behavior.

shiftkey:
> The previous Unity.gitignore couldn't handle it, but we can now place Unity.gitignore in Root/UnityProject to properly ignore the Library directory.

One of the nice benefits of how `gitignore` files work is that they don't need to live at the root, and you can have different `.gitignore` files at different levels in the repository.

@LucasSteer are you able to add a comment at the top of the file to recommend, if you are using a Unity project in a subdirectory, to use this template at the root of your Unity project?

matzar:
I made a repository ([https://github.com/matzar/slashOrUnslash](url)) where I tested both versions of .gitignore: 
- `[Ll]ibrary` etc. (This one I call unslashed) 
- `/[Ll]ibrary` (This one I call slashed)

I made two cases:
- un-nested Unity project 
- nested Unity project

It seems to me that both slashed and unslashed versions work the same for un-nested projects.
The difference begins when the project is nested. Then it seems to me that the slashed version is not ignoring .mets files while the unslashed version is ignoring them.

I just quickly skimmed through the commits. Please have a look for yourself and tell us what you think @shiftkey @Flate @shana @johnsoncodehk @FutureClemens @ksmit799

matzar:
I made a repository ([https://github.com/matzar/slashOrNoslash](https://github.com/matzar/slashOrUnslash)) where I tested both versions of .gitignore: 
- `[Ll]ibrary` etc. (This one I call unslashed) 
- `/[Ll]ibrary` (This one I call slashed)

I made two cases:
- un-nested Unity project ( https://github.com/matzar/slashOrNoslash/commit/257e4722a647f6da87e044118d0254484f65dbbd , https://github.com/matzar/slashOrNoslash/commit/91b90d8842e77083e688b108004bc0e3d9ab1646 )
- nested Unity project ( https://github.com/matzar/slashOrNoslash/commit/72a717539424a9a613f6b5a8bbc6a180cbc6e3f0 , https://github.com/matzar/slashOrNoslash/commit/5e35fafee5175677364d5e5125fc7dc771541a96 )

If you peek at the nested commits you can see that the slashed version has many additions while the unslashed version has many removes.

It seems to me that both slashed and unslashed versions work the same for un-nested projects.
The difference begins when the project is nested. Then it seems to me that the slashed version is not ignoring .meta files while the unslashed version is ignoring them.

I just quickly skimmed through the commits. Please have a look for yourself and tell us what you think @shiftkey @Flate @shana @johnsoncodehk @FutureClemens @ksmit799

johnsoncodehk:
@matzar Of course... because you still put .gitignore in the root directory.

We are care about the unexpected behavior of the noslashed version.

FutureClemens:
@shiftkey I think we all pretty much agreed, that the gitignore file is (and needs to be) placed within the project directory and therefore, the slashed version (as changed in this PR) is perfectly fine.

I personally have tons of Unity projects, and each contains its own gitignore file, that might be slightly modified, depending on what I need additionally in this project.

If it is in question, if users may put the gitignore file in the wrong directory (e.g. outside of the project dir), we might add a comment to the very first line, saying something like:
`# This .gitignore file needs to be placed inside the Unity project directory (besides the Assets and ProjectSettings directories)`
You might find a better phrasing though ;)

Also, but this is actually nitpicking, we could add the slashes to the lines below, shouldn't we?
```
# Never ignore Asset meta data
!/[Aa]ssets/**/*.meta

# Uncomment this line if you wish to ignore the asset store tools plugin
# /[Aa]ssets/AssetStoreTools*
```

shiftkey:
This looks like it's good to merge - thanks @LucasSteer and to everyone else who was involved with the review!

PR #2988, opened 2019-03-10
Unity: ignore TextMesh Pro

qCring:
TextMesh Pro is fully integrated in Unity since 2017 and all related resources are managed by Unity's Package Manager. So it would make sense to ignore them by default
Check out [the official blog post](https://blogs.unity3d.com/2017/03/20/textmesh-pro-joins-unity/) from early 2017

### DIFF
--- a/Unity.gitignore
+++ b/Unity.gitignore
@@ -11,6 +11,9 @@
 # Uncomment this line if you wish to ignore the asset store tools plugin
 # [Aa]ssets/AssetStoreTools*
 
+# TextMesh Pro files
+[Aa]ssets/TextMesh*Pro/
+
 # Visual Studio cache directory
 .vs/
 


shiftkey:
@qCring thanks!

heatblayze:
@qCring I think this may be incorrect.. While TextMesh Pro has indeed moved to the package manager, the TextMesh Pro folder is still created when importing the "TMP Essential Assets".
As the name suggests, these assets are essential for using TMP, as stated in the dialogue that appears when you first add a TMP object into a scene.

shiftkey:
@heatblayze if you feel like this should be reverted or changed, I'm open to a PR which helps improve things

qCring:
sorry for the delayed response!

@heatblayze I agree, you are absolutely right

I expected all related dependency resources to be managed by Unity's package manager without double-checking. NuGet (or gradle or other comparable tools) perform package-related operations like restore/update/upgrade on the client-side. Dependecies are referenced in some manifest instead of actually pushing files to the VCS. Unfortunately, Unity handles things differently in this case. But this is no excuse, I really should have checked. Sorry for the inconvenience!

So yes, this should be reverted!

shiftkey:
> So yes, this should be reverted!

@qCring @heatblayze I'm not able to quickly revert this PR (because of subsequent changes to the file) but I'd love to see a PR to close this out now we're on the same page.

qCring:
@heatblayze do you want to create a PR since you pointed out the issue?

heatblayze:
@qCring Sure thing! I'll get onto it now

PR #2989, opened 2019-03-10
Rename VisualStudio.gitignore to DotNet.gitignore

shanselman:

The VisualStudio.gitignore is more correctly a DotNet.gitignore. The lack of a .NET/C#/F#/VB one is a constant source of confusion. We'd like to change this. Source: I work on the .NET team


KexyBiscuit:
Then we'll need another one for MSVC, and all other project type supported by Visual Studio.

shiftkey:
@shanselman the last time this came up I wanted to call out that there's a lot of non-.NET technologies inside the `VisualStudio` template, and suggested a different approach https://github.com/github/gitignore/pull/2816#issuecomment-422432117

> What if we went a slightly different approach:
> 
> * `DotNet.gitignore` - a new template fresh with whatever current .NET requires (leaner)
> * `Global/VisualStudio.gitignore` - the current template, moved alongside the other IDEs

Thoughts?

shanselman:
@shiftkey that is a great idea and a solid compromise! I'll update.

PR #2990, opened 2019-03-10
Updated jetbrains' gitignore

longnguyen2306:
Updated jetbrains' gitignore. Just ignore the whole .idea since nobody wants it. Also ignore virtual enviroment of pycharm

I have to ignore .idea, *.iml and venv for every new project

https://github.com/intellij-rust/intellij-rust/issues/1570
https://github.com/joeblau/gitignore.io/issues/360

### DIFF
--- a/Global/JetBrains.gitignore
+++ b/Global/JetBrains.gitignore
@@ -1,42 +1,14 @@
 # Covers JetBrains IDEs: IntelliJ, RubyMine, PhpStorm, AppCode, PyCharm, CLion, Android Studio and WebStorm
 # Reference: https://intellij-support.jetbrains.com/hc/en-us/articles/206544839
 
+# Iml
+*.iml
+
 # User-specific stuff
-.idea/**/workspace.xml
-.idea/**/tasks.xml
-.idea/**/usage.statistics.xml
-.idea/**/dictionaries
-.idea/**/shelf
-
-# Generated files
-.idea/**/contentModel.xml
-
-# Sensitive or high-churn files
-.idea/**/dataSources/
-.idea/**/dataSources.ids
-.idea/**/dataSources.local.xml
-.idea/**/sqlDataSources.xml
-.idea/**/dynamic.xml
-.idea/**/uiDesigner.xml
-.idea/**/dbnavigator.xml
-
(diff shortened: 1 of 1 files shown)


shiftkey:
Thoughts @citizenmatt?

longnguyen2306:
@shiftkey While working with JetBrains, when the initial commit forgets to ignore <code>.idea</code> and <code>*.iml</code> or <code>venv</code> that always leads to complaints of other members, since nobody wants those file/folder on git, because they will be generated automatically every time you open/create a new project with any JetBrains IDE. 

I think it therefore will be a good idea to ignore those.

Global/JetBrains.gitignore
```diff
@@ -50,17 +22,12 @@ out/
...
 # Crashlytics plugin (for Android Studio and IntelliJ)
 com_crashlytics_export_strings.xml
 crashlytics.properties
 crashlytics-build.properties
 fabric.properties
 
-# Editor-based Rest Client
-.idea/httpRequests
+# Pycharm virtual environment
+venv
```
shiftkey: A broad set of `venv`-related rules is covered in `Python.gitignore` which should avoid the need for this rule:

https://github.com/github/gitignore/blob/7079791ab6058c544dd8dde60294b69e02a6dcf8/Python.gitignore#L98-L105

shiftkey:
Some extra context:

> I have to ignore .idea, *.iml and venv for every new project

You can configure a [global gitignore](https://help.github.com/en/articles/ignoring-files#create-a-global-gitignore) which will be used when working with any local Git repository without needing to commit `.gitignore` file changes to each repository.

> Just ignore the whole .idea since nobody wants it.

This template has 40 different contributors who have contributed improvements to the template work out what should and should not be included in version control. This is intended to be a reference template based on the [JetBrains documentation](https://intellij-support.jetbrains.com/hc/en-us/articles/206544839), and I don't see the value in being less granular here.

shiftkey:
I'm going to close this out, but I'll revisit this if the JetBrains staff feel like a simpler template here is the way to go.

BrianMSheldon:

PR #2991, opened 2019-03-11
Remove Rider (idea) ignore statements from Visual Studio template

CoskunSunali:
Rider has its own ignore file, so does Visual Studio. The ignore statements for Rider (idea) IDE should be removed from Visual Studio .gitignore file template.

Discussion: https://github.com/github/gitignore/commit/e429db11802448b3d7e7305ee78df1ee918b8fc6

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -293,10 +293,6 @@ paket-files/
 # FAKE - F# Make
 .fake/
 
-# JetBrains Rider
-.idea/
-*.sln.iml
-
 # CodeRush personal settings
 .cr/personal
 


CoskunSunali:
Thanks for the quick reaction @shiftkey

potatman:
This change seems really odd to me as that JetBrains ignore isn't really great for dotnet repos.  I think my core concern stems from the fact that the vast majority of what is in the VisualStudio ignore has nothing to do with VS and is more just for dotnet in general.  If the intention is to have the ignores be more modular (like java, which has an ignore for the language and one ignore each for the common ides), then the dotnet stuff should really

PR #2992, opened 2019-03-11
Add more standard ignored files for Julia

ararslan:

The Julia template here hasn't been updated in a while, and since its last update, new ignorable files have entered the Julia project workflow, including:

* Documentation build artifacts generated by [Documenter.jl](https://github.com/JuliaDocs/Documenter.jl), which the vast majority of Julia packages use.
* Manifest.toml, which is like Julia's equivalent to other languages' package manager's "lock files." It is generated automatically by the package manager, Pkg, and typically appears in docs/ and at the top level, though it can appear anywhere a corresponding Project.toml file is present.

* Documenter: https://juliadocs.github.io/Documenter.jl/dev/man/hosting/#.gitignore-1
* Pkg: No relevant link, but see most Julia packages' `.gitignore` files


shiftkey:
Only question I have is about this bit:

> Pkg: No relevant link, but see most Julia packages' .gitignore files

Is the community consensus that lock files shouldn't be committed? My experiences with other ecosystems (Node/Yarn and Rust being the most recent) is that the lockfile is good to commit to ensure contributors are using the same version of dependencies (where the tooling supports some sort of range syntax). Any links to discussions about that?

ararslan:
Well, it depends on the purpose of the repository. Currently all of the items in this example .gitignore are relevant to Julia _packages_. For Julia packages, committing the Manifest file causes big headaches for running CI, since when a dependent package updates to fix a bug, it's not reflected in CI. For Julia-based _projects_, the Manifest should be committed, as you want a static environment. Since this example .gitignore is currently targeted toward packages, I've opted to include the Manifest file here. Unfortunately I don't have a canonical reference for discussions on this topic, but I can say that I am a developer of Julia itself and have been lightly involved in the development of the package manager. We should probably add some information about this to Julia's documentation.

shiftkey:
>  Since this example .gitignore is currently targeted toward packages, I've opted to include the Manifest file here. 

This is the extra bit of context I needed. Would you be able to add a comment before the rule to indicate:

 - what rule represents
 - why this rule is included and enabled by default
 - when you might want to remove it

ararslan:
Yes, absolutely.

ararslan:
Does this next commit provide sufficient information, or is there anything else you think would be valuable to mention?

ararslan:
Thanks for suggesting they be there! I think that will be helpful for people wondering what these things are, and I wouldn't have thought to comment on them myself.

PR #2993, opened 2019-03-12
fix gradle ignore

hellokent:

Some java file that located at  'src/com/company/build/' would also be ignored.
So I add exclude expression.

_TODO_

_TODO_

If this is a new template:

### DIFF
--- a/Gradle.gitignore
+++ b/Gradle.gitignore
@@ -1,6 +1,9 @@
 .gradle
 /build/
 
+# Ignore some java file
+!/build/**/*.java
+
 # Ignore Gradle GUI config
 gradle-app.setting
 


drothmaler:
As far is I understand the [.gitignore documentation](https://git-scm.com/docs/gitignore) this should not be necessary. It states: 

> A leading slash matches the beginning of the pathname. For example, "/*.c" matches "cat-file.c" but not "mozilla-sha1/sha1.c".

So the existing rule `/build/` should only match the `build` folder in the root of the repo.

shiftkey:
I agree with what @drothmaler said.

Without more context about the change I'm inclined to close this out.

PR #2997, opened 2019-03-13
add Docker.ignore

kobiam:

adding Dockerfile for docker users

https://docs.docker.com/engine/reference/builder/#dockerignore-file

If this is a new template:

docker.com

### DIFF
--- a/Docker.gitignore
+++ b/Docker.gitignore
@@ -1,0 +1,19 @@
+.git
+.idea
+.DS_Store
+.gitignore
+README.md
+env.*
+logs/
+
+# Docker files
+.dockerignore
+Dockerfile
+Dockerfile-dev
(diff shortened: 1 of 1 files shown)


polarathene:
You linked to docs about the `.dockerignore` file.. That is not a `.gitignore` file, it serves a different purpose, it's for ignoring filse when building Docker images rather than files to ignore when working with git..

This PR makes little sense and should be rejected. You don't appear to understand what you're contributing.

shiftkey:
@kobiam while `.dockerignore` looks similar to a `.gitignore` file, their use cases are different enough and a potential source of confusion that I think this should not be added. 

Without more information about what this template represents and why it's important, I cannot take in this PR.

PR #3001, opened 2019-03-14
Update SCons.gitignore

prs513rosewood:
Hi,

I've added some files that SCons generates when its tries to configure libraries.

Missing files that SCons automatically generates when configuring libraries.

**Documentation supporting change**

[Autoconf functionality](https://scons.org/doc/production/HTML/scons-user.html#chap-sconf) does not list generated files, but basically:
- `config.log` contains the compilation logs from SCons trying to link against libraries
- `.sconf_temp` contains the files and executables that SCons tries to compile to test library validity.

### DIFF
--- a/SCons.gitignore
+++ b/SCons.gitignore
@@ -1,2 +1,6 @@
 # for projects that use SCons for building: http://http://www.scons.org/
 .sconsign.dblite
+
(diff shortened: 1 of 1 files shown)


shiftkey:
@prs513rosewood thanks!

PR #3003, opened 2019-03-17
Add Selenium for Python

ignatiusmb:
* Add webdrivers for Python selenium

This adds the executable drivers to be ignored

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -121,3 +121,9 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+
+# Selenium
+chromedriver.*
+MicrosoftWebDriver.*
+geckodriver.*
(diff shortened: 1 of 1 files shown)


shiftkey:
@ignatiusmb given that Selenium can be used in other environments to Python, do you think this might be worth maintaining as it's own `Selenium.gitignore` template?

ignatiusmb:
@shiftkey right, I was thinking that it only uses multiple webdrivers that could be added to each environment. But now that you mention it, I realize it also has its own IDE that could be used with a lot of other APIs and frameworks.

I think creating its own `Selenium.gitignore` is probably the right step, considering it's one of the most popular browser automation tool. Personally, I only use it with Python so I can't really say much about it maintaining it.

PR #3004, opened 2019-03-19
[Android] Change to Android Studio 3

Neargye:

1) There is no need to add cache files one by one if we already ignore all cache files.
2) Since Android Studio 3.0 update, the apk generated will be in `{project-folder}/app/release/app-release.apk`

1) https://github.com/github/gitignore/pull/2644
2) https://stackoverflow.com/questions/46949719/android-studio-3-0-does-not-generate-signed-apk


shiftkey:
@Neargye thanks!

PR #3005, opened 2019-03-19
Update Autotools.gitignore

stevenleeS0ht:
Add ignore-rule for Makefiles generated by configure
(directly by config.status)

Makefiles are also unnecessary to be tracked because they are also be generated.

_TODO_

### DIFF
--- a/Autotools.gitignore
+++ b/Autotools.gitignore
@@ -41,3 +41,6 @@ m4/ltoptions.m4
 m4/ltsugar.m4
 m4/ltversion.m4
 m4/lt~obsolete.m4
+
+# Generated Makefile
+Makefile
--- a/Autotools.gitignore
+++ b/Autotools.gitignore
@@ -42,5 +42,8 @@ m4/ltsugar.m4
 m4/ltversion.m4
 m4/lt~obsolete.m4
 
-# Generated Makefile
+# Generated Makefile 
+# (meta build system like autotools, 
+# can automatically generate from config.status script
+# (which is called by configure script))
 Makefile


shiftkey:
@stevenleeS0ht any links to StackOverflow discussions or other documentation about this change?

alepinio:
you can read in

[https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.69/autoconf.html#Making-configure-Scripts](url)

that `configure` creates the `Makefile`s

and so as `configure` is actually included `Autotools.gitignore`, the `Makefile`s should be also included

stevenleeS0ht:
It would be a common knowledge that Makefile is generated without any handwritten part by Meta-build system. (CMake, SCons and etc.)

Autotools.gitignore
```diff
@@ -41,3 +41,6 @@ m4/ltoptions.m4
 m4/ltsugar.m4
 m4/ltversion.m4
 m4/lt~obsolete.m4
+
+# Generated Makefile
```
shiftkey: Can we make this comment a bit more descriptive? I worry about users pulling in templates and potentially not understanding _why_ an important file like this is actually safe to ignore.

stevenleeS0ht: @shiftkey , I don't think it is necessary. Because, people who use Autotools are usually the people who know why.

In general, people who use those hand-write meta-build system are supposed to know how they work.

stevenleeS0ht: Fine description added.

shiftkey:
@stevenleeS0ht thanks!

PR #3006, opened 2019-03-20
add alternative sftp settings file

GlassGruber:

Adding alternative SFTP config file

### DIFF
--- a/Global/SublimeText.gitignore
+++ b/Global/SublimeText.gitignore
@@ -12,7 +12,7 @@
 
 # SFTP configuration file
 sftp-config.json
-sftp-config-alt.json
+sftp-config-alt*.json
 
 # Package control specific files
 Package Control.last-run
--- a/Global/SublimeText.gitignore
+++ b/Global/SublimeText.gitignore
@@ -12,6 +12,7 @@
 
 # SFTP configuration file
 sftp-config.json
+sftp-config-alt.json
 
 # Package control specific files
 Package Control.last-run


shiftkey:
@GlassGruber any links to StackOverflow discussions or other documentation about this change?

GlassGruber:
Hey there, honestly I can't find much about proper docs on this, the best info I found is the [changelog ](https://wbond.net/sublime_packages/sftp/changelog)for SFTP at version number 1.9.0. 

Basically if you use the SFTP plugin it's possible to create multiple alternative configs

image

Experimenting with this I found also that my actual rule is _naive_ since you can add more than 1 alternative configuration file, creating `sftp-config-alt2.json`, `sftp-config-alt3.json` and so on... 

I'm not big in gitignore shenenigans, but I don't think it's possible to use a regex to target directly multiple digits like `sftp-config-alt\d+\.json`, can similar be achieved with globs?

shiftkey:
Oops, apologies for the delay in replying @GlassGruber:

> I don't think it's possible to use a regex to target directly multiple digits like `sftp-config-alt\d+\.json`, can similar be achieved with globs?

gitignore syntax doesn't support full regex, but you can use `[0-9]` to represent a digit. Does that help to make this rule flexible? Alterntively, if it'll continue beyond 9 and into two digit numbers, maybe a `*` here to say "anything is fine" and it could be merged into the previous rule.

GlassGruber:
> maybe a -alt* here to say "anything is fine" if it matches the format.

Hey there, well I guess it is indeed a sensed way to do it.

PR #3007, opened 2019-03-20
Ignore Local History folder

ofthelit:

This will add support for the default location the Local History extension for Visual Studio Code uses.

See https://marketplace.visualstudio.com/items?itemName=xyz.local-history


shiftkey:
@ofthelit thanks for the suggestion.

This should be submitted against `Global/VisualStudioCode.gitignore`, but an earlier discussion about this in #2651 ended up being closed out because I couldn't validate the rule myself - any external docs or links would be greatly appreciated.

PR #3008, opened 2019-03-20
Ignore Local History folder for VisualStudioCode

ofthelit:

This will add support for the default location the Local History extension for Visual Studio Code uses.

The default path is at the same level as the `.vscode` folder in the workspace.

```
    "local-history.path":     // Specify another location for .history folder (null: use workspaceFolder)
        This settings must be an abolute path. You can also start your path with:
        ${workspaceFolder}: current workspace folder
            e.g. ${workspaceFolder}/.vscode to save in each workspace folder .vscode/.history
        ${workspaceFolder: 0}: first workspace folder
            e.g. workspace folders A, B, C. But save always in A/.history
```

See https://marketplace.visualstudio.com/items?itemName=xyz.local-history

Earlier PRs were #2651 and #3007.


Coeur:
Please update your Pull Request title to mention that it's affecting "VisualStudioCode", as otherwise it's hard to browse through more than a hundred pending pull requests. Thank you.

ofthelit:
I fixed the conflict. This change is still applicable when using the default settings of this extension.

PR #3009, opened 2019-03-20
Remove wwwroot/lib exclusion for ASP.NET Core

jimmylewis:
ASP.NET Core projects no longer use Bower by default (since Bower is now deprecated), and instead create static files in the wwwroot/lib path.  This path is can also be used by convention for ASP.NET Core developers, and since it's no longer populated by Bower, it is unituitive to be excluded by default.

This change removes the lines added by #2307.

VS default flow is now broken by excluding files required to run an ASP.NET Core project.

The changes to the ASP.NET Core templates was tracked by https://github.com/aspnet/templating/issues/48.


shiftkey:
@jimmylewis thanks!

PR #3011, opened 2019-03-21
Add VS Code directory into ignore

cooperj:
Added the directory .vscode into the git ignore, it provides settings about the client, e.g. folder location... This post on Stack Overflow (https://stackoverflow.com/questions/32964920/should-i-commit-the-vscode-folder-to-source-control) says it's you shouldn't commit .vscode because it holds personal settings.

_"you shouldn't commit .vscode because it holds personal settings."_

_https://stackoverflow.com/questions/32964920/should-i-commit-the-vscode-folder-to-source-control_

If this is a new template:

 - **Link to application or project’s homepage**: _n/a_

### DIFF
--- a/Unity.gitignore
+++ b/Unity.gitignore
@@ -17,6 +17,9 @@
 # Visual Studio cache directory
 .vs/
 
+# VS Code directory 
+.vscode/
+
 # Gradle cache directory
 .gradle/
 


shiftkey:
The VSCode team submitted [this accepted answer](https://stackoverflow.com/a/32979933/1363815) which differs to the previous statement:

> Check in the `.vscode` folder if you want to share settings, task configuration and debug configuration with the team. I think generally it makes sense to share settings (e.g. whitespace vs tabs) with the team if you want to enforce settings in a team. We in the VS Code team share debug and task specific settings as well because we want our team to have the same set of debug targets and task targets for VS Code

We have an existing template for VSCode [here](https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore), showing that some files inside `.vscode` can be committed to version control and shared.

PR #3012, opened 2019-03-23
[Android] ignore .idea/navEditor.xml

nistix:

By default, avoid storing the positions of your elements in the Navigation Editor.

[Get started with the Navigation component](https://developer.android.com/guide/navigation/navigation-getting-started)
[Deep dive into .idea folder in Android Studio](https://proandroiddev.com/deep-dive-into-idea-folder-in-android-studio-53f867cf7b70)


shiftkey:
@nistix thanks!

PR #3015, opened 2019-03-25
add OS generated files to Unity.gitignore

Arashk-A:

Removing unwanted OS generated files inside git project.


shiftkey:
Thanks, but these are maintained in [`Global/macOS.gitignore`](https://github.com/github/gitignore/blob/master/Global/macOS.gitignore). If you have any additional rules that can be added there, please submit a PR against that template.

PR #3016, opened 2019-03-26
Ignore transpiled .java and .class files

kevin51jiang:

Ignore the transpiled .java and .class files since they aren't related to processing (they're autogenerated at runtime since Processing is a JVM language)

These files aren't don't influence the source code, and these files are only generated at compile time. Other folders which are generated at compile time (e.g. `application.windows32`) are also ignored by default.


shiftkey:
@kevin51jiang thanks!

PR #3017, opened 2019-03-30
[Dart] Ignore some Flutter-related files

urmilshroff:

Google's Flutter mobile app SDK is used to make Android and iOS apps written in Dart, in IDEs such as Android Studio, IntelliJ and VS Code. That's why it creates many additional useless files that are not ignored by the regular `Dart.gitignore` file, so I've appended all those extra rules in my updated `Dart.gitignore` file itself.

- Here's the [official site](https://flutter.dev/) of Flutter, as well as the [documentation](https://flutter.dev/docs).
- The new rules I added were taken directly from the `.gitignore` file from the [official Flutter repo](https://github.com/flutter/flutter/blob/master/.gitignore).

### DIFF
--- a/Dart.gitignore
+++ b/Dart.gitignore
@@ -19,3 +19,100 @@ doc/api/
 *.js_
 *.js.deps
 *.js.map
+
+# Files and directories generated for Flutter apps: https://flutter.dev/
+# This part of the .gitignore is taken from: https://github.com/flutter/flutter/blob/master/.gitignore
+
+# Miscellaneous
+*.class
+*.lock
+*.log
+*.pyc
+*.swp
+.DS_Store
+.atom/
+.buildlog/
+.history
+.svn/
+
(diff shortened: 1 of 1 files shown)


urmilshroff:
Hey @shiftkey, friendly reminder that this PR is still open. I saw from another similar PR from a while ago that you were interested in a merge of Flutter-ignored files into the regular `Dart.gitignore`, so here it is. Thanks

shiftkey:
@urmilshroff thanks for the contribution. Unfortunately I can't accept this in it's current form for a few reasons:

 - it duplicates OS-specific and editor-specific rules that are captured in other templates that are maintained in this project
 - it excludes editor-specific files that we recommend committing to version control (check out the `JetBrains` and `VisualStudioCode` templates for example)

I'm not sure how to proceed here, and will think more about the options available.

PR #3022, opened 2019-04-06
Fix: python.gitignore ignore unnecessary

atomse:
only directory build in basedir should be ignored.

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -8,7 +8,7 @@ __pycache__/
 
 # Distribution / packaging
 .Python
-build/
+./build/
 develop-eggs/
 dist/
 downloads/


shiftkey:
@atomse any links to StackOverflow discussions or other documentation about this change?

atomse:
No, but I was doing a python project and found a submodule called build does not exist any more and it turned to be .gitignore. Generally, the build in the root dir is what python made and should be ignored, but the build inside the project seems should not.

The tree of my project
.
├── atomse
│   ├── bin
│   ├── build
│   │   └── __pycache__
│   ├── calculators
│   │   └── __pycache__
│   ├── cli
│   │   └── __pycache__
│   ├── debug
│   ├── generator
│   │   └── test_cases
│   ├── io
│   │   ├── base_format
│   │   ├── format_string
│   │   │   └── __pycache__
│   │   ├── __pycache__
│   │   └── testcases
│   ├── main
│   │   └── __pycache__
│   ├── network
│   ├── __pycache__
│   ├── test
│   ├── ts
│   └── visualize
└── build

The last build is in the root dir which is generated by python and should be ignored, but the first build in in the project. If build/ in .gitignore, then the first one is ignored as well, but I think it's unnecessary.

Of course, that's what I meet in my project. If you think this PR is unnecessary, please close it.

atomse:
```bash
.
├── atomse
│   ├── bin
│   ├── build
│   │   └── __pycache__
│   ├── calculators
│   │   └── __pycache__
│   ├── cli
│   │   └── __pycache__
│   ├── debug
│   ├── generator
│   │   └── test_cases
│   ├── io
│   │   ├── base_format
│   │   ├── format_string
│   │   │   └── __pycache__
│   │   ├── __pycache__
│   │   └── testcases
│   ├── main
│   │   └── __pycache__
│   ├── network
│   ├── __pycache__
│   ├── test
│   ├── ts
│   └── visualize
└── build
```
Tree of my project.

Anaxilaus:
There could be more packages in a complex project. Their purposes differs greatly and the imports becomes complicated, unnecessarily long. 

There could be also a time-wise concern while rebuilding for testing etc.

If someone wants to commit build directory, they can just customize .gitignore according to their own needs. Not the other way around, IMO.

shiftkey:
> If build/ in .gitignore, then the first one is ignored as well, but I think it's unnecessary.

In that case it contains a `__pycache__` entry (directory?) which should be ignored due to an earlier rule.

I think the convention here of `build` containing build artifacts at any level in the repository is still reasonable to keep around, but as @Anaxilaus points out this can be changed locally by the user if they have a personal preference.

atomse:
All right. I'll close the PR.

PR #2957, opened 2019-02-07
Update VisualStudio.gitignore

happydude:
Ignore backup files from editing business intelligence reports.

https://feedback.azure.com/forums/908035-sql-server/suggestions/32897485-ssrs-2016-creates-a-new-rdl-backup-file-each-time

PR #3024, opened 2019-04-06
Remove *- Backup*.rdl from VisualStudio.gitignore

ElanHasson:

This probably should not be here. A .rdl file is a valid thing to version in source control.

Having the name _*Backup*_ could mean anything.

n/a


drothmaler:
It was added in #2957 because SSRS 2016 seems to create these backup files automatically. So it is not just any random Backup file... So I think it should not be removed...

ElanHasson:
I haven't actually encountered that, but then again I'm not using report builder, I'm in Visual Studio.

I agree it should stay unfortunately.

PR #3025, opened 2019-04-07
KiCAD: the -rescue.lib file is also precious!

jnavila:

the -rescue.lib file is also precious! 

https://forum.kicad.info/t/lost-the-schematic-components/12026/11
https://forum.kicad.info/t/how-to-clone-a-project/3878/23

### DIFF
--- a/KiCad.gitignore
+++ b/KiCad.gitignore
@@ -9,7 +9,6 @@
 *~
 _autosave-*
 *.tmp
-*-rescue.lib
 *-save.pro
 *-save.kicad_pcb
 


shiftkey:
@jnavila any links to StackOverflow discussions or other documentation about this change?

jnavila:
Well, first of all my own prose :laughing: : https://jnavila.github.io/plotkicadsch/ , but more seriously, there's an explanation for the life cycle of `-cache.lib` and `-rescue.lib` files [here](https://forum.kicad.info/t/lost-the-schematic-components/12026/11) or [here](https://forum.kicad.info/t/how-to-clone-a-project/3878/23)

bartlettD:
This is the right move, the `-rescue.lib` file is needed for schematics drawn with the Eeschema tool in KiCAD to resolve symbols when those symbols dissapear or are modified from the Global or Project libraries.

_Ideally,_ a user should never have a `-rescue.lib` because symbols will never dissapear or wont be changed, but sometimes a change to the library may break a design an so a "rescue" is a safer option.

[The KiCAD docs explain this in a bit more detail.](
http://docs.kicad-pcb.org/5.1.2/en/eeschema/eeschema.html#rescuing-cached-components)

shiftkey:
@jnavila @bartlettD thanks!

PR #3026, opened 2019-04-08
[Unity] ignore Jetbrains plugin folder

LotteMakesStuff:
When a Unity project is opened in Jetbrains Rider, it installs a editor plugin into the Assets/Plugins/Editor/Jetbrains folder to manage Unity->Rider integration. This plugins life cycle is managed by your local Rider install and is automatically updated by Rider. It should not be committed to source control.  This is documented by Jetbrains here https://www.jetbrains.com/help/rider/Unity.html

This change should ignore both the Jetbrains plugin folder and its .meta file correctly.


shiftkey:
@LotteMakesStuff thanks!

PR #3027, opened 2019-04-09
Ignoring Visual Studio Code workspace files.

W1M0R:

Unreal Engine can generate Visual Studio Code workspace project files.

_TODO_

If this is a new template:

### DIFF
--- a/UnrealEngine.gitignore
+++ b/UnrealEngine.gitignore
@@ -40,6 +40,7 @@
 *.sdf
 *.VC.db
 *.VC.opendb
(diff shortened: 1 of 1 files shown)


shiftkey:
@W1M0R thanks for the contribution!

I'm not sure I want to ignore these for every Unity project, as some people might want to use them. Are you able to elaborate on why this should be the default?

shiftkey:
Closing this out due to inactivity, but #3216 is also discussing this proposed change.

PR #3028, opened 2019-04-10
Modifies Unity gitignore to keep saved backup scenes

CarolineClark:

This PR is to save backup scenes in Unity. 
On playing an unsaved scene, Unity saves this in `Temp/__Backupscenes/0.backup`. However, on opening Unity after the crash (i.e., at the point where one realises they have lost work), it deletes this file.
My suggestion is to add it to the version control. Thus if a crash happens, one can commit their changes, and be secure they have a backup in their commit history.

https://support.unity3d.com/hc/en-us/articles/210224803-Can-I-restore-my-unsaved-scene-after-Unity-crash-

### DIFF
--- a/Unity.gitignore
+++ b/Unity.gitignore
@@ -1,11 +1,14 @@
 # This .gitignore file should be placed at the root of your Unity project directory
 
 /[Ll]ibrary/
-/[Tt]emp/
+/[Tt]emp/*
 /[Oo]bj/
 /[Bb]uild/
 /[Bb]uilds/
 /[Ll]ogs/
+
+# Keep backup scenes
+![Tt]emp/__Backupscenes
 
 # Never ignore Asset meta data
 !/[Aa]ssets/**/*.meta


shiftkey:
@CarolineClark thanks for the contribution!

> Thus if a crash happens, one can commit their changes, and be secure they have a backup in their commit history.

I'm a bit nervous about committing backup files to Git (particularly when they're stored in a temporary directory), because if they're large or binary files they have the potential to bloat the repository history. I'd love a sanity check on this from other Unity contributors to the template, cc @FritzsHero @LucasSteer @qCring @Flate @shana 

I also don't have access to that link from the Unity forums - it requires a login to see the page.

FritzsHero:
Hello CarolineClark,

thank you for your suggestion.
I can not really see a reason why this Scene backup should be included into git commits.
Whenever you enter the Playmode of Unity, Unity will create a Backup of the Scenes you are playing.
For the most parts you never make drastic changes inside the Playmode, since you have to manually copy these changes over anyway, into the original Scene. Whenever you leave the Playmode you'll find the untouched version of the original Scene.

So it depends on how someone is working, but it might be a better solution do to this for your specific projects.

Greetings
Fritzs

qCring:
I agree with @FritzsHero. It really is temporary user/machine-specific data. There might be some scenarios where it makes perfect sense to track such files in your VCS. But, objectively, I'd say those are edge cases. By default I would expect such files to be ignored as implied by their location in the Temp folder.

LucasSteer:
I agree with @qCring, @FritzsHero, and @shiftkey. Changes made within Playmode don't seem important enough that the average user would want them in the commit history. In addition, these backups could bloat the git history unnecessarily. Lastly, retaining the backup is done by a manual action (committing, something you wouldn't normally do after a crash) with this solution; since a manual action is being performed anyway, it would make more sense to manually make a copy of the backup files. Overall, this doesn't seem like something that the average user would want by default

shiftkey:
It looks like there's lots of feedback here that this doesn't feel like the right change to have available by default. @CarolineClark are you fine with me closing this out, or is there more context here that we've not covered?

shiftkey:

PR #3030, opened 2019-04-11
Python.gitignore: editor .swp files

zzhou1:
Editor .swp files is annoying  ;)

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -104,6 +104,9 @@ ENV/
 env.bak/
 venv.bak/
 
+# Editor stuff
+*.swp
+
 # Spyder project settings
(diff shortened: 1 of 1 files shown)


shiftkey:
@zzhou1 these are from Vim, right? [We have a separate template for Vim-specific rules](https://github.com/github/gitignore/blob/master/Global/Vim.gitignore)...

zzhou1:
> @zzhou1 these are from Vim, right? [We have a separate template for Vim-specific rules](https://github.com/github/gitignore/blob/master/Global/Vim.gitignore)...

Aha, nice, perfect, thanks the great work !

PR #3031, opened 2019-04-16
added Pycharm/IntelliJ files

gkontos89:

JetBrains makes a Python IDE called PyCharm.  PyCharm projects generate ".idea" folders that contain meta data including version control history, workspace and other environment specific files that should no be version controlled.

[PyCharm IDE](https://www.jetbrains.com/pycharm/)

If this is a new template:

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -121,3 +121,6 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+
+# IntelliJ Pycharm
+*.idea


mmhyamin:
Check: https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore

shiftkey:
@gkontos89 thanks for the suggestion, but we have a [`Global\JetBrains.gitignore`](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore) template that documents what files can be committed to version control and shared. We recommend that over a blanket ignore rule.

BrianMSheldon:

PR #3034, opened 2019-04-16
Update R.gitignore

zy26:

Ignore knitr and R markdown default cache directories in subdirectories as well


shiftkey:
@zy26 thanks!

PR #3035, opened 2019-04-20
Ignore .DS_Store which only Apple generates

wilsonmar:

VisualStudio 2017 and 2019 for Mac runs on macOS. 
And macOS creates .DS_Store files that are not used by other file systems.

If this is a new template:

This change is not a new template.


shiftkey:
@wilsonmar thanks, but see [`Global/macOS.gitignore`](https://github.com/github/gitignore/blob/master/Global/macOS.gitignore) for these OS-specific rules. We recommend composing templates together based on your tooling needs.

PR #3036, opened 2019-04-22
JetBrains CLion ignore for CMake

Pythonic-Rainbow:

`.idea` and `cmake-build-debug` are IDE generated files and varies from developers

https://www.jetbrains.com/help/clion/managing-cmake-project-files.html

If this is a new template:
N/A


shiftkey:
Thanks, but see [`Global\JetBrains.gitignore`](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore) for these rules and others...

PR #3039, opened 2019-04-23
Add ignore rules for IntelliJ and IntelliJ like IDEs

judepereira:

The IntelliJ settings frequently differ across environments, and since other IDEs such as BlueJ is present, I figured that it might be a good idea to add these.

https://www.jetbrains.com/help/idea/creating-and-managing-projects.html#project-formats

If this is a new template:

NA.

### DIFF
--- a/Java.gitignore
+++ b/Java.gitignore
@@ -10,6 +10,9 @@
 # IntelliJ and IntelliJ like IDEs (JetBrains)
 .idea
 *.iml
+*.ipr
+*.iws
+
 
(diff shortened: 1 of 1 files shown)


shiftkey:
Thanks, but we have an existing [`Global\JetBrains.gitignore`](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore) template that can be added to projects, instead of duplicating rules across templates.

judepereira:
Got it. Thanks for reviewing!

PR #3040, opened 2019-04-23
Delete .ist from TeX.gitignore

Anton-Sakovich:

.ist files are makeindex style files, which determine how a
makeindex-generated index will look like. Therefore, they must be
included in source control.

A link to an example .ist file for customizing the look of an index:
https://tex.stackexchange.com/questions/245368/how-to-create-a-ist-file-for-the-purpose-of-customizing-the-index

### DIFF
--- a/TeX.gitignore
+++ b/TeX.gitignore
@@ -144,7 +144,6 @@ acs-*.bib
 *.idx
 *.ilg
 *.ind
-*.ist
 
 # minitoc
 *.maf


cdlm:
👍

At the time I probably thought that it was a generated file; I don't even recall working on a document with an index 🤷🏻‍♂️

shiftkey:
Thanks for the follow up @cdlm, and thanks for the contribution @Anton-Sakovich!

PR #3041, opened 2019-04-25
Update VisualStudio.gitignore to ignore NuGet Symbol Packages

alefranz:
This is to ignore the new Nuget Symbol Packages, which are now the recommended approach for debug symbols of public NuGet packages.

https://docs.microsoft.com/en-us/nuget/create-packages/symbol-packages-snupkg


shiftkey:
@alefranz thanks!

alefranz:
That was really quick! :) thanks!

PR #3042, opened 2019-04-25
Changes to match the Unreal Engine .gitignore with the real one

Chkoupinator:

It's in the default Unreal Engine .gitignore when creating a new project on GitHub 

Just try making a new empty project and select Unreal Engine as a .gitignore template

### DIFF
--- a/UnrealEngine.gitignore
+++ b/UnrealEngine.gitignore
@@ -1,6 +1,9 @@
 # Visual Studio 2015 user specific files
 .vs/
 
+# Visual Studio 2015 database file
+*.VC.db
+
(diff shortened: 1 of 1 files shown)


shiftkey:
This was previously removed in https://github.com/github/gitignore/pull/2741 because it's already duplicated. The list on GitHub is updated from here but maybe it needs to be refreshed?

Chkoupinator:
I guess it does need to be refreshed, I tried loading the UE .gitignore on
GitLab and I didn't get the duplicate, I assume there is an issue with
GitHub somehow

PR #3046, opened 2019-04-30
Updating Swift.gitignore with accio folders

pogramos:

This PR includes both the config and dependency folders for [Accio](https://github.com/JamitLabs/Accio)

https://github.com/JamitLabs/Accio#usage

### DIFF
--- a/Swift.gitignore
+++ b/Swift.gitignore
@@ -58,6 +58,10 @@ playground.xcworkspace
 
 Carthage/Build
 
+# Accio dependency management
+Dependencies/
+.accio/
+
 # fastlane
 #
 # It is recommended to not store the screenshots in the git repo. Instead, use fastlane to re-generate the


shiftkey:
In a nutshell, these are the default rules from Accio:

https://github.com/JamitLabs/Accio/blob/c96e78d0e08dcdb2b3a073246a66f27666f318b1/Sources/AccioKit/Globals/Constants.swift#L6-L7

```swift
    static let buildPath: String = ".accio"
    static let dependenciesPath: String = "Dependencies"
```

And this is the code that adds them to the `.gitignore` when you `accio init`:

https://github.com/JamitLabs/Accio/blob/ee33988927bfcef09b76e29dabef4640b3f781ba/Sources/AccioKit/Services/GitIgnoreIntegrationService.swift#L32-L33

```swift
 gitignoreContents += "# Accio dependency management\n\(Constants.dependenciesPath)/\n\(Constants.buildPath)/\n"
 try gitignoreContents.write(toFile: gitignoreUrl.path, atomically: true, encoding: .utf8)
```

shiftkey:
@pogramos thanks!

PR #3047, opened 2019-05-02
Update VisualStudio.gitignore to add a safer rule for rdl backups

aruanoguate:

In #3024 there was a discussion about the rule for SSRS rdl backups that can match something by mistake. I'm updating it to have a more specific version.

https://feedback.azure.com/forums/908035-sql-server/suggestions/32897485-ssrs-2016-creates-a-new-rdl-backup-file-each-time


shiftkey:
@aruanoguate thanks!

PR #3048, opened 2019-05-02
Update VisualStudio.gitignore for VS family & .NET

metoly:
Update VisualStudio.gitignore to cover all products in the VS family and .NET including:
VS IDE
VS4Mac
VS Code
.NET CLI
Supporting mixed development teams on different IDEs and different platforms.

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -346,3 +346,92 @@ healthchecksdb
 
 # Backup folder for Package Reference Convert tool in Visual Studio 2017
 MigrationBackup/
+
+##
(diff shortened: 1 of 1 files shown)


shiftkey:
@metoly thanks for the contribution!

These templates are designed to be composed together based on your setup and technologies, and merging multiple templates together like this adds a significant maintenance burden that I can't take on currently.

metoly:
@shiftkey,
Thanks. We need these changes to support cross platform projects in Visual Studio. I already verified the changes with VS Code, VS4Mac and VS IDE.

PR #3049, opened 2019-05-09
Exclude memory profiler captures from git

FritzsHero:

Excluding memory captures from git since they use 80-500MB or even more.

https://docs.unity3d.com/Packages/com.unity.memoryprofiler@0.1/manual/index.html
Note: The memory profiler is still a preview package!
Also it may or may not make sense to exclude folders used by specific packages.


shiftkey:
@FritzsHero thanks!

PR #3050, opened 2019-05-09
Replace non-ASCII character in comment

jrc:

The curly apostrophe in the pipenv comment (added in commit 7079791) is causing AWS EB CLI to fail. https://stackoverflow.com/a/55988700/594211


shiftkey:
@jrc thanks! Definitely not needed if that's the sort of problem it's introducing...

PR #3051, opened 2019-05-09
Add macOS to Visual Studio .gitignore

RLittlesII:

As a .NET Developer that uses a Mac I add `.DS_Store` to every .gitignore file.

None.  I'd just like macOS as part of the main VS template.

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -346,3 +346,6 @@ healthchecksdb
 
 # Backup folder for Package Reference Convert tool in Visual Studio 2017
 MigrationBackup/
+
+# macOS
+.DS_Store


shiftkey:
> None. I'd just like macOS as part of the main VS template.

This repository is intended to be a reference, so that consumers can compose together the templates how they need. I've been pushing back on these sorts of "convenience PRs" because they become a maintenance burden that I cannot take on.

In this case, we've duplicates a rule from `Global/macOS.gitignore` into this template. It's a small thing, but I'm still trying to undo years of these PRs being merged.

I've got two suggestions to achieve what you're looking for:

 - [Create a global gitignore](https://help.github.com/en/articles/ignoring-files#create-a-global-gitignore) on your machine, which is applied to any Git repository you use. The downside that this isn't committed with the repository, but maybe that's fine.
 - Try out [`gitignore.io`](https://gitignore.io/) that helps you compose together a gitignore file using this set of templates as well as some others.

PR #3052, opened 2019-05-09
Add .tsbuildinfo files to Node.gitignore

SimonSiefke:

Typescript 3.4 has added `.tsbuildinfo` files for caching. 

"These .tsbuildinfo files can be safely deleted and don’t have any impact on our code at runtime - they’re purely used to make compilations faster"

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html

### DIFF
--- a/Node.gitignore
+++ b/Node.gitignore
@@ -43,6 +43,9 @@ jspm_packages/
 # TypeScript v1 declaration files
 typings/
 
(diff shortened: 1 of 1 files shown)


shiftkey:
@SimonSiefke thanks!

PR #3054, opened 2019-05-13
updated vscode ignore, now it's works

danbailo:

the old version of .gitignore for vscode folder did not works for me, so I've change this file and add new line and it's works now

_TODO_

If this is a new template:

### DIFF
--- a/Global/VisualStudioCode.gitignore
+++ b/Global/VisualStudioCode.gitignore
@@ -1,3 +1,4 @@
+.vscode/
 .vscode/*
 !.vscode/settings.json
 !.vscode/tasks.json


shiftkey:
> the old version of .gitignore for vscode folder did not works for me, so I've change this file and add new line and it's works now

Can you elaborate on why this is the case, showing examples?

danbailo:
of course, but I will need to show images for a better explanation

danbailo:
[Image to Explain here](https://drive.google.com/open?id=1J-Ar_Gzr41TmRz8zzznIqN3du2OoJTq1)

Ok, so, following the order from 1 until 4. In the image 1 I've executed `$ ls -al ` on terminal in my local repo to show all files and folders. After this, I executed `$ git status` and `$ cat .gitignore`, to show that my repo are up to date and you can see my actual version of .gitignore .
In the final of image 1, I've modified the .gitignore and execute again `$ git status` to show that now, my repo are seeing the `.vscode/` how we can see, in image 2. After this I executed `$ cat .gitignore` to show the modified .gitignore and, how we can see, it's same that the older version before I update it and this didn't work!
In image 3 and 4, I've modified the .gitignore and comment the older version, in line `.vscode/*` and see that this line not is necessary, but I decided keep it because will be guaranteed that will works for everyone.

shiftkey:
@danbailo it's not clear to me what's inside your `.vscode` directory, and the rule you've added is to say "ignore the directory" which then overwrites all of the subsequent rules. The default template here indicates there are files in the directory that can be committed, and I think that's where the confusion is coming in.

If you run `git status --untracked-files=all` in your repository without this change you'll see the full list of files that Git would like to add. If it matches these files:

https://github.com/github/gitignore/blob/74643fe2c8a8ca02384f6367be2393c00cf8cd24/Global/VisualStudioCode.gitignore#L2-L5

Then I believe the template is working as intended, and the files found can be committed and shared.

danbailo:
Got it! I was thinking that the purpose it was ignore the directory `.vscode/` inside a repository... Thanks for your explanation and sorry for the inconvenience.

shiftkey:
@danbailo no problem, and thanks for walking me through it from your side!

PR #3055, opened 2019-05-14
add model checkpoint, dataset, image, font

max-yue:

some files should be ignored

_TODO_

If this is a new template:

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -121,3 +121,44 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+
+# tensorflow checkpoint files
+*.data-*
+*.index
+*.meta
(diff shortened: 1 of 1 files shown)


shiftkey:
> some files should be ignored

I need more context to understand this change, because it currently replicates at least this template:

 - [`Global/Images.gitignore`](https://github.com/github/gitignore/blob/master/Global/Images.gitignore)

And if you're crafting some rules for using TensorFlow, which can be used in a bunch of different situations, maybe that could live in it's own template?

shiftkey:

@max-yue if you're still interested in submitting this contribution, please open a fresh pull request and address the outstanding questions from this review.

PR #3056, opened 2019-05-14
add the ist file generated by glossaries-extra package

MohHizzani:
*.ist files that are generated by glossaries-extra package


shiftkey:
@MohHizzani thanks!

PR #3057, opened 2019-05-14
Updated Craft CMS .gitignore to recommendation

brandonkelly:
Now based on https://github.com/craftcms/craft/blob/master/.gitignore

The current Craft CMS .gitignore is for an old version of Craft. This is now updated for Craft 3.

This is the same .gitignore you’d get if you created a new Craft project using the [installation instructions](https://docs.craftcms.com/v3/installation.html).

### DIFF
--- a/CraftCMS.gitignore
+++ b/CraftCMS.gitignore
@@ -1,4 +1,4 @@
-# Craft 2 Storage (https://craftcms.com/support/craft-storage-gitignore)
-# not necessary for Craft 3 (https://github.com/craftcms/craft/issues/26)
-/craft/storage/*
-!/craft/storage/rebrand
+/.env
+/.idea
+/vendor
+.DS_Store


shiftkey:
@brandonkelly thanks for the contribution!

> This is the same .gitignore you’d get if you created a new Craft project using the installation instructions.

Is this worth keeping around if there's now an included `.gitignore` with the project? Given how small this template is, I'm curious the value of continuing to maintain it here.

It also contains rules that are overlap with other templates (and the other templates are more refined or have differing opinions on what can be committed):

 - `Global/macOS.gitignore`

brandonkelly:
I’d be fine with just removing it.

shiftkey:
@brandonkelly let's do that then

shiftkey:

@brandonkelly feel free to submit a fresh PR which addresses the feedback provided if this is still something you'd like to see land.

PR #3058, opened 2019-05-15
vim: add extra session file Sessionx.vim

wilhelmtell:
from vim's documentation on `:mksession` (:help :mksession):

 ...

 10. If a file exists with the same name as the Session file, but ending
     in "x.vim" (for eXtra), executes that as well.  You can use *x.vim
     files to specify additional settings and actions associated with a
     given Session, such as creating menu items in the GUI version.

we already have Session.vim ignored.  the Sessionx.vim file, like
Session.vim, is a user file.  a user would generally want that file kept
private or for themselves, and the public or a team fetching from or
sharing the repository generally have no interest in a file relevant
only to a particular user.  so it's a good idea to get git to help us
avoid mistakenly sharing the file.

### DIFF
--- a/Global/Vim.gitignore
+++ b/Global/Vim.gitignore
@@ -7,6 +7,7 @@
 
 # Session
 Session.vim
+Sessionx.vim
 
 # Temporary
 .netrwhist


wilhelmtell:
It's unfortunate that Vim's session filenames are indistinguishable from any other Vim script. A user may create a session with `:mksession yesterday.vim`, and there is no reasonable way for a gitignore to capture that, because, as far as telling by name, that file could be a shared vimscript as well as a user's session file. Session files can take any of a gazillion names, and so so do their accompanying *x.vim file...

shiftkey:
@wilhelmtell thanks!

PR #3059, opened 2019-05-16
Update Unity.gitignore

nicocavi:

_TODO_

_TODO_

If this is a new template:

### DIFF
--- a/Unity.gitignore
+++ b/Unity.gitignore
@@ -24,6 +24,7 @@
 
 # Visual Studio cache directory
 .vs/
+.vscode/
 
 # Gradle cache directory
 .gradle/


Unity.gitignore
```diff
@@ -24,6 +24,7 @@
 
 # Visual Studio cache directory
 .vs/
+.vscode/
```
shiftkey: We have an existing VSCode template at [`Global\VisualStudioCode.gitignore`](https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore) that shows what you can be committed and shared.

shiftkey:
@nicocavi thanks for the contribution!

I'm going to close out this PR as we have an existing [`Global/VisualStudioCode.gitignore`](https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore) for our VSCode template which shows what can be committed and shared.

PR #3060, opened 2019-05-17
 Add MS-DOS Device Driver names for Global/Windows.gitignore

nehcdahc:

Some colleagues will submit files named MS-DOS Device Driver names on Linux or other non-Windows systems. After git repo is cloned on Windows, there may be problems.

MS-DOS Device Driver names cannot be used as file names in Windows. 
As a result, we are unable to name folders as con, aux, nul, etc.

https://www.thewindowsclub.com/create-restricted-files-folders-con-aux-nul-click-windows

### DIFF
--- a/Global/Windows.gitignore
+++ b/Global/Windows.gitignore
@@ -22,3 +22,16 @@ $RECYCLE.BIN/
 
 # Windows shortcuts
 *.lnk
+
+# MS-DOS Device Driver names cannot be used as file names in Windows. 
+# As a result, we are unable to name folders as con, aux, nul, etc.
+CON
+PRN
+NUL
+COM1
+COM2
+COM3
+COM4
+LPT1
+LPT2
+LPT3


shiftkey:
@nehcdahc thanks for the PR!

I'm not sure this is the entire solution, because these file names can also contain extensions, and be lowercase (gitignore rules are case-sensitive). My favourite example in this space is cloning the [Linux source code repository](https://github.com/torvalds/linux) on Windows, which contains a number of files which are caught by these file patterns:

```
...
error: unable to create file drivers/gpu/drm/nouveau/nvkm/subdev/i2c/aux.c: No such file or directory
error: unable to create file drivers/gpu/drm/nouveau/nvkm/subdev/i2c/aux.h: No such file or directory
...
error: unable to create file include/soc/arc/aux.h: No such file or directory
```

This is not likely to ever be fixed, because these file names have a very specific meaning for the project, and the value of this is likely to override requests for compatibility with other OSes.

Even if we make these rules more adaptable to support the different scenarios, the second problem here is the impact to the workflow of users because of how `gitignore` rules are applied. Consider two users working in a repository with these rules enabled - one working on Linux, the other working on Windows:

 - the user on Linux creates a file named `LPT1`, as part of a bunch of changes
 - the Linux user creates a commit, **but the file is silently excluded** from the commit
 - the Linux user pushes their changes to the remote
 - the Windows users sees there are commits, pulls down the changes, but it doesn't work because the required file was excluded
  - meanwhile the Linux user does a `git clean -xdf` to remove any untracked files and switch to some other work, and they too lost the file with the problem name

I believe these `.gitignore` templates should be used to outline what things can be safely excluded from a Git repository. Excluding these files because of an OS-specific limitation feels very different to that, and could lead to data loss on unsuspecting users because of how Git works. These rules can also be overridden by users with a `git add -f` in the event that a user notices this problem but doesn't understand why.

I believe alternatives should be used when a repository needs to work across different OSes, such as code review or better guidance for teams in this situation. For example, I refer to this link a bunch whenever this situation comes up, as it outlines the various restrictions for the Windows OS around filesystem support for filenames: https://docs.microsoft.com/en-us/windows/desktop/fileio/naming-a-file

nehcdahc:
@shiftkey 
Thanks!
Yes, what you said is indeed problems that I have not considered carefully.
My original intention was to let this situation be avoided by the program, not the developer. Code reviews or programming manuals are still too dependent on developers.
So, I thought about it, adding these rules to the project's build script, rather than the gitignore file might be my solution?

PR #3061, opened 2019-05-18
style: (Golang) Remove the non-essentials

erdaltsksn:

We don't have to state that this is for Hugo. It is very clear from filename and [generator](https://www.gitignore.io/api/hugo) will add `###Hugo###` at the top.

These changes are style related so I'm not sure I can find any documentation about that. `CONTRIBUTING.md` file doesn't include any information about styling.


shiftkey:
@erdaltsksn thanks!

PR #3066, opened 2019-05-22
Make ignoring .ist files for glossaries-extra ...

Anton-Sakovich:
There has been an accepted pull request recently which added `*.ist` pattern to the bottom of the `TeX.gitignore`. I see two issues with this pull request:

1. __The major one:__ `*.ist` files are also custom style files for `makeindex`. Therefore, simply adding `*.ist` pattern under `#glossaries-extra` section can be misleading for those who use `makeindex` but not `glossaries-extra`. This problem is similar to the `*.pdf` problem. So, I suggest solving it the same way: let's add a commented line for `*.ist` files under `#glossaries-extra` section with an extended comment.

2. __The minor one__: there is already `#glossaries` section in `TeX.gitignore`. I suggest moving `*.ist` pattern to this section for easier navigation.

### DIFF
--- a/TeX.gitignore
+++ b/TeX.gitignore
@@ -110,6 +110,9 @@ acs-*.bib
 *.gls
 *.glsdefs
 
+# uncomment this for glossaries-extra (will ignore makeindex's style files!)
+# *.ist
+
 # gnuplottex
 *-gnuplottex-*
 
@@ -259,6 +262,3 @@ TSWLatexianTemp*
 
 # standalone packages
 *.sta
-
-# glossaries-extra
-*.ist


shiftkey:
cc @MohHizzani who submitted #3056 - do those changes make sense for you?

MohHizzani:
glossaries-extra is a different package, but I thinks its okay to move it to glossary section, and yes it may cause a problem form who uses makeindex, but you comment it out

Anyway, these changes are good

shiftkey:
Thanks @Anton-Sakovich and @MohHizzani!

PR #3068, opened 2019-05-23
Include Library/*.asset files for Unity3D

der-hugo:

For me it made sense to also include the [Ll]ibrary/*.asset files.

This includes e.g. the build target and build settings so after cloning on a different device I don't have to go to the Build settings and switch platform or re-add all required scenes.

There isn't really one, this is just my experience.

### DIFF
--- a/Unity.gitignore
+++ b/Unity.gitignore
@@ -61,3 +61,6 @@ sysinfo.txt
 # Crashlytics generated file
 crashlytics-build.properties
 
+# keep the *.asset files from Library
+# including e.g. the current build target and build settings
+!/[Ll]ibrary/*.asset


Coeur:
Please update your Pull Request title to mention that it's affecting `Unity`: it's hard enough to browse through more than a hundred pending pull requests. Thank you.

jdpoenisch:
@Coeur Done. Sorry for the inconvenience!

shiftkey:
@jdpoenisch thanks for the contribution!

> There isn't really one, this is just my experience.

Does anyone else agree or disagree with this change? 

cc @Flate @shana @johnsoncodehk @FutureClemens @ksmit799 @LucasSteer @matzar @cscharfen

ksmit799:
The library folder is a cache folder, and as such would only serve to clutter commits. Rather then relying on cached configuration options, I would recommend setting up your own build scripts to simplify the building process across devices (see [here](https://support.unity3d.com/hc/en-us/articles/115000368846-How-do-I-support-different-configurations-to-build-specific-players-by-command-line-or-auto-build-system-)).

jdpoenisch:
> 
> 
> The library folder is a cache folder, and as such would only serve to clutter commits. Rather then relying on cached configuration options, I would recommend setting up your own build scripts to simplify the building process across devices (see [here](https://support.unity3d.com/hc/en-us/articles/115000368846-How-do-I-support-different-configurations-to-build-specific-players-by-command-line-or-auto-build-system-)).

That's true in general for the Library folder .. that's why it is ecluded by the gitignore already.

Just as said for me it made sense to keep at least some of the `*.asset` files and rather to implement additional methods for something that actually already exists in the UnityEditor I decided to simply keep them and never had to worry about switching the platform or adding all scenes again to the build settings when cloning to a new device.

ksmit799:
You're free to make your own local changes, however in general I think many developers would not want their build configurations being tracked by git. This is especially true when working on multi-team projects, as there will often be several different developers building with different configurations simultaneously.

johnsoncodehk:
In a simple case, osx and windows build settings are not common, like ios build only exists in osx and not in windows.

zipzit:
> der-hugo commented on Jul 25, 2019 • 
> The library folder is a cache folder, and as such would only serve to clutter commits. Rather then relying on cached configuration options, I would recommend setting up your own build scripts to simplify the building process across devices (see here).
> 
> That's true in general for the Library folder .. that's why it is ecluded by the gitignore already.

Not.  That's NOT what the current .gitignore for unity says.   

Currently top of .gitignore for Unity says:
  
```
# This .gitignore file should be placed at the root of your Unity project directory
#
# Get latest from https://github.com/github/gitignore/blob/master/Unity.gitignore
#
/[Ll]ibrary/
/[Tt]emp/
/[Oo]bj/
/[Bb]uild/
/[Bb]uilds/
/[Ll]ogs/
/[Uu]ser[Ss]ettings/
```

Everything in the root of `Library` is included in the repo.    Why isn't that first line (and the rest of 'em)   `*/[Ll]ibrary`   simple and pure?

cscharfen:
> You're free to make your own local changes, however in general I think many developers would not want their build configurations being tracked by git. This is especially true when working on multi-team projects, as there will often be several different developers building with different configurations simultaneously.

Absolutely true!
Plus: Generated files shall not be submitted to the git repo. Use shared Cache Servers to share cached & generated files, but not git.

matzar:
> > You're free to make your own local changes, however in general I think many developers would not want their build configurations being tracked by git. This is especially true when working on multi-team projects, as there will often be several different developers building with different configurations simultaneously.
> 
> Absolutely true!
> Plus: Generated files shall not be submitted to the git repo. Use shared Cache Servers to share cached & generated files, but not git.

Yes, and because it would require some extra set-up for `.gitignore` to work properly I wouldn't add it to the official one; which should work, without any extra set-up. 

Mark, I've put `Library/*.asset` into my `.gitignore` in the past; used that `.gitignore` from the first commit onward; did not store `Library/*.asset`  anywhere else; lost all my local copies and effectively lost all of my `Library/*.asset`  files.

So I'd say, there might be an argument against putting this one into the official `.gitignore`.

cscharfen:
@der-hugo It is probably to late to tell you now or you already figured it out yourself, but in case not or someone else comes across this, you might be interested in [Unity Cache Server](https://docs.unity3d.com/Manual/CacheServer.html) / [Unity Accelerator](https://docs.unity3d.com/Manual/UnityAccelerator.html) depending on your [Asset Import Pipeline](https://blogs.unity3d.com/2019/10/31/the-new-asset-import-pipeline-solid-foundation-for-speeding-up-asset-imports/).

bdougie:
This has gotten a bit stale and now has conflicts. Moving forward there will be easier merged with links to community documentation. If anyone wants to pick this up in a new PR, please provide links and context.

PR #3069, opened 2019-05-24
Added pycharm folder: ./idea

xanjay:

This folder (.idea/) in PyCharm (IDE for python) project contains files which store user-specific settings which can cause project desynchronization.

https://stackoverflow.com/questions/24516814/should-i-ignore-the-idea-folder-when-using-pycharm-with-git

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -54,6 +54,10 @@ coverage.xml
 *.mo
 *.pot
 
+#pycharm
+this folder contains files which store user-specific settings
+./idea
+
 # Django stuff:
 *.log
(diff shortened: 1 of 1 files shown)


shiftkey:
@xanjay thanks for the contribution!

However we currently use [`Global/JetBrains.gitignore`](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore) to track what files PyCharm and other JetBrains IDEs should include or exclude from version control. Please check it out!

BrianMSheldon:

PR #3072, opened 2019-05-27
added gitignore for .NET Core

nbarkhina:

.NET Core is a popular cross platform framework. This file covers the common build folders and front end scenarios

### DIFF
--- a/DotnetCore.gitignore
+++ b/DotnetCore.gitignore
@@ -1,0 +1,7 @@
+# .NET Core build folders
+/bin
+/obj
+
+# Common node modules locations
+/node_modules
+/wwwroot/node_modules


WeihanLi:
maybe it could be merged with `VisualStudio.gitignore`

nbarkhina:
> maybe it could be merged with `VisualStudio.gitignore`

You could though I think the Visual Studio one is more general.

SanderDeclerck:
I don't think this gitignore file is sufficient. Consider following two scenario's:

- Your project is not in the root folder (eg: you might have multiple projects) - in that case the bin and obj folders are not ignored.
- You open the project (csproj file) in Visual Studio - this opens a .vs folder that does not need to be in a repository.

I believe the `VisualStudio.gitignore` file contains everything that's needed for .NET Core repositories.

shiftkey:
This request has come up a few times, and my current thinking on how to manage these is best summarized in https://github.com/github/gitignore/pull/2989#issuecomment-471509901:

> the last time this came up I wanted to call out that there's a lot of non-.NET technologies inside the `VisualStudio` template, and suggested a different approach [#2816 (comment)](https://github.com/github/gitignore/pull/2816#issuecomment-422432117)
> 
> > What if we went a slightly different approach:
> > 
> > * `DotNet.gitignore` - a new template fresh with whatever current .NET requires (leaner)
> > * `Global/VisualStudio.gitignore` - the current template, moved alongside the other IDEs

PR #3043, opened 2019-04-28
KiCad: add footprint cache

syzzer:
**Reasons for making this change:**

As described in commit c5a2ea1dd8b8f85f7524bbdaa17844f2324d02c5 in the
kicad repository itself, the fp-info-cache file should not be checked
in to version control.

**Links to documentation supporting these rule changes:**

https://git.launchpad.net/kicad/commit/?id=c5a2ea1dd8b8f85f7524bbdaa17844f2324d02c5

PR #3075, opened 2019-05-27
Revert "KiCad: add footprint cache"

bartlettD:
**Reverts github/gitignore#3043**

Recently filed pull request #3013 adds the `fp-info-cache` line in duplicate but in a more sensible location.
Propose we revert #3043 and remove the duplicate line as it's just clutter.


shiftkey:
@bartlettD thanks!

PR #3076, opened 2019-05-28
Update Java.gitignore

2540461056:
I found that there is no ignore for springboot+eclipse, so I added some.

_TODO_

If this is a new template:

### DIFF
--- a/Java.gitignore
+++ b/Java.gitignore
@@ -21,3 +21,12 @@
 
 # virtual machine crash logs, see http://www.java.com/en/download/help/error_hotspot.xml
 hs_err_pid*
+
+#for springboot+gradle
+.settings
+gradlew
+gradlew.bat
+HELP.md
+.classpath
+.project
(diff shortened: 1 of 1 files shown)


drothmaler:
This repository tries to keep things modular to avoid the overhead of managing rules in multiple redundant places... You can just combine the individual rules to fit the needs of you project (personally I usually use [gitignore.io](https://www.gitignore.io) to do that). 
As Spring Boot/Initializr projects can be generated for either Gradle or Maven, and also do not contain any Eclipse/IDE specific files, it would not be good to just throw all of this into `Java.gitignore`.

So there is `Gradle.gitignore` (or `Maven.gitignore`) and `Global/Eclipse.gitignore` (or `Global/Jetbrains.gitignore` / `Global/Netbeans.gitignore`)... The only "missing" rule would be for `HELP.md`, but personally I would not ignore that by default.

PR #3077, opened 2019-05-28
unity 2019 package directory

lizelive:
these packages are auto downloaded if they are missing.

becuse unity 2019 has a package manager

https://docs.unity3d.com/Packages/com.unity.package-manager-ui@1.8/manual/index.html

If this is a new template:

### DIFF
--- a/Unity.gitignore
+++ b/Unity.gitignore
@@ -9,6 +9,7 @@
 /[Bb]uilds/
 /[Ll]ogs/
 /[Mm]emoryCaptures/
+/[Pp]ackages/
 
 # Never ignore Asset meta data
 !/[Aa]ssets/**/*.meta


FritzsHero:
Hello and thank you for your Pull Request,

some things I have to mention here.
The packages should never be excluded from the git system.
If you remove the manifest.json, which stores the informations about the installed packages, the following things will happen:

- Unity will write a new manifest.json including the default packages
- Unity will install all default packages

Yes the missing packages are auto downloaded, so your project is synced for every member in the team, using the same packages. This file will be checked by Unity at startup for the installed packages.
Unity will download and install them, if they are not present in the packages of the project.

Each time someone installs a new package via the Package Manager, this package gets written with it's specific version into the manifest.json. Giving an team the option to work with the same dependencies.

It may be an better option to never exclude packages from git.

shiftkey:
@lizelive have you had a chance to read over the feedback from @FritzsHero here?

It seems like this change isn't suitable in it's current form, and may introduce other problems down the track.

alexanderzimmerman:
I came here because I am setting up my .gitignore for a Unity project.

It seems to me that the only file that Unity 2019 will put into `Packages/` is `Packages/manifest.json`. For me as a new Unity user, the structure is un-intuitive, leading me to think that the package contents will also go into `Packages/` (which they apparently do not). Anyway, I definitely want to version control `Packages/manifest.json` as @FritzsHero says above.

PR #3078, opened 2019-05-29
Update the Unity Gitignore

Falme:
Added ignore line for VSCode users (.vscode cache directory)

If the developer uses a light VSCode to write the source code, the cache folder will be ignored (just like ".vs/" )

### DIFF
--- a/Unity.gitignore
+++ b/Unity.gitignore
@@ -25,6 +25,9 @@
 # Visual Studio cache directory
 .vs/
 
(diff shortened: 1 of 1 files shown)


shiftkey:
The `.vscode/` directory contains things that can be shared between users.

We also maintain a template for it in [`Global/VisualStudioCode.gitignore`](https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore), so please check that out.

PR #3081, opened 2019-05-30
Updating gitignore.visualstudio to ignore proper assets from other VS tools

metoly:

The VS IDE, VS4Mac and the .NET Core CLI team need to use the same .gitignore file.

### DIFF
--- a/KiCad.gitignore
+++ b/KiCad.gitignore
@@ -13,6 +13,9 @@ _autosave-*
 *-save.pro
 *-save.kicad_pcb
 
+# Footprint cache
+fp-info-cache
+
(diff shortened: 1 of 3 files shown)


shiftkey:
@metoly as I mentioned in #3048, I am not willing to take in this change because of the additional maintenance burden that results from combining these templates (almost 250 lines of rules when you exclude whitespace and comments).

Please look into alternatives, such as:

 - managing a fork of this repository with the changes required to support your use case
 - combining the templates as part of the build process for the product/s
 - combining templates when the project is initialized based on the setup chosen by the user

PR #3082, opened 2019-06-05
Added ".idea/" for PyCharm

jieggii:

I added .idea/ folder to Python.gitignore because it contains only personal and sensitive informaiton 

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -1,3 +1,6 @@
+# PyCharm
+.idea/
+
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]


shiftkey:
Thanks, but we have a shared [`Global\JetBrains.gitignore`](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore) for our template for JetBrains IDEs (including PyCharm).

jieggii:
That's cool, but why don't you add it into repository creation menu?
image

shiftkey:
@jieggii that's a separate area to the maintenance of these templates, and you'll have to speak to [GitHub Support](https://github.com/contact) about that feature.

Baneeishaque:
try [gitignore.io](https://www.gitignore.io) @jieggii

BrianMSheldon:

PR #3083, opened 2019-06-07
Swift: ignore .swiftpm

maxxfrazer:

SPM now supported with Xcode 11; .swiftpm is generated by Xcode and is not needed on git

Basing off Apple's swift-package-manager .gitignore
https://github.com/apple/swift-package-manager/blob/master/.gitignore

### DIFF
--- a/Swift.gitignore
+++ b/Swift.gitignore
@@ -38,7 +38,9 @@ playground.xcworkspace
 # Packages/
 # Package.pins
 # Package.resolved
+# *.xcodeproj
 .build/
+.swiftpm
 
 # CocoaPods
 #
--- a/Swift.gitignore
+++ b/Swift.gitignore
@@ -39,8 +39,12 @@ playground.xcworkspace
 # Package.pins
 # Package.resolved
 # *.xcodeproj
+#
+# Xcode automatically generates this directory with a .xcworkspacedata file and xcuserdata
+# hence it is not needed unless you have added a package configuration file to your project
+# .swiftpm
+
 .build/
-.swiftpm
 
 # CocoaPods
 #


Swift.gitignore
```diff
@@ -38,7 +38,9 @@ playground.xcworkspace
 # Packages/
 # Package.pins
 # Package.resolved
+# *.xcodeproj
```
Coeur: Really no need to give the wrong idea here: xcodeproj files are very needed in almost all the iOS/macOS projects of the world. So I wouldn't add this commented line.

Coeur:
@FabrizioBrancati @natestedman Can you review to tell if `.swiftpm` is needed or not?

Swift.gitignore
maxxfrazer: This is specifically for swift packages; you open the Package.swift file with Xcode and all your code is brought in through that. The xcodeproj file can generated using `swift package generate-xcodeproj`.

If you're still not convinced though I'll remove the line.

Coeur: Currently, I'm not used enough to Xcode 11 beta SPM to answer if there are better alternatives. I've ping previous committers.

FabrizioBrancati:
I found some interesting links:
- https://github.com/apple/swift-evolution/blob/master/proposals/0219-package-manager-dependency-mirroring.md
- https://github.com/apple/swift-package-manager/commit/839474f16fe7c88825c41e7c9fe6bce561a3eaea
- https://github.com/apple/swift-package-manager/pull/2200

It seems that adding this file to `.gitignore` is up to the developer or based on the project, but even the Swift Package Manager Apple team said to do not add this folder to `.gititnore`, because it helps the team to use the same dependencies, etc.

Maybe if it's commented by default, we can keep it.

maxxfrazer:
Ok, seems like something a lot of developers might want to ignore, especially if they're creating a package without dependencies - as Ankit has added it to the SPM Package.
But would be best to just have it as a commented line, allowing people to turn it on easily.

I'll move this up a line to group it with the other commented suggestions & comment it out. Do you think leaving a comment message would be necessary or just leave it there?

Based on the [InitPackage.swift](https://github.com/apple/swift-package-manager/blob/master/Sources/Workspace/InitPackage.swift) file I think it would be appropriate to leave `*.xcodeproj` as a commented addition to the file also RE @Coeur's concerns.

FabrizioBrancati:
Yes, leave a comment to explain when and why the developer has to uncomment that line.
Also, I think that leaving the `*.xcodeproj` line is ok, the developers that don't want to commit it will uncomment that line.

maxxfrazer:
@FabrizioBrancati should be all set if the added comment is to satisfaction 🚀

shiftkey:
@maxxfrazer thanks for the contribution, and thanks @Coeur @FabrizioBrancati for the reviews!

Coeur:
@shiftkey I wouldn't mind a merge on another PR that only changes the comments: #3133

shiftkey:
@Coeur no problem - just catching up on notifications during a quiet morning, so I'll prioritize that next

PR #3084, opened 2019-06-12
Update Unity.gitignore

ceceomer:
Visual Studio Code cache folder


shiftkey:
@ceceomer thanks for the contribution!

This directory is not quite a cache, and our [`Global/VisualStudioCode.gitignore`](https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore) template illustrates what can be committed and shared.

PR #3086, opened 2019-06-14
Don't ignore *.pot files.

idgserpro:

Although dynamically generated from i18n tools, in some situations (specially in the Plone world) you need to add a manual.pot with manual entries that need to be versioned.

https://github.com/plone/documentation/blob/0ea1b1f715e9031a434d72f57c189f25aa1f4b73/develop/plone/i18n/internationalisation.rst#manual-po-entries

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -52,7 +52,6 @@ coverage.xml
 
 # Translations
 *.mo
-*.pot
 
 # Django stuff:
 *.log


idgserpro:
@shiftkey can you take a look here please?

Python.gitignore
```diff
@@ -52,7 +52,6 @@ coverage.xml
 
 # Translations
 *.mo
-*.pot
```
shiftkey, line 55: This rule has been largely unchanged for 7+ years, so I have several concerns about dropping this:

 - Most `.pot` files are generated, right? Would this be noisy to now have all `.pot` files in version control?
 - For situations where include/exclude files from version control is situational, we like to have a comment with context about when and why you should care about this rule. Here's an example:

https://github.com/github/gitignore/blob/218a941be92679ce67d0484547e3e142b2f5f6f0/Rails.gitignore#L24-L25

Can we do that for this case, to explain for Plone usage what they should be aware of?
 
 - if we're trying to exclude a specific file by convention, could we keep this rule but add a `!manual.pot` or `!lib/manual.pot` rule to opt-in for this specific file?

shiftkey:
@idgserpro the other concern I have about this change is interest from the community - I'm not familiar with Plone so I'm not sure how many people out there agree with this change.

idgserpro:
Your concerns are valid. Actually, I don't know the motivation to version control *.pot files, but this has been for ages in Plone world.

@mauritsvanrees would you have more information about this?

mauritsvanrees:
I am one of the Plone Release managers. I am indeed used to having the `.pot` files in version control. I encountered a Plone project at a customer recently where the `.pot` file was ignored, and it surprised me, wondering why I *did* see changes in the `.po` files, but not in the `.pot` file.

Some reasons for versioning the `.pot` files:

- Yes, you can regenerate the `.pot` file, but if someone uses a different tool for this, or a different version, I would like to see this difference in version control.
- I find it useful to be able to update only the central `.pot` file to check what influence my code changes have there.
- It makes it easy to copy the `.pot` file to a `.po` file for adding a new language.
- I am simply used to it.

Something like `!*manual.pot` may work. 
I am not using this `gitignore` project though.

PR #3089, opened 2019-06-20
starting section for editors and IDEs

MartinTrat:

contributing to repo with ideas from daily struggle; first public contribution / acquiring productive pull request experience

n/a

If this is a new template:

n/a

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -105,6 +105,9 @@ ENV/
 env.bak/
 venv.bak/
 
+# IDE / Editor
+.vscode/
+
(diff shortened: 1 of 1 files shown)


shiftkey:
@MartinTrat thanks for the contribution!

However we have a separate template for VSCode at [`Global/VisualStudioCode.gitignore`](https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore) that outlines what can be shared in version control. We recommend adding that to your projects.

PR #3093, opened 2019-06-22
Update Wordpress.gitignore

samuraijane:
This file is missing `/*` to begin with ignoring everything. I have also added a line to ignore Mac's `.DS_Store`, several of which appear within the directories that belong to `wp-content`.

_TODO_

_TODO_

If this is a new template:

### DIFF
--- a/WordPress.gitignore
+++ b/WordPress.gitignore
@@ -1,5 +1,9 @@
 # ignore everything in the root except the "wp-content" directory.
+/*
 !wp-content/
+
+# ignore Mac OS file that stores custom attributes for its containing folder
+.DS_Store
 
 # ignore everything in the "wp-content" directory, except:
 # "mu-plugins", "plugins", "themes" directory


WordPress.gitignore
```diff
@@ -1,6 +1,10 @@
 # ignore everything in the root except the "wp-content" directory.
+/*
 !wp-content/
 
+# ignore Mac OS file that stores custom attributes for its containing folder
+.DS_Store
```
shiftkey: These rules are covered by [`Global/macOS.gitignore`](https://github.com/github/gitignore/blob/master/Global/macOS.gitignore).

Please omit this change as it is not specific to Wordpress.

WordPress.gitignore
```diff
@@ -1,6 +1,10 @@
 # ignore everything in the root except the "wp-content" directory.
+/*
```
shiftkey: > This file is missing `/*` to begin with ignoring everything. 

Are you able to share some context about why this is a safe change to make?

stkrzysiak: Without /* the negated lines(like !wp-content) are irrelevant. This whole file is built on the concept that only the negated lines are to be included, for that to work everything must be excluded initially.

WordPress.gitignore
NerdsvilleCEO: Agreed, also !.gitignore should be added so we can add the .gitignore to the repo :P

WordPress.gitignore
Triloworld: Yep, @stkrzysiak is right. Right now part with "!wp-content/" don't work ;/ 
We need that "/*" . Otherwise files of core wordpress are now added to change

WordPress.gitignore
pedrosanta: +1 on both. I think we should have clear separation of concerns on these files—hence I have some serious doubts if, for example, `*.sql` files exclusion should be here in the first place if it's not something that occurs/is generated naturally on a typical WP install. (Want to bring everyone to attention that own project/personal rules, should probably be applied by oneself individually on each project `.gitignore` alongside these general contrib rules, to avoid the risk of possible workflow enforcement onto others.)

Having each file handle the specific files of each library/component/platform, allows for better combination of several of these files, but also for utils like [gitignore.io](http://gitignore.io).

WordPress.gitignore
pedrosanta: I want to pick up on @shiftkey comment, namely about the _safety_ aspect of that specific `/*` rule—I assume that was something along this @shiftkey was referring to.

Wasn't better (and safer, more below) if instead of a catch-all approach like that, we should opt for a more **declarative** way of stating the folders we want to see ignored instead?

I'm concerned that using the 'catch-all' approach we might raise some conflict with other .gitignores, or hinder if one wants to mix-and-match several of these .gitignore files like on [gitignore.io](http://gitignore.io) and some technology requires some folder on the root to be added that would be caught by this rule, etc.

Does this makes sense?

pedrosanta:
Also, should we close this one in favor of #3282?

WordPress.gitignore
pedrosanta: FYI, added the #3295 PR that addresses my concern above.

WordPress.gitignore
Triloworld: ".DS_Store" line will be removed from this push as @shiftkey and @NerdsvilleCEO is right

WordPress.gitignore
Triloworld: I will redo this to change to previous to enable this pull request to be merge and close some issues. 
This need to be fixed in #3295 with another pull request after all paths are declarative and battle tested for errors
Thank you @pedrosanta for helping understand better context

Triloworld: Declarative way is better imho. Need to be discuss on approach and tests

pedrosanta: Yes #3295 has a declarative approach, I based it mostly on folders from a fresh WP install, and tested it, but a couple common folders on WP ecosystem might or nor be missing, but then again I have the strong feeling we should only add into **.gitignore** what is generally assumed as widely as possible as a common practice/setup on WP ecosystem. Because of course YMMV and it's supposed for you to branch of the .gitignore, and then add rules for your particular setup.

335121 (approved):

bdougie:
> Also, should we close this one in favor of #3282 or #3295?

PR #3095, opened 2019-06-23
Create full.gitignore

Backup-eric645:
I want FULL+ GitIgnore template to be present.


shiftkey:
@Backup-eric645 thanks for the contribution!

Unfortunately I'm not interested in supporting this template currently for a couple of reasons:

 - it's not clear to me how valuable this is
 - there's a significant maintenance overhead to keep this up to date every time a different file is updated

PR #3097, opened 2019-06-26
Ignoring AppCode local configs

3pehrbehroozi:
Added .idea/ to Objective-C Xcode and Swift for ignoring AppCode local config files

_TODO_

_TODO_

If this is a new template:

### DIFF
--- a/Global/Xcode.gitignore
+++ b/Global/Xcode.gitignore
@@ -5,6 +5,9 @@
 ## User settings
 xcuserdata/
 
+## AppCode local configs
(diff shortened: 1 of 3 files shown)


shiftkey:
@3pehrbehroozi thanks for the contribution!

I'm going to close this out because we have a [`Global/JetBrains.gitignore`](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore) that covers AppCode, and we recommend adding that to your projects when you need it.

PR #3098, opened 2019-06-27
Update Laravel.gitignore

alkhachatryan:
The most of Laravel developers are using dedicated/VPS servers. So it's difficult to have a Laravel project with PUBLIC path there, as apache entry point is PUBLIC_HTML.

Added PUBLIC_HTML/ some folders in gitignore


shiftkey:
@alkhachatryan thanks!

PR #3103, opened 2019-07-04
Update Python.gitignore

1lin24:
Ignore temporary file generated by vi/vim


shiftkey:
Thanks, but these rules are already part of our [`Global/Vim.gitignore`](https://github.com/github/gitignore/blob/master/Global/Vim.gitignore) template.

We want to avoid duplicating rules wherever possible as it adds a maintenance burden to the project to keep things in sync.

PR #3104, opened 2019-07-04
Added more glossaries auxiliary files and makeindex logs to TeX template

vkusvody:

Discovered more auxiliary files that don't seem to need tracking: 'lzo' and 'lzs' for glossaries with glossary type=leipzig, and 'lpz' is generated in between by makeindex.


shiftkey:
@vkusvody thanks!

PR #3106, opened 2019-07-05
Create SSRS.gitignore (#3106)

gonesoft:
Added SQL Server Reporting Services (SSRS) gitignore for those who use git to versioning one or more projects of SSRS. Proyects like this autogenerate files that no need to be stage.

This is a new template.

### DIFF
--- a/SSRS.gitignore
+++ b/SSRS.gitignore
@@ -1,0 +1,18 @@
+# gitignore template for SQL Server Reporting Services (SSRS)
+#
+# Recommended: SSRS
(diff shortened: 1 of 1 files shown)


paulsmithkc:
Missing *.rsuser

paulsmithkc:
*.sln and *.rptproj should not be ignored

gonesoft:
> *.sln and *.rptproj should not be ignored
That's correct. Should not be ignored.

PR #3107, opened 2019-07-08
Update Unity.gitignore

jvin042:
Added "vscode" folder for ide Visual Studio Code

Many people use Visual Studio Code to develop on Unity3D

### DIFF
--- a/Unity.gitignore
+++ b/Unity.gitignore
@@ -24,6 +24,7 @@
 
 # Visual Studio cache directory
 .vs/
+.vscode/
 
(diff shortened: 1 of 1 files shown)


shiftkey:
@jvin042 thanks for the contribution!

I'm going to close out this PR as we have an existing [`Global/VisualStudioCode.gitignore`](https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore) for our VSCode template which shows what can be committed and shared.

PR #3108, opened 2019-07-08
Ignore the .ionide working folder (VisualStudio.gitignore)

steveh-ivendi:

If you have the [Ionide](http://ionide.io/) tools installed you will get an `.ionide` directory created in each directory that you open with VS Code, regardless of whether or not you are using F#.

Ignored at line 75 of the `ionide-vscode-fsharp` .gitignore: https://github.com/ionide/ionide-vscode-fsharp/blob/cc4fdbf90a33a91b305f74c36ec3634cb5744e95/.gitignore#L75


shiftkey:
@steveh-ivendi thanks!

PR #3116, opened 2019-07-13
feat: Add microbundle cache directories

Richienb:

[Microbundle](https://github.com/developit/microbundle) creates cache directories.

[Official microbundle gitignore](https://github.com/developit/microbundle/blob/master/.gitignore#L4-L7)

### DIFF
--- a/Node.gitignore
+++ b/Node.gitignore
@@ -53,6 +53,12 @@ typings/
 # Optional eslint cache
 .eslintcache
 
+# Microbundle cache
+.rpt2_cache/
+.rts2_cache_cjs/
+.rts2_cache_es/
+.rts2_cache_umd/
+
 # Optional REPL history
 .node_repl_history
 


Richienb:
Any update on this?

shiftkey:
@Richienb is this actually required?

The only mention of the cache I see is in [here](https://github.com/developit/microbundle/blob/b170137f8df4eae74beeab22289ea576cdbb144a/src/index.js#L537) which is a path inside `node_modules` (so it should be handled by the previous `node_modules` rule):

```js
cacheRoot: `./node_modules/.cache/.rts2_cache_${format}`
```

The other usage is part of [cleanup inside a test harness](https://github.com/developit/microbundle/blob/8142733c753e8b497337c30156f2e2846c2c6ece/tools/build-fixture.js#L52-L54):

```js
	await rimraf(resolve(`${fixturePath}/.rts2_cache_cjs`));
	await rimraf(resolve(`${fixturePath}/.rts2_cache_es`));
	await rimraf(resolve(`${fixturePath}/.rts2_cache_umd`));
```

The `git blame` for the linked `.gitignore` doesn't really explain why these were added, so I'm not clear on the benefits here of taking in this change...

Richienb:
@shiftkey That doesn't seem right. Here's a screenshot of a typical repository:

Typical Repository

PR #3118, opened 2019-07-13
.travis.yml: Trim trailing whitespace and indent

jayvdb:

https://github.com/moremoban/pypi-mobans uses this repo as a git submodule, and want to run our yaml lint tools without excluding this repo.  We dont mind if it breaks occasionally; we'll come help fix, like we have in the past https://github.com/github/gitignore/pull/2728.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ sudo: false
 
 branches:
   only:
-  - master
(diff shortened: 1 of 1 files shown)


shiftkey:
@jayvdb thanks!

PR #3120, opened 2019-07-16
Add *.hie to Haskell.gitignore

openjck:
This new file type can be generated by GHC 8.8.

https://www.haskell.org/ghc/blog/20190626-HIEFiles.html


shiftkey:
@openjck thanks!

PR #3123, opened 2019-07-17
Update Android.gitignore

tr3v0r5:

_TODO_

_TODO_

If this is a new template:

### DIFF
--- a/Android.gitignore
+++ b/Android.gitignore
@@ -17,6 +17,7 @@ release/
 
 # Gradle files
 .gradle/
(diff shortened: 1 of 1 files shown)


shiftkey:
@tr3v0r5 thanks for the contribution, but this clashes with our [`Global\JetBrains.gitignore`](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore) template.

I recommend using that in your project as it's more detailed.

PR #3124, opened 2019-07-18
Added VS Code support to UE4's gitignore

Maldonacho:
Simple, safe change.
VS Code creates some files just like its big brother. These files can be safely ignored as they are user specific and can be generated by the engine. It's beneficial to ignore these files as they contain lots of absolute directory paths that will likely be different across computers, meaning that whenever someone loads up the project, VS code will "fix" them, causing a bunch of unnecessary changes.

I also changed "VS 2015" to just "VS" since it applies to all versions of VS supported by UE4.

 - Increase feature parity / consistency between VS and VS Code
 - Save others from having to make this change every time

### DIFF
--- a/UnrealEngine.gitignore
+++ b/UnrealEngine.gitignore
@@ -1,5 +1,6 @@
-# Visual Studio 2015 user specific files
+# Visual Studio & VS Code user specific files
 .vs/
+.vscode/
 
 # Compiled Object files
 *.slo
@@ -35,6 +36,7 @@
 *.xcodeproj
 *.xcworkspace
 *.sln
+*.code-workspace
 *.suo
 *.opensdf
 *.sdf


UnrealEngine.gitignore
```diff
@@ -1,5 +1,6 @@
-# Visual Studio 2015 user specific files
+# Visual Studio & VS Code user specific files
 .vs/
+.vscode/
```
shiftkey: I don't agree with this change, as the `.vscode` directory inside a repository will contain workspace settings - specific to the project - and I believe these can be shared with users. Our [`Global/VisualStudioCode.gitignore`](https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore) shows what can be committed and shared currently.

This is different to the `.vs` directory, which is used as a cache.

Maldonacho: Hey, thanks for taking the time to review this!
I checked your claim and tried sending these files to another computer. The project would not build at all, UBT (Unreal Build Tool) would log a bunch of errors about invalid paths & missing files. To fix the problem, I just needed to regenerate these files using UE4. I did not make any other change. GitHub Desktop recognized 5 changed files, all of which would be excluded by my changes above.
Committing these changes and opening them back on my main PC would cause the same issue there - the project would not build until I regenerated the files, now it's in the same state as before. An endless loop.

I also tried making some workspace specific changes (I disabled a couple of plugins for this workspace). No files were changed.
Even if some settings won't be shared, I think the disadvantage of having to regenerate project files every time (not to mention having +4 changed files with every commit) greatly outweighs the benefit of having shared settings.
UE4 generates these files for you, and I've never needed to make a change to them.

shiftkey: I can't really follow what you're trying to say here, and I'm not familiar with the specifics of the UE toolchain.

I'm fine with leaving the `.vs` rule as-is, but it's not clear to me what's in your `.vscode` directory here that would cause issues in UE - can you share more information about that?

Maldonacho: What I'm trying to say is that UE4 generates these files for the user, and they contain a lot of absolute directory paths to the installation directory of UE4, the location of the project itself, and some external tools like the Windows 10 API.

Trying to use these files on another machine that doesn't have the exact same setup will not work - they are only useful on the machine that created them, hence why I'm suggesting they should be ignored by git.

Furthermore, another user of the workspace may not even choose to use VS Code, so those files would be even more useless to them.

I have validated all of these claims myself, I can even send you an example .vscode folder of one of my unreal projects so you can see for yourself if you wish.
I can assure you that it would benefit us unreal devs to have these files ignored.

bdougie:
I stand by what was mentioned already. We have an existing [Global\VisualStudioCode.gitignore](https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore) that covers more than just this rule.

PR #3125, opened 2019-07-19
Ignore configuration file

chougron:
Database credentials and security keys should not be commited.

The wp-config.php file should not be commited, otherwise sensitive credentials would be commited too.

https://wordpress.org/support/article/editing-wp-config-php/


pedrosanta:
Should we close this one in favor of #3282?

Even with this change, it continues to allow the complete WP install to be added/staged since #2972 merge.

PR #3128, opened 2019-07-22
Scrivener: Added new checksums and allows for relative directories

jzucker2:

Found that when I had the Scrivener file in the same root directory as my README, I needed relative paths to the gitignore files

Also noticed that Scrivener 3 has slightly different paths then Scrivener 2

If this is a new template:


shiftkey:
@jzucker2 thanks!

PR #3130, opened 2019-07-23
ignore DS_Store file

haydarKarkin:

The .DS_Store file is a file created by OS that holds meta-information about a directory. You can expect to always find it and can safely ignore it. The file is not shown by the Finder.

### DIFF
--- a/Swift.gitignore
+++ b/Swift.gitignore
@@ -80,3 +80,12 @@ fastlane/test_output
 # https://github.com/johnno1962/injectionforxcode
 
 iOSInjectionProject/
+
+# DS_Store Files
+#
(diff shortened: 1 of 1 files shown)


Coeur:
I disagree with this change: this `.DS_Store` should be in your `.gitignore_global` file on any macOS system: it's totally unrelated to Swift or other programming languages.

shiftkey:
To add to @Coeur's comment - we also have a [`Global/macOS.gitignore`](https://github.com/github/gitignore/blob/master/Global/macOS.gitignore) template that contains this rule.

haydarKarkin:
@shiftkey you are absolutely right. I couldn't realize global one. I close the PR! Thanks.

PR #3131, opened 2019-07-23
[Python] Ignore coverage annotations reports

thedrow:

There's no need to include these reports in the repository

**Links to documentation supporting these rule changes:

**https://coverage.readthedocs.io/en/coverage-4.2/cmd.html#text-annotation

If this is a new template:

N/A

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -47,6 +47,7 @@ htmlcov/
 nosetests.xml
 coverage.xml
 *.cover
+*.py,cover
 .hypothesis/
 .pytest_cache/
 


Python.gitignore
```diff
@@ -47,6 +47,7 @@ htmlcov/
 nosetests.xml
 coverage.xml
 *.cover
+*.py,cover
```
shiftkey: @thedrow is this the right extension? The `,` here seems like it'd be a huge problem on Windows, and I can't seem to find a mention of this in the documentation you linked to.

thedrow: That's the extension `coverage` generates.
Try it yourself.

Here's the code that says so BTW: https://github.com/nedbat/coveragepy/blob/502ecfc1767480ceefd0745d258a945163b61024/coverage/annotate.py#L21

shiftkey: TIL: this actually works

shiftkey:
@thedrow thanks!

PR #3132, opened 2019-07-23
Partial revert of 837181e2f60198de3701a1f87e66aaf8e69d0e07

Coeur:

Previous commit on this file (#2566) was incorrectly supported by a 10-years old Stack Overflow answer. There aren't any demonstrated extra files (apart from `xcuserdata/` already listed on top) that need to be excluded from the `.xcodeproj/` folder, and as such this folder shouldn't be added to .gitignore.

I'm leaving the change on `/*.gcno`, but updating the description about it. We could note that most Xcode users are using Clang (default compiler) instead of Gcc nowadays.

### DIFF
--- a/Global/Xcode.gitignore
+++ b/Global/Xcode.gitignore
@@ -22,9 +22,5 @@ DerivedData/
 *.perspectivev3
 !default.perspectivev3
 
-## Xcode Patch
-*.xcodeproj/*
-!*.xcodeproj/project.pbxproj
-!*.xcodeproj/xcshareddata/
-!*.xcworkspace/contents.xcworkspacedata
+## Gcc Patch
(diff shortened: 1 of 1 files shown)


shiftkey:
@Coeur thanks!

Coeur:
That was fast.

shiftkey:
> That was fast.

You caught me at the start of my work day 😉

Coeur:
For the record, I'd also like to share [my Stack Overflow answer](https://stackoverflow.com/a/53039267/1033581) from a few months ago:
>Regarding **Xcode 9 and newer**, the only line you really need in your .gitignore is:
>
>     xcuserdata/
>
>Nothing else for Xcode 9+.

PR #2410, opened 2017-06-08
clarifying when ignoring is not required

Coeur:
**Reasons for making this change:**

clarifying when ignoring is not required anymore

**Links to documentation supporting these rule changes:** 

* https://www.gravitywell.co.uk/latest/mobile/posts/useful-.gitignore-file-for-xcode-explained/
* https://developer.apple.com/library/content/documentation/Xcode/Conceptual/RN-Xcode-Archive/Chapters/xc4_release_notes.html#//apple_ref/doc/uid/TP40016994-CH3-SW2

PR #3133, opened 2019-07-23
Improving the documentation of Objective-C and Swift files

Coeur:
This doesn't change anything at all to the rules themselves (except for adding an explicit `/` to `Carthage/Build/`).

This change simply applies the re-ordering and documentation improvement that I already made in 2017 on the `Xcode.gitignore` file: #2410.


shiftkey:
@Coeur thanks!

PR #3134, opened 2019-07-24
Revert of #2988. TextMesh Pro files are not auto managed

heatblayze:
Unlike other packages, TextMesh Pro is not entirely managed by Unity's new Package Manager, and still imports some assets directly into the project. Removing these files can cause issues when using Unity's Cloub Build service (which can utilise git to build your project).

The [Package Manager guide for TextMesh Pro](https://docs.unity3d.com/Packages/com.unity.textmeshpro@2.0/manual/index.html) also states that these files are required.

You can also find the conversation on the original PR #2988


shiftkey:
@heatblayze thanks!

PR #3137, opened 2019-07-27
[Swift] Added DS_Store and GoogleService-Info.plist

WikipediaBrown:
The DS_Store file isn't necessary to the purpose of most repositories so I remove it. Also, best practices would be to remove sensitive information from source control. Google Service Accounts are often used by iOS engineers as well as Vapor and Kitura engineers building on GCP/Firebase so I remove them from source control as a rule too.

I wanted to be able to use the swift.gitignore file from GitHub out of the box. I have to make these changes every time.

Wait what?

If this is a new template:

### DIFF
--- a/Swift.gitignore
+++ b/Swift.gitignore
@@ -80,3 +80,13 @@ fastlane/test_output
 # https://github.com/johnno1962/injectionforxcode
 
 iOSInjectionProject/
+
+# .DS_Store
+#
+# This file records the visual position of the images of the files in the finder window. You don't need this.
+.DS_Store
+
+# GoogleService-Info.plist
+#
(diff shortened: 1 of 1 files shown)


Coeur:
See the global macOS gitignore for `DS_Store`.

As for the plist:
- it's not dependent of a programming language (it would be the same filename for Objective-C)
- Hiding API keys to developers is a concern that could apply for a **thousand of different third-parties, not just Firebase**: a tech architect would eventually put all the sensitive data in a folder of its own.

Coeur:
@shiftkey no reply from the author, so I suggest that we close the pull request now

PR #3139, opened 2019-07-31
Add .idea/ to gitignore on new section Integrated Development Environ…

neosergio:
…ments settings to ignore Pycharm settings files

A new section for IDE settings would be useful

If this is a new template:

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -122,3 +122,6 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+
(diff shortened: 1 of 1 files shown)


shiftkey:
@neosergio thanks for the contribution, but we have an existing [`Global/JetBrains.gitignore`](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore) that should cover PyCharm and the other JetBrains IDEs.

BrianMSheldon:

PR #1415, opened 2015-02-20
Update VisualStudio.gitignore

wisitsaks:
ignore all files in Log/log folder

PR #3143, opened 2019-08-06
[VisualStudio] add Logs folder

paulsmithkc:

* The singular form of the Log folder is included in the ignore file, but not the plural form.

https://stackoverflow.com/questions/9745119/git-ignore-file-still-commits-logs

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -32,6 +32,7 @@ bld/
 [Bb]in/
 [Oo]bj/
 [Ll]og/
+[Ll]ogs/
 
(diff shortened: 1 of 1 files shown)


shiftkey:
@paulsmithkc can we document which tool is generating the `Logs` folder? We did this for `Log` back in #1415 and I want to confirm this is related to an external tool, and isn't something that users have configured themselves...

paulsmithkc:
When using nlog, both "Log" and "Logs" (plus lowercase versions) depending on which tutorial/textbook you follow.

shiftkey:
@paulsmithkc thanks!

PR #3144, opened 2019-08-06
Ignore gummi previews

qogn9daohs1:

The LaTeX editor [gummi](https://github.com/alexandervdm/gummi) creates hidden files during compilation. This rule ignores these temporary files.

[Documentation in gummi source code](https://github.com/alexandervdm/gummi/blob/master/src/editor.c#L166)


shiftkey:
@qogn9daohs1 thanks!

PR #3145, opened 2019-08-11
add pycharm project setting dir to Python.gitignore

ygjia:
add pycharm project setting dir ".idea/" to the Python.gitignore.


BrianMSheldon:

PR #3152, opened 2019-08-19
Update Qt.gitignore

firelightlong:
Add "Qt VS Tools" generate files to the Qt.gitignore

_TODO_

_TODO_

If this is a new template:


shiftkey:
@firelightlong thanks for the contribution, but I can't take in this change because of how it duplicates rules currently maintained in [`VisualStudio.gitignore`](https://github.com/github/gitignore/blob/master/VisualStudio.gitignore).

PR #3156, opened 2019-08-21
modify swift.gitignore to have .DS_Store

shahianshu:

In mac , .DS_Store file is a hidden files, which is added to the github repo whenever we try to upload our projects on github. We can remove .DS_Store file using the methods expressed in this [link](https://stackoverflow.com/questions/107701/how-can-i-remove-ds-store-files-from-a-git-repository) . But it will be very good and hassel free if we already have .DS_Store file present in .gitignore template of swift.

### DIFF
--- a/Swift.gitignore
+++ b/Swift.gitignore
@@ -5,6 +5,8 @@
 ## User settings
 xcuserdata/
 
+.DS_Store
+
 ## compatibility with Xcode 8 and earlier (ignoring not required starting Xcode 9)
 *.xcscmblueprint
 *.xccheckout


shahianshu:
can anyone please tell me when my pull request will be accepted or declined ? I am new to open source so anyone can please tell me where to contact the repo admin.

aruanoguate:
Hi @shahianshu, 

Regards,

Alvaro

shahianshu:
Hi @aruanoguate 

I was not aware of this. So , thank you for telling me.

Regards, 
Anshu Shahi

PR #3158, opened 2019-08-22
Ignore ide settings

samuelterra22:
Just ignore *.ide/* folder.


shiftkey:
@samuelterra22 thanks for the contribution, but please check out [`Global/JetBrains.gitignore`](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore) which covers the setup for all JetBrains editors.

PR #3159, opened 2019-08-22
[Qt]Add generated translation binaries to the .gitignore

martonmiklos:
The *.qm binary translation files are generated from the translation *.ts files and should not be tracked by git since they are generated.

qm files are autogenerated (on each build if the translation sources are added to the .pro file)

https://wiki.qt.io/Automating_generation_of_qm_files

If this is a new template:

N/A


shiftkey:
@martonmiklos thanks!

PR #3163, opened 2019-08-25
Update Dart.gitignore

Purukitto:
## Changes in reference to Dart v2

- With Dart 2 all relevant js files are built inside /build directory which are already handled so there is no need to add .js files in the .gitignore
- Added files that are generated by other development environments

## References:
- *Removal of .js*
dart-lang/site-www#250
>I think we can remove Avoid committing generated JavaScript files: – all of our recommended flows have folks using our build system – which puts everything in the build directory. And that's already handled.

Addressed in dart-lang/site-www#1850

- *Addition of files by other development environments*
https://dart.dev/guides/libraries/private-files


kevmoo:
`.DS_Store` really bothers me. This should be in every mac developers `~/.gitignore_global`

Dart.gitignore
```diff
@@ -11,11 +11,9 @@ pubspec.lock
...
 
-# Avoid committing generated Javascript files:
-*.dart.js
-*.info.json      # Produced by the --dump-info flag.
-*.js             # When generated by dart2js. Don't specify *.js if your
-                 # project includes source files written in JavaScript.
-*.js_
-*.js.deps
-*.js.map
+# Directories and files created by other development environments
```
shiftkey: These rules below are covered by existing templates:

 - `Global/macOS.gitignore`

I'm 👍 to the cleanup here, but I need to prevent duplicating rules covered elsewhere because they're a maintenance burden I can't support currently.

Purukitto: The above `gitignore`s might not be relevant to someone developing a flutter app and if I'm developing on a windows device then having to add the MacOS `gitignore` seems like a very unnecessary nuance.

And I believe it can be a good way to just allow the developer to add the dart ignore and be able to manage the same

kevmoo: TIL https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore – Nice!

We should train devs to learn to global ignore their OS- and tool-specific bits IMHO.

There is disagreement here, though!

Purukitto: Let me if I should just remove the same or they would be fine in the setup

shiftkey: > Let me if I should just remove the same or they would be fine in the setup

Yes please. The other templates will get rule changes that are not synchronized here, so for the moment this is my current recommendation.

PR #3164, opened 2019-08-27
Exclude VS Code cache directory in Unity.gitignore

mjrousos:
Added .vscode/ as a directory that may need to be ignored depending on how C# files are edited.

Aligning the unity.gitignore file with other .NET .gitignore files like [dotnet/corefx/.gitignore](https://github.com/dotnet/corefx/blob/master/.gitignore).

### DIFF
--- a/Unity.gitignore
+++ b/Unity.gitignore
@@ -22,6 +22,9 @@
 # Visual Studio cache directory
 .vs/
 
+# Visual Studio Code cache directory
+.vscode/
+
 # Gradle cache directory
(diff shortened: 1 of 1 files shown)


shiftkey:
@mjrousos thanks for the contribution, however the `.vscode` directory can be version controlled. 

We have an existing template at [`Global/VisualStudioCode.gitignore`](https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore) that you can see in action.

PR #3165, opened 2019-08-28
add maven into gitignore

HGuedes:
add maven into gitignore

_TODO_

_TODO_

If this is a new template:


shiftkey:
@HGuedes thanks for the contribution, but we already have a template for Maven at [`Maven.gitignore`](https://github.com/github/gitignore/blob/master/Maven.gitignore).

PR #3166, opened 2019-08-29
Ignore macOS resource files

Kucashu:

Ignore .DS_Store. If using Visual Studio for Mac, this file will exist every folder, so ignore it!

https://github.com/lijiejie/ds_store_exp

If this is a new template:

 - **Link to application or project’s homepage**: https://www.apple.com/macos/

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -350,3 +350,6 @@ MigrationBackup/
 
 # Ionide (cross platform F# VS Code tools) working folder
 .ionide/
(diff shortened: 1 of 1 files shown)


shiftkey:
@Kucashu thanks for the contribution, but we currently have a [`Global/macOS.gitignore`](https://github.com/github/gitignore/blob/master/Global/macOS.gitignore) template that covers this and many other rules - I recommend pulling that into your projects!

PR #3167, opened 2019-08-29
Upgrate to rails 6.0.0

schweigert:

Rails 6 generates multiple databases for each test thread.

https://blog.bigbinary.com/2019/04/29/rails-6-adds-parallel-testing.html


shiftkey:
@schweigert thanks for the contribution!

PR #3168, opened 2019-08-29
Add OpenSSL-related file extensions

wesley-dean-flexion:
There are a number of OpenSSL-related file extensions (e.g., .pem, .crt,
etc..) that contain data that are generally best not committed to
repositories.  This file contains several common file extensions that
often correlate to these types of files.

Often times keys, certificates, etc., when they're committed, it's on accident.  This file excludes several common extensions used by OpenSSL, a popular tool for working with SSL / TLS certificates and such.  As I couldn't find resources for excluding these files, I'm submitting a new .gitignore file.

https://datatypes.net/openssl-file-types

### DIFF
--- a/community/OpenSSL.gitignore
+++ b/community/OpenSSL.gitignore
@@ -1,0 +1,34 @@
+# OpenSSL-related files best not committed
+
+## Certificate Authority
+*.ca
+
+## Certificate
+*.crt
+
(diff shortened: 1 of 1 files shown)


shiftkey:
@wesley-dean-flexion thanks!

PR #3169, opened 2019-08-31
Update Vim.gitignore

mazunki:

.svg files are more often than not vector files. You need to exceed 36 swap files for a .svg file to come from Vim.

Vim source code regarding naming of swap files: https://github.com/vim/vim/blob/8c6173c7d3431dd8bc2b6ffc076ef49512a7e175/src/memline.c#L5076


shiftkey:
@mazunki great find - thanks for the fix!

PR #3172, opened 2019-09-05
Updated Nim.gitignore

Akito13:
* added further cache directory
* added auto-generated documentation location

Updates to the Nim language.

Create a Nim project and see for yourself. It is basic caching and auto generation. Nothing special or noteworthy.

### DIFF
--- a/Nim.gitignore
+++ b/Nim.gitignore
@@ -1,1 +1,3 @@
 nimcache/
+nimblecache/
+htmldocs/


shiftkey:
@Akito13 any links to StackOverflow discussions or documentation about this change?

Akito13:
@shiftkey https://nim-lang.org/docs/docgen.html

I don't see a mention of nimblecache, probably because it is just a cache folder i. e. nothing noteworthy.

shiftkey:
@Akito13 thanks!

PR #3175, opened 2019-09-11
Add Rider files to. gitignore

KrzysztofDabek:
It could be a good idea to add project specific files from JetBrains Rider to .gitignore as this IDE is currently supported in Unity for Windows, Linux and iOS

### DIFF
--- a/Unity.gitignore
+++ b/Unity.gitignore
@@ -22,6 +22,9 @@
 # Visual Studio cache directory
 .vs/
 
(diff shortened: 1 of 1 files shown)


shiftkey:
@KrzysztofDabek thanks for the contribution, but we have a [`Global/JetBrains.gitignore`](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore) currently which I recommend using.

PR #3183, opened 2019-09-17
Update Python.gitignore with .DSStore for macOS

gatto:

Ignore .DS_Store. This file will exist in every folder if working on a Mac. Let's put it in the python.gitignore file so that we can use this one .gitignore file on a Mac.

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -122,3 +122,6 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+
+# macOS exclusions
+.DSStore


shiftkey:
@gatto thanks for the contribution, but we have an existing [`Global/macOS.gitignore`](https://github.com/github/gitignore/blob/master/Global/macOS.gitignore) template that contains this and many other macOS-specific rules.

We recommend incorporating that into your projects.

gatto:
Thanks @shiftkey, perhaps I tried to solve my issue in a way that wasn't appropriate. I created an issue https://github.com/desktop/desktop/issues/8284#issue-494649823 that is hopefully in a better spot to solve the problem I see.

PR #3188, opened 2019-09-25
Update VisualStudio.gitignore

jekuer:
Added Visual Studio Code workspace files to the user-specific files list.

One might include this somewhere in the git folder while working with Visual Studio Code, but it should never be included with git. It is a core file type, so should be included here.

https://code.visualstudio.com/docs/editor/multi-root-workspaces#_workspace-file

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -9,6 +9,7 @@
 *.user
 *.userosscache
 *.sln.docstates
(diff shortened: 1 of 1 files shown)


shiftkey:
@jekuer I think this is a better fit for `Global/VisualStudioCode.gitignore` - would you mind opening a PR up against that file instead?

PR #3189, opened 2019-09-27
Update Android.gitignore

gavinliu:
Fix Gradle buildSrc ignored

The custom local Gradle plugin will use the buildSrc directory, and `build/` will exclude the `buildSrc/src/**` directory and change it to `/build`.

If this is a new template:

### DIFF
--- a/Android.gitignore
+++ b/Android.gitignore
@@ -18,7 +18,7 @@ out/
 
 # Gradle files
 .gradle/
-build/
+/build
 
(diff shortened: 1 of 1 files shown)


friederbluemle:
With this change, it will no longer ignore `app/build/` (or any other build output directories in other subprojects). If you use a custom local Gradle plugin, please consider just adding a new line in your local project: `!buildSrc/` _below_ `build/`

PR #3192, opened 2019-09-29
Visual Studio Code cache directory added

gerardreches:

Visual Studio Code can be used instead of Visual Studio for Unity development. This change adds the VSC cache folder to be ignored, same as it was already done for Visual Studio.


shiftkey:
@gerardreches thanks for the contribution, but we have an existing template for [`Global/VisualStudioCode.gitignore`](https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore) which we consider to be handled differently to `.vs/`.

PR #3193, opened 2019-10-01
Add apt_generated_test/

koppor:

The annotation processing creates a separate sub-directory for test cases.

Did not find it. A developer tested it in Eclipse and manually added that directory.

### DIFF
--- a/Global/Eclipse.gitignore
+++ b/Global/Eclipse.gitignore
@@ -48,6 +48,7 @@ local.properties
 
 # Annotation Processing
 .apt_generated/
+.apt_generated_test/
 
 # Scala IDE specific (Scala & Java development for Eclipse)
 .cache-main


shiftkey:
Found some links on the Eclipse bugtracker:

 - https://bugs.eclipse.org/bugs/show_bug.cgi?id=539288

> Since upgrading to Eclipse Photon, my Maven submodules get unwanted .gitignore files in their roots with an ignore added for .apt_generated_tests/
>
> I've excluded this already in my global git configuration (.gitignore_global) and don't want my to find .gitignore files in my Maven submodules as all ignores are either handled globally or in the root of my Maven project, not at submodule level.

 - https://bugs.eclipse.org/bugs/show_bug.cgi?id=531072

> The folder for files generated for test sources can be configured in the project settings "Java Compiler" > "Annotation Processing" with the new setting "Generated test source directory:" (default value ".apt_generated_tests"). If no test code is present in the project, no corresponding folder is created or added to the classpath.

This seems like a reasonable inclusion.

PR #3194, opened 2019-10-03
Update Go.gitignore

iGusev:

improvements

### DIFF
--- a/Go.gitignore
+++ b/Go.gitignore
@@ -13,3 +13,4 @@
 
 # Dependency directories (remove the comment below to include it)
 # vendor/
+# Godeps/


shiftkey:
@iGusev any links to StackOverflow discussions or other documentation about this change?

drothmaler:
Seems to be related to the [godep - dependency tool for go](https://github.com/tools/godep). But the problem is, that Godep has been archived/deprecated and suggests using [Dep](https://github.com/golang/dep) instead - so I don't think it makes sense to include rules for an outdated tool.
Also I'm not sure if it would be correct to ignore the directory all at once, because it also contains the file `Godeps/Godeps.json`, which is kind of a lock file that contains the  the used dependency versions - and should probably be committed to git in any case. I would probably only ignore the directory `Godeps/_workspace` - where they stored the dependencies, before they also moved them to the `vendors` directory for newer versions of Go.

shiftkey:
@drothmaler thanks for the reminder. To summarize a bunch of old discussions:
 
 - `dep` uses `vendor/` which we decided in #2266 to remove (and make opt-in if users wish to ignore it)
 - if `godep` pre-dates `dep` and is now considered an outdated tool, it doesn't make sense to add it's folder

Closing this out as it doesn't make sense to pull it in.

PR #3203, opened 2019-10-14
Create Angular.gitignore file

jacekmusial:

When working with angular frontend application,  this gitignore will be useful.

https://gitignore.io/api/angular

If this is a new template:

 - **Link to application or project’s homepage**:  http://angular.io

### DIFF
--- a/Angular.gitignore
+++ b/Angular.gitignore
@@ -1,0 +1,27 @@
+dist/
+tmp/
+app/**/*.js
+app/**/*.js.map
+
+# dependencies
+node_modules/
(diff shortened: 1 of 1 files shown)


bdougie:
How is this different from using the https://github.com/github/gitignore/blob/main/Node.gitignore? And can those changes be made there?

bdougie:
After a quick review, this is mostly duplicated things already represented in the [Node.gitignore](https://github.com/github/gitignore/blob/main/Node.gitignore)

PR #3204, opened 2019-10-15
Comment out / add note about conflicting gitignore lines

karlhorky:

Next.js recently introduced a feature to be able to use the `public` directory (should **not** be ignored), which conflicts with Gatsby's generate-only `public` directory (which should be ignored).

Since the Node.js gitignore file should be unopinionated when it comes to frameworks, I made the decision to move the files out to `community/JavaScript`. I also did this for Nuxt.js, since this is a similar type of framework.

Ref (introduction of a new "public" folder in Next.js):
https://nextjs.org/blog/next-9-1#public-directory-support

Ref: #2562 by @rgbkrk
Ref: #3180 by @ImedAdel 

If this is a new template:

- Next.js https://nextjs.org/
- Gatsby https://www.gatsbyjs.org/

**Alternatives considered:**

An alternative would be to **by default** ignore all of the **overlapping** build directories (this would **not** ignore `public`) and then add a note for Gatsby (potentially with a commented-out line with `public` on it).


karlhorky:
Actually, the more I think about it, the more I like the simplicity of my other option - just to comment out `public` and leave everything in `Node.gitignore`.

I will change it to this now.

karlhorky:
Ok, I have changed it to the second version - a single Node.js `gitignore` file with all of the frameworks in it (I've also moved over some Nuxt.js things), that ignores only the overlapping items.

The `public` line is commented out and a note is added above for Gatsby users.

karlhorky:
I am also fine with going back to my original solution, in case that is more desirable for the project team.

karlhorky:
@shiftkey Added an updated solution here, in case somehow this information got lost in the notifications.

nerdybeast:
Ignoring a generic "dist" folder should not have been included in the node js ignore file. That is not unique to Next.js as a lot of projects build to a "dist" folder. I think that's too generic of a folder to be automatically ignored:
image

shiftkey:
@nerdybeast if you'd like this reverted please open a PR so we can discuss it further

karlhorky:
@nerdybeast I would argue that in most cases you will want to ignore the build output in `dist` - regardless of framework.

I think it's a reasonable default. You're going to need to go through this `.gitignore` anyway, in case you have special, uncommon requirements.

If this conflicts with some concrete, common use case that I haven't considered, then I would be all for making it optional!

nerdybeast:
@karlhorky you bring up a fair point. I was thinking of few edge cases but I agree we should cater to the majority. Thank you for the kind response!

PR #3208, opened 2019-10-16
# IntelliJ IDEA .idea/

odykyi:

_TODO_

_TODO_

If this is a new template:


shiftkey:
@odykyi thanks for the contribution, but we have an active JetBrains template available at [`Global/JetBrains.gitignore`](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore).

PR #3209, opened 2019-10-20
Ignore .vscode generated folder in Unity project

sakitkepala:

VS Code now generates .vscode folder in the root of project folder containing some custom workspace settings only for that project and it doesn't need to be tracked.

https://code.visualstudio.com/docs/getstarted/settings#_settings-file-locations

### DIFF
--- a/Unity.gitignore
+++ b/Unity.gitignore
@@ -21,6 +21,7 @@
 
 # Visual Studio cache directory
 .vs/
+.vscode/
 
 # Gradle cache directory
 .gradle/


OscarAbraham:
Hi, I just wanted to say that this makes a lot of sense. Unity is used with a lot of different editors, and people that use it with VS Code with it can have very different preferences for working with it. Sometimes one might even try different settings for the same project, or have different settings in different computers with different screens. The files included in .vscode almost never have any useful relation with the rest of the project, in the context of versioning or sharing. And, in the rare case where one might want to share the settings with other people, for educational or organizational reasons, it is usually shared in a separate way, with a set of instructions that include how to install a series of plugins necessary for Unity to work with VS Code. So the .vscode folder is almost always a nuisance in a git repository.

shiftkey:
@sakitkepala @OscarAbraham thanks for the contribution, but we have a  [`Global/VisualStudioCode.gitignore`](https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore) template that tracks files that can be shared between developers. 

For the moment I've been declining PRs that add this folder to language-specfic templates, but I have a discussion open in #3110 about the template itself. Please follow along with that.

PR #3210, opened 2019-10-21
Add Vim.gitignore

xqyww123:
Ignore for vim temporary files, which may enter repository if commit when Vim opening.


shiftkey:
Thanks, but see [`Global/Vim.gitignore`](https://github.com/github/gitignore/blob/master/Global/Vim.gitignore) for our current template.

PR #3212, opened 2019-10-24
# Business Intelligence projects - add *.dtproj.user,

pmcunha:
# Business Intelligence projects
 *.dtproj.user contain information about your preferences for working with the project, and for that reason should be ignored.

_TODO_

_TODO_

If this is a new template:


shiftkey:
@pmcunha I'm curious why this rule to catch every `*.user` file isn't satisfactory:

https://github.com/github/gitignore/blob/a14296fa82ae0be1ddeaf39b12fce1e2aed2c5b9/VisualStudio.gitignore#L9

Can you provide more context?

PR #3213, opened 2019-10-24
exclude *.tps files from TeXnicCenter

Aikhjarto:
TeXnicCenter produces a status file named *.tps which holds information on currently open *.tex files and window positions. This most likely should not be checked in.

http://texniccenter.sourceforge.net/placeholders.html#placeholder-TXC

### DIFF
--- a/TeX.gitignore
+++ b/TeX.gitignore
@@ -259,6 +259,9 @@ TSWLatexianTemp*
 # KBibTeX
 *~[0-9]*
 
+# TeXnicCenter
+*.tps
+
 # auto folder when using emacs and auctex
 ./auto/*
 *.el


TeX.gitignore
```diff
@@ -259,6 +259,9 @@ TSWLatexianTemp*
 # KBibTeX
 *~[0-9]*
 
+# TeXnicCenter
+*.tps
```
shiftkey: The linked document mentions other extensions. Should these also be added?

> **TXC**
> All files generated by TeXnicCenter regarding the project. Currently only the `.tcp`, `.tps` and `.tiw` files.

Aikhjarto:
The other two files are project dependent.
* `*.tpc` is the project defintion, e.g. main tex file to compile.
* `*.tiw` file contain spell check a white lists

As I usually check those other two files in, IMHO it is sufficient to exclude the `*.tps` file.

PR #3215, opened 2019-10-25
Ignore IDE files

zehao-sean-huang:

Ignore Ruby on Rails IDE files (RubyMine, VSCode, Atom) 

This is natural in most web development projects.

### DIFF
--- a/Rails.gitignore
+++ b/Rails.gitignore
@@ -65,3 +65,9 @@ yarn-debug.log*
 # Ignore uploaded files in development
 /storage/*
 !/storage/.keep
+
+# Ignore IDE files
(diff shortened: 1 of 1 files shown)


shiftkey:
@zehao-sean-huang thanks for the contribution, but this adds rules that diverge from some of our IDE-specific templates:

  - [`Global/JetBrains.gitignore`](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore)
 - [`Global/VisualStudioCode.gitignore`](https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore)

I don't want to disable these by default, as there are files for these IDEs that can be shared between contributors to a repository.

PR #3216, opened 2019-10-26
Add vscode user specific and workspace files to UnrealEngine

ccarvalheiro:

Unreal Engine 4 supports vscode project generation since 4.18

https://docs.unrealengine.com/en-US/Support/Builds/ReleaseNotes/4_18/index.html#new:visualstudiocodesupportedonwindows,macandlinux

### DIFF
--- a/UnrealEngine.gitignore
+++ b/UnrealEngine.gitignore
@@ -1,6 +1,9 @@
 # Visual Studio 2015 user specific files
 .vs/
 
+# Visual Studio Code user specific files
+.vscode/
+
 # Compiled Object files
 *.slo
 *.lo
@@ -40,6 +43,7 @@
 *.sdf
 *.VC.db
 *.VC.opendb
+*.code-workspace
 
 # Precompiled Assets
 SourceArt/**/*.png


UnrealEngine.gitignore
```diff
@@ -1,6 +1,9 @@
 # Visual Studio 2015 user specific files
 .vs/
 
+# Visual Studio Code user specific files
+.vscode/
```
shiftkey: For the moment I want to drop this and see where this discussion ends up, because VSCode also gets a mention there: https://github.com/github/gitignore/pull/3218/files#r347336080

UnrealEngine.gitignore
```diff
@@ -40,6 +43,7 @@
 *.sdf
 *.VC.db
 *.VC.opendb
+*.code-workspace
```
shiftkey: Can we move this out to it's own section and explain why this is needed? If this is being generated by the UE tooling, and it can be ignored, let's add that context. If it's not, then I'm curious why we should special-case it here over what's in `Global/VisualStudioCode.gitignore`

UnrealEngine.gitignore
SuperWig: Unreal generates those files which contain absolute paths to the project and Unreal Engine directories.

2020-07-07 09_32_40-tasks json -  vscode - Visual Studio Code

UnrealEngine.gitignore
SuperWig: That is also generated by Unreal, yes.

wirecat:
Closing for now as I believe this PR has gone stale.

PR #3217, opened 2019-10-26
Ignore visual studio 'enc_temp_folder' in UnrealEngine

ccarvalheiro:

It's usual to have files inside the folder when working with hot-reload. Visual Studio will delete it most of the times, but sometimes if a crash occurs or something similar happens the folder won't be deleted.

### DIFF
--- a/UnrealEngine.gitignore
+++ b/UnrealEngine.gitignore
@@ -1,5 +1,6 @@
-# Visual Studio 2015 user specific files
+# Visual Studio 2015+ user specific files
 .vs/
+enc_temp_folder/
 
 # Compiled Object files
 *.slo


shiftkey:
@ccarvalheiro any links to StackOverflow discussions or other documentation about this problem and why it's safe to ignore this directory?

ccarvalheiro:
Hi @shiftkey,

I couldn't find any discussion related to this problem. Probably because it isn't an issue, I would consider this more like a quality of life improvement.

This is the directory that holds temporary versions/backups of source files created by the [Edit And Continue](https://devblogs.microsoft.com/cppblog/c-edit-and-continue-in-visual-studio-2015/) feature of Visual Studio ("**enc**" stands for **E**dit a**n**d **C**ontinue) ([other link](https://docs.microsoft.com/en-us/visualstudio/debugger/edit-and-continue-visual-cpp)). Those files are artifacts of development that are unique to each environment, so they shouldn't be committed and that's why I think it's safe to ignore them.

The resources on this feature seem sparse so I couldn't find any more helpful information.

martinwoodward:
Closed due to lack of activity. LMK if anyone else wants this.

PR #3218, opened 2019-10-26
Ignore cmake build and JetBrains user specific files in UnrealEngine

ccarvalheiro:

Similar to https://github.com/github/gitignore/pull/3216
Support for CLion IDE in Unreal Engine 4 projects

### DIFF
--- a/UnrealEngine.gitignore
+++ b/UnrealEngine.gitignore
@@ -1,6 +1,9 @@
 # Visual Studio 2015 user specific files
 .vs/
 
+# JetBrains user specific files
+.idea/
+
 # Compiled Object files
 *.slo
 *.lo
@@ -40,6 +43,7 @@
 *.sdf
 *.VC.db
 *.VC.opendb
+CMakeLists.txt
 
 # Precompiled Assets
 SourceArt/**/*.png


UnrealEngine.gitignore
```diff
@@ -40,6 +43,7 @@
 *.sdf
 *.VC.db
 *.VC.opendb
+CMakeLists.txt
```
shiftkey: We have an existing [`CMake.gitignore`](https://github.com/github/gitignore/blob/master/CMake.gitignore) file that includes this rule, and I feel like duplicating the rule here isn't helpful in the long-term.

UnrealEngine.gitignore
```diff
@@ -1,6 +1,9 @@
 # Visual Studio 2015 user specific files
 .vs/
 
+# JetBrains user specific files
+.idea/
```
shiftkey: We have an existing [`Global\JetBrains.gitignore`](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore) that covers more than just this rule,  and I've been declining PRs that add this folder to avoid confusion.

shiftkey (commented):
@ccarvalheiro thanks for the contribution. Here's some initial thoughts of mine.

UnrealEngine.gitignore
ccarvalheiro: Hi @shiftkey, thank you for your review.

I understand your concerns on the confusion that this may create and totally understand the modular approach that generally is followed here.

However, I feel that Unreal Engine (UE) may be kind of an exception as a UE project doesn't need IDE files, only a custom file descriptor. Then it has mechanisms to re-generate IDE project files and metadata (i.e. IntelliSense bindings in Visual Studio ignored with the _.vs/_ rule) for a couple of IDEs that the engine supports. In the same way, project building is handled by engine tools and not by the standard IDE tools. These mechanisms are used fairly frequently which means that ideally most of the things that are related to the IDE should be ignored.

VSCode is an excellent example of this where [I ignored everything inside _.vscode/_ folder](https://github.com/github/gitignore/pull/3216) because the [VisualStudioCode.gitignore](https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore) doesn't ignore i.e. _launch.json_, which is generated by the engine tools. So, for UE specifically one "cannot" use UnrealEngine and VisualStudioCode _.gitignore_ files at the same time. We actually have this problem as we work with both Visual Studio and Visual Studio Code on the same project, we always have to edit the _.gitignore_ to set up it correctly.

As it stands it already has rules for Visual Studio and I thought that we could follow that model and more for other supported IDEs. I feel like the _.gitignore_ for UE would be more useful if it contained a basic set of rules ready to use by regular users for the most common IDEs, and "power users" would still be able to extend on it as they see fit. Otherwise, less experienced users may combine multiple _.gitignore_ rulesets and end up with conflicting, inconsistent or less efficient rules for their environment without really knowing why.

What do you think?

shiftkey: I don't entirely agree here, and it might be down to not having enough context about how UE works.

> Then it has mechanisms to re-generate IDE project files and metadata (i.e. IntelliSense bindings in Visual Studio ignored with the `.vs/` rule) for a couple of IDEs that the engine supports.

This directory has been around since VS2017 and has never needed to be committed to version control - that's the big difference here, and why I'm not inclined to handle JetBrains and VSCode in the same way.

> because the VisualStudioCode.gitignore doesn't ignore i.e. launch.json, **which is generated by the engine tools**.

This part is puzzling to me - why are the engine tools generating these files if they're not used or needed? If we need to exclude files for specific reasons, let's focus on those rules there. Blanket rules for the `.idea` and `.vscode` directories will prevent other files created by users from being added, which is something I want to avoid.

> As it stands it already has rules for Visual Studio and I thought that we could follow that model and more for other supported IDEs.

I want to be clear that Visual Studio until recently itself hasn't really had a good story about sharing project-specific preferences or settings within a project (think editorconfig and the like). It _has_ had a rich history of creating temporary files and directories, which we want to avoid bloating the repository history. That's basically why these rules have been duplicated over time.

I guess this is my guidance for how I'd like to proceed with this PR: what *generated* files must we exclude from the template for a UE repository, and why?

From #3216 the `*.code-workspace` is something that I want to see merged because it's part of the IDE template (just needs more context about why it's safe to ignore), but the other cases are not clear to me because they are more broad than what the IDEs recommend can be ignored.

bronsoncox: > I guess this is my guidance for how I'd like to proceed with this PR: what _generated_ files must we exclude from the template for a UE repository, and why?

The why part is probably subjective since UE supports multiple IDE/Editors via Plugin support (VS, VSC, CLion, Rider, CodeLite, etc.. ) and presumably generates a hierarchy for each accordingly.

When using the latest Rider version for UE, it generates a handful of XML files (including workspace) and an IML, but it nests them in a folder structure based off the UE project name:

`.idea/.idea.PROJECTNAME/.idea/`

YawLighthouse:
Is there going to be any future updates to this PR? 
As an Unreal developer I very much welcome this change to the template as EVERY Unreal Engine project I've worked on that uses both Visual Studio and Rider has had to ignore the `.idea` folder and the `Plugins/Developer/RiderLink` folder as both of these contain auto generated files that nobody wants to share across their Unreal Engine project's due to different configurations per project and the different tooling that is used per project.

PR #3219, opened 2019-10-27
Ignore all except wp-contents

akshay-kh:
* add ignore all before wp-content exception
@shiftkey 

Since comment says ignore all files except wp-contents the correct rule should also ignore all files except wp-contents.


davidmsibley:
:+1: the current template does nothing without this addition

PR #3221, opened 2019-11-01
added untracked SSIS file patterns;

philo-ng:

added new file patterns for SSIS

_TODO_

If this is a new template:

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -259,14 +259,19 @@ ServiceFabricBackup/
 *.ldf
 *.ndf
 
-# Business Intelligence projects
+# SQL Server Integration Services
+*.dtproj.user
+
+# SQL Server Reporting Services
 *.rdl.data
-*.bim.layout
-*.bim_*.settings
 *.rptproj.rsuser
 *- [Bb]ackup.rdl
 *- [Bb]ackup ([0-9]).rdl
 *- [Bb]ackup ([0-9][0-9]).rdl
+
+# SQL Server Analysis Services
+*.bim.layout
+*.bim_*.settings
 
 # Microsoft Fakes
 FakesAssemblies/


VisualStudio.gitignore
```diff
@@ -259,15 +259,20 @@ ServiceFabricBackup/
 *.ldf
 *.ndf
 
-# Business Intelligence projects
+# SQL Server Integration Services
+*.dtproj.user
```
shiftkey: This should be unnecessary given the `*.user` rule on line 9, unless I'm missing some context.

VisualStudio.gitignore
```diff
@@ -259,15 +259,20 @@ ServiceFabricBackup/
...
+# SQL Server Reporting Services
 *.rdl.data
-*.bim.layout
-*.bim_*.settings
 *.rptproj.rsuser
 *- [Bb]ackup.rdl
 *- [Bb]ackup ([0-9]).rdl
 *- [Bb]ackup ([0-9][0-9]).rdl
 
+# SQL Server Analysis Services
```
shiftkey: 👍 to extracting this to it's own section (if it makes sense, I'm not familiar with SSAS)

philo-ng: Yes, SSRS, SSIS, and SSAS should have it's own section.

VisualStudio.gitignore
philo-ng: Somehow, my project need this line to work.

shiftkey: Can you confirm that this isn't needed? `*.user` should also handle `*.dtproj.user` based on this quick test I did locally:

```shellsession
$ git init test-user
Initialized empty Git repository in /Users/shiftkey/src/test/test-user/.git/
$ cd test-user
$ touch foo.user
$ touch foo.dtproj.user
$ touch foo
$ git status
On branch master

No commits yet

Untracked files:
	foo
	foo.dtproj.user
	foo.user

$ echo "*.user" >> .gitignore
$ git  status
On branch master

No commits yet

Untracked files:
	.gitignore
	foo

```

If we don't need this rule, we can drop this section to keep this template a bit neater.

shiftkey:
Closing this out as I believe the new rule is already covered by `*.user`.

@philo-ng please open a fresh pull request and address the feedback above if you believe this is still needed.

PR #3224, opened 2019-11-05
Update UnrealEngine.gitignore

Jieiku:
Update UnrealEngine.gitignore to include Codelite generated files. Codelite workspace files are generated by Unreal Editor. I am a Linux user and primarily use Codelite for developing c++ to use in my Unreal Projects. The *.tags in particular are massive files.

I used this gitignore as a template for my Unreal Project. The *.tags files are so huge that they actually resulted in an error:

```
error: RPC failed; HTTP 413 curl 22 The requested URL returned error: 413 Request Entity Too Large
fatal: The remote end hung up unexpectedly
```

### DIFF
--- a/UnrealEngine.gitignore
+++ b/UnrealEngine.gitignore
@@ -1,6 +1,10 @@
 # Visual Studio 2015 user specific files
 .vs/
 
+# Codelite Generated files
+*.session
+*.tags
+
 # Compiled Object files
 *.slo
 *.lo


shiftkey:
@xekon is the the Codelite editor you're referring to? https://codelite.org/

I only ask because it looks like it supports working with different languages, and would be an ideal candidate for it's own editor file under `Global/` if you're interested.

xekon:
Yes, I use codelite, but I initiated the repository using the Gitea web interface and chose Create Repository and from there selected Unreal from the .gitignore templates.

JoeBaker1:
Get your sexy ass over here and hand me the panties on the way in

PR #3225, opened 2019-11-06
Ignore the logs folder

TheEquity:
In my opinion i don't see nobody wanting to push log files to their repo or having anything else then log files in it.

It doesn't really make sense to write code in the `logs/` directory, nor does it make sense to upload log files to github.

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -37,6 +37,9 @@ MANIFEST
 pip-log.txt
 pip-delete-this-directory.txt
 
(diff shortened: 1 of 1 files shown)


sam-ha:
Storing program logs in a folder called `logs` isn't universal, and it'd be trivial for people who do store logs in `logs` to add it to their own project's .gitignore. `*.log` is also already in Python.gitignore, which is a common suffix for log filenames.

PR #3226, opened 2019-11-06
Visual Studio - Ignore Subversion source control directory

kkam907:
Ignore Subversion source control directory.  Helpful when a developer works in an environment with a remote SVN Repository but wants to use a GIT repository for more frequent commits

My team works in a mixed environment (Subversion and GIT).  Currently we have several projects that require a remote SVN commit per feature completion (stable build only).  However, more frequent commits are left to the developer.  We use a local GIT repository for progressive check-ins and sometimes forget to add the ".svn" ignore to the file created by Visual Studio.

n/a

If this is a new template: n/a

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -351,3 +351,6 @@ MigrationBackup/
 
 # Ionide (cross platform F# VS Code tools) working folder
 .ionide/
+
+# Subversion Source Control (for mixed environments i.e. GIT local SVN Remote)
+.svn


shiftkey:
@kkam907 thanks for submitting this PR. These templates are supposed to capture the files and folders that should be ignored when working with a specific tool or language, so that user can bootstrap a new repository easily.

This case doesn't feel like a good fit for this, for a couple of reasons:

 - we have an existing [`Global/SVN.gitignore`](https://github.com/github/gitignore/blob/master/Global/SVN.gitignore) template that represents things in a Subversion repository that should be ignored in Git
 - it's not clear to me how common this scenario is

I'm going to keep an eye out for other requests for this in the future, but for now I'm going to close this out so I can focus my attention on the other outstanding PRs. Thanks again!

simonbuehler:
just want to give a +1 on adding this, pops up often when migrating old projects to git

PR #3233, opened 2019-11-10
Added aar file type to gitignore

MichalDanielDobrzanski:
Instead of regular APK, we can build AAR that are library modules. They are similar to the APK, but they are used by them as dependency modules that contain resources.

Link:

https://developer.android.com/studio/projects/android-library

### DIFF
--- a/Android.gitignore
+++ b/Android.gitignore
@@ -1,5 +1,6 @@
 # Built application files
 *.apk
+*.aar
 *.ap_
 
 # Files for the ART/Dalvik VM


shiftkey:
@MichalDanielDobrzanski thanks!

friederbluemle:
TBH, I'm not sure if the link in the description provides enough evidence for this change...

In fact, some people might have to add aar files to the default `app/libs` directory. If aar files are excluded by default, they won't show up in `git status`.

On the other hand, if you're talking about _build output_, then those files are contained in the `build/` folder - which is _already_ part of this `.gitignore`, meaning there is no real value in adding `*.aar`. This (amongst many other things) is taken care of in #2402

shiftkey:
@friederbluemle thanks for the feedback. @MichalDanielDobrzanski do you agree with this?

MichalDanielDobrzanski:
I agree that some aar files land in libs folder. 

However, the usual flow does not assume adding to Git compiled dependencies (due to their size or specific versioning). They are downloaded by default and checked out from VCS.

We can revert this, though.

shiftkey:
> We can revert this, though.

I have no strong feelings either way on this, so I'll leave it up to @friederbluemle and @MichalDanielDobrzanski to figure this out and submit a follow-up PR (if necessary).

friederbluemle:
Yes, also agree with @MichalDanielDobrzanski that compiled dependencies should _normally_ not be added to version control. Normally these files are all contained in `build/` (which is already ignored). So we can revert this. It is already taken care of in #2402 - @shiftkey please take a look. And @MichalDanielDobrzanski as well, in case there is anything to add. Thanks!!

PR #3234, opened 2019-11-11
Update Rust.gitignore

lazysoundsystem:

This echoes changes proposed in https://github.com/rust-lang/rustfmt/pull/3919 - and includes editor/OS specific conditions that are already added in that file.

The rustfmt .gitignore was originally generated via https://github.com/toptal/gitignore, (as mentioned in the first line of https://github.com/rust-lang/rustfmt/blob/master/.gitignore)

That tool uses this repo as a source (See section 2 `Patch` in https://github.com/toptal/gitignore#files). 

This is just to keep things consistent.


lazysoundsystem:
Updated to just include the differences already shown with https://github.com/rust-lang/rustfmt/blob/master/.gitignore - comment on https://github.com/rust-lang/rustfmt/pull/3919 states the other files aren't needed here.

shiftkey:
@lazysoundsystem thanks for the contribution, but these rules overlap with existing templates in the `Global` directory:

 - `Global/macOS.gitignore`
 - `Global/VisualStudioCode.gitignore`
 - `Global/Vim.gitignore`

I've been declining PRs like this in the past because the other templates are more complete, and the duplication of rules, while convenient for users, has become a maintenance burden that I'm not interested in supporting currently...

lazysoundsystem:
@shiftkey - fair enough! Thanks for the explanation and sorry for the noise.

PR #3236, opened 2019-11-15
exclude compiler.xml and artifacts when using Gradle

davidkron:
In our IntelliJ projects where we are using Gradle, the file .idea/compiler.xml and files inside the .idea/artifacts folder are automatically generated by IntelliJ based on the Gradle build model.

Automatically generated files should be ignored in version control.

I did comment out my changes though, as the existing Gradle/Maven related ignore rules are commented out as well.

https://intellij-support.jetbrains.com/hc/en-us/articles/206544839?page=3#comment_360000710180

Answer of Serge Baranov to my comment:
_The files generated on import and not having any manual customizations can be safely excluded from the version control._


shiftkey:
@davidkron thanks!

PR #3238, opened 2019-11-15
.gitignore .idea/

jtmoon79:
Git ignore PyCharm project directory `.idea/`

PyCharm is one of the most common IDEs for Python projects.  It's established project directory is always at `project/.idea`.

I've used PyCharm for several years.

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -81,6 +81,9 @@ target/
 profile_default/
 ipython_config.py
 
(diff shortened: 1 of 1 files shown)


jtmoon79:
This is already covered in https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore

BrianMSheldon:

PR #3239, opened 2019-11-15
add Node.js Diagnostic Report files

boneskull:

Node.js diagnostic report files can contain sensitive information from the environment (e.g., cloud provider secrets), and should not be committed to version control.

- [Node.js Diagnostic Report](https://nodejs.org/api/report.html#report_diagnostic_report)
- [process.report.writeReport()](https://nodejs.org/api/process.html#process_process_report_writereport_filename_err)

Per the docs,

> If filename is not provided, the default filename includes the date, time, PID, and a sequence number

An example looks like `report.20191115.141412.94056.0.001.json`.

### DIFF
--- a/Node.gitignore
+++ b/Node.gitignore
@@ -102,3 +102,6 @@ dist
 
 # TernJS port file
 .tern-port
+
+# Node.js Diagnostic Report files
+report.*.json


shiftkey:
@boneskull thanks for the contribution! It looks like this overlaps with an earlier PR #2971 - could you update this rule so that we're not duplicating it?

https://github.com/github/gitignore/blob/cb0c6ef7ac68f2300409ee85501d9ad432cb4c7e/Node.gitignore#L9-L10

boneskull:
@shiftkey I’ll close this PR in lieu of that one.

PR #3242, opened 2019-11-17
Explanation for __pypackages__

mrg0029:
I was very confused by [this](https://github.com/github/gitignore/blob/cb0c6ef7ac68f2300409ee85501d9ad432cb4c7e/Python.gitignore#L94).  I'd heard of `__pypackages__/` before but not pyflow.

A Google search for "pyflow" found several other projects with the same name ([the project](https://github.com/David-OConnor/pyflow) that `Python.gitignore` was referring to didn't show up until the 5th result for me). I only figured it out after finding [this PR](https://github.com/github/gitignore/pull/3205).

`__pypackages__` comes from [this PEP](https://www.python.org/dev/peps/pep-0582/). It's currently in draft status i.e. the proposal hasn't been accepted and the directory is not used within CPython.

I didn't know if pyflow's use of `__pypackages__` was in conflict with the PEP or if it's trying to be some kind of early implementation of it.  I'm guessing the latter since it [mentions the PEP 
 in its readme](https://github.com/David-OConnor/pyflow#references).

### tl;dr
The `__pypackages__` directory isn't used by Python yet.  But a somewhat obscure package is using the directory. It probably doesn't hurt to leave the ignore rule as is but a comment would be helpful.

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -91,7 +91,7 @@ ipython_config.py
 #   install all needed dependencies.
 #Pipfile.lock
 
-# pyflow
+# PEP 582; used by e.g. github.com/David-OConnor/pyflow
 __pypackages__/
 
 # Celery stuff


shiftkey:
cc @thedrow who added this in https://github.com/github/gitignore/pull/3205 - does this make sense to you?

mrg0029:
more research:

#### the big projects related to packaging and virtual environments
- Poetry had an issue about `__pypackages__` [here](https://github.com/sdispater/poetry/issues/872).  A lot of +1s but no replies.
- Flit had [zero matches](https://github.com/takluyver/flit/search?q=__pypackages__&unscoped_q=__pypackages__) when I searched its repo.
- Searching the whole github.com/pypa org found just [this one comment](https://github.com/pypa/pip/issues/7309#issue-519011747).

#### less-widely used projects
- pipx, which I had heard of but never used, added [this](https://github.com/pipxproject/pipx/commit/7c3b3653bad11b82806dc792f0b61fdbcfd0158a)
- pythonloc, which I had not heard of but appears related to pipx, [mentions it in their readme](https://github.com/cs01/pythonloc)

PR #3244, opened 2019-11-20
Adding Terragrunt ignore file.

lazzurs:
Terragrunt uses local caching directories as part of the run in their
repo. Ignoring the cache directory.

https://github.com/gruntwork-io/terragrunt

### DIFF
--- a/Terragrunt.gitignore
+++ b/Terragrunt.gitignore
@@ -1,0 +1,2 @@
+# Terragrunt
+.terragrunt-cache


drothmaler:
The github site states:
> Terragrunt is a thin wrapper for Terraform that provides extra tools [...]

So couldn‘t this just be added to the existing `Terraform.gitignore`, instead of creating a new one?

lazzurs:
I would be happy to add this to the existing Terraform file if you think that is best but not everyone that uses Terraform will use Terragrunt and so would end up with lines in their git ignore they don't need.

drothmaler:
Well a lot of files here contain lines for optional adding/extension/whatever, so I don’t see a reason to create a new file for just 2 lines. 

I think it also depends on how common terragrunt is. Not sure if this is representative, but Terragrunt has >3000 stars on GitHub vs. >19.000 for Terraform. For me this seems to be common enough to put it into the ‘terraform.gitignore‘...

But I‘m just a contributor like you; maybe @shiftkey can give some guidance about it.

lazzurs:
If someone with merge access can provide guidance either way happy to provide another PR if the guidance is to put it into the file with Terraform.

lazzurs:
@shiftkey should I be making this just an addition to Terraform or are you happy with it as a new file for a new tool?

yyrjlw:
。。。



---Original---

lazzurs:
Closing as it appears this is not getting merged :(

Would open another for the Terraform file but there are other PRs for that file waiting.

PR #3246, opened 2019-11-24
ignore vscode settings in python.gitignore

hainesm6:

Ignore [VS Code](https://code.visualstudio.com/) settings.

>A VS Code "workspace" is usually just your project root folder. Workspace settings as well as debugging and task configurations are stored at the root in a .vscode folder. You can also have more than one root folder in a VS Code workspace through a feature called Multi-root workspaces.

If this is a new template:

Not applicable.

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -127,3 +127,6 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+
+# VScode settings
+.vscode


shiftkey:
@hainesm6 thanks for the contribution, but we have a [`Global/VisualStudioCode.gitignore`](https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore) template which documents things that can be shared from that directory.

The multi-root workspace file `*.code-workspace` is definitely not for sharing, because it's a machine-specific and user-specific file, but there are things in here that can be shared:

https://github.com/github/gitignore/blob/db0ef78b5ad1d3caf4cab3240467562360885613/Global/VisualStudioCode.gitignore#L1-L6

PR #3250, opened 2019-11-29
Update Python.gitignore

5StevenWu:
#add pycharm  .idea/
#add django settings.py

_TODO_

_TODO_

If this is a new template:

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -58,6 +58,7 @@ coverage.xml
 # Django stuff:
 *.log
 local_settings.py
+settings.py
 db.sqlite3
 db.sqlite3-journal
 
@@ -127,3 +128,6 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+
+#Pycharm
+.idea/


Python.gitignore
```diff
@@ -58,6 +58,7 @@ coverage.xml
 # Django stuff:
 *.log
 local_settings.py
+settings.py
```
kryskool: settings.py need to be keep on the repository

Python.gitignore
```diff
@@ -127,3 +128,6 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+
+#Pycharm
+.idea/
```
kryskool: It's ok for PyCharm

shiftkey:
@5StevenWu thanks for opening this PR

I'm going to second the feedback from @kryskool here, and without more context about the reason behind the changes I plan to close this PR in a few days.

RNag:
I didn't actually submit the PR, but I think its a useful change 👍 

Pycharm editor for Python leaves behind an `.idea/` folder, but ideally there's no reason to commit it to Github i believe.

5StevenWu:
perfect

BrianMSheldon:

PR #3252, opened 2019-12-03
added Coverlet result files as a Code Coverage Tool

lkurzyniec:

[Coverlet](https://github.com/tonerdo/coverlet) is cross-platform code coverage for .NET

https://github.com/tonerdo/coverlet/blob/master/Documentation/MSBuildIntegration.md#coverage-output

By default, tool gives such files as a result (`p:CoverletOutputFormat` property):
* coverage.info
* coverage.json
* coverage.cobertura.xml
* coverage.opencover.xml

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -137,6 +137,9 @@ _TeamCity*
 .axoCover/*
 !.axoCover/settings.json
 
+# Coverlet is a free, cross platform Code Coverage Tool
+coverage*[.json, .xml, .info]
+
 # Visual Studio code coverage results
 *.coverage
 *.coveragexml


VisualStudio.gitignore
```diff
@@ -137,6 +137,9 @@ _TeamCity*
 .axoCover/*
 !.axoCover/settings.json
 
+# Coverlet is a free, cross platform Code Coverage Tool
+coverage*[.json, .xml, .info]
```
shiftkey: I'm not 100% sure this is how to define a rule to cover multiple extensions. Here's how I tested it:

```shellsession
$ git init coverlet-test
Initialized empty Git repository in /Users/shiftkey/src/test/coverlet-test/.git/
$ cd coverlet-test
$ echo "coverage*[.json, .xml, .info]" >> .gitignore
$ cat .gitignore
coverage*[.json, .xml, .info]
$ git add .
$ git commit -am "add rule"
[master (root-commit) 5240fe7] add rule
 1 file changed, 1 insertion(+)
 create mode 100644 .gitignore
$ touch coverlet.json
$ git status
On branch master
Untracked files:
	coverlet.json

```

I think this needs to be split out to multiple lines to achieve the effect you're interested in.

lkurzyniec: @shiftkey it works fine. you created a file with wrong name, that does not match the rule - you created `coverlet.json`, but the rule is `coverage.json`. Could be typo ;)

image

shiftkey: > you created a file with wrong name



Well that was silly of me. Serves me right for doing things before I'm properly caffeinated.

phoogenb: @lkurzyniec: @shiftkey is correct.  The syntax is wrong.  I discovered this because the pattern is causing source files named "Coverage*.cs" to be ignored.

shiftkey: @phoogenb do you think this is something that needs to be revisited? I've not heard any issues about this since it was merged...

lkurzyniec: @shiftkey
unfortunately @phoogenb is right, here is PR to fix that #3415.

PR #3256, opened 2019-12-05
Ignore files inside `.vscode-test`

SimonSiefke:

[vscode-test](https://github.com/microsoft/vscode-test) is a testing framework for vscode extensions. Inside the `vscode-test` folder are stored one or more versions of vscode, which are used for testing a vscode extension.

https://github.com/microsoft/vscode-test/blob/master/sample/.gitignore

### DIFF
--- a/Node.gitignore
+++ b/Node.gitignore
@@ -103,5 +103,5 @@ dist
 # TernJS port file
 .tern-port
 
-# VSCode test files
+# Stores VSCode versions used for testing VSCode extensions
 .vscode-test
--- a/Node.gitignore
+++ b/Node.gitignore
@@ -102,3 +102,6 @@ dist
 
 # TernJS port file
 .tern-port
+
+# VSCode test files
+.vscode-test


Node.gitignore
```diff
@@ -102,3 +102,6 @@ dist
 
 # TernJS port file
 .tern-port
+
+# VSCode test files
```
shiftkey: I'm still reading up on the details in here

https://github.com/microsoft/vscode-test/blob/487a134ea86da96c28c7ea7093b99b71f21b83d5/lib/download.ts#L15

```ts
const vscodeTestDir = path.resolve(extensionRoot, '.vscode-test');
```

But this description feels like it's missing some context. At first glance it looks like this directory is used to download a version of VSCode for use with testing a VSCode extension. This feels more specific than what's mentioned here, and doesn't feel (to me at least) like it should be part of the Node template.

Thoughts?

SimonSiefke: > But this description feels like it's missing some context. At first glance it looks like this directory is used to download a version of VSCode for use with testing a VSCode extension. 

You are right. The description doesn't quite fit. How about 
- `# Stores VSCode versions used for end-to-end testing VSCode extensions` or 
- `# Stores VSCode versions used for testing VSCode extensions` or 
- `# Stores downloaded versions of VSCode for use with testing a VSCode extension`

> and doesn't feel (to me at least) like it should be part of the Node template.

I disagree. I would say that VSCode extensions are NodeJS applications just as Gatsby Apps and Nuxt Apps are NodeJS applications.

shiftkey: > `# Stores VSCode versions used for testing VSCode extensions`

I like this option

> I would say that VSCode extensions are NodeJS applications just as Gatsby Apps and Nuxt Apps are NodeJS applications.

One of the things that I weigh up with changes like these is how impactful they are. We have [community templates](https://github.com/github/gitignore/tree/master/community/JavaScript) for some Node-based projects, which is the eventual model I'd like to get to.

But that's not a blocker on this PR, and there's other things I'd like to get in place before the Gatsby's and Nuxt's of the world are outside this core template.

PR #3258, opened 2019-12-06
Create PHP.gitignore

vinnyfs89:

Creates a starter template for PHP applications.

:sparkles: This is a new template. 

**Link to project’s homepage:** https://www.php.net

### DIFF
--- a/PHP.gitignore
+++ b/PHP.gitignore
@@ -1,0 +1,3 @@
+/vendor/
+composer.phar
+.env


shiftkey:
@vinnyfs89 thanks for opening this PR. The rules here feel like they are covered by the [`Composer.gitignore`](https://github.com/github/gitignore/blob/master/Composer.gitignore).

`.env` is not specific to PHP, but because we haven't figured out a good story for it has made an appearance in some other templates. If you can think of a better home for that I'd be open to reviewing a PR for that.

vinnyfs89:
@shiftkey Thanks for reply.

I agree with the questions asked. Despite having identical files to those already in Composer.gitignore, this is currently a minimal framework for PHP projects or modules that most current frameworks already use.

Initially the purpose of this template is to make it possible to create a PHP repository that ignores **minimally** what is most common in most PHP applications, which are files linked to the PHP dependency manager (Composer) and .env environment variables.

but I understand what you said and using [Python.gitignore template](https://github.com/github/gitignore/blob/master/Python.gitignore) as a reference, i added most of the files that are generated by PHP most used Frameworks, ORMs and Environment Variables.

Please check that you are now in agreement.

vasilvestre:
.env is required by Symfony

PR #3263, opened 2019-12-14
Update Node.gitignore

anishagg17:

DS_Store files are automatically created in Macs by the operating system to store Finder visibility information (the way icons are displayed) .They are therefore specific for a pc and shouldn't be pushed to repository 

https://github.com/ebeshero/DHClass-Hub/wiki/Banishing-DS_Store,-Thumbs.db,-and-other-pesky-files-we-don%27t-need


shiftkey:
@anishagg17 thanks for the contribution, but as this isn't a rule specific to Node we have a [`Global\macOS.gitignore`](https://github.com/github/gitignore/blob/master/Global/macOS.gitignore) that defines a set of rules for macOS that can safely be ignored. We recommend using that in your projects.

PR #3265, opened 2019-12-14
Added ReactNative gitignore file

anishagg17:

On React Native project There are many files that don’t need to be checked in to Github

https://github.com/facebook/react-native/blob/master/.gitignore


anishagg17:
Why is't my pr merged?

PR #3267, opened 2019-12-15
Update Python.gitignore

rk4bir:

This change is for [PyCharm IDE](https://www.jetbrains.com/pycharm/). I feel that **PyCharm IDE**'s project settings should be included in the `Python.gitignore`. The directory `.idea` is automatically generated by PyCharm for its project.

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -127,3 +127,7 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+
+# PyCharm settings
+.idea/
+


shiftkey:
@rk4bir thanks for the contribution, but we have an existing template for the JetBrains IDEs at [`Global/JetBrains.gitignore`](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore) that we recommend using in your projects.

BrianMSheldon:

PR #3269, opened 2019-12-17
[Godot] Ignore imported translations

Calinou:

The Godot editor automatically creates those binary files from translations in CSV format. There's no need to add them to version control.

https://docs.godotengine.org/en/latest/getting_started/workflow/assets/importing_translations.html

### DIFF
--- a/Godot.gitignore
+++ b/Godot.gitignore
@@ -4,6 +4,9 @@
 export.cfg
 export_presets.cfg
 
+# Imported translations (automatically generated from CSV files)
+*.translation
+
 # Mono-specific ignores
 .mono/
 data_*/


Xrayez:
Also see [my comment](https://godotengine.org/qa/22993/3-0-gitignore?show=22993#q22993) regarding this at one of the Godot's Q&A answers.

Godot.gitignore
```diff
@@ -3,6 +3,8 @@
 .import/
 export.cfg
 export_presets.cfg
+# Imported translations (automatically generated from CSV files)
```
shiftkey: It looks like this can be it's own section, so should we add a newline before this comment header?

```suggestion

# Imported translations (automatically generated from CSV files)
```

PR #3271, opened 2019-12-21
Update Autotools.gitignore

lromor:

When generating an hello world project the folder .deps is generated in src/

I can't find in the documentation details about the .deps folder but there are multiple resources on the internet that indeed add to the .gitignore the .deps folder.

### DIFF
--- a/Autotools.gitignore
+++ b/Autotools.gitignore
@@ -6,6 +6,7 @@ Makefile.in
 /py-compile
 /test-driver
 /ylwrap
+.deps/
 
 # http://www.gnu.org/software/autoconf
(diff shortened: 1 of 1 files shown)


shiftkey:
> I can't find in the documentation details about the .deps folder but there are multiple resources on the internet that indeed add to the .gitignore the .deps folder.

@lromor can you link to them here for future reference?

lromor:
Seems to be an autoconf autogenerated folder.

https://stackoverflow.com/questions/16404066/which-files-of-an-autoconf-project-to-put-into-gitignore

PR #3272, opened 2019-12-22
[Python] Ignore VS Code's 'Container - Remote extension'

imadtoubal:

Support developers who use Remote - Containers [Preview] extension for VS Code (https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers)

https://code.visualstudio.com/docs/remote/containers
https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers

If this is a new template:

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -127,3 +127,6 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+
+# VS Code - Container - Remote
+.devcontainer


shiftkey:
@imadtoubal thanks for the contribution. As I understand it, this feature of VSCode isn't tied to a specific languages - could this be added to [`Global/VisualStudioCode.gitignore`](https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore) instead?

shiftkey:
Also, any documentation or links about the fact that this folder can be ignored in version control would be greatly appreciated.

martinwoodward:
Agree this should be the in `Global/VisualStudioCode.gitignore`

PR #3273, opened 2019-12-27
Added Rider (IDEA) support

cwoodruff:
Adding JetBrains Rider support since the reason why it was removed is false. 

There is no ignore file for Rider. Also since Rider supports .NET and .NET Core why should it not be included in this ignore file. Would just replicate almost all of the entries present in this ignore file.

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -351,3 +351,7 @@ MigrationBackup/
 
 # Ionide (cross platform F# VS Code tools) working folder
 .ionide/
+
+# JetBrains Rider
+.idea/
+*.sln.iml


shiftkey:
@cwoodruff thanks for the contribution!

> There is no ignore file for Rider.

We've been maintaining a [`Global/JetBrains.gitignore`](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore) for a while in this repository, which should cover the documented rules for Rider alongside all the IntelliJ family.

This was dropped from `VisualStudio.gitignore` in #2991 so I'm curious what's changed to motivate re-adding them. Can you elaborate?

cwoodruff:
> This was dropped from `VisualStudio.gitignore` in #2991 so I'm curious what's changed to motivate re-adding them. Can you elaborate?

@shiftkey thanks for looking at this PR. I have two reasons to add the ability to ignore the .idea folder from the Visual Studio .gitignore.

1. There may be a JetBrains.gitignore that is maintained but this is not present in GitHub when creating new repositories.
2. Most importantly, we must remember that the Visual Studio family of IDE's are not the only .NET IDE currently or in the future. Rider is a first-class .NET IDE and so needs to have the same intention as VS has. Lumping in Rider to the JetBrains.gitignore will not allow files that VS gitignore covers to be ignored for .NET and .NET Core generated solutions and projects with Rider.

I do not see the big deal is having 2 more lines to ignore the .ideafolder and .sln.iml files. To be honest the Visual Studio gitignore file should be renamed since it covers a development framework. I don't see any other IDE specific gitignore other than Delphi. Similar to the Java gitignore file.

I hope you understand my view and can see my reasoning.

shiftkey:
> There may be a JetBrains.gitignore that is maintained but this is not present in GitHub when creating new repositories.

I'd love to review and clean up a bunch of very old templates that I bet are never used, but that's something that needs to be raised with GitHub itself - I barely have the bandwidth to review these contributions these days.

> To be honest the Visual Studio gitignore file should be renamed since it covers a development framework.

This is something I've asked for input from community members several times in the past, with little movement or feedback about what I've proposed:

 - https://github.com/github/gitignore/pull/2989#issuecomment-471509901
 - https://github.com/github/gitignore/pull/2816#issuecomment-422432117
 - https://github.com/github/gitignore/pull/2720#issuecomment-429573217

There's a rich history about how we ended up with `VisualStudio.gitignore` at the root, and while I do agree that things should change, I'm looking for people who are more across modern .NET dev to propose a simpler template for .NET so the defaults can be editor agnostic.

cwoodruff:
@shiftkey I understand the history but why was the Rider support removed last year? I am curious about that really. Did it affect anything with Visual Studio?

shiftkey:
> I understand the history but why was the Rider support removed last year?

Looking back at #2991 I think the feeling at the time was we wanted to avoid duplicating rules in different templates, even if it wasn't the exact same rules. I also like ensuring these templates represent what can be version controlled, and the JetBrains template comes from a knowledge-base article that JetBrains maintain.

cwoodruff:
@shiftkey I am open to letting this PR go. It would be good to look at things in detail for what Rider needs from both the JetBrains side and the Visual Studio. Let's just let this PR be ignored :-)

serialseb:
Seeing as there's not been changes coming up towards a nirvana solution, and considering forcing one to chose between Jetbrains or Visual Studio for .net projects is antithetical to how dotnet is being used out there, and the low cost of that pull request, maybe reintroducing this until such a time as that perfect solution is found would alleviate this false dichotomy of Visual Studio or Jetbrains dotnet devs?

worldbeater:
+1 for adding `.idea/` to the Visual Studio `.gitignore` file, at least until a separate `.gitignore` template is introduced for Rider. A bit tired adding those lines by hand every time I create a separate repo with cross-platform .NET projects, would be happy if this could be automated. The PR which removed those lines [states](https://github.com/github/gitignore/pull/2991), that

> Rider has its own ignore file, so does Visual Studio.

But as for now, this is not true.

PR #3275, opened 2019-12-30
Pytest-Benchmark

Better-Boy:
Adding pytest-benchmark support

When you run pytest-benchmark, it creates a folder called - ".benchmarks" which is not needed to be in the repository


shiftkey:
@Better-Boy any links to StackOverflow discussions or other documentation about this change?

martinwoodward:
Closing due to lack of activity.

PR #3277, opened 2020-01-04
Create PlatformIO.gitignore

bamapookie:
Add default ignores from PlatformIO in VSCode

There currently is not a gitignore template for PlatformIO

Can't find in documentation, but this is the default from the PIO template.

This is a new template:

 - https://platformio.org/


wirecat (approved):
Thank you for contributing!

PR #3278, opened 2020-01-05
Update PureScript adding .spago

kayhide:

Recently, [spago](https://github.com/spacchetti/spago) is getting more common as a build tool for PureScript.
This PR adds `.spago` which is to be generated when we use this tool.

https://github.com/spacchetti/spago

### DIFF
--- a/PureScript.gitignore
+++ b/PureScript.gitignore
@@ -1,5 +1,6 @@
 # Dependencies
 .psci_modules
+.spago
 bower_components
 node_modules
 


cdepillabout:
cc @f-f and @justinwoo

justinwoo:
my guess is that they'll ask you to move `.spago` up to the `# Dependencies` section

shiftkey:
> my guess is that they'll ask you to move `.spago` up to the `# Dependencies` section

I like this idea, as it looks like `spago` is a package manager. Am I understanding that right @kayhide?

kayhide:
@shiftkey `spago` does both of package management and build management.
Though, `.spago` directory has dependent packages.

Okay, I will move it to `# Dependencies` section.

kayhide:
@shiftkey I moved it to the `#Dependencies`. Please take a look.

PR #3280, opened 2020-01-07
Adding .DS_Store to list

Gr8Warrior:

On macOS development environment for android, .DS_Store is created on every folder. **.DS_Store** is a file that stores custom attributes of its containing folder, such as the position of icons or the choice of a background image, which isn't required to be placed in VCS. Many android devs work on mac machine for them it's a must to be placed in gitignore.

https://gist.github.com/iainconnor/8605514
https://coderwall.com/p/piwu4w/gitignore-for-android-studio-projects

### DIFF
--- a/Android.gitignore
+++ b/Android.gitignore
@@ -83,3 +83,6 @@ lint/generated/
 lint/outputs/
 lint/tmp/
 # lint/reports/
+
+# macOS Generated
+**.DS_Store


JerryGreen:
For macos users there's [macOS.gitignore](https://github.com/github/gitignore/blob/master/Global/macOS.gitignore)

Gr8Warrior:
> For macos users there's [`macOS.gitignore`](https://github.com/github/gitignore/blob/master/Global/macOS.gitignore)

For every Android Project on Mac OS, Do We need to integrate macOS.gitignore and Android.gitignore?

aruanoguate:
Hi @Gr8Warrior 

Thanks for your contribution.

Regards,

Alvaro

PR #3281, opened 2020-01-07
Update JetBrains.gitignore (venv default folder (created by Pycharm) to be ignored)

dexXxed:

_Pycharm by default makes folder venv/ to keep there virtual environment_


drothmaler:
`venv` Is already part of `Python.gitignore` so there is no need to add it here...

dexXxed:
Ok, thank\`s for the feedback! I`m closing this pull request

BrianMSheldon:

PR #3282, opened 2020-01-07
Update WordPress.gitignore

Triloworld:
Hello all and thanks to create great file.
There is problem to find proper meaning of content for this file in this repo.
My propose need review as I have different, more direct approach with fix expectation.

1. Header. To know what rules are used, and be apply to project after using template. Errors can be removed, updates won't be broken and clearly show why I using that file. 
Now: Any changes are confused.
After: Rules are in header

2. Introduce link to Github where template is created. Like a template, this can be modified so no big deal. Used as link where to watch. For this pull request nobody will update their gitignore file just because that link don't preexist from copy & paste.
Now: Manually fixing and no connection to place where file come from.
After: Can be used to update if some bugs will be introduce. 

3. Core Files need to be *ignored* (clue from comment) but there was broken part:
Solution apply and work (thx. @stkrzysiak) : https://github.com/github/gitignore/pull/3093/files/1115dec039582df7242acd588811357f76d610ce#diff-f2cf96854b528e9828c4ca34a4917e54
Now: All root catalog was added ;/
After: Just some will be excluded

If core file need to be *included*, then we need fix other parts as and new pull request:
- /wp-content/language
- /wp-content/index.php (security bug when someone use apache listings)

4. Thanks @chougran
https://github.com/github/gitignore/pull/3125
This is most important as gitignore provide solution to sensitive credentials. 
Now: wp-config leak
After: Configuration is safe :)

5. License
As they can be manually edited for adding more software (bundle of license) then is out of scope for ignoring it in this particular template. Rule is to add wordpress context stuff that are repeated every new repository is created. Maybe add clue that need to be handled manually? or just header that Wordpress use particular one.

It is tested on: WP 5.3.2
Used: Custom theme + custom plugin (they language are ignore but can be download from dashboard)

Comments appreciate as my first post here :)

### DIFF
--- a/WordPress.gitignore
+++ b/WordPress.gitignore
@@ -12,7 +12,9 @@ wp-config.php
 wp-content/*
 !wp-content/mu-plugins/
 !wp-content/plugins/
+wp-content/plugins/index.php
 !wp-content/themes/
+wp-content/themes/index.php
 
 # Example plugins
 wp-content/plugins/hello.php
--- a/WordPress.gitignore
+++ b/WordPress.gitignore
@@ -1,23 +1,29 @@
-# ignore everything in the root except the "wp-content" directory.
+# Wordpress - ignore core, configuration, examples, generated data and logs.
+# https://github.com/github/gitignore/blob/master/WordPress.gitignore
+
+# Core
+/*
 !wp-content/
 
-# ignore everything in the "wp-content" directory, except:
-# "mu-plugins", "plugins", "themes" directory
+# Configuration
+wp-config.php
+
+# Generated data without custom one
 wp-content/*
 !wp-content/mu-plugins/
 !wp-content/plugins/
 !wp-content/themes/
 
-# ignore these plugins
+# Example plugins
 wp-content/plugins/hello.php
 
-# ignore specific themes
+# Example themes
 wp-content/themes/twenty*/
 
-# ignore node dependency directories
+# Example themes use node dependency
 node_modules/
 
-# ignore log files and databases
+# Log files and databases
 *.log
 *.sql
 *.sqlite
--- a/WordPress.gitignore
+++ b/WordPress.gitignore
@@ -29,3 +29,6 @@ node_modules/
 *.log
 *.sql
 *.sqlite
+
+# Self
+!*.gitignore


Triloworld:
5. License
As they can be manually edited for adding more software (bundle of license) then is out of scope for ignoring it in this particular template. Rule is to add wordpress context stuff that are repeated every new repository is created.  Maybe add clue that need to be handled manually? or just header that Wordpress use particular one.

It is tested on: WP 5.3.2
Used: Custom theme + custom plugin (they language are ignore but can be download from dashboard)

WordPress.gitignore
```diff
@@ -1,23 +1,34 @@
...
 wp-content/themes/twenty*/
 
-# ignore node dependency directories
+# Example themes use node dependency
 node_modules/
 
-# ignore log files and databases
+# Log files and databases
 *.log
 *.sql
```
pedrosanta: Why are we ignoring `.sql` files? Do they naturally come/are generated on a typical WordPress installation? Aren't we over-assuming this?

Triloworld: Wordpress require SQL database. I don't find that they pre-exist on first installation. 
Remove it? I don't see value to have it in this ignore file template

pedrosanta: Yes, I think the `*.sql` and `*.sqlite` rules should be removed. I know some people might have them as part of their workflows, but I see this gitignore repository more as **base/starter gitignore files**, that can be mixed together (like using [gitignore.io](http://gitignore.io))

I see these files only containing (1) the specific files for each technology/tool and (2) and _possibly_ rules for widely common configurations of such technologies/tools (ex: `*.log` line).

Also, they shouldn't have redundant lines that are covered by other specific gitignores (ex: [`Global/macOS.gitignore`](https://github.com/github/gitignore/blob/master/Global/macOS.gitignore)).

Project/workflow specific rules should be added/maintained individually on each project, IMHO.

pedrosanta: Btw, added my own contribution addressing the above and more on #3295.

shiftkey:
Closing this out in favour of #3295 which seems to most detailed discussion

Triloworld:
Why part of solution was rejected when really it was accepted?
This PR's can be merged, some small fix needed to be done before.
Discussion of #3295 can be really long when my fix just are in place of some problem when I use that template.

Triloworld:
I propose fix to pull request:
- revert "catch-all" after community decide what can be done there and better, tested declarative solution will be
- Remove logs, added after next pull request when there will be evidence that starter need that

shiftkey:
@Triloworld I'm not opposed to interim fixes while #3295 goes on, it just wasn't clear to me whether this PR was active based on the discussion. 

> I propose fix to pull request:
> 
> * revert "catch-all" after community decide what can be done there and better, tested declarative solution will be
> * Remove logs, added in other pull request when there will be evidence that starter template need that

Please open up PRs for these changes, as they feel like they are a step in the right direction.

PR #3283, opened 2020-01-08
[Java] i think these 4 lines will helpful

Daibv:
maybe many diffurent eclipse or IDEA have diffurent special file ,so these files or directory need to be ignored.

### DIFF
--- a/Java.gitignore
+++ b/Java.gitignore
@@ -21,3 +21,8 @@
 
 # virtual machine crash logs, see http://www.java.com/en/download/help/error_hotspot.xml
 hs_err_pid*
+
+.classpath
+.project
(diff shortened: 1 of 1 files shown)


shiftkey:
@Daibv thanks for the contribution, but we have existing templates that are more suitable for (or already cover) these changes:

 -  [`Global/Eclipse.gitignore`](https://github.com/github/gitignore/blob/master/Global/Eclipse.gitignore)
 -  [`Global/JetBrains.gitignore`](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore)

Please check those templates and suggest changes that you think would be relevant for them.

PR #3284, opened 2020-01-08
[Python] Add PyCharm and VSCode project/setting folders

lpozo:

PyCharm and VSCode are very popular tools in the Python Community.

https://code.visualstudio.com/docs#vscode
https://www.jetbrains.com/pycharm/

If this is a new template:

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -114,6 +114,12 @@ venv.bak/
 .spyderproject
 .spyproject
 
+# PyCharm project settings
+.idea/
+
+# VSCode project settings
+.vscode/
+
 # Rope project settings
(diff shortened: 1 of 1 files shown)


shiftkey:
@lpozo thanks for the contribution, but we have existing templates for both cases that are actively maintained:

 -  [`Global/JetBrains.gitignore`](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore)
 - [`Global/VisualStudioCode.gitignore`](https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore)

BrianMSheldon:

PR #3285, opened 2020-01-09
Ignore .DS_Store

marcusturewicz:
I always have to ignore this when developing .NET applications on Mac.

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -351,3 +351,6 @@ MigrationBackup/
 
 # Ionide (cross platform F# VS Code tools) working folder
 .ionide/
+
+# Mac files
+.DS_Store


aruanoguate:
Hi @marcusturewicz 

Thanks for your contribution.

Regards,

Alvaro

marcusturewicz:
I don’t think either of those are appropriate alternatives. .NET Core is
cross platform and therefore one should be able to check out a .NET Core
repository on Windows, Mac and Linux and not have to have any local config
on their machine in order to contribute to it. Perhaps the Visual Studio
gitignore is no longer appropriate for .NET Core development, and there
should be a dedicated gitignore file for that, which includes handling of
DS_Store?

shiftkey:
This repository is intended to be a set of reference templates, and if we went down this path (combining templates for convenience) it would add a maintenance burden to updating (when a rule is added, changed, or removed) that I'm not able to support currently.

Maybe in the future we'll get there, but for now I need to keep these separate.

> Perhaps the Visual Studio gitignore is no longer appropriate for .NET Core development, and there
should be a dedicated gitignore file for that, which includes handling of DS_Store?

I mostly agree with this sentiment because of how complex `VisualStudio.gitignore` already is, and have offered to help contributors figure out a specific template:

 - https://github.com/github/gitignore/pull/2989#issuecomment-471509901
 - https://github.com/github/gitignore/pull/2720#issuecomment-429573217
 - https://github.com/github/gitignore/pull/2816#issuecomment-422432117

But I don't agree with pulling in files like `.DS_Store`, and have been declining contributions like this for other templates like Python and Ruby.

PR #3286, opened 2020-01-10
TeX: Ignore REVTeX generated Notes.bib files

hmenke:

The REVTeX documentclass [<a href="#footnote1">1</a>,<a href="#footnote2">2</a>] for LaTeX is the default style for submitting papers to the journals published by the American Physical Society (APS).

By default this class is loaded with the option `footinbib` which displays footnotes in the bibliography.  To do so, the class generates a file with the name `<jobname>Notes.bib` in the working directory.  Since this file is generated from footnotes that are present in the input, it essentially duplicates information and should therefore not be placed under version control.

See section VII. “Footnotes” or just search for `Notes.bib`.

> Please note that even if  BibTeX is not being used for the references, you may have to run BibTeX if you are using footnotes without the `nofootinbib` option. The log file will contain errors about missing references such as `Note1` in this case and a file ending in `Notes.bib` will have been produced during the processing of the TeX file.

Source: http://mirrors.ctan.org/macros/latex/contrib/revtex/doc/auguide/auguide4-2.pdf#section*.47

---

[<a id="foonote1">1</a>] https://ctan.org/pkg/revtex
[<a id="foonote2">2</a>] https://journals.aps.org/revtex

### DIFF
--- a/TeX.gitignore
+++ b/TeX.gitignore
@@ -274,3 +274,8 @@ TSWLatexianTemp*
 
 # Makeindex log files
 *.lpz
+
+# REVTeX puts footnotes in the bibliography by default, unless the nofootinbib
+# option is specified. Footnotes are the stored in a file with suffix Notes.bib.
+# Uncomment the next line to have this generated file ignored.
+#*Notes.bib


Anton-Sakovich:
I would like to note that including the `*Notes.bib` pattern into `TeX.gitignore` as it is will mean that a bibliography file for _any_ document whose name ends in `Notes` will be ignored unless specifically renamed, most likely unexpectedly for those who is using  `TeX.gitignore`. It doesn't sound good. I'd suggest commenting the line out by default. Although it is easy not to spot lines commented out, IMHO, it is better to risk including something unnecessary than losing something required.

hmenke:
Fair enough, I've updated my branch.

Anton-Sakovich:
Now it seems OK to me.

Well, a tiny remark: what about a more detailed description? Something which will suggest why the files are excluded even for those who don't know what `footinbib` is, e.g. "Bibliography files generated by REVTeX to support its footnotes system".

hmenke:
I have added a longer description.

PR #3290, opened 2020-01-15
ignore any git mergetool files

jasond-s:

Prevent git mergetool files from being added to the git repo

https://stackoverflow.com/questions/1251681/git-mergetool-generates-unwanted-orig-files

or 

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -232,6 +232,7 @@ ClientBin/
 *.pfx
 *.publishsettings
 orleans.codegen.cs
+*.orig
 
 # Including strong name files can present a security risk
 # (https://github.com/github/gitignore/pull/2483#issue-259490424)


shiftkey:
@jasond-s thanks for the contribution, but this rule is already present in two global templates:

 - [Global/Backup.gitignore](https://github.com/github/gitignore/blob/cd872a98f0faf94f3472645e9f8cacfbcd463b71/Global/Backup.gitignore)
 - [Global/Patch.gitignore](https://github.com/github/gitignore/blob/cd872a98f0faf94f3472645e9f8cacfbcd463b71/Global/Patch.gitignore)

We prefer global templates for situations that aren't related to a specific tool or technology, and recommend adding that to your global config or project gitignore.

jasond-s:
Ah. Saw it was in the rails file and a few others, so thought I'd do the same, didn't realise there was a global one.

shiftkey:
> Saw it was in the rails file and a few others, so thought I'd do the same

That's a vestige of days gone by, and should be cleaned up at some stage:

https://github.com/github/gitignore/commit/2f219bd07e6a65ef6cd51012d678482b564fa171

jasond-s:
Good to know there's something I might be able to help with. Give something back.

PR #3292, opened 2020-01-21
ignore files generated by Metals, a vscode extension for Scala.

goatwu1993:

Metals, an extension for Scala, will create files.

<https://scalameta.org/metals/docs/editors/vscode.html#gitignore-projectmetalssbt-metals-and-bloop>

~~If this is a new template:~~

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -351,3 +351,8 @@ MigrationBackup/
 
 # Ionide (cross platform F# VS Code tools) working folder
 .ionide/
+
+# Scala extensions
+.bloop/
+.metals/
+.scalafmt.conf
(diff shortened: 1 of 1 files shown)


ckipp01:
I didn't see this when I added my [pr](https://github.com/github/gitignore/pull/3296) in, but I think it may be best not include these here in the vscode extension, but rather in a more general one. These aren't specific to vs code, but rather to all editors that are choosing to use Metals as a language server.

goatwu1993:
Somehow I misunderstood metals as an vscode extension. Thanks for answering.

shiftkey:
Thanks for the contribution @goatwu1993!

Given #3296 is the approach that I recommend here for defining rules for editors and tooling needs, I'm going to close this out. Let's discuss our options there.

PR #3293, opened 2020-01-22
Wordpress gitignore should ignore everything in root

Ahuahuachi:

Comment in file, mentions that everything in root directory should be ignored, but there is no line doing it

### DIFF
--- a/WordPress.gitignore
+++ b/WordPress.gitignore
@@ -1,4 +1,5 @@
 # ignore everything in the root except the "wp-content" directory.
+/*
 !wp-content/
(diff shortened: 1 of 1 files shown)


pedrosanta:
Duplicate of #3093 .

Other open PRs around this file/WordPress gitignore are #3125 and #3282 .

@Ahuahuachi possible to ping in/join the above mentioned duplicate and close this?

shiftkey:
Closing this out in favour of #3295 which seems to most detailed discussion

PR #3294, opened 2020-01-22
Add .dtproj.user

Liturgist:
Added *.dtproj.user for SSIS projects.

Individual user configuration and settings are not shared among multiple developers.

https://docs.microsoft.com/en-us/sql/integration-services/integration-services-ssis-projects-and-solutions?redirectedfrom=MSDN&view=sql-server-ver15

"The *.dtproj.user file contains information about your preferences for working with the project"

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -267,6 +267,7 @@ ServiceFabricBackup/
 *- [Bb]ackup.rdl
 *- [Bb]ackup ([0-9]).rdl
 *- [Bb]ackup ([0-9][0-9]).rdl
+*.dtproj.user
 
 # Microsoft Fakes
 FakesAssemblies/


shiftkey:
@Liturgist it's not clear to me why we need this rule if we have `*.user` near the top of the file:

https://github.com/github/gitignore/blob/7ab549fcae8269fdd4004065470176f829d88200/VisualStudio.gitignore#L9

I've asked about this in #3221 and I'll repost the test I did here for reference:

Can you confirm that this isn't needed? `*.user` should also handle `*.dtproj.user` based on this quick test I did locally:

```shellsession
$ git init test-user
Initialized empty Git repository in /Users/shiftkey/src/test/test-user/.git/
$ cd test-user
$ touch foo.user
$ touch foo.dtproj.user
$ touch foo
$ git status
On branch master

No commits yet

Untracked files:
	foo
	foo.dtproj.user
	foo.user

$ echo "*.user" >> .gitignore
$ git  status
On branch master

No commits yet

Untracked files:
	.gitignore
	foo

```

Liturgist:
@shiftkey - I think you are correct; *.dtproj.user is not needed since *.user is already in the file.

My apologies for wasting your time.

PR #3295, opened 2020-01-23
Update and fix WordPress gitignore

pedrosanta:
(**Reasons for making this change:**)

### Background for this PR

#2972 seems to have broken the WordPress gitignore—using this file with a brand new install, allows to add pretty much all the install, because it's missing a catch-all rule, since it operates with excludes mostly. So WordPress gitignore wasn't seemingly working at all.

#3125 it's an update to add configuration file to gitignore, but it doesn't fix all the other underlying issues with this gitignore.

#3093 as noted by @NerdsvilleCEO [here](https://github.com/github/gitignore/pull/3093/files#r322507507) is missing a rule to allow for gitignore itself. (#3293, #3378, #3438 are duplicates.)

#3282, from which this PR is based, is IMHO the better attempt at fixing this gitignore, but I see two potential issues with it:

1. Ignores `*.sql` files, and I'm afraid we might be **over assuming** the ignores, as a typical WordPress install doesn't seem to generate `*.sql` files, and I think it might be more linked to a specific workflow that generates `*.sql` files (see [here](https://github.com/github/gitignore/pull/3093/files#r369312284) and [here](https://github.com/github/gitignore/pull/3282#discussion_r370250802)).

2. Employs a 'catch-all' approach for all root files with exclude rules, and I think that might not be the safest or best approach—it might conflict and break other gitignore files you mix with this one, and force users to **always** exclude other non-WP related paths that otherwise they might want to include in the repository. IMHO, a **declarative approach** of all the paths and files one needs to ignore is the best way to go.

### Proposal/discussion for a WordPress gitignore

This PR adds my proposal for a gitignore for WordPress alongside with a couple questions up for discussion.

But before those, I want to frame my approach.

**Objectives of this gitignore**

Since Git is mostly useful for tracking code, I feel that the uploads folder, usually full of images/binaries should be ignored.

One might have other folders with binaries on their project/workflows (caches, etc), but I consider these files to be starters/base files from which you can expand your own gitignores, so if there are more of these folders on `wp-content` you should maintain them on your own gitignore.

**Q1 — Commit WP Core or not?**

While for some this might be an obvious response, I'm still not convinced of either approach as a base, as I can see both being useful:

- **Committing core.** **Pros:** to have a code-ready deploy of one site and, help lock WP version for a specific site. **Cons:** Committing plugins and WP core code can make repo out-of-sync when one instance is upgraded through Admin, additional repo size.

- **Keeping core out.** **Pros:** smaller repo, WP updates don't need to be committed to have consistency. **Cons:** additional setup for dev workflow, deploys, etc. **Also,** if we keep WP core out shouldn't we also keep third-party plugins and/or parent-themes out of repo too? (Q3)

My proposal ignores WP core, but has a note for those that want to commit it, they can do so by deleting/commenting one section of the gitignore. But I'm still not convinced of which approach this gitignore should take. (Before #2972 the approach was committing core.)

**Q2 — Catch-all or declarative approach?**

Here, I think a **declarative approach** to be the best. Catch-all has a big risk of conflicting with other gitignore files of this repo if you mix several together, as I explained before/above.

**Q3 — Should we allow all themes and plugins, or ignore all and whitelist our own?**

The gist of this proposed gitignore and all the others mentioned before is to **allow all the themes and plugins by default**.

But this can lead to having 3rd party plugins and parent custom themes—that aren't our custom code and of similar nature of WP core code—being added and tracked. Shouldn't these also to be ignored if we go with having the WP core ignored? If so, wasn't better to ignore all themes and plugins, and have each one **whitelist** their own custom themes and plugins on their gitignore files (like [Sal Ferrarello one](https://salferrarello.com/wordpress-gitignore))?

This proposal **continues to allow all themes and plugins by default**, but adds lines one can uncomment if they want to follow an whitelist approach per above.

### Final words

I know we already have several open PRs for the WordPress gitignore, but since we don't have Issues enabled on this repo, I felt it was the best way for me to contribute to the discussion around this gitignore.

### DIFF
--- a/WordPress.gitignore
+++ b/WordPress.gitignore
@@ -1,23 +1,29 @@
-# ignore everything in the root except the "wp-content" directory.
-!wp-content/
+# Wordpress - ignore core, configuration, examples, uploads and logs.
+# https://github.com/github/gitignore/blob/master/WordPress.gitignore
 
-# ignore everything in the "wp-content" directory, except:
-# "mu-plugins", "plugins", "themes" directory
-wp-content/*
-!wp-content/mu-plugins/
-!wp-content/plugins/
-!wp-content/themes/
+# Core and configuration
+/wp-admin/
+/wp-content/index.php
+/wp-content/plugins/index.php
+/wp-content/themes/index.php
+/wp-includes/
+/index.php
+/license.txt
+/readme.html
+/wp-*.php
+/xmlrpc.php
 
-# ignore these plugins
-wp-content/plugins/hello.php
+# Example themes
+/wp-content/themes/twenty*/
 
-# ignore specific themes
-wp-content/themes/twenty*/
+# Example plugin
+/wp-content/plugins/hello.php
 
-# ignore node dependency directories
-node_modules/
+# Uploads
+/wp-content/uploads/
 
-# ignore log files and databases
+# Log files
 *.log
-*.sql
-*.sqlite
+
+# htaccess
+/.htaccess
--- a/WordPress.gitignore
+++ b/WordPress.gitignore
@@ -7,6 +7,7 @@
 # you can delete this whole section/until Configuration.
 /wp-admin/
 /wp-content/index.php
+/wp-content/languages
 /wp-content/plugins/index.php
 /wp-content/themes/index.php
 /wp-includes/
@@ -33,3 +34,15 @@ wp-config.php
 
 # htaccess
 /.htaccess
+
+# All plugins
+#
+# Note: If you wish to whitelist plugins,
+# uncomment the next line
+#/wp-content/plugins
+
+# All themes
+#
+# Note: If you wish to whitelist themes,
+# uncomment the next line
+#/wp-content/themes
--- a/WordPress.gitignore
+++ b/WordPress.gitignore
@@ -1,7 +1,10 @@
 # Wordpress - ignore core, configuration, examples, uploads and logs.
 # https://github.com/github/gitignore/blob/master/WordPress.gitignore
 
-# Core and configuration
+# Core
+#
+# Note: if you want to stage/commit WP core files
+# you can delete this whole section/until Configuration.
 /wp-admin/
 /wp-content/index.php
 /wp-content/plugins/index.php
@@ -12,6 +15,9 @@
 /readme.html
 /wp-*.php
 /xmlrpc.php
+
+# Configuration
+wp-config.php
 
 # Example themes
 /wp-content/themes/twenty*/


shiftkey:
@pedrosanta thanks for the great writeup about the state of the template. I've skimmed the linked PRs and I get the consensus here that #2972 was too aggressive.

Two options to help us get back to a good place with this template:

 - merge this PR and then refine it with follow-up changes
 - use this PR as a chance to discuss the changes before merging

Happy to go with either option here, and I'll cc everyone who has been active in the linked PRs to get their feeling for the situation: @rohanvakharia @malithmcr @kolorafa @hdwebpros @chougron @samuraijane @NerdsvilleCEO @Triloworld @stkrzysiak

If we're in agreement with using this PR as a baseline, then I'll close out the other open PRs and funnel them into this PR.

rohanvakharia:
In general, WordPress plugins and themes create files and directories in distributed places. For example, they do create cache files/directories in the `wp-content` directory, but log files in the root directory. That is the precise reason I have excluded all the files and directories.

In the WordPress ecosystem users create following things custom made:
1. Plugins (`wp-content/plugins`)
2. Must use plugins (`wp-content/mu-plugins`)
3. Themes (`wp-content/themes`)

We must allow only the above three directories with a selective approach. The reason for being selective is to exclude the pre-installed plugins and themes. Besides, we must exclude log files, DB backup files, cache directories, etc which people keep dangling the root directory and other places.

I am not against the declarative approach, it is just that it will not work here because of the too many unknowns.

pedrosanta:
> In general, WordPress plugins and themes create files and directories in distributed places. For example, they do create cache files/directories in the `wp-content` directory, but log files in the root directory. That is the precise reason I have excluded all the files and directories.
> (...)
> We must allow only the above three directories with a selective approach. The reason for being selective is to exclude the pre-installed plugins and themes. Besides, we must exclude log files, DB backup files, cache directories, etc which people keep dangling the root directory and other places.
>
> I am not against the declarative approach, it is just that it will not work here because of the too many unknowns.

Like I said on the PR and previous comments, I have a strong feeling that this `.gitignore` should only take care of the **overly common files and overly common setups** and not delve into particular setups or try to do everything. For that one should use this `.gitignore` and add in/expand it with the rules specific to their own workflow or setup—for instance, I don't consider root `.sql` to be a common file of a typical WordPress installation. I think there might be a risk of forcing all the others/consider that all the others should adhere to our particular setup or rules, if we go that way.

(Perhaps such an approach might be something that fits better on the `community` folder, idk.)

About the declarative approach not working, because of the many unknowns, I think my reasoning above effectively makes this a not-problem, because I don't think we should be striving to address all the unknowns in the first place, I don't think that's a sensible approach. This `.gitignore` should only be a sensible default.

(Though if there's is a provenly hugely used cache plugin, I admit we could ignore that folder, etc.)

And, for the record, I do stress I think this `.gitignore` should definitely follow a **declarative** approach, as I've explained above/on the PR.

pedrosanta:
Also, and the more I think of it, the more I feel there are two major approaches to this (related to my Q1 and Q3).

I'm starting to get the sense that ignoring WP core and blanket allowing only all the `themes` and `plugins` (even 3rd party ones) with the purpose if ignoring common/downloadable files, is also valid for 3rd party plugins, which leads for one to whitelist themes and plugins to keep common/downloadable code at minimum (as I say on Q3), but then **we would have a `.gitignore` that ignores everything** and that one should add their themes and plugins before using, and that doesn't seem to make much sense or be clear.

If the above is the case, I think it probably makes more sense—an approach that a good number of people use too—**to have a repository for each custom theme and custom plugin**—and have the repos cloned inside WP file structure.

So, I feel there might be two **major** approaches to this:

- **Commit everything** (including WP Core) except for **`wp-content/uploads`** (binaries) and **`wp-config.php`**—this makes the repo basically closer to a deployable payload, but keeps all the core/3rd party plugins files tracked too (😕something to live with).

- Avoid tracking any common core code or 3rd party plugins by basically having **a repository for each custom theme and plugin**—on which a common `.gitignore` like this one is basically useless.

So, I'm not sure if we shouldn't model this `.gitignore` around the first approach above...

Triloworld:
I go for theme as starter, not solution to all problems.
This is for "**overly common files and overly common setups** "

What scenario is common and proper to be in this template for EVERYONE?
- user generated content
- settings with passwords
- cache in wp-content
- themes that are included by default and not the "twentytwenty" as WP recommend to be there
- plugins that are included by default ("akismet", "Hello Dolly")

99% just create plugins/themes so this is the case. 3rd-party must be case by case different.
Propose to add two optional (and by default commented) lines:
- wp-content/plugins/* #Uncomment to ignore plugins
- wp-content/themes/* #Uncomment to ignore themes

More? You can customize. After all, is starter and need to be for:
- Core team? No, they have and don't need this repo.
- Theme creators
- Plugins creators

More?
Create new file, why only one if there is different scenarios.
Main file is for any scenario. 
Propose if necessary:
- WordPress.core.gitignore
- WordPress.theme.gitignore
- WordPress.plugins.gitignore

This fix any discussion in this field as files name distinguish scenario. 
Any fix is in proper line and cover some use examples.
What you thinks?
@pedrosanta - this resolve to ``` Avoid tracking any common core code or 3rd party plugins ```

Triloworld:
In one line: 
What community encourage, to be in Git for production environment as proper start. 
Other thinks: Case by case. 3rd party plugins/themes isn't in our radar in this place. Just default Wordpress and how is properly used in GitHub

pedrosanta:
@Triloworld great contribution, thanks. Will follow through with your suggestions (those two plugin/themes lines and comments) and update the PR.

As for the multiple files that is certainly interesting. Perhaps we can keep a sensible default on root folder and add the other files to the `community` folder, not sure if that's entirely inline with the [contributions guidelines](https://github.com/github/gitignore#folder-structure) but I'll look at/think about it. But I would imagine that's better to go forward with those two lines and keep discussing the multi-files on `community` folder approach.

pedrosanta:
Also @Triloworld didn't get your take regarding ignoring WP core or not, etc—see [my comment on that](https://github.com/github/gitignore/pull/3295#issuecomment-586724282).

Based on that comment, what's your take about ignoring WP core or not?

If we went with multi-files approach perhaps default could have WP core committed/not-ignored and then you could add the `WordPress.core.gitignore` contents to it if you wish to ignore, etc.

Because, this list of things to ignore you make below, it's super on point, well-thought and consise, and you haven't enumerated WP core here, so I assume you're considering committing/not-ignoring WP core with this one, yes?

I like this scope/list and I think we should follow it, as it complies with all the requisites of a 'starter sensible default' for WordPress.

> What scenario is common and proper to be in this template for EVERYONE?
> 
> * user generated content
> * settings with passwords
> * cache in wp-content
> * examples that are for novice, not in production env:
>   -- themes that are included by default and not the "twentytwenty"
>   -- plugins that are included by default ("akismet", "Hello Dolly")

Triloworld:
Example **What I think about core files**:
```
# Core
/wp-admin/
/wp-content/index.php
/wp-content/plugins/index.php
/wp-content/themes/twentytwenty/ #Required theme
/wp-content/uploads/
/wp-content/upgrade/
/wp-includes/
/index.php
/license.txt
/readme.html
/wp-*.php
/xmlrpc.php

#Core - examples that can be safety removed
/wp-content/plugins/hello.php
/wp-content/plugins/akismet/
/wp-content/themes/twentyseventeen/
/wp-content/themes/twentynineteen/

```
My description what I ask myself about this file:
What You think when downloading this file? 
Ignore core file but like a list of places to be easily customizable - important than just ignore.
You need them? just add comment. You, as *me*, need VERIFIED list of core files? get this one template like it is and have comment to guide you.

My example:
I, about work on theme/plugin without touching core, important file. I mix with proper architecture (Apache, PHP, what on server) to have elegant and clean approach. After me somebody need to be sure that I don't touch important piece of WP system. You need that "security" feel as this file I add to Git. People tend to do mistake, this template must be list of predefined and verified directories/files as a good start to work on.

pedrosanta:
Hi @Triloworld just noticed these two suggestions you made were already built in:

> Propose to add two optional (and by default commented) lines:
> 
> * wp-content/plugins/* #Uncomment to ignore plugins

Here: https://github.com/github/gitignore/pull/3295/files#diff-f2cf96854b528e9828c4ca34a4917e54R42

> * wp-content/themes/* #Uncomment to ignore themes

And here: https://github.com/github/gitignore/pull/3295/files#diff-f2cf96854b528e9828c4ca34a4917e54R48

Like so?

Triloworld:
> Hi @Triloworld just noticed these two suggestions you made were already built in:
> 
> > Propose to add two optional (and by default commented) lines:
> > 
> > * wp-content/plugins/* #Uncomment to ignore plugins
> 
> Here: https://github.com/github/gitignore/pull/3295/files#diff-f2cf96854b528e9828c4ca34a4917e54R42
> 
> > * wp-content/themes/* #Uncomment to ignore themes
> 
> And here: https://github.com/github/gitignore/pull/3295/files#diff-f2cf96854b528e9828c4ca34a4917e54R48
> 
> Like so?
> 
> So if I understand correctly the PR goes along your position/suggestions, yes?

Yep. This is great but this will ignore all plugins/themes that wasn't great. By default Wordpress now recommend having one "twentytwenty" theme in dir. So this need to be revalidated

> Q1 — Commit WP Core or not?
No commit. Give line by line list of files.

> Q2 — Catch-all or declarative approach?
Declarative. Just to have starter, no "guessing"

> Q3 — Should we allow all themes and plugins, or ignore all and whitelist our own?
Ignore proper one so now is better to have only some commented as they are added by default:
```
/wp-content/themes/twentyseventeen/
/wp-content/themes/twentynineteen/
```

pedrosanta:
> Yep. This is great but this will ignore all plugins/themes that wasn't great. By default Wordpress now recommend having one "twentytwenty" theme in dir. So this need to be revalidated

You're absolutely right—site health freaks out when there isn't a default one installed, yes. I'll add an exclude for `twentytwenty`, like so:

```
# Example themes
/wp-content/themes/twenty*/
!/wp-content/themes/twentytwenty/
```

pedrosanta:
> Yep. This is great but this will ignore all plugins/themes that wasn't great. By default Wordpress now recommend having one "twentytwenty" theme in dir. So this need to be revalidated

Hi @Triloworld , but now that I come think of it, even if Health Check gives a warning, since this is an example theme, aren't we overdoing it by commiting `twentytwenty` code into the repository? Shouldn't we handle that outside of versioning control, as in, it's an issue we should handle on each instance, etc—making sure they have default themes? Or else we would be just committing `twentytwenty` on each update etc/having `twentytwenty` code changes on our history log?

Wdyt?

pedrosanta:
Any further thoughts/discussion on this?

pedrosanta:
@shiftkey , how to best proceed with this?

pedrosanta:
I really don't want to be 'that guy', but I think this repository (that depend a lot of popular tools, like for instance [toptal/gitignore.io](https://github.com/toptal/gitignore.io/pull/494)) should really have a working WordPress `.gitignore`. So, I don't know what's the best way to move forward with this, but I feel it nonetheless should be done, even if it means reverting this all the way back before the broken and controversial PR that updated this `.gitignore`, #2972, and sort it out afterwards.

@shiftkey any thoughts?

pedrosanta:
#3655 is another WordPress PR added to this repository. It's starting to be too many, meanwhile this issue goes on and on and on. @shiftkey let me know if I should close this because this is taking far too long.

bdougie:
This PR has more than enough documentation and links to justify this change. Thanks for your patience and contribution @pedrosanta

bdougie (approved):
LGTM

PR #3296, opened 2020-01-25
Add in global metals

ckipp01:

This is useful for Scala developers that are using Metals as a language server with various editors. 

You can on various places of our website that we recommend the following to be ignore. I've linked one of those places [here](https://scalameta.org/metals/docs/editors/vscode.html#gitignore-projectmetalssbt-metals-and-bloop).

If this is a new template:

[Homepage of Metals](https://scalameta.org/metals/)

https://github.com/scalameta/metals/issues/1336

### DIFF
--- a/Global/Metals.gitignore
+++ b/Global/Metals.gitignore
@@ -1,0 +1,5 @@
+ # Generated Metals (Scala Language Server) files
+ # Reference: https://scalameta.org/metals/
+.metals/
+.bloop/
+project/metals.sbt


Global/Metals.gitignore
```diff
@@ -0,0 +1,4 @@
+# Metals specific
```
shiftkey: Is it worth writing a bit in the header here about what Metals is, why it's relevant, and linking to the project page? These templates are consumed in many unexpected places, and this context often helps users understand the template when they stumble upon it and they're far removed from the original PR.

The example that I keep pointing people too is probably `JetBrains.gitignore`:

https://github.com/github/gitignore/blob/7b5783ace752279104f7bf7a812c6f6225f30cfd/Global/JetBrains.gitignore#L1-L4

ckipp01: Sure, I've added in a short description and a link to the Metals website.

PR #3297, opened 2020-02-01
Update Unity.gitignore

danielheim:
Unity 2020 introduces a new "UserSettings" top-level project folder. This folder should be ignored from version control, as it will contain per-user configuration files.
See https://forum.unity.com/threads/whats-the-usersettings-directory.754436/ for more information.

### DIFF
--- a/Unity.gitignore
+++ b/Unity.gitignore
@@ -8,6 +8,7 @@
 /[Bb]uild/
 /[Bb]uilds/
 /[Ll]ogs/
+/[Uu]ser[Ss]ettings/
 
 # MemoryCaptures can get excessive in size.
 # They also could contain extremely sensitive data


shiftkey:
Quoting the relevant bits from that linked discussion:

> I noticed Unity creates an "UserSettings" directory next to "ProjectSettings". This seems to be new in 2020.1.
>
> But, parts of editor settings were in Library/EditorUserSettings.asset file (things like user's perforce connection settings etc.). So now it's in that new location. The folder should not be put into version control, since it's per-user things there.

@danielheim thanks for the contribution!

danielheim:
No probs 😊

paulsinnett:
When does this get integrated into the latest version of Github Desktop? I have Version 2.9.4 (x64) but it still doesn't include these changes from last year.

PR #3302, opened 2020-02-08
Add Umbraco V8 packages folder location

Mivaweb:
In Umbraco v8 we have a new packages folder located under Umbraco/views/packages/...

This gets ignored by the current version. Adding this new line prevents this and includes all files and subfolders under this new location.

### DIFF
--- a/Umbraco.gitignore
+++ b/Umbraco.gitignore
@@ -18,6 +18,7 @@
 # Make sure to include details from VisualStudio.gitignore BEFORE this
 !**/App_Data/[Pp]ackages/*
 !**/[Uu]mbraco/[Dd]eveloper/[Pp]ackages/*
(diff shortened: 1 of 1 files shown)


shiftkey:
@Mivaweb any links to StackOverflow discussions or other documentation about this change?

Mivaweb:
@shiftkey The only documentation I can find is in the source code of Umbraco itself. See the repo of the packages folder on the new location:

https://github.com/umbraco/Umbraco-CMS/tree/v8/contrib/src/Umbraco.Web.UI.Client/src/views/packages

Is this enough or do you need some more info?

Mivaweb:
I found a thread on the Umbraco forum where I have added a new post now pointing to this pull request with the required change so others can see this and report back if correct.

https://our.umbraco.com/forum/using-umbraco-and-getting-started/98159-umbraco-8-gitignore-file

PR #3303, opened 2020-02-08
Ignore .idea/ jetbrains rider cache folder

gustavopsantos:
The JetBrains Rider IDE creates a cache folder named as `.idea`, it should be ignored.

### DIFF
--- a/Unity.gitignore
+++ b/Unity.gitignore
@@ -25,6 +25,9 @@
 # Visual Studio cache directory
 .vs/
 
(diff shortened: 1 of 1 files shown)


shiftkey:
@gustavopsantos thanks for the contribution, but we maintain a [`Global/JetBrains.gitignore`](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore) template that outlines what things can be shared when working with the JetBrains suite of IDEs.

PR #3304, opened 2020-02-08
Added ignore for blob crash files

thechubbypanda:

I'm sure nobody wants crash reports in their repos :ok_hand:

### DIFF
--- a/Godot.gitignore
+++ b/Godot.gitignore
@@ -10,3 +10,4 @@ export_presets.cfg
 # Mono-specific ignores
 .mono/
 data_*/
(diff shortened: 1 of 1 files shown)


shiftkey:
@thechubbypanda any links to StackOverflow discussions or other documentation that explain why this change is a safe one to make?

thechubbypanda:
I cannot find any such documentation sadly.
/blob files contain mono crash reports and nothing else from what I can tell. I found no problems with my Godot project after adding the suggested line to my gitignore.

PR #3307, opened 2020-02-13
Add ignore to umbraco folder

bakkiung:
umbraco folder should be git ignore. 

The content indside this folder should be added by Install-Package UmbracoCms or updated by Update-Package UmbracoCms instead of the developers who is using the CMS. 

https://our.umbraco.com/documentation/getting-started/setup/upgrading/general

### DIFF
--- a/Umbraco.gitignore
+++ b/Umbraco.gitignore
@@ -14,6 +14,9 @@
 # Ignore Umbraco content cache file
 **/App_Data/umbraco.config
 
+## git ignore this folder as it will be added by Install-Package UmbracoCms or updated by Update-Package UmbracoCms
+**/[Uu]mbraco/ 
+
 # Don't ignore Umbraco packages (VisualStudio.gitignore mistakes this for a NuGet packages folder)
 # Make sure to include details from VisualStudio.gitignore BEFORE this
 !**/App_Data/[Pp]ackages/*
--- a/Umbraco.gitignore
+++ b/Umbraco.gitignore
@@ -14,10 +14,12 @@
 # Ignore Umbraco content cache file
 **/App_Data/umbraco.config
 
-## this umbraco folder should be created by cmd like `Install-Package UmbracoCms -Version 8.5.3` 
+## this [Uu]mbraco/ folder should be created by cmd like `Install-Package UmbracoCms -Version 8.5.3` 
 ## you can find your umbraco version at your Web.config. (i.e. <add key="Umbraco.Core.ConfigurationStatus" value="8.5.3" />)
-## or you can remove this line if you think it fits the way you work on your project.
-**/[Uu]mbraco/ 
+## Uncomment this line if you think it fits the way you work on your project.
+## **/[Uu]mbraco/ 
+
+
 
 # Don't ignore Umbraco packages (VisualStudio.gitignore mistakes this for a NuGet packages folder)
 # Make sure to include details from VisualStudio.gitignore BEFORE this
--- a/Umbraco.gitignore
+++ b/Umbraco.gitignore
@@ -14,7 +14,7 @@
 # Ignore Umbraco content cache file
 **/App_Data/umbraco.config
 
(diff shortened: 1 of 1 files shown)


shiftkey:
Tentative :thumbsup: here but I'd like a confirmation from others who may be familiar with this.

cc @Mivaweb @mbukejlovic @richarth who have contributed previously to this template

Mivaweb:
Normally its ok to exclude this `Umbraco` folder. But in Umbraco v7+ you also have the `Umbraco_client` folderwhat about that one? In v8 its only the `Umbraco` folder, the `Umbraco_client` has been removed. Some people exclude it, others not. Maybe we could create a poll on twitter for this.

Mivaweb:
I have created a poll on twitter for 1 day, will reply with the results and answers about this thread.

richarth:
It's true that the folder can be created by installing the NuGet package and updated with the update package command.

I'm not sure if you open a solution without the Umbraco folder whether the folder would get recreated automatically?  If it isn't recreated automatically then a team mate would have to run a NuGet command when first working on a project.  If you have to run the commands manually, this could potentially lead to a different version of Umbraco being installed which is probably not what you want.

In the agency I work for we deploy everything straight from Git, so we have to have this folder checked in.

In short I think this one depends on the way you work.  I'll let others have their say and we can go with the consensus.

Mivaweb:
The poll on twitter is also concluding that not everyone is ignoring the folder. So I would leave the ignore file like it is and let the users decide to ignore it or not.

shiftkey:
@bakkiung thoughts on the feedback raised above? ☝️

bakkiung:
Hi, @shiftkey, @Mivaweb and @richarth , thanks for the discussion.  Here are my thoughts:

1. The files insides the  **/[Uu]mbraco/ should be modified by Umbraco team thus we should not include in our repo as it will introduce more than 1k files to our repo which should not be modified by us. Including it is like including node_modules/ in your repo in which you should not.  You should get youre dependencies from 'npm install' so does you should get your Umbraco from Install-Package UmbracoCms. Maybe there are some special cases you want to break the best practices and do want to include it in your repo. However, I think this git ignore is to help developers to deal with common cases. You can free feel to follow it or not if it does not suit your special cases. 

2.  I admit that if we do not include it in the repo, then when other developers check out the repo, they do need to run NuGet command to get that.  But I do not think "this could potentially lead to a different version of Umbraco being installed which is probably not what you want."  Since you can do "Install-Package UmbracoCms -Version 8.5.3"  in which you can find your Umbraco version easily in your Web.config. Besides, I noticed that downloading from NuGet is faster than checking them out from the repo in terms of 1k+ files.

3.  I also agree that "let the users decide to ignore it or not." However, I think it will be better if we can add something like this below to be more informative to let the users decide to ignore it or not :

## this folder should be added via Install-Package UmbracoCms or you can remove this line if you think it fits the way you work on your project.
**/[Uu]mbraco/ 

Otherwise, it may be a missing piece of information for developers who are new to this and want to look up some useful info regarding git ignore in Umbraco.

 
Thanks,

Adrian

richarth:
I agree, if your systems/processes allow you to not check in the umbraco folder then it's best not to.

I like adding the comment. I think it might be worth adding the version parameter to it though and adding instructions on how to find the version number. This would help those new to Umbraco to not end up on the wrong version.

bakkiung:
@richarth , thanks for the advice and I've updated my pull request as you suggested.

shiftkey:
The only extra bit of advice I have around this is whether it should be ignored by default, or commented out. If there are downsides to having this enabled by default, i.e. the user should "opt-in" to this setting rather than have it applied for everyone, I recommend commenting it out.

bakkiung:
Hi, @shiftkey , the pull request is updated as you recommended. 

thanks,
Adrian

shiftkey:
@bakkiung thanks!

PR #3311, opened 2020-02-18
Add Unity's Addressables and Android's temp files.

JasperCiti:

Current Unity.gitignore does not compenstate for Unity's new Addressables package and the auto-generated files by an Android platform.

Trail and error.

### DIFF
--- a/Unity.gitignore
+++ b/Unity.gitignore
@@ -62,8 +62,9 @@ sysinfo.txt
 # Crashlytics generated file
 crashlytics-build.properties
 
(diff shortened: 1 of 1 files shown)


Soren025:
Noticed the content state .bin files were ignored and tracked down the PR where that was added. Wanted to mention here that [Unity actually recommends including those in version control](https://docs.unity3d.com/Packages/com.unity.addressables@1.19/manual/ContentUpdateWorkflow.html#planning-for-content-updates).

PR #3314, opened 2020-02-19
fix(node): add .DS_Store

ameerthehacker:
.DS_Store is not ignored by default and it is bad for mac users


shiftkey:
@ameerthehacker thanks for the contribution, but our [`Global/macOS.gitignore`](https://github.com/github/gitignore/blob/master/Global/macOS.gitignore) covers this and plenty of other macOS-specific rules.

PR #3316, opened 2020-02-20
[Node] Use real conventions

frederikhors:
Use real conventions (from the creators of dotenv)! https://github.com/bkeepers/dotenv#what-other-env-files-can-i-use


bdougie:
Fixed merge conflicts. This is a convention used by modern JS frameworks as well.

latipun7:
uhh... what happen with this: https://github.com/motdotla/dotenv#should-i-have-multiple-env-files? From the creator of JS `dotenv` too, and I agree with him.

noinkling:
Why was plain `.env` removed? Isn't it better to err on the safe side? The cited conventions say "depends" for `.env`

Even then, that's the Ruby package. The guidance in the readme for the [Node package](https://github.com/motdotla/dotenv#faq) is different:

> ### Should I commit my `.env` file?
>
> No. We **strongly** recommend against committing your `.env` file to version
control.

> ### Should I have multiple `.env` files?
>
> No. We **strongly** recommend against having a "main" `.env` file and an "environment" `.env` file like `.env.test`.

People and frameworks use different conventions - .gitignore should reflect how things are in practice, it's not the place to impose a convention.

PR #3317, opened 2020-02-21
[Python] adding IDE & static related stuff

meysam81:

VSC should not track IDE generated files, nor should it track statics; so that's why this PR is created.

I couldn't find any official doc, but it was annoying to me so I thought it may be useful for others as well.

If this is a new template:

 - I don't know if it's gonna be of any use but I started the work from [here](https://github.com/meysam81/PyQuotes)


shiftkey:
> VSC should not track IDE generated files,

meysam81:
When `python manage.py collectstatic` is invoked, a static directory is created. So that's that.

Python.gitignore
```diff
@@ -133,3 +133,11 @@ dmypy.json
 
 # pytype static type analyzer
 .pytype/
+
+# IDE
+.vscode
+.idea
```
shiftkey: We have existing templates for [`Global/VisualStudioCode.gitignore`](https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore) and [`Global/JetBrains.gitignore`](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore). Please drop these rules.

```suggestion

```

Python.gitignore
```diff
@@ -133,3 +133,11 @@ dmypy.json
...
+
+# static
```
shiftkey: > When `python manage.py collectstatic` is invoked, a static directory is created. So that's that.

Given this is the first mention of adding these directories, it'd be great to hear that others are also encountering this and would benefit from the change. Are you sure there's no other resources or discussions about this out there?

Also, are we able to better document what this group of rules is about?

meysam81: Can't say for sure about the resources or discussions actually.

But about the documentation, I guess what we can say about them is this:

    static files generated from Django application using `collectstatic`

shiftkey: The other concern I have is that these names are rather broad (in particular `media`), and might inadvertently catch a user-created directory with the same name. Can these be opt-in (and commented out) or do you not feel this is a concern?

meysam81: Commenting them out might just arise the need for a developer to rewrite them at the end of the file as it might be the case that one might not read the whole `.gitignore` file; I know I don't.

And anyway, the name `media` is not usually used for a source code related stuff and is probably general enough for one to realize to not use.

shiftkey: Let's not forget this suggestion:

```suggestion
# static files generated from Django application using `collectstatic`
```

meysam81: yes it seems fine indeed

Honeybyte:
> nor should it track statics

Static also includes css and js files. Why shouldn't a VCS track them? By adding static to gitignore, app specific static folders are ignored which shouldn't happen. [`collectstatic`](https://docs.djangoproject.com/en/3.0/ref/contrib/staticfiles/#django-admin-collectstatic) will collect from these specific folders:

> Files are searched by using the enabled finders. The default is to look in all locations defined in STATICFILES_DIRS and in the 'static' directory of apps specified by the INSTALLED_APPS setting.

meysam81:
> > nor should it track statics
> 
> Static also includes css and js files. Why shouldn't a VCS track them? By adding static to gitignore, app specific static folders are ignored which shouldn't happen. [`collectstatic`](https://docs.djangoproject.com/en/3.0/ref/contrib/staticfiles/#django-admin-collectstatic) will collect from these specific folders:
> 
> > Files are searched by using the enabled finders. The default is to look in all locations defined in STATICFILES_DIRS and in the 'static' directory of apps specified by the INSTALLED_APPS setting.

You're right; we should consider css and js files. But I'd say that those static source code files are not (or better should not) be in the root of the project and therefore, probably a better solution is to ignore the root `static` and track the app related statics like the following:

```
static
!*/static
```

meysam81:
And if for any reason someone liked the idea of having the `static` in the root directory, there is always the option for that too: `git add --force`

Honeybyte:
Not sure about this either. As Django states in its [guide](https://docs.djangoproject.com/en/3.0/howto/static-files/#configuring-static-files):

> Your project will probably also have static assets that aren’t tied to a particular app. In addition to using a static/ directory inside your apps, you can define a list of directories (STATICFILES_DIRS) in your settings file where Django will also look for static files. For example:
`STATICFILES_DIRS = [
    os.path.join(BASE_DIR, "static"),
    '/var/www/static/',
]`

where `BASE_DIR` is normally the root directory. I think we should better not include static and media into the gitignore as both are really dependent on the project structure and configuration. I would rather define one rule more on my own in my gitignore instead of finding out that my changes haven't been tracked (as it happened to me 😉 ).

meysam81:
Fair enough.
But are you sure about `media`? Is it a good idea to track production site's binaries?

Honeybyte:
As [MEDIA_ROOT](https://docs.djangoproject.com/en/3.0/ref/settings/#std:setting-MEDIA_URL) doesn't have a default value set, I am not sure whether to assume `media` as a name. `uploads` would be rational as well. Additionally we don't work with VCS and commit in production environments (rather deploy to them) so we don't track production site binaries. The only reason I could think of why this rule could be useful is for testing the file upload feature locally right?

meysam81:
or maybe for something like favicon; hence 'production site binaries'

Honeybyte:
Don't these just count as static files too and thus be placed in `static`?

meysam81:
Yes you're right.
I think the `media` (or even `uploads`) would better be there then, or wouldn't you agree? I can't think of anything that will need the tracking of VCS in there; cause everything that'll end up there, is probably auto generated or even data that should be backed up using other means.

Honeybyte:
Yes that's right, auto generated stuff doesn't have to be tracked. But in case someone only has access to the result of generated stuff (e.g when using third party assets), they again would like to track them. Also I still would use `media` and `static` for their actual use cases, user uploaded files and static files. In case people want parts of static not be tracked, they can just add project specific gitignore rules :)

meysam81:
@shiftkey, maybe revert this commit then?

shiftkey:
@meysam81 done: #3363

PR #3318, opened 2020-02-21
Update Python.gitignore

dbrova15:
Adding the .idea/ directory

_TODO_

_TODO_

If this is a new template:

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -7,6 +7,7 @@ __pycache__/
 *.so
 
 # Distribution / packaging
+.idea/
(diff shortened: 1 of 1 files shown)


shiftkey:
@dbrova15 thanks for the contribution, but we maintain a [`Global/JetBrains.gitignore`](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore) template that covers the JetBrains suite of tools (including PyCharm).

BrianMSheldon:

PR #3321, opened 2020-02-25
Unity: Added Mac OS .DS_Store files (folder view cache)

ManeFunction:

Excluded file contains folder view for the Finder (folders positioning, sizes, grid, etc.) what is obviously personal settings.

https://en.wikipedia.org/wiki/.DS_Store (sorry for the wiki, but it's actually legit)

### DIFF
--- a/Unity.gitignore
+++ b/Unity.gitignore
@@ -62,3 +62,5 @@ sysinfo.txt
 # Crashlytics generated file
 crashlytics-build.properties
 
+# MacOS folder attributes cache
+.DS_Store


shiftkey:
@ManeFunction thanks for the contribution, but we maintain a separate macOS-specific template in [`Global/macOS.gitignore`](https://github.com/github/gitignore/blob/master/Global/macOS.gitignore) that we recommend using in your projects.

ManeFunction:
Thank you, I've missed it somehow...

PR #3322, opened 2020-02-27
[Qt] Extend Qt.gitignore with qmlcache qrc files

martonmiklos:

Projects using QML files generating a temporary qrc file if a qrc file present in the project.

image

### DIFF
--- a/Qt.gitignore
+++ b/Qt.gitignore
@@ -50,3 +50,5 @@ compile_commands.json
 
 # QtCreator local machine specific files for imported projects
 *creator.user*
+
+*_qmlcache.qrc


shiftkey:
@martonmiklos any links to StackOverflow discussions or other documentation about this change?

martonmiklos:
It's a generated file, I doubt that there are any discussions about it on StackOverflow or anywhere.

Sourcecode wise I think this is the source file which is responsible for it's naming:
https://github.com/therecipe/env_linux_amd64_513/blob/master/5.13.0/gcc_64/mkspecs/features/qtquickcompiler.prf#L12

For  indirect proof you could do a websearch * "_qmlchache.qrc" and you will other projects also adding the same pattern to their gitignore:
https://github.com/mavlink/qgroundcontrol/blob/master/.gitignore#L2

PR #3323, opened 2020-02-28
Include Rider as a supported IDE

bitbonk:
As far as I can tell this `.gitignore` also applies to Rider. Looking at the somewhat official `.gitignore` for Rider [here](https://github.com/JetBrains/resharper-rider-samples/blob/master/.gitignore), it seems compatible.

There should be an official and definitive `.gitignore` reference that applies to Rider too.

https://github.com/JetBrains/resharper-rider-samples/blob/master/.gitignore


cja-github:
The file is unsuitable for Rider as-is.

Firstly, according to https://rider-support.jetbrains.com/hc/en-us/articles/207097529, modules.xml and riderModule.iml should be excluded.

Also, there are significant differences between this file and https://github.com/JetBrains/resharper-rider-samples/blob/master/.gitignore with respect to Rider, primarily that in the latter all the lines referencing files in .idea/ start with "**/", for example "**/.idea/**/workspace.xml". This is important because a Rider project can contain nested .idea directories.

Either this file should be made compatible with Rider or Rider should be removed from the comment at the start of the file.

PR #3324, opened 2020-02-28
ReactJS

mr2rm:

Recently I need common gitignores to create and publish my own ReactJS projects. So I tried to find and collect some of them which can be usable for everyone and it can be updated during the time.

https://www.npmjs.com/
https://yarnpkg.com/
http://nodejs.org/api/addons.html
https://nodejs.org/api/modules.html
https://jspm.org/docs/guide
https://bower.io/docs/api/
https://eslint.org/
https://www.typescriptlang.org/
http://gruntjs.com/creating-plugins#storing-task-files
https://www.npmjs.com/package/jscoverage
https://tntim96.github.io/JSCover/
https://istanbul.js.org/
https://www.npmjs.com/package/node-waf

If this is a new template:

 - **Link to application or project’s homepage**: At the moment, there is no public or published application which I can mention it; but in the future it can be turned into a new template.

### DIFF
--- a/ReactJS.gitignore
+++ b/ReactJS.gitignore
@@ -1,40 +1,35 @@
-# Logs
-logs
-*.log
-npm-debug.log*
-yarn-debug.log*
-yarn-error.log*
-/.changelog
-
-# Runtime
-*.seed
-pids
-*.pid*
-
-# Testing (jscover, istanbul etc.)
-lib-cov
-coverage
-.nyc_output
-
-# Dependency directories (node, bower and etc.)
+# Dependency directories (node, bower, jspm and etc.)
 node_modules/
 bower_components/
 jspm_packages/
+
+# Package managers (yarn, npm and etc.)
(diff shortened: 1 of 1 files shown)


deemoding:
Maybe you can use node.ignore rather than add a new ignore file.

nickmccurdy:
I don't see any major advantages to this over using the existing Node gitignore or the one provided by create-react-app. Besides, the exact files you're using really depend on your build tooling.

bdougie:
This seems like a lot of duplicated work. Also, the majority of React starters and templates work off of [Node.gitignore](https://github.com/github/gitignore/blob/main/Node.gitignore). I don't see any reason to merge this, unless I am missing something.

bdougie:
Closing, since the Node.gitignore already ignores framework and library-specific files. Also a template for a Framework should be in the community folder per the guidelines.

PR #3325, opened 2020-03-03
add IntelliJ config

oxcz:
Java development mostly uses IntelliJ


shiftkey:
Thanks for the contribution, but we maintain a separate [`Global/JetBrains.gitignore`](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore) that covers IntelliJ. 

Please incorporate that into your projects.

PR #3326, opened 2020-03-04
[Global/JetBrains] Ignore AWS specific file with IDEA

dsrini-open:
AWS plugin generates AWS specific files which are usually not needed in the repo.

### DIFF
--- a/Global/JetBrains.gitignore
+++ b/Global/JetBrains.gitignore
@@ -8,6 +8,9 @@
 .idea/**/dictionaries
 .idea/**/shelf
 
+# AWS User-specific
+.idea/**/aws.xml
+
 # Generated files
 .idea/**/contentModel.xml
 


shiftkey:
@dsrini-open any links to StackOverflow discussions or other documentation about this change?

dsrini-open:
Unfortunately, no at this point as I don't find any. 

- But, the file is generated by IntelliJ Idea's plugin - AWS toolkit in the .idea folder locations to store the AWS configuration used for testing applications using the plugin. 

- This config file is particular to User based on their [AWS config file](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html#cli-configure-files-where) (usually saved at `~/home/<user_name>/.aws` and it is not common for a group or all git users.

Some pointers - 
[Sample file](https://gist.github.com/dsrini-open/2467e82ccf593b7516b4adb1cf627901)
[Source code pointing how the file is used by the plugin](https://github.com/aws/aws-toolkit-jetbrains/search?q=%22aws.xml%22&unscoped_q=%22aws.xml%22)

Do you recommend me to create an issue in the plugins repo or open a Q in stackoverflow for discussion?

shiftkey:
> Do you recommend me to create an issue in the plugins repo or open a Q in stackoverflow for discussion?

I think a StackOverflow question would be a better bet, and might get a faster answer.

dsrini-open:
Created - [Q](https://stackoverflow.com/questions/60713870/ignoring-aws-plugin-configuration-file-for-jetbrains-intellij)
Please feel free to add your comments.

dsrini-open:
Got only one comment so far - 
[Q](https://stackoverflow.com/questions/60713870/ignoring-aws-plugin-configuration-file-for-jetbrains-intellij)

stepio:
**Question to the project's owners.**

Why don't you simply ignore the whole `.idea/` folder ? Reasons:
- the whole folder is for IntelliJ IDEA-specific project configurations: you either source control all of them, or none;
- extra lines just make `.gitignore` bulky and less readable.

drothmaler:
@stepio Because the `.idea` folder also contains both

- user specific settings/stuff, that you ALWAYS want to ignore, AND
- project settings, that you often DO want to share. E. g. Code Style settings...

Please see the [official Jetbrains documentation](https://intellij-support.jetbrains.com/hc/en-us/articles/206544839), references at the top of the file.

So the choice is basically, either use this gitignore or ignore the whole `.idea` folder - but adding the whole folder is never an option (at least not, if you are working at a project together with a team)!

stepio:
thanks @drothmaler, this makes sense now.

martinwoodward:
LGTM

PR #3328, opened 2020-03-05
[VisualStudio] Add rule for ScaffoldingReadMe.txt

plr108:
Add rule to ignore the auto-generated ScaffoldingReadMe.txt discussed here: 
https://docs.microsoft.com/en-us/aspnet/core/security/authentication/scaffold-identity

_TODO_

_TODO_

If this is a new template:

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -60,6 +60,7 @@ BenchmarkDotNet.Artifacts/
 project.lock.json
 project.fragment.lock.json
 artifacts/
+**/ScaffoldingReadMe.txt
 
 # StyleCop
 StyleCopReport.xml
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -60,6 +60,8 @@ BenchmarkDotNet.Artifacts/
 project.lock.json
 project.fragment.lock.json
 artifacts/
+
+# ASP.NET Scaffolding
 ScaffoldingReadMe.txt
 
 # StyleCop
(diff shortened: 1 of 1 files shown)


VisualStudio.gitignore
```diff
@@ -60,6 +60,7 @@ BenchmarkDotNet.Artifacts/
 project.lock.json
 project.fragment.lock.json
 artifacts/
+**/ScaffoldingReadMe.txt
```
shiftkey: Two suggestions:

First, I don't think we need the `**` here unless we get into more complex expressions using directories - this can just be the file name itself:

```suggestion
ScaffoldingReadMe.txt
```

Second, can we break this out into it's own section with a comment about what this is for?

The project related to this file seems to be https://github.com/aspnet/Scaffolding and it seems to be just writing to the directory root.

https://github.com/aspnet/Scaffolding/blob/f409eb883a60aa0399c7dcf8540aabe45f7cfb61/src/VS.Web.CG.Mvc/Identity/IdentityGeneratorFilesConfig.cs#L38-L46

```c#
        private static IdentityGeneratorFile ReadMe = new IdentityGeneratorFile()
        {
            Name = "ScaffoldingReadme",
            SourcePath = "ScaffoldingReadme.cshtml",
            OutputPath = $"./{Constants.ReadMeOutputFileName}",
            IsTemplate = false,
            ShouldOverWrite = OverWriteCondition.Always,
            ShowInListFiles = false
        };
```

plr108: Sounds good to me!

PR #3331, opened 2020-03-07
update parcel cache

SimonSiefke:

In parcel v2 the cache directory is named `.parcel-cache`.

https://github.com/parcel-bundler/parcel/blob/v2/packages/core/core/src/resolveOptions.js#L15


shiftkey:
@SimonSiefke thanks for the contribution!

PR #3332, opened 2020-03-08
Improved JENKINS_HOME example

samrocketman:

After years of use I've come up with some improvements to the `JENKINS_HOME.gitignore` example.

- Major performance improvement: On very large Jenkins installations that have been running for more than one year, there tends to be many builds (hundreds of thousands of builds).  The `builds` directory of these jobs contain millions of files which would cause Git to hang for several minutes on simple commands like `git status` and longer for committing changes.  `strace` was used on Git to figure out the performance impact and this proposed change includes the optimization.  I also added a clear comment explaining the line's purpose.
- There's an example for how to include Jenkins encryption keys, and there's a disclaimer informing the user why they shouldn't but still giving an example.
- Comments have been reworded and slightly reformatted to be a little more clear.

No docs, this change comes from real world experience using the template as the original author across several companies.  I included a thorough commit message documenting reasoning for changes.

### DIFF
--- a/JENKINS_HOME.gitignore
+++ b/JENKINS_HOME.gitignore
@@ -1,25 +1,50 @@
-#Learn more about Jenkins and JENKINS_HOME directory for which this file is intended.
+# Learn more about Jenkins and JENKINS_HOME directory for which this file is
+# intended.
+#
 #  http://jenkins-ci.org/
 #  https://wiki.jenkins-ci.org/display/JENKINS/Administering+Jenkins
+#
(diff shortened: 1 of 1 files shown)


samrocketman:
@shiftkey this pull requests builds on my original contribution #1763 as an enhancement.  It doesn't fundamentally change the configuration purpose.  It just includes some performance fixes and more commented out examples for people with varying needs.  i.e. those who do not care about the secrets being backed up with their configuration.  The default is still to be secure.

PR #3340, opened 2020-03-17
[Global/VisualStudioCode] Directories does not require expansion with *

davivcgarcia:

According Git documentation [1], asterisks (`*`) on folders are not necessary. Actually it causes the opposite, not matching the folder but only sub-folders.

[1]  -https://git-scm.com/docs/gitignore

### DIFF
--- a/Global/VisualStudioCode.gitignore
+++ b/Global/VisualStudioCode.gitignore
@@ -1,4 +1,4 @@
-.vscode/*
+.vscode/
 !.vscode/settings.json
 !.vscode/tasks.json
 !.vscode/launch.json


shiftkey:
@davivcgarcia thanks for opening this up!

I believe this is necessary because of the explicit inclusions after this line:

```
$ cat .gitignore
# ignore every path inside .vscode
.vscode/*
# but explicitly track these files
!.vscode/settings.json
!.vscode/tasks.json
!.vscode/launch.json
$ git status --untracked-files=all --porcelain
?? .gitignore
?? .vscode/launch.json
?? .vscode/settings.json
?? .vscode/tasks.json
```

If we ignore the entire directory (rather than all paths within the directory) we invalidate the explicit inclusions:

```
$ cat .gitignore
# ignore .vscode directory
.vscode/
# but explicitly track these files
!.vscode/settings.json
!.vscode/tasks.json
!.vscode/launch.json
$ git status --untracked-files=all --porcelain
?? .gitignore
```

Does that help provide context about why the `*` is necessary in this situation?

davivcgarcia:
@shiftkey Got it! I thought the idea was to ignore all `.vscode` content. Thanks!

PR #3345, opened 2020-03-23
added Sublime Text project extensions

afcuttin:

_TODO_

_TODO_

If this is a new template:


shiftkey:
@afcuttin thanks for the contribution, but we maintain a separate template for Sublime Text rules that covers scenarios beyond this change: [`Global/SublimeText.gitignore`](https://github.com/github/gitignore/blob/master/Global/SublimeText.gitignore).

PR #3346, opened 2020-03-24
added pycharm .idea and .iml files to python git ignore

michalslowikowski00:
There are missing files for PyCharm IDE: `.idea` and `*.iml`.

I couldn't find docs about why I should ignore these files, but here are two links with discussion about it. 
https://intellij-support.jetbrains.com/hc/en-us/community/posts/115000154070-How-to-gitignore-idea-files
https://github.com/intellij-rust/intellij-rust/issues/1570

If this is a new template:

https://www.jetbrains.com/pycharm/


mik-laj:
We maintain a [Global/JetBrains.gitignore](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore) template here that covers everything you should need.

potiuk:
But maybe it's worth to add it to assume that a lot of people want to use it ? there is no harm in it.

mik-laj:
I don't manage this project, so I don't have an opinion.Here is similar PR about Python + VSCode: https://github.com/github/gitignore/pull/3317

turbaszek:
I think it would be good to add it also to Golang .gitignore

drothmaler:
The problem is, once you go down that road, you would have to add it to many more files... 

- WebStorm -> `Node.gitignore`
- Rider -> `VisualStudio.gitignore` (that is actually more a `DotNet.gitignore`)
- CLion -> `C.gitignore`, `C++.gitignore`, and probably more languages like `Rust` or `D`
- Ruby -> `Ruby.gitignore`
- AppCode -> `Swift.gitignore`, `Objective-C.gitignore`
- GoLand -> `Go.gitignore`
- PhpStorm -> to alle the PHP related gitignores (probably `CakePHP`, `Composer`, `FuelPHP` and many more...)
- IntelliJ -> `Java.gitignore` - and all the other JVM-Languages...

And why not also add `VSCode`, or `<nextGreatIdeOrEditor>` ignores to all the files - because everybody uses it these day...? Same goes for the Operating System gitignore files...  
But who is going to keep all of those in sync, when the new version of the tool is released?

... As you can see, just because 'a lot of people want to use it', doesn't mean it's a good idea, to mix it all together.

This repos is modular. There are small, maintainable gitignore files for Tools, Languages, BuildSystem, Editors, IDEs, OperatingSystems and probably more. And you can combine them, to tailor a custom gitignore specific for all your Projects needs (e.g. using [gitignore.io](http://gitignore.io)).  
I really like this principle, so I will continue to vote agains all the LetsCombineAllThingsTogether-PRs, popping up from time to time - because they will make it harder for us all in the long run...

potiuk:
So why don't we describe this principle and link to gihub.io prominently in the README ? That would save a lot of hassle and there will be far less LetsCombineAllThingsTogether-PRs, I created PR for that https://github.com/github/gitignore/pull/3348

shiftkey:
Thanks everyone for the discussion on this. I've been meaning to rewrite the README for a while to address some things but it keeps getting bumped down my priority queue. Hearing this context from everyone helps me to understand what I need to communicate to everyone.

I'm going to close out this PR given we have #3348 open for improving the README and #3349 to improve how the JetBrains products are listed here, stay tuned!

BrianMSheldon:

PR #3351, opened 2020-03-26
Visual Studio. Ignore ".orig" backup of merge conflict files

alfonso-martin-csa:

As mentioned in issue #546, when using Visual Studio with Fork (https://fork.dev) and kdiff3 (http://kdiff3.sourceforge.net), this files remain in version control.

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -355,3 +355,6 @@ MigrationBackup/
 
 # Ionide (cross platform F# VS Code tools) working folder
 .ionide/
+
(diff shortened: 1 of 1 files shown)


shiftkey:
@alfonso-martin-csa thanks for the contribution, but `*.orig` is not an extension specific to Visual Studio.

Please check out [`Global/Backup.gitignore`](https://github.com/github/gitignore/blob/master/Global/Backup.gitignore) as that's where we keep our curated files about backups.

PR #3353, opened 2020-03-27
Update Python.gitignore

wind-zrf:
add pyc

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -61,6 +61,7 @@ cover/
 local_settings.py
 db.sqlite3
 db.sqlite3-journal
+*.pyc
 
 # Flask stuff:
 instance/


Python.gitignore
```diff
@@ -61,6 +61,7 @@ cover/
 local_settings.py
 db.sqlite3
 db.sqlite3-journal
+*.pyc
```
shiftkey: This should be covered by line 3

https://github.com/github/gitignore/blob/8170c0ed25ecf564389332896f44b5aec359282f/Python.gitignore#L3

shiftkey:
@wind-zrf can you confirm that the existing rule works for you?

Python.gitignore
michaelotty: Agreed, this is covered in line 3 according to the git docs shown below, and works for me at ignoring .pyc files.

> [An asterisk "*" matches anything except a slash. The character "?" matches any one character except "/". The range notation, e.g. [a-zA-Z], can be used to match one of the characters in a range. See fnmatch(3) and the FNM_PATHNAME flag for a more detailed description.](https://git-scm.com/docs/gitignore)

pradyunsg:
I'm pretty sure this can be closed. :)

shiftkey:
Closing this out due to inactivity and unresolved questions.

PR #3355, opened 2020-03-27
Added file extensions when glossaries is loaded with 'symbols' option

afcuttin:

files created when the `glossaries` package is loaded with the `symbols` option (see section 2.6 of the manual)

_TODO_

> **2.6 Other Options**
> Other available options that don’t fit any of the above categories are:
> `symbols` This option defines a new glossary type with the label symbols via
> ```\newglossary[slg]{symbols}{sls}{slo}{\glssymbolsgroupname}```

_TODO_

### DIFF
--- a/TeX.gitignore
+++ b/TeX.gitignore
@@ -111,6 +111,9 @@ acs-*.bib
 *.glsdefs
 *.lzo
 *.lzs
+*.slg
+*.slo
+*.sls
 
 # uncomment this for glossaries-extra (will ignore makeindex's style files!)
 # *.ist
--- a/TeX.gitignore
+++ b/TeX.gitignore
@@ -279,3 +279,7 @@ TSWLatexianTemp*
 # option is specified. Footnotes are the stored in a file with suffix Notes.bib.
 # Uncomment the next line to have this generated file ignored.
 #*Notes.bib
+
+# Sublime Text
+# *.sublime-project
+*.sublime-workspace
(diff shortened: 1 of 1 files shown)


TeX.gitignore
```diff
@@ -279,3 +282,7 @@ TSWLatexianTemp*
 # option is specified. Footnotes are the stored in a file with suffix Notes.bib.
 # Uncomment the next line to have this generated file ignored.
 #*Notes.bib
+
+# Sublime Text
```
shiftkey: We maintain a separate [`Global/SublimeText.gitignore`](https://github.com/github/gitignore/blob/master/Global/SublimeText.gitignore) template containing these rules. Please drop these changes from the template.

TeX.gitignore
```diff
@@ -111,9 +111,12 @@ acs-*.bib
 *.glsdefs
 *.lzo
 *.lzs
+*.slg
+*.slo
+*.sls
 
 # uncomment this for glossaries-extra (will ignore makeindex's style files!)
-# *.ist
+*.ist
```
shiftkey: Do we need to enable this rule by default? The comment above indicates it'll now break another extension, so I want to understand this change a bit better and ensure it's correctly documented...

TeX.gitignore
afcuttin: will close this pull request and submit a new one

PR #3356, opened 2020-03-28
added file extensions when glossaries is loaded with "symbols" option

afcuttin:

Files created when the `glossaries` package is loaded with the `symbols` option (see section 2.6 of the manual)

_TODO_

From the [`glossaries`](http://ctan.mirror.garr.it/mirrors/CTAN/macros/latex/contrib/glossaries/glossaries-user.pdf) manual:

> **2.6 Other Options**
> [...]
> Other available options that don’t fit any of the above categories are:
> `symbols` This option defines a new glossary type with the label symbols via
> ```\newglossary[slg]{symbols}{sls}{slo}{\glssymbolsgroupname}```

_TODO_

### DIFF
--- a/TeX.gitignore
+++ b/TeX.gitignore
@@ -111,6 +111,9 @@ acs-*.bib
 *.glsdefs
 *.lzo
 *.lzs
+*.slg
+*.slo
+*.sls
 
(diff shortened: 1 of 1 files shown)


bdougie (approved):
LGTM

PR #3361, opened 2020-04-01
Add PyCharm/Jetbrains IDE project settings

Legorooj:

I use PyCharm, and always have to update my gitignore. It'd be useful not to have to do that for every repo.

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -141,3 +141,6 @@ cython_debug/
 # static files generated from Django application using `collectstatic`
 media
 static
+
+# PyCharm/JetBrains IDE project settings
+.idea/


drothmaler:
Well, [it's](https://github.com/github/gitignore/pull/3346) [groundhog](https://github.com/github/gitignore/pull/3318) [day](https://github.com/github/gitignore/pull/3284)... [again](https://github.com/github/gitignore/pull/3267).

Please use `Global/JetBrains.gitignore` instead. This repo tries to be modular. There are small, maintainable, separate ignore-files for each tool, language, framework, IDE and more. Please combine them according to your projects needs (you could also use [gitignore.io](https://gitignore.io) to help you with that).

Legorooj:
@drothmaler I believe that seeing as PyCharm is one of the top IDEs for python, it warrants its own entry in the gitignore - and Spyder has an entry. I'll close this, but I still think this is needed. (PS I already knew about both those PRs

BrianMSheldon:

PR #3362, opened 2020-04-02
Adding a gitignore for the Crystal language

arschles:

There is no current gitignore option for the [Crystal language](https://crystal-lang.org). 

The [`crystal init`](https://crystal-lang.org/reference/using_the_compiler/#crystal-init) command generates a `.gitignore` file, among others. Out of the box, it looks like the file I've added in this PR. I've copied that `.gitignore` source into here.

The source code for how the `crystal init` file is created is [here](https://github.com/crystal-lang/crystal/blob/a691846bb4b989962ba349a57535079652731a59/src/compiler/crystal/tools/init/template/gitignore.ecr).

If this is a new template:

 - **Link to application or project’s homepage**:  https://crystal-lang.org

### DIFF
--- a/Crystal.gitignore
+++ b/Crystal.gitignore
@@ -1,0 +1,5 @@
+/docs/
+/lib/
+/bin/
+/.shards/
+*.dwarf


hahwul:
Shards.lock is also included in the ignore when running with `crystal init lib`, so wouldn't it be good to include it as an comment in `.gitignore`?

e.g
```bash
/docs/
/lib/
/bin/
/.shards/
*.dwarf

# Write this if you create a library.
# /shard.lock
```

.gitignore from `crystal init lib`
```bash
# ...

# Libraries don't need dependency lock
# Dependencies will be locked in applications that use them
/shard.lock
```

Apart from this, I hope gitignore for crystal project will be applied soon 🙏🏼

PR #3364, opened 2020-04-04
Remove pip-wheel-metadata/ from Python.gitignore

pradyunsg:

pip generated this folder for a few versions, as part of it's initial
implementation of PEP 517.

pip has not generated this folder for a few versions now, so it should
be OK to remove this from the standard gitignore file.

https://pip.pypa.io/en/latest/news/ I guess?

<sub>If y'all deem this to be "too early" for some reason, I'd appreciate it if you suggest a timeline and keep this PR open till then.</sub>

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -20,7 +20,6 @@ parts/
 sdist/
 var/
 wheels/
-pip-wheel-metadata/
 share/python-wheels/
 *.egg-info/
 .installed.cfg


shiftkey:
> If y'all deem this to be "too early" for some reason, I'd appreciate it if you suggest a timeline and keep this PR open till then.

Knowing which versions are "good" and documenting them here would be great. Do you know when this was fixed?

pradyunsg:
This was fixed in pip 19.3. That's > 6 months ago.

> The pip-wheel-metadata directory does not need to persist between invocations of pip, use a temporary directory instead of the current setup.py directory.

pradyunsg:
A gentle nudge on this.

shiftkey:
> This was fixed in pip 19.3. That's > 6 months ago.

Great, thanks for the specificity!

pradyunsg:
Whee! Thanks @shiftkey! ^>^

PR #3366, opened 2020-04-06
Added Rider cache directory and Mac .DS_Store files.

valyard:
Ignoring JetBrains Rider specific files. These are per developer per platform, as well as not all developers use Rider for Unity C# code.
Ignoring folder specific attributes stored in .DS_Store on OSX.


bitbonk:
.idea is not a cache folder. It contains settings that may be shared with the team. Some files in there therefore need to be added to source control.

valyard:
I had a chat with Rider developers on this topic last week. Basically, the only things this folder contains you'd like to commit to git are:
1. Custom run configurations
2. In-editor VCS settings for custom issue tracker patterns and stuff like that
(1) is not used with Unity, (2) is rarely used.

I ran a poll in twitter on this topic https://twitter.com/valyard/status/1253373196042211328
_Those of you using @JetBrainsRider and Git to develop #unity3d projects: do you add  ".idea" folder to ".gitignore", commit the whole folder to the repo, or commit a part of it?_
Adding to .gitignore  **82.1%**
Committing to repo **2.4%**
Committing a part **15.5%**

Most of the Unity devs add the whole folder to .gitignore with every project. I am pretty sure the majority of the others just don't know what the .idea folder is about and commit it just in case.

I'd like to discuss this of course, but I am certain that if the vast majority of people can benefit from adding .idea folder to the default .gitignore — we should do it.

bitbonk:
We share the following settings from the .idea folder in many project company wide 
1. VCS settings
2. run configurations
3. database dialect settings
4. database connection setting 
4. JavaScript version settings 
5. HTTP scratch files 
6. information about external folders attached to the solution

All of the above can only be stored inside the .idea folder. Adding this folder to .gitignore will brake us and works against the the design of that folder.

Rider will automatically add and update separate .gitignore inside the .idea folder for all the stuff inside the .idea folder that should never be added to VCS. You can also add things there manually.

Rider also has a UI to add thing to .gitignore:
https://twitter.com/bitbonk/status/1252850414191427586?s=21

So generally speaking, adding something to .gitignore manually in a specific project is much easier than removing something from .gitignore that you don’t want.

In my opinion, only those things should be added to the common .gitignore baseline file that **100%** of the user never ever want to have in VCS.

citizenmatt:
I oppose adding the `.idea` folder to `.gitignore`.

Disclaimer, I'm a developer advocate at JetBrains, and work on the Unity plugin. It might even have been me that gave the advice that it's fine to do so, but this wasn't intended as advice for all Unity developers, more like confirmation that it won't break anything mission critical if you choose to do so. It is safe to ignore the `.idea` folder, but you're throwing the baby out with the bathwater here, and it is not the advice we want to give to ALL of our users.

As @bitbonk points out, there are lots of benefits to having properly shared settings, and for those that set it up, making this the default will break their workflow for new projects.

The official advice is this: let Rider handle it. Rider will create a `.idea/.gitignore` file for you, if you're not already ignoring the files. It will correctly exclude the files and folders that should be excluded, and you can commit everything else.

Secondary advice is this: if you don't want to let Rider handle this, use the [JetBrains.gitignore](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore) file, and add it to your `.gitignore`. This file is correct, albeit more complex than I would like.

It is very unfortunate that the `.idea` folder has evolved over the years and across many products into something where it is now very difficult to know if a file is per-user or shareable. Personally, I'm very frustrated by the situation, and would love to see a single, simple rule in `.gitignore`. The good news is that we also recognise this internally, and are trying to fix it. But it's like trying to turn a supertanker, with many things to juggle, such as features, products, plugins and compatibility/migration, so it might be a while before we get this nice and clean.

In the meantime, the recommended advice is either let Rider handle it for you, or include the correct patterns from `JetBrains.gitignore`.

valyard:
I don't know if @bitbonk does any development in Unity, because it is not correct to comment on the issue from a purely C# point of view. It is clear that there might be use cases when the `.idea` folder shouldn't be ignored but in the real world of Unity development it is just an extra step to add it to `.gitignore` manually or use Rider to do that. One way or another it shows up in repository sparkling those discussions alike "tabs vs. spaces" wasting everyone's time. I believe that if the majority of users have to manually ignore the folder it should be ignored by default. The approach that a minor amount of devs might want to use this feature so it has to stay this way making the life of the majority a little bit harder — this is what's dragging our society backward. Let it be then.

PR #3367, opened 2020-04-07
Add generated timing files to Coq.gitignore

JasonGross:

Ignoring more autogenerated files

These generated files were added in https://github.com/coq/coq/pull/745


drothmaler:
The timing files are also documented here:  
https://coq.inria.fr/refman/practical-tools/utilities.html#timing-targets-and-performance-testing

PR #3374, opened 2020-04-16
Added mono crash logs and dumps to ignored files

Mek101:
Removing otherwise huge files related to the local machine only from tracking.

### DIFF
--- a/Godot.gitignore
+++ b/Godot.gitignore
@@ -10,3 +10,5 @@ export_presets.cfg
 # Mono-specific ignores
 .mono/
 data_*/
+mono_crash.*.json
+*.blob


shiftkey:
@Mek101 any links to StackOverflow discussions or other documentation about this proposed change?

Mek101:
I don't have other complaints regarding mono and Godot specifically other than my personal experience, but I have the [documentation](https://www.mono-project.com/docs/debug+profile/debug/telemetry/) regarding why those blobs are generated (under "Normal Crash Reporting", step 5) and threads of users mentioning mono crashing and generating those blobs.

Godot.gitignore
```diff
@@ -10,3 +10,5 @@ export_presets.cfg
 # Mono-specific ignores
 .mono/
 data_*/
+mono_crash.*.json
+*.blob
```
iamacup, line 14: Change this to `mono_crash.*.blob`

iamacup:
Hi, i was just about to submit a similar PR for this. 

They are mono telemetry files (https://www.mono-project.com/docs/debug+profile/debug/telemetry/) which will always drop in the root of the project on Mono crash.

image

I have suggested an update as I don't think we want to blanked ignore all .blob files as, you could import one as part of your game project.

There is no 'official' documentation within the mono project about these files to point to as they really are related to the runtime.

iamacup:
Infact, i found the same ignore in https://github.com/github/gitignore/blob/master/VisualStudio.gitignore - it might be that we just copy that line over and not have two seperate things.

PR #3378, opened 2020-04-18
Ignore everything in root directory for WordPress.gitignore

loteoo:
I think that first line was missing in the .gitignore, am I wrong?

This first line in the current repository points to this ignoring everything in the root directory:
https://github.com/github/gitignore/blob/master/WordPress.gitignore#L1

Thanks

### DIFF
--- a/WordPress.gitignore
+++ b/WordPress.gitignore
@@ -1,3 +1,4 @@
+/*
 # ignore everything in the root except the "wp-content" directory.
 !wp-content/
 
--- a/WordPress.gitignore
+++ b/WordPress.gitignore
@@ -1,5 +1,5 @@
-/*
 # ignore everything in the root except the "wp-content" directory.
+/*
 !wp-content/
 
 # ignore everything in the "wp-content" directory, except:


LucasLarson:
> I think that first line was missing

`.gitignore` files don’t use `/* PHP-style comments */`

    

> am I wrong?

Probably. Your change uses valid `.gitignore` syntax, but not in the way I imagine you intended:
1. the leading `/` refers to the directory separator, and
2. the `*` matches everything except `/`, 

so adding this `.gitignore` would permit Git to ignore any file in the same directory whose name was a single character, but wouldn’t
> Ignore everything in root directory for WordPress.gitignore

loteoo:
@LucasLarson Thanks for the feedback!

I was not trying to add a php style comment, but trying to 
> Ignore everything in root directory for WordPress.gitignore

This configuration (with the modification I added) works great on my windows machine with gitbash, but maybe my machine is a edge-case.

Without it, all the files of WordPress code are commited to the repo. With it, only the plugin / theme files are commited (which is what I was looking for in my case)

LucasLarson:
I don’t think this is production-ready yet.

`.gitignore` is already built to ignore or skip everything in a directory unless otherwise specified. Adding the `/*` line to an existing `.gitignore` does nothing to your repo unless you also run `git rm -rf --cached`: “The files/folder in your version control will not just delete themselves just because you added them to the `.gitignore`” and if they’re “already in the repository” then “you have to remove them” yourself (via [Ahmad on StackOverflow](https://stackoverflow.com/a/25436481)).

If the user will already have to `git rm -rf --cached`, then there is no need to add this line, because those entries will already be skipped unless they’re in this `.gitignore`’s other rules.

tywmick:
> `.gitignore` is already built to ignore or skip everything in a directory unless otherwise specified.

I'm on Git 2.23.0, and using this `.gitignore` without a `/*` at the beginning, in a freshly downloaded WordPress folder, after `git init` and before any commits, VS Code tells me there are 1,694 pending changes:

.gitignore without /*

After adding `/*` (and `!.gitignore`), VS Code tells me there are 29 pending changes:

.gitignore with /*

Also, [`gitignore`'s documentation](https://git-scm.com/docs/gitignore) says that with the `!`, "any matching file excluded *by a previous pattern* will become included again", and their example at the end for excluding everything except a specific directory includes a `/*` at the top:

> Example to exclude everything except a specific directory `foo/bar` (note the `/*` - without the slash, the wildcard would also exclude everything within `foo/bar`):
>
> ```
> $ cat .gitignore
> # exclude everything except directory foo/bar
> /*
> !/foo
> /foo/*
> !/foo/bar
> ```

bdougie:

TyMick:
In favor of #3295, that is. 👌🏼

PR #3381, opened 2020-04-21
Idea creates .iml files on multi module projects

fotisp:
I believe *.iml are also part of the Intellij configuration strategies, and should be ignored when present in a module directory.

_TODO_

_TODO_

If this is a new template:

### DIFF
--- a/Global/JetBrains.gitignore
+++ b/Global/JetBrains.gitignore
@@ -69,3 +69,6 @@ fabric.properties
 
 # Android studio 3.1+ serialized cache file
 .idea/caches/build_file_checksums.ser
+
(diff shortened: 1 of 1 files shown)


shiftkey:
@fotisp thanks for the contribution. We have an existing section which is commented out by default that feels relevant here:

https://github.com/github/gitignore/blob/105ce45ee9b3ac4a9de282752737269b0adc0b93/Global/JetBrains.gitignore#L27-L38

Does this make sense? Can this be improved at all?

fotisp:
Oh sorry, didn't see that, really sorry

PR #3385, opened 2020-04-28
Add .vscode folder for Visual Studio Code users

ciro-unity:
Visual Studio Code is a popular IDE for Mac users who use Unity.
Unity's Package Manager includes a package to deal with the integration, which pushes more users to use it every day.

Package: https://docs.unity3d.com/Packages/com.unity.ide.vscode@1.0/manual/index.html
https://code.visualstudio.com/

### DIFF
--- a/Unity.gitignore
+++ b/Unity.gitignore
@@ -25,6 +25,7 @@
 
 # Visual Studio cache directory
 .vs/
+.vscode/
 
 # Gradle cache directory
 .gradle/


shiftkey:
@ciro-unity thanks for the contribution, but we have a separate [`Global/VisualStudioCode.gitignore`](https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore) that should cover your use case.

PR #3388, opened 2020-04-29
Fix ignoring of everything in the root folder (WordPress)

tywmick:

It appears `!` can only be used to create exclusions for explicit patterns on previous lines and can't infer "all items" when used at the beginning of a document.

Also we want to still commit `.gitignore` itself.

From [`gitignore`'s documentation](https://git-scm.com/docs/gitignore):

> An optional prefix "`!`" which negates the pattern; any matching file excluded by a previous pattern will become included again.

> Example to exclude everything except a specific directory `foo/bar` (note the `/*` - without the slash, the wildcard would also exclude everything within `foo/bar`):
>
> ```
> $ cat .gitignore
> # exclude everything except directory foo/bar
> /*
> !/foo
> /foo/*
> !/foo/bar
> ```

### DIFF
--- a/WordPress.gitignore
+++ b/WordPress.gitignore
@@ -1,4 +1,6 @@
 # ignore everything in the root except the "wp-content" directory.
+/*
+!.gitignore
 !wp-content/
 
 # ignore everything in the "wp-content" directory, except:


tywmick:
Whoops this is already being discussed several other places—I am embarrassed for not having checked first 🤦🏼‍♂️ I'll close this one and discuss on other PRs instead.

PR #3391, opened 2020-04-30
[add] empty gitignore

s1mpleTEK:
It is a new template:

I think that will be good to make a empty gitignore for people who use Github's api for create us own Github repository without any "template" (empty gitignore).


drothmaler:
When you create a new Git repo in GitHub, there is a „None“ option for „Add .gitignore:“ (actually the default). 

So if I create a repo I can either select one of the templates as a starting point, or just add no gitignore at all - and maybe add one later, if I need it. 

But why would I ever want to add an empty `.gitignore` to my repo? It just creates a useless commit... So it looks like [YAGNI](https://www.martinfowler.com/bliki/Yagni.html) to me... But maybe I‘m missing a point. Could you please further explain, why you‘d want to do that?

s1mpleTEK:
Why do I want to do this ? I'm in college, and my college has set up a new collection system for our projects on github. So I decided to create a unix shell script, to facilitate the creation of our repositories on github as well as on the old version of the collections that are still effective. 

I create the GitHub repositories from an unix shell script thanks to the GitHub api and I would like to make it as usable as possible for my comrades since it is not used case to be used for college but also for personal projects.

Since I do not know what language my comrades are going to work with, so I can not define the appropriate `.gitignore` that is why I make this request, to have an empty `.gitignore` that I can implement directly with the creation of the repository, without it having to do it.

That's why I make this request near this repository that manages all the database of the `"gitignore_template"` feature available with the GitHub api.

shiftkey:
> I create the GitHub repositories from an unix shell script thanks to the GitHub api 

I'm still thinking about this PR, but if you're looking for reuse with scripting there is the ability to create [repository templates](https://github.blog/2019-06-06-generate-new-repositories-with-repository-templates/) that represents _all_ files a new repository should contain. This can then be used to [create a repository based on a template](https://developer.github.com/v3/repos/#create-a-repository-using-a-template) through the GitHub API.

It's not clear to me whether your script is making a commit with additional files, but if you are then maybe the repository template is a more elegant solution to this problem.

s1mpleTEK:
Is it possible to make changes with a model ? Because I plan to add the possibility to choose the license that we want which I do not know in advance, and I'm afraid that to do with a template this blocks me later. Because, to put a license or a gitignore on a repository using GitHub APIs it is necessary to do the creation of the repository otherwise we will not be able to add it later.

PR #3392, opened 2020-04-30
Added Drop, Out to VisualStudio.gitignore

abatishchev:

Stands for Drop/drop.

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -32,6 +32,7 @@ bld/
 [Ll]og/
 [Ll]ogs/
 [Dd]rop/
+[Oo]ot/
 
 # Visual Studio 2015/2017 cache/options directory
 .vs/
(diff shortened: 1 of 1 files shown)


shiftkey:
@abatishchev any links to StackOverflow discussions or other documentation about this change?

abatishchev:
I think it should be both /out and /drop.

I don't have a discussion on Stack Overflow but internally at Microsoft it's a common practice to move the build drop folder to the root of the project and l these two folders are the most popular folder names for that.

abatishchev:
@shiftkey, ping.

shiftkey:
> I don't have a discussion on Stack Overflow but internally at Microsoft it's a common practice to move the build drop folder to the root of the project and l these two folders are the most popular folder names for that.

@abatishchev this is still rather unsatisfying for me, because if I were to merge this and someone asks why we made this change I don't have anything to point to beyond this quote. Does that help clarify my situation?

abatishchev:
Incredible level of collaboration and cooperation between Microsoft and GitHub

PR #3393, opened 2020-05-02
Update Android.gitignore

SaifurRahmanMohsin:
Updated google-services.json rule to have a wildcard prefix and also added the new google_maps_api file as it contains the development/production key when using maps API.

Uncommenting the rule doesn't exclude it on account of it being inside the app/ folder than the root directory (as clearly mentioned in the google documentation as well). So the wildcard has been appended. The google_maps_api rule is self explanatory.

https://developers.google.com/android/guides/google-services-plugin#adding_the_json_file
https://developers.google.com/maps/documentation/android-sdk/get-api-key#getting-the-certification-information-from-android-studio

### DIFF
--- a/Android.gitignore
+++ b/Android.gitignore
@@ -60,7 +60,8 @@ captures/
 .cxx/
 
 # Google Services (e.g. APIs or Firebase)
-# google-services.json
+# **/google-services.json
+# **/google_maps_api.xml
 
 # Freeline
 freeline.py


drothmaler:
+1 for adding `google_maps_api.xml`
But you should remove the leading `**/` from you rules. Those are note necessary. Gitignore rules are always matched to subdirectories, unless the rule contains a `/`separator.

The [official gitignore documentation](https://git-scm.com/docs/gitignore#_pattern_format) states:
>If there is a separator (_note by me: `/`_) at the beginning or middle (or both) of the pattern, _then the pattern is relative_ to the directory level of the particular `.gitignore` file itself. **Otherwise the pattern may also match at any level below the `.gitignore` level** [...]  
> A leading "\*\*" followed by a slash means match in all directories. For example, "\*\*/foo" matches file or directory "foo" anywhere, _**the same as pattern "foo"**_.

drothmaler:
Now it looks good to me :-D

PR #3395, opened 2020-05-03
Adding gitignore for Meson Build system

ineshbose:

Meson is an extremely fast build system used for building projects written in C++; it also uses Ninja backend. While building my projects in C++, I had to get rid of the files created by Meson and since there is no gitignore for it, I'm proposing one.

While there is no concrete documentation regarding the files creating during the build, the following link gives an idea.
https://mesonbuild.com/Running-Meson.html

If this is a new template:

 - **Link to application or project’s homepage**: https://mesonbuild.com/
- **GitHub**: https://github.com/mesonbuild/meson/

### DIFF
--- a/Meson.gitignore
+++ b/Meson.gitignore
@@ -1,0 +1,17 @@
+# Meson Directories
+meson-logs
+meson-private
+
+# Meson Files
+meson_benchmark_setup.dat
+meson_test_setup.dat
+sanitycheckcpp.cc
+sanitycheckcpp.exe
+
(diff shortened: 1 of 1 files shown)


eax0r:
@shiftkey It would be nice if we get some reviewers for this template.

ItzSwirlz:
Yeah.. it would.

ineshbose:
You know, I can tell you guys already have too much on your plate so I'll just make this easy for you.

PR #3398, opened 2020-05-06
Dart: Ignore .flutter-plugins

saschpe:
Generated files that appear when using [google_fonts](https://pub.dev/packages/google_fonts) or other Dart packages designed as Flutter plugins.

Those files are not meant to be versioned in any Flutter project.

Screenshot from 2020-05-06 15-06-05

### DIFF
--- a/Dart.gitignore
+++ b/Dart.gitignore
@@ -19,3 +19,6 @@ doc/api/
 *.js_
 *.js.deps
 *.js.map
+
+.flutter-plugins
+.flutter-plugins-dependencies


shiftkey:
@saschpe I've not kept up with the latest in the Dart/Flutter ecosystem - what do these files contains?

saschpe:
Some cached stuff, i.e. for **.flutter-plugins**:
```
# This is a generated file; do not edit or check into version control.
path_provider=$HOME/.pub-cache/hosted/pub.dartlang.org/path_provider-1.6.7/
path_provider_macos=$HOME/.pub-cache/hosted/pub.dartlang.org/path_provider_macos-0.0.4+2/
```
And for **.flutter-plugins-dependencies**:
```
{"_info":"// This is a generated file; do not edit or check into version control.","dependencyGraph":[{"name":"path_provider","dependencies":["path_provider_macos"]},{"name":"path_provider_macos","dependencies":[]}]}
```
I also don't get why Flutter doen't create a directory `$HOME/.flutter/`and put it's caches there instead of littering projects and home directories with a bunch of loose files. My `$HOME` currently sports **.flutter**, **.flutter_settings**, **.flutter_tool_state**, **.dart/**, **.dartServer/**  and **.pub-cache/** ;-)

shiftkey:
@saschpe thanks for the extra context - seeing the `This is a generated file; do not edit or check into version control` message in both files makes this an easy 👍 from me.

Suarhae:
versmisse/libgit2@8ee19a2

to test the solution and it works fine with git_tree_entry_byname. But it is not very clean (i change a private member (_cmp)) and incomplete (we must find a solution for all bsearch2 in the code

PR #3401, opened 2020-05-08
ignore projectSettingsUpdater.xml from Rider

bitbonk:

This file is generated by Rider and should never be added to VCS.

To qote @citizenmatt who is one of Rider developers at JetBrains:

> This is a file to record when certain settings migrations have occurred. Right now, it only makes changes to VCS configuration, but could be expanded in the future. The VCS changes are to make it the default to show the diff preview in the Commit tool window, and to clear the initial commit message. Once the changes have been made, this file is updated with the current version (“2”) and as long as the version matches, the settings are not re-applied the next time the project is loaded.
It should not be added to source control - it controls defaults to per-user settings ...

### DIFF
--- a/Global/JetBrains.gitignore
+++ b/Global/JetBrains.gitignore
@@ -10,6 +10,7 @@
 
 # Generated files
 .idea/**/contentModel.xml
+.idea/**/projectSettingsUpdater.xml
 
 # Sensitive or high-churn files
 .idea/**/dataSources/
--- a/Global/JetBrains.gitignore
+++ b/Global/JetBrains.gitignore
@@ -10,7 +10,6 @@
 
 # Generated files
 .idea/**/contentModel.xml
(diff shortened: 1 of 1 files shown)


Global/JetBrains.gitignore
```diff
@@ -10,6 +10,7 @@
 
 # Generated files
 .idea/**/contentModel.xml
+.idea/**/projectSettingsUpdater.xml
```
citizenmatt, line 13: I'd add a comment that this is a Rider specific file, rather than JetBrains wide.

You could also make the pattern a little more Rider specific with `.idea/*/.idea/projectSettingsUpdater.xml`, as Rider has a nested `.idea` folder to handle multiple solution files in the same folder. (.NET solutions are file based, and other JetBrains IDEs' projects are folder based)

PR #3409, opened 2020-05-16
ignore the `.idea` directory

vighnesh153:
`.idea` directory is added by the Pycharm IDE.

_TODO_

_TODO_

If this is a new template:


Hipska:
Already present, see https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore

BrianMSheldon:

PR #3410, opened 2020-05-16
Added pycharm project settings

chrisbeardy:

pycharm is a widely popular used python IDE so it's project settings should be included in the gitignore

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -120,6 +120,9 @@ venv.bak/
 # Rope project settings
 .ropeproject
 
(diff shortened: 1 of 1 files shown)


shiftkey:
@chrisbeardy thanks for the contribution, but we maintain a separate [`Global/JetBrains.gitignore`](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore) that covers what can be included in version control.

BrianMSheldon:

PR #3412, opened 2020-05-18
ignore .xwm generated by xwatermark package in TeX

bebuch:

The Latex Package [xwatermark](https://ctan.org/pkg/xwatermark) generates files with the ending `.xwm` during compilation. These should be ignored by Git since they are generated.

- https://ctan.org/pkg/xwatermark
- http://mirrors.ctan.org/macros/latex/contrib/xwatermark/doc/xwatermark-guide.pdf

### DIFF
--- a/TeX.gitignore
+++ b/TeX.gitignore
@@ -275,6 +275,9 @@ TSWLatexianTemp*
 # Makeindex log files
 *.lpz
 
+# xwatermark package
+*.xwm
+
 # REVTeX puts footnotes in the bibliography by default, unless the nofootinbib
 # option is specified. Footnotes are the stored in a file with suffix Notes.bib.
 # Uncomment the next line to have this generated file ignored.


shiftkey:
@bebuch thanks for the contribution.

That second link seems to error - is there a better resource that mentions this new extension?

bebuch:
Probably a bad mirror, here is one that works for me:

- http://mirror.physik-pool.tu-berlin.de/pub/CTAN/macros/latex/contrib/xwatermark/doc/xwatermark-guide.pdf

From a quick scan it seems to me that the creation of the `.xwm' file is not documented.

Here is a minimal example that demonstrates the generation:

```latex
\documentclass{article}

\usepackage{xwatermark}

\begin{document}
    Generates \texttt{xwatermarks.xwm}.
\end{document}
```

Compile by:

```bash
pdflatex your-file-name.tex
```

PR #3413, opened 2020-05-19
KiCad extension to ignore backup

longos:
Now KiCad gitignore bakcup files (also rescue backup)

In a previous version, gitignore does not contain an entry that ignores the folder of the automatically created backup.
Such a folder is created automatically, e.g. when updating libraries or an element conflict.

### DIFF
--- a/KiCad.gitignore
+++ b/KiCad.gitignore
@@ -24,3 +24,6 @@ fp-info-cache
 # Exported BOM files
 *.xml
 *.csv
+
+# Backup (also rescue backup)
+*backup/


shiftkey:
@longos thanks for the contribution!

Can you confirm examples of the paths that this rules is intended to ignore? I worry about the `*` catching other files or folders unnecessarily, and want to ensure that's the best we can do here...

longos:
In my case, it looks like this:

```
project/
... (some project files)
├── rescue-backup/
│   ├── <project-name>-<date>.pro
│   ├── <project-name>-<date>.sch
│   ├── <project-name>-<date>.lib
... (other project files)
```

shiftkey:
Could `backup/` and `rescue-backup/` then be "good enough" rules to add here?

longos:
This is also possible. Adding `*` would eliminate committing folders that sens, GIT takes over anyway.

PR #3414, opened 2020-05-19
Add *.aab

claw88:

Unity can build *.aab(Android App Bundle) instead of APK.

### DIFF
--- a/Unity.gitignore
+++ b/Unity.gitignore
@@ -57,6 +57,7 @@ sysinfo.txt
 
 # Builds
 *.apk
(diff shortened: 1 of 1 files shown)


shiftkey:
This seems like a good explainer of the differences https://stackoverflow.com/questions/52059339/difference-between-apk-apk-and-app-bundle-aab (and mentions the format).

GedAmbra:
Hello, I'm using Unity Bolt. Are there any additional gitignore file entries in this case?

PR #3415, opened 2020-05-19
Fix for Coverage* filename issue

lkurzyniec:
As @phoogenb properly notice, there was issue with Coverlet ignore rules. This PR fix the issue.

image

fix
image

Results received after execution of `dotnet test /p:CollectCoverage=true /p:CoverletOutputFormat=\"json,opencover,lcov,cobertura\"`.


lkurzyniec:
@shiftkey here is the PR to fix coverlet files result.

PR #3418, opened 2020-05-23
Add in .DS_Store (macOS Specific)

Nirma:
Usually I and other developers need to manually add this line later.
.DS_Store should be specific to everyone and there is almost no value in sharing it in a project.


Hipska:
This should be in your global `.gitignore`, It is not relevant for swift.

Nirma:
@Hipska That is true, `.DS_Store` is OS wide if you are on macOS so maybe I should close this. 

Thanks for the feedback!

PR #3419, opened 2020-05-26
Visual Studio Code

laper32:

Add Visual Studio code cache/option directory to ignore.
(I think vs and vsc are similar, so i add it to VisualStudio.gitignore.


shiftkey:
@laper32 thanks for the contribution, but we have an existing [`Global/VisualStudioCode.gitignore`](https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore) template that covers what VSCode can share.

PR #3424, opened 2020-05-30
Add text editors specific settings (.idea, .vscode) in Python.gitignore

SadiaAfrinPurba:

Different text editors create some hidden files, for example, VS Code creates .vscode folder. These folders should be ignored during a git push.

_TODO_

If this is a new template:

This is not a new template

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -136,3 +136,7 @@ dmypy.json
 
 # Cython debug symbols
 cython_debug/
+
+# Text editors
+.idea
+.vscode


Hipska:
Already present, see https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore and https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore

bdougie:
We have an existing [Global\JetBrains.gitignore](https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore) that covers more than just this rule.

PR #3428, opened 2020-06-02
Add .idea folder of PyCharm to Python

leonidpodriz:
Many python programmers using PyCharm in develop.


Hipska:
Already present, see: https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore

BrianMSheldon:

PR #3431, opened 2020-06-09
This is not Delphi-specific folder

Arvur:

This is just another extension for Delphi. 
No reason to ignore folder named "modules" for everybody.

Please keep generic gitignore-template as universal as possible. 
You can use specialized template for your extension as written in Readme:
> If you have a small set of rules, or want to support a technology that is not widely in use, and still believe this will be helpful to others, please read the section about specialized templates for more details.

### DIFF
--- a/Delphi.gitignore
+++ b/Delphi.gitignore
@@ -64,6 +64,3 @@ __recovery/
 
 # Castalia statistics file (since XE7 Castalia is distributed with Delphi)
 *.stat
-
-# Boss dependency manager vendor folder https://github.com/HashLoad/boss
-modules/


viniciussanchez:
Boss is the best dependency manager for Delphi, and also the most used.
Removing this folder would be bad for many Delphi developers.

viniciussanchez:
https://github.com/HashLoad/boss

dliocode:
Removing this folder would be bad for many Delphi developers.

juliosenha:
Removing this folder would be bad for many Delphi developers who use the **Boss** to manage their project dependencies.

nickolasdeluca:
I was very surprised to see Boss support on Delphi's template, I don't see a reason to remove it.

joaogirade:
Removing this folder would be bad for many Delphi application, especially for Boss users.  Boss is the best dependency manager for Delphi.

juliomar:
Very used and very good by the way used by BOSS. please do not remove. this is a mistake!

victorgonzales:
I see the removal as an error. 
please don't remove!

alanvcb:
Removing this folder would be bad for many Delphi developers who use the Boss to manage their project dependencies.

brutodelphi:
Boss is the best dependency manager for Delphi, and also the most used.
Removing this folder would be bad for many Delphi developers.

antoniojmsjr:
Removing this folder would be bad for many Delphi developers who use the Boss to manage their project dependencies.

Arvur:
A kind of flashmob here? Huh! )))

1. Boss is not a part of Delphi. It will make template non-universal.
2. It's folder name is too general. Someone can lose his sources subfolder because of this line.

The only solution I see is to keep this line commented. Will it be Ok?

viniciussanchez:
Have you ever seen someone lose their files because they program in Angular and put the name of a folder as node_modules?

bdougie:
Based on the responses, I don't see enough evidence to remove this folder from the ignore. If there were links to StackOverflow or similar forum comments supporting this change it would be considered. However, I am going to close this for now.

wlandgraf:
> Have you ever seen someone lose their files because they program in Angular and put the name of a folder as node_modules?

`node_modules` is part of NPM. Boss is not part of Delphi.
I once had a product with a folder named `Modules` with lots of .pas and .dfm files for TDataModule. My git client simply ignored that folder and I (almost) missed those files. Only after a while when cloning the repo I realized my `Modules` folder were not there.

So, Boss users should just add `Modules` folder to their .gitignore and that's it.

viniciussanchez:
I disagree!
`node_modules` is part of NPM, ok. But NPM is not part of Node. Just like the `modules` folder is part of Boss and Boss is not part of Delphi.
When installing Node, it is even optional whether or not to install NPM. You can work with Node using NPM, Yarn, JSPM...
Just like you can or cannot use Boss with Delphi.

image

wlandgraf:
But the % of people who use npm in Node.js. That's not the same with Boss.

You consider the situation I faced is normal and desired?

PR #3432, opened 2020-06-09
Add (Mac) .DS_Store file to .gitignore

OliverPearmain:

The .DS_Store file is auto generated in directories by the OS on MacOS to hold user specific properties about the directory. 


Hipska:
This should be in your global `.gitignore`, It is not relevant in any way for VisualStudio.

OliverPearmain:
You are quite right.

PR #3434, opened 2020-06-10
Add code editors

daveusa31:


Hipska:
Already present, see https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore and https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore

PR #3436, opened 2020-06-12
Rust only had top-level ignores, and was missing a major one

theJollySin:
One bug fix and one important addition to the Rust `.gitignore`.

1. BUG: One of the `.gitignore` lines was broken. It should have been `target/` but was written as top-level-only `/target/`.  Very easy to fix.
2. Improvement: If we are excluding Cargo-standard build `target/` folders, we should also exclude the (more commonly used) Cargo-standard build `debug/` folder.  Also easy to fix.

If you are not familiar with Rust (and Cargo the Rust build system that comes installed with Rust as a sort of standard library), you can see the official documentation for `debug/` and `target/` [here](https://doc.rust-lang.org/stable/book/ch01-03-hello-cargo.html).


theJollySin:
I'm sure the dev team is busy.  But I believe the Rust community would find this to be a small change and an obvious improvement.

Tell me if you need anything to expedite this PR.

mathstuf:
I question the addition of `debug` since it is always under `target` anyways.

phi-fell:
I'm also a little confused by this change.  `/target/` -> `target/` makes sense.  I could see projects wanting the inverse, but I think this is a reasonable default.  I'm not aware of any case that cargo generates a `debug` folder that is not a child of `target`.  The linked documentation only ever references `debug/` as `target/debug/` so this seems redundant.  If a debug directory is ever created outside `target/` could you point out some documentation that supports this? - and if not, shouldn't `release/` also be added?  I don't see a reason to add useless clutter, especially as this could confuse newcomers.

PR #3438, opened 2020-06-12
Wordpress.gitignore exclude all files in root folder

mircea-n:
!wp-content/ was not enough on my setup.
adding /* solved it for me

https://stackoverflow.com/a/8594971


LucasLarson:
Good call, but this pull request is already open in various forms in #3074, #3093, #3125, #3295, and #3378.

bdougie:

PR #3440, opened 2020-06-15
Add JetBrains familiy IDEs

mostafasolati:
This gitignore adds .idea folder that JetBrains family IDEs use


drothmaler:
Please see  `Global/JetBrains.gitignore`...

PR #3446, opened 2020-06-19
adding more ignore elements

pankajsingh143:

_TODO_

_TODO_

If this is a new template:

### DIFF
--- a/Java.gitignore
+++ b/Java.gitignore
@@ -1,23 +1,84 @@
-# Compiled class file
-*.class
-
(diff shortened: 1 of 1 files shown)


shiftkey:
@pankajsingh143 thanks for the contribution!

I can't accept this PR in its current form due to the rules which are covered by other templates in this repository.

Are there any rules here that are specific to the Java setup? Or was this more about adding additional tools that are covered by other templates?

PR #3448, opened 2020-06-21
Adding TeXstudio.gitignore

mxa:
This gitignore can be useful for LaTeX XeTeX, LuaTex in general and TeXstudio projects in particular.
*TeX creates a plethora of temporary files which will be re-generated when compiling the .tex, bibilography and style files. All these files should not be included in versioning systems like git.
You may wonder why .pdf files are not ignored, as the compiled result of a TeX project is a pdf. This is because illustrations in a TeX document can be pdf files which you will need for compilation.

It was missing!

No changes, this is new!

If this is a new template:

 - **Link to application or project’s homepage**: https://www.texstudio.org/ https://github.com/texstudio-org/texstudio/


Anton-Sakovich:
We already have `TeX.gitignore` for TeX. Could you have a look at it and maybe just add patterns missing from `TeX.gitignore` which are important for TeXstudio? Patterns like `*.dvi`, for example, are already there.

bdougie:
Per the feedback provided by @Anton-Sakovich. Please edit the https://github.com/github/gitignore/blob/main/TeX.gitignore

PR #3452, opened 2020-06-25
Exclude *.tfvars files

kevincloud:

Contrary to the current .gitignore explanation about .tfvars files, these are data points abtracted away from the Terraform variables due to the potentially sensitive nature of the values. For example, AWS, Azure, and GCP credentials are oftentimes assigned in the .tfvars files. I personally have had my credentials exposed several times due to the oversight of not excluding the .tfvars files in this particular .gitignore file.

The values in .tfvars files are, by nature, personal to the individual running the script. Data points such as these should not be included by default. If values are to be checked into version control, the variable definition includes a default attribute. 

https://learn.hashicorp.com/terraform/getting-started/variables.html#from-a-file
https://www.linode.com/docs/applications/configuration-management/terraform/secrets-management-with-terraform/#keeping-secrets-out-of-tf-files
https://secrethub.io/blog/secret-management-for-terraform/#secrets-in-tfvars

If this is a new template:

Not new, just a slight but important change

### DIFF
--- a/Terraform.gitignore
+++ b/Terraform.gitignore
@@ -8,11 +8,12 @@
 # Crash log files
 crash.log
 
-# Ignore any .tfvars files that are generated automatically for each Terraform run. Most
-# .tfvars files are managed as part of configuration and so should be included in
-# version control.
+# Exclude all .tfvars files, which are likely to contain sentitive data, such as
+# password, private keys, and other secrets. These should not be part of version 
+# control as they are data points which are potentially sensitive and subject 
+# to change depending on the environment.
 #
-# example.tfvars
+*.tfvars
 
 # Ignore override files as they are usually used to override resources locally and so
 # are not checked in


shiftkey:
I have no horse in this race, but I'll ask for feedback from the people who were involved with #2650 to see if they agree with the changes proposed...

cc @jkodroff @apparentlymart

jkodroff:
@kevincloud You really don't want to put sensitive values in any TF file, vars or otherwise.  Those values should come from environment variables.

kevincloud:
Hi @jkodroff, I'm a Sr. Solutions Engineer at HashiCorp. Our [official stance](https://www.terraform.io/docs/cloud/workspaces/variables.html#terraform-variables) is that tfvars files should never be put into VCS. Users are given the choice of using tfvars files, environment variables, etc, and there are many valid reasons why users need to use tfvars files for secrets as opposed to environment variables. Environment variables actually take the [lowest precendence](https://www.terraform.io/docs/configuration/variables.html#variable-definition-precedence) for Terraform values.

jkodroff:
Ok.  That's good enough for me.

apparentlymart:
Please note that the advice @kevincloud linked to is specifically for Terraform Cloud/Enterprise and does not apply to folks using Terraform CLI directly. What I said in [my previous comment](https://github.com/github/gitignore/pull/2650#issuecomment-383168330) still holds for the general case. It's a shame that the Terraform Cloud documentation is making assertions about usage of Terraform outside of Terraform Cloud.

jkodroff:
@shiftkey Oh no!  Now I'm back on the fence!  (@apparentlymart writes a _lot_ of Terraform and I would not take his opinion lightly.)

kevincloud:
Hi @apparentlymart! Actually, if you read the reference for our [official stance](https://www.terraform.io/docs/cloud/workspaces/variables.html#terraform-variables), the very last sentence in that section specifically calls out:

> (Note that you shouldn't check in `terraform.tfvars` even when running Terraform solely on the command line.)

This is how we train our customers in our workshops. As I mentioned before, the `.tfvars` files are designed to abstract values away from the code. If you want to check values into your repo with Terraform, use the `default` attribute in the variable definition. Or, you can add a `terraform.tfvars.example` file if you want to provide a template.

`.tfvars` files do not have the same context as, say, Java `.properties` files, where values are set post-build/post-compile. The Terraform code in your repo is production-ready. Many users have exposed sensitive data to the public by accidentally adding their `.tfvars` files to their repo. Because of that, they should never be checked in as the default option.

mikecmpbll:
I'm relatively new to this whole space, but I'm using [Azure CAF](https://github.com/aztfmod/terraform-azurerm-caf) supermodule and you configure it with tfvars, so near enough every file in my repo is a tfvars file. Is this a common pattern in terraform usage? If not, then no problem, but I found it weird when tfvars were gitignored using this gitignore file 🤷🏻 .

PR #3456, opened 2020-06-27
ignore HPROF files

maherzaidoune:
ignore massive file .hprof

.hprof is a massive file that contains the memory profiling info. 

https://developer.android.com/studio/profile/memory-profiler

### DIFF
--- a/Android.gitignore
+++ b/Android.gitignore
@@ -83,3 +83,6 @@ lint/generated/
 lint/outputs/
 lint/tmp/
 # lint/reports/
+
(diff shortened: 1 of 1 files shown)


Android.gitignore
```diff
@@ -83,3 +83,6 @@ lint/generated/
 lint/outputs/
 lint/tmp/
 # lint/reports/
+
+#Android Profiling
```
shiftkey, line 87: Tidy up documentation

```suggestion
# Android Profiling
```

PR #3458, opened 2020-07-02
Support Rider properly

cja-github:
As per https://github.com/JetBrains/resharper-rider-samples/blob/master/.gitignore and https://rider-support.jetbrains.com/hc/en-us/articles/207097529

### DIFF
--- a/Global/JetBrains.gitignore
+++ b/Global/JetBrains.gitignore
@@ -2,23 +2,25 @@
 # Reference: https://intellij-support.jetbrains.com/hc/en-us/articles/206544839
 
 # User-specific stuff
-.idea/**/workspace.xml
-.idea/**/tasks.xml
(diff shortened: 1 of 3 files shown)


scscgit:
Could we have this moved from Global to a root directory, just like VisualStudio.gitignore? For about a year there hasn't been any way to create a JetBrains-compatible project using https://github.com/new due to this removal of Rider compatibility: https://github.com/github/gitignore/commit/96baa6309f0c8e79b7daa3899edd8d786a0e5593

scscgit:
Also is there any plan to support Shared Run Configurations? In Rider it can be like `.idea/.idea.NameOfTheProject/.idea/runConfigurations/` with a variable project name (created using "Store as a project file" in the right-top corner of Run Configurations window), while in IDEA it's like `.idea/runConfigurations`

PR #3463, opened 2020-07-09
Added MacOS files that should not be checked in

anton-roos:
Added MacOS files that should not be checked in like .DS_Store for example.

We need to ignore MacOS specific files too.

https://en.wikipedia.org/wiki/AppleSingle_and_AppleDouble_formats

### DIFF
--- a/Swift.gitignore
+++ b/Swift.gitignore
@@ -88,3 +88,31 @@ fastlane/test_output
 # https://github.com/johnno1962/injectionforxcode
 
 iOSInjectionProject/
+
+
+# MacOS General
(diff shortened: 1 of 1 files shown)


ninest:
You should use the global gitignore for Mac OS files: https://github.com/github/gitignore/blob/master/Global/macOS.gitignore

This guide may help: http://egorsmirnov.me/2015/05/04/global-gitignore-file.html

anton-roos:
Thanks for the clarification.

anton-roos:
Done 👍

PR #3464, opened 2020-07-09
Add .env.local to Node

redshoga:

Next.js recommends that to add the `.env.local` to the `. gitignore`, as described in the documentation below.

https://nextjs.org/docs/basic-features/environment-variables#loading-environment-variables

### DIFF
--- a/Node.gitignore
+++ b/Node.gitignore
@@ -79,6 +79,7 @@ web_modules/
 # Next.js build output
 .next
 out
+.env.local
 
 # Nuxt.js build / generate output
 .nuxt


Node.gitignore
```diff
@@ -79,6 +79,7 @@ web_modules/
 # Next.js build output
 .next
 out
+.env.local
```
bdougie, line 82: Understand this is a change requested in Next.js documentation but this should live [here](Note: .env, .env.development, and .env.production files should be included in your repository as they define defaults. .env.*.local should be added to .gitignore, as those files are intended to be ignored. .env.local is where secrets can be stored.)

bdougie (changes requested):
in addition to my requested changes. From the Next.js docs.

> Note: .env, .env.development, and .env.production files should be included in your repository as they define defaults. .env.*.local should be added to .gitignore, as those files are intended to be ignored. .env.local is where secrets can be stored.

bdougie:
nextjs already installs with a gitingore. I think the .env.local should be the only change here.

bdougie:
This was actually fixed in this merge https://github.com/github/gitignore/pull/3316

PR #3477, opened 2020-07-20
Update Node.gitignore to incorporate .DS_Store

Godzilla5111:
.DS_Store is missing from all templates.

_TODO_

_TODO_

If this is a new template:


Godzilla5111:
.DS_Store is missing in all gitignore templates. Please try to include it in each and every template as they can be commonly encountered while working with forked and shared projects. Doing this once for all will benefit everyone.

mys721tx:
Use `Global/macOS.gitignore`.

PR #3478, opened 2020-07-20
Ignore Gradle bin directory

ewpratten:
Gradle generates `bin/` directories containing `.class` files that shouldn't be checked in to git. This PR ignores these directories

### DIFF
--- a/Gradle.gitignore
+++ b/Gradle.gitignore
@@ -1,6 +1,7 @@
 .gradle
 **/build/
 !src/**/build/
+**/bin/
 
 # Ignore Gradle GUI config
 gradle-app.setting


drothmaler:
As far as I know, the [default Gradle buildDir](https://docs.gradle.org/current/userguide/writing_build_scripts.html#sec:standard_project_properties) is called `"build"`. Is there any plugin / task that uses `"bin"` by default? If so, please provide a link to the documentation.

But to me it looks more like you are dealing with the Eclipse default output directory (which actually is `"bin"` - and is listed in `Global/Eclipse.gitignore`)

Ewpratten:
Every project and subproject I build with gradle's `java-library` plugin generates a `build` and a `bin` directory. I encounter this all the time, but I now wonder if this actually has something to do with the VSCode gradle plugin?

jfcostta:
I'm using VSCode too and realized it creates a hidden `.settings` folder containing the `org.eclipse.buildship.core` file. Maybe it's related to Java Extension plugin for the VSC editor.

Anyway, `bin` was manually added to my `.gitignore` since it equals to a `build` dir.

PR #3479, opened 2020-07-21
Update Yarn modern gitignore based on documentation

nickserv:

The ignored Yarn modern files are out of date and don't support zero-install.

https://yarnpkg.com/advanced/qa#which-files-should-be-gitignored

### DIFF
--- a/Node.gitignore
+++ b/Node.gitignore
@@ -109,13 +109,13 @@ dist
 .vscode-test
 
 # Yarn modern
-# Comment for zero-install
-.pnp.*
 .yarn/*
-# Uncomment for zero-install
-# !.yarn/cache
 !.yarn/patches
 !.yarn/plugins
 !.yarn/releases
 !.yarn/sdks
 !.yarn/versions
+# Swap the comments on the following lines if you don't wish to use zero-installs
+# Documentation here: https://yarnpkg.com/features/zero-installs
+!.yarn/cache
+#.pnp.*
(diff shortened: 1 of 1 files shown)


redxtech:
I came here to make the exact same pull request. I'm surprised they haven't merged this yet.

nickmccurdy:
I find it tricky to make sure I have the right gitignore when using yarn 2 (berry) with GitHub since it's still out of date. Is there anything we can do to help get this merged?

Node.gitignore
```diff
@@ -108,9 +108,14 @@ dist
...
-.yarn/install-state.gz
+# Yarn modern
+# Comment for zero-install
 .pnp.*
+.yarn/*
+# Uncomment for zero-install
+# !.yarn/cache
+!.yarn/patches
+!.yarn/plugins
+!.yarn/releases
```
SiiiTschiii, line 119: Does it really make sense to add those files:
```
ll .yarn/releases/yarn-3.2.4.cjs
-rwxr-xr-x 1 christof christof 2.1M Okt 19 21:43 .yarn/releases/yarn-3.2.4.cjs
```

nickmccurdy: This is what's currently recommended in the docs. However, Yarn 4 seems like it will have better support for managing versions with corepack, so this may change in the future.

nickmccurdy: Also you can use `.gitattributes` if you don't want these files showing up in diffs.

nickmccurdy:
@Maxime-p Thanks for the update, now using this.

PR #3486, opened 2020-07-30
Add Storybook static build directory to Node.gitignore

yanneves:

Storybook's build command `build-storybook` outputs a static build to this directory, `storybook-static/` by default.

https://storybook.js.org/docs/basics/exporting-storybook/

The documentation refers to setting the static directory to `.out/` using the `-o` flag. The default (`-o` flag is not required) outputs to `storybook-static/`.

### DIFF
--- a/Node.gitignore
+++ b/Node.gitignore
@@ -114,3 +114,6 @@ dist
 .yarn/build-state.yml
 .yarn/install-state.gz
 .pnp.*
+
+# Storybook static build (https://storybook.js.org/)
+storybook-static


krfong916:
For clarity, does storybook-static belong in the gitignore file?

yanneves:
Yeah, because you would likely sideload Storybook with another build tool (e.g. create-react-app, Next, Gatsby) that would output your application code to a directory like `public/` or `dist/`, Storybook by default targets a specific build directory, `storybook-static/` to avoid conflicts. We wouldn't want to commit the contents of the `storybook-static/` directory so it belongs in `.gitignore`.

For clarity, Storybook configuration and code that we _would_ want to commit lives in `.storybook/` by default.

Linking this again https://storybook.js.org/docs/basics/exporting-storybook/

AndysonDK:
what about `build-storybook.log` which gets generated when you run `yarn chromatic`?

bdougie:
Building commands can be run by deployment platforms. My assumption is most folks are using local servers to view their storybook changes i.e. `npm run storybook.` 

This change seems unnecessary. Can you expand on the situation in where you are running a build to generate this file?

yanneves:
That's a fair point @bdougie, I can see why we'd want to assess that difference to avoid cluttering the template

You may need to run `build-storybook` to debug a production vs development Storybook build, but not as part of the prescribed local development workflow

bdougie:
I have needed that folder to debug my GitHub Action logs in the past, which is why I raised the question.

rbong:
It may not be part of the local development workflow, but that doesn't change the fact that you don't want to commit it.

Plenty of production-only folders are already in this repo.

What if I am for instance debugging the production build of storybook?

Can this be reopened?

PR #3487, opened 2020-07-31
Update VisualStudio.gitignore

weitson:
These three files need to be ignored when using Visual Studio 2019.
Here they are: *.sln、*.vcxproj、*.vcxproj.filters
Thanks.

_TODO_ These three files need to be ignored when I using Visual Studio 2019,
they will be uploaded to the repositories

_TODO_

If this is a new template:

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -9,6 +9,7 @@
 *.user
 *.userosscache
 *.sln.docstates
+*.sln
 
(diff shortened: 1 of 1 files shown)


erikyuzwa:
I agree with the `.vcxproj` and `.vcxproj.filters`, but *not* the `.sln`. Thought this is / was a core VS project file? 😕

PR #3501, opened 2020-08-13
ignoring .aar files may cause to loose the files if libs are manually imported

AbhinayMe:
Desc: Comment next line if you have manually imported libs with .aar files.

ignoring .aar files may cause to loose the files if libs are manually imported

[https://developer.android.com/studio/projects/android-library](https://developer.android.com/studio/projects/android-library)

If this is a new template:


jkwe7777:
it' sounds so good,man

kbakdev (approved):

PR #3505, opened 2020-08-20
Add .idea ignore

quockhanhng:

Add ignore to folder .idea/

_TODO_

If this is a new template:

### DIFF
--- a/Android.gitignore
+++ b/Android.gitignore
@@ -38,6 +38,7 @@ captures/
 
 # IntelliJ
 *.iml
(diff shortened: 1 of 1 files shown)


redxtech:
It appears as though the android template already ignores some of the files within the .idea folder. The ones that are not ignored are useful to include in a project as they deal with project configuration, such as scripts, linters, vcs, & etc. Essentially, there's a reason why the entire folder isn't ignored, and only the "personal" files are.

PR #3510, opened 2020-08-26
Add *.zst to ArchLinuxPackages.gitignore

aspenluxxxy:
[Arch Linux now uses `.pkg.tar.zst` for packages by default](https://www.archlinux.org/news/now-using-zstandard-instead-of-xz-for-package-compression/)


53jk1 (approved):

aspenluxxxy:
Well it's obvious that Github ignores this repo, so imma close it

PR #3523, opened 2020-09-16
[Global/macOS] Prevent corruption of "Icon\r" rule

i-to-r:
Fixed #2203 PR.

Make `Icon\r` pattern less vulnerable to editors and to [auto `CRLF` → `LF` conversion on commit](https://github.com/github/gitignore/commit/c02a4743f9c66a1d70d9ffb0da3e2156f5cc797e#commitcomment-25924422). As @spencermathews wrote in #2337 :
> Since this ignore rule may be vulnerable to editors that automatically standardize line endings, it seems imperative that updates to this file be done with care or else this rollercoaster seems likely [to continue](https://github.com/github/gitignore/pulls?q=is%3Apr+Icon+%5Cr).

Note: the Start..End comment has been added because most editors display `Icon[␍]` pattern in two lines:
```
# Start of Icon[\r] pattern
Icon[
]
# End of Icon[\r] pattern
```

- [How to ignore Icon? in git](https://stackoverflow.com/questions/17556250/how-to-ignore-icon-in-git/33974028#33974028) (`Icon[\r]`)

### DIFF
--- a/Global/macOS.gitignore
+++ b/Global/macOS.gitignore
@@ -3,8 +3,9 @@
 .AppleDouble
 .LSOverride
 
-# Icon must end with two \r
-Icon
+# Start of Icon[\r] pattern
+Icon[]
+# End of Icon[\r] pattern
 
 # Thumbnails
 ._*


53jk1 (approved):

revolter:
Bump

gimbo:
Unfortunately, I can report that at least one relatively popular tool doesn't handle this well either.

I just tried this vs the [`mixed-line-endings`](https://github.com/pre-commit/pre-commit-hooks#mixed-line-ending ) hook for the popular (at least in the python world) [pre-commit](https://pre-commit.com/) framework, and sadly it does still replace the CR with an LF, rendering the rule useless.

Now, it's clear that this behaviour is _a bug in that hook_ (as from its name it should obviously only be fixing line _endings_!), but the fact that this was literally the first tool I tried this with suggests to me that this is still a fragile solution, and that anything involving embedding actual CR characters in a `.gitignore` file is going to be prone to this issue, leaving the "rollercoaster" in place — at least until/unless proper support for unicode code points is added to gitignore.

So, while this PR is a clear improvement on the current situation, may I suggest that it might be worth considering a less precise but more robust solution not involving actual CR characters?

For example, in the same Stack Overflow thread mentioned by @i-to-r in their comment above, we also have [this answer](https://stackoverflow.com/a/65429032/392743), which suggests an approach like this:

```
Icon?
![iI]con[_a-zA-Z0-9]
```

As the author says in that answer, it would be worth expanding the `_a-ZA-Z0-9` part for more completeness, but you get the idea.

I'd actually started working on a PR suggesting this when I realised I should check if something similar was already in flight, which led me to this PR.

Leif-W:
> For example, in the same Stack Overflow thread mentioned by @i-to-r in their comment above, we also have [this answer](https://stackoverflow.com/a/65429032/392743), which suggests an approach like this:
> 
> ```
> Icon?
> ![iI]con[_a-zA-Z0-9]
> ```
> 
> As the author says in that answer, it would be worth expanding the `_a-ZA-Z0-9` part for more completeness, but you get the idea.

What about `Icon\r?`, does that work?  I was thinking about this last night, and a human or tool may conceivably make a file with other characters besides `_a-zA-Z0-9`.  Rather than go down the rabbit hole of chasing what might be valid, from a set of all available valid characters on HFS+(except possibly `NUL`, `/`, and maybe `:`), what about the relatively safe assumption that ANYTHING (except nothing, hence `?` not `*`) after an `Icon\r` is considered fair game for ignoring?  That way, the `\r` isn't at the EOL, the `?` is.  That seemed to work for me.  I don't have MacOS for testing, but was able to ignore in WSL2 Ubuntu, and Android Termux, both of which allow creation of `Icon\r\r` files and directories (with `touch Icon\r\r/foo` to ensure git would see non-empty dir).  In both cases, despite having `Icon\r\r` in .gitignore, the `Icon\r\r` file and folders were NOT being ignored, in either WSL2 Ubuntu, or Android Termux.  I am not sure why that was.

Alongazhar:
Nyahlanggan

Pada Jum, 30 Mei 2025, 12:35 PG github-actions[bot] <
***@***.***> menulis:

wirecat:
This appears to be a duplicate of https://github.com/github/gitignore/pull/4596 which was accepted. Closing for now.

PR #3527, opened 2020-09-19
Update Maven.gitignore add flattened-pom.xml for ci friendly

iRecursion:
add .flattened-pom.xml for ci friendly and module version control etc

### DIFF
--- a/Maven.gitignore
+++ b/Maven.gitignore
@@ -9,3 +9,6 @@ buildNumber.properties
 .mvn/timing.properties
 # https://github.com/takari/maven-wrapper#usage-without-binary-jar
 .mvn/wrapper/maven-wrapper.jar
+
+# for maven ci friendly
+.flattened-pom.xml


drothmaler:
Seems to be related to the [Flatten Maven Plugin](https://www.mojohaus.org/flatten-maven-plugin/index.html)... `.flattened-pom.xml` is actually the default for [`flattenedPomFilename`](https://www.mojohaus.org/flatten-maven-plugin/flatten-mojo.html#flattenedPomFilename). I cant find an explicit statement about whether to ignore the file or not, but as it is generated by the plugin it seems reasonable to ignore it. On the other hand the plugin seems to also update the original POM, to point to the flattened one - which (to me) would be a reason to actually commit it. 

Can you please provide us with some more detail about the usage. Or ideally an „official“ statement that indicates if the file should be ignored or added...

iRecursion:
> 
> 
> Seems to be related to the [Flatten Maven Plugin](https://www.mojohaus.org/flatten-maven-plugin/index.html)... `.flattened-pom.xml` is actually the default for [`flattenedPomFilename`](https://www.mojohaus.org/flatten-maven-plugin/flatten-mojo.html#flattenedPomFilename). I cant find an explicit statement about whether to ignore the file or not, but as it is generated by the plugin it seems reasonable to ignore it. On the other hand the plugin seems to also update the original POM, to point to the flattened one - which (to me) would be a reason to actually commit it.
> 
> Can you please provide us with some more detail about the usage. Or ideally an „official“ statement that indicates if the file should be ignored or added...

https://github.com/iRecursion/eco-starter
yes,i used flatten maven plugin just want to manage modules version as a centeralization way,
i can not see Required attribute of tag <flattenedPomFilename> before,
Thank you for your prompt reply,it helps me!

PR #3528, opened 2020-09-20
Add .vscode/ VSCode cache folder

omerhijazi404:

Many Unity developers use Visual Studio Code as their primary code editor for their scripts (C#)
and using Code with projects creates a cache directory called .vscode/ which is not needed like the .vs/ directory.

### DIFF
--- a/Unity.gitignore
+++ b/Unity.gitignore
@@ -26,6 +26,9 @@
 # Visual Studio cache directory
 .vs/
 
+# Visual Studio Code cache directory
+.vscode/
+
 # Gradle cache directory
 .gradle/
 


66Gramms:
Agreed

AdityaGupta150:
> 
> 
> 
> Many Unity developers use Visual Studio Code as their primary code editor for their scripts (C#)
> and using Code with projects creates a cache directory called .vscode/ which is not needed like the .vs/ directory.

But the whole directory shouldn't be added to .gitignore I think, as it also stores configurations, so if another user clones the repo, vs code will set it up from the configs.
Better left to the user, or prefer what MOST would prefer, though i can't say about that.

For example, i can have intellisense settings in my `settings.json` like this, so that it maintains code conventions throughtout, like 

>    "files.trimFinalNewlines": true,
>    "files.trimTrailingWhitespace": true,

many repos say to trim whitespaces, and using the original `settings.json` as the author, shows the intent itself, and vs code will maintain it.

```json
    "editor.formatOnPaste": true,
    "workbench.editor.highlightModifiedTabs": true,
    "explorer.sortOrder": "type",
    "files.trimFinalNewlines": true,
    "files.trimTrailingWhitespace": true,
    "editor.acceptSuggestionOnEnter": "off",
    "deepcode.uploadApproved": true, //accepting suggestions using enter turned off
    "workbench.colorCustomizations": {
        "activityBar.background": "#2B3105",
        "titleBar.activeBackground": "#3C4407",
        "titleBar.activeForeground": "#F8FCE1"
    }
```

Also `launch.json` can include specific args to be included for debug mode, that makes it simple, to just head over to the debugger in VS Code, rather than configure the commands yourself, or using another better debugger.

`c_cpp_properties.json` has include paths, so when vscode intellisense isn't using cmake as the provider, it uses this, though it can be generated easily

omerhijazi404:
Well then like this, even the .vs directory is needed when talking that way

66Gramms:
It's not needed, as those are developer settings. A developer's personal settings shouldn't be in the version control!!! So, the .vs directory is not needed at all.

PR #3529, opened 2020-09-22
Add rider cache directory to Unity ignore list

Peaj:

Many Unity developers work with JetBrains Rider.
As the Unity.gitignore considers VSCode files it should also consider Rider files.

Riders .idea folder contains a mix of cache, user/shared configuration and settings files.
Similar to Visual Studios .vs folder it should be ignored.
There may be some files (e.g shared dictionaries and settings) in this folders that could be shared.
Sadly JetBrains does not separate local from shared files via folder structure.
Thus it is easier to ignore the folder and later on explicitly state exceptions for files one would consciously share.

This is JetBrains recommendation on which of the files in the .idea folder to ignore:
The comments are already mentioning though that some files are missing from this official recommendation.
This is their recommended gitigonre file
It is quite complex though and is not future proof as it may miss Rider generated files added in the future.

As long as Rider does not clearly separate their local files from their shared files I would recommend to ignore this folder completely.

### DIFF
--- a/Unity.gitignore
+++ b/Unity.gitignore
@@ -26,6 +26,9 @@
 # Visual Studio cache directory
 .vs/
 
+# Rider cache directory
+.idea/
+
 # Gradle cache directory
 .gradle/
 


kbakdev (approved):

thatrobotdev:
It looks like there are some duplicate PRs with this same change: #3543, #3665, #3681, and #3687.

dustingraham:
This is a good change. I do this manually for each project right now.

SabinT:
This would be very useful indeed! I do this manually for every project as well.

PR #3530, opened 2020-09-22
Update Go.ignore to add Delve compiled output

dvelitchkov:
Add Go Delve debugger compiled output. Default is "./ __debug_bin" per https://github.com/go-delve/delve/blob/master/Documentation/usage/dlv_debug.md#options

The Go Delve debugger is the de-facto standard Go debugger. Add its default compiled binary to the ignore list. 

https://github.com/go-delve/delve/blob/master/Documentation/usage/dlv_debug.md#options

### DIFF
--- a/Go.gitignore
+++ b/Go.gitignore
@@ -13,3 +13,6 @@
 
 # Dependency directories (remove the comment below to include it)
 # vendor/
+
+# Delve debugger compiled binary
+/__debug_bin


kbakdev (approved):

radeksimko (approved):

jollyrogue:
Is there a way to bump this? I'd like this added too.

Go.gitignore
```diff
@@ -13,3 +13,6 @@
 
 # Dependency directories (remove the comment below to include it)
 # vendor/
+
+# Delve debugger compiled binary
+/__debug_bin
```
BananZG, line 18:
```suggestion
__debug_bin*
```

There's a new change to delve to include timestamp in output
source: https://github.com/go-delve/delve/commit/801a9109c7fd7be8b060b466cc041a3695455c77

PR #3543, opened 2020-10-09
Update Unity.gitignore - Added ignore for JetBrains Rider

MickL:
Rider from JetBrains is a commonly used C# IDE. It will add a `.idea` folder which should be ignored in git.


53jk1 (approved):

MickL:
Yeah lets close unresolved issues, thats how to solve things the right way!

thorrsson:
I'd point out the Contribution Guidelines https://github.com/github/gitignore?tab=readme-ov-file#contributing-guidelines (We do have a rewrite of this coming but the guidance for Tool specific exclusions in a Language file will likely remain the same)

Specifically 

>Please consider the scope of your change. If your change is specific to a certain language or framework, then make sure the change is made to the template for that language or framework, rather than to the template for an editor, tool, or operating system.

Since the .idea folder isn't specific to Unity, nor Unity development I don't think this change belongs in this file. The .idea directory also stores Shared Run Configurations and some other artifacts that folks benefit from including in their committed changes. Ignoring this entire directory shouldn't be the standard, but optional for teams who choose not to include it.

** I acknowledge the .vs line in this file as well... it should not have been pushed through in the first place if we were to apply the Contribution guidelines **

MickL:
@wirecat I strongly disagree with this decision. I work for 20+ years as a web developer and since then I worked with countless popular frameworks, lately e.g. Angular, Nuxt, React, Vite, etc. etc. and each and every project ignores the `.idea` folder as well as the vscode folder because they contain personal configuration of each developers machine. Run config is not shared by sharing the `.idea` folder.

Unity is the one and only annoying project who doesnt ignore `.idea` folder

PR #3544, opened 2020-10-09
Add svg temporary files

schtandard:

The [`svg`](https://ctan.org/pkg/svg) package creates a large number of auxiliary files when used (two for each `.svg` graphic that is included). These are conveniently placed into a folder `svg-inkscape` by default and should be ignored.

[`svg`'s documentation](http://mirrors.ctan.org/graphics/svg/doc/svg.pdf)

### DIFF
--- a/TeX.gitignore
+++ b/TeX.gitignore
@@ -185,6 +185,9 @@ _minted*
 # scrwfile
 *.wrt
 
+# svg
+svg-inkscape/
+
 # sympy
 *.sout
 *.sympy


408261668:
I can to learning to svg, because the doctor can help anymore.

53jk1 (approved):

schtandard:
@shiftkey Mentioning you because you merged the last TeX PR: I am kind of confused as to the state of this. This PR is now almost four weeks old with no interaction (except a Spam message?). About two weeks ago I received a message that it was approved, but there is no trace of that now. Maybe it's been retracted..?

Is it normal for this to take so long and I just have to be patient or what's going on?

bdougie:
Hi @schtandard, we get a high volume of PR's here. Can you provide context about why this is not needed in the git history? The TeX PRs all link to white papers with no context as to where to look. 

@53jk1 did not provide any context for the approval and this frankly sat for all of the above reasons.

Please do provide more context on this change as my understanding is that svg files are useful to story in git due to it being a common file structure.

schtandard:
Hi @bdougie, glad to see that this isn't dead after all. I'll happily provide some more context.

> The TeX PRs all link to white papers with no context as to where to look.

I feel like this hints at general confusion about the LaTeX ecosystem (feel free to correct me if I'm wrong), so I'll try to explain. Unlike other programming languages (if we decide to categorize TeX/LaTeX as such), there is no central organization steering its development and deciding what becomes part of it. There is a small team of volunteers developing the kernel, but most of the functionality used in any LaTeX project is provided by packages maintained by the community. Some of those packages create auxiliary files that are (or can be) recreated on every LaTeX run and should not be included in the version history (they are redundant and they may change on every run, even when the actual content of the project hasn't changed). Accordingly, most of this `.gitignore` file is concerned with excluding those temporary files of commonly used packages (and related tools).

Unlike with, say, Python packages, the documentation to those packages is not made accessible in some standardized web format but instead each package author documents their package in a typeset document (after all, (La)TeX is a typesetting system), usually published as a PDF file. I suspect that those are the "white papers" you are referring to. Packages are published to [CTAN] where their source and documentation can be accessed (e.g. [here for the `svg` package][svg], which this PR is concerned with). This is the "most official" place one could refer to and I suspect that many of the TeX PRs do that.

[CTAN]: https://ctan.org/
[svg]: https://ctan.org/pkg/svg

> Please do provide more context on this change as my understanding is that svg files are useful to story in git due to it being a common file structure.

The `svg` package is used to include SVGs in documents typeset using LaTeX. Due to PDF limitations (a LaTeX run produces a PDF document), these graphics can't be directly included, so they have to be converted into PDF graphics first. [Inkscape] (an SVG editor) can perform this conversion, seperating the "drawing part" and the "text part" of the SVG image, converting the "image part" to PDF while storing the "text part" in a text file in such a way that LaTeX can typeset it in the correct position on top of the "image part". This means that for every `svg` file two temporary files have to be created (one `pdf` file and one text file, usually carrying the extension `pdf_tex`).

[Inkscape]: https://inkscape.org/

Now, the `svg` package automatically calls Inkscape to do all this during a LaTeX run and stores all of these temporary files in a new folder called `svg-inkscape`. This folder _does not_ include the `svg` files, only the `pdf` and `pdf_tex` files generated from the `svg` files. The `svg` files themselves should of course be included in the version history.

----

I hope this clears up any confusion and provides the necessary context. If not, please let me know.

bdougie:
>I hope this clears up any confusion and provides the necessary context. If not, please let me know.

This is extremely helpful and unlocks a few other TeX PRs. Thanks for taking the time in responding.

PR #3549, opened 2020-10-12
Update Node.gitignore

heclack:
Reason:
Personal IDE settings do not belong in code base.

documentation of ".vscode" file:

If this is a new template: 

### DIFF
--- a/Node.gitignore
+++ b/Node.gitignore
@@ -107,6 +107,7 @@ dist
 
 # Stores VSCode versions used for testing VSCode extensions
 .vscode-test
+.vscode
 
 # yarn v2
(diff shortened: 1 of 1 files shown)


Jabster28:
Personal IDE settings would be in the user settings, no? I feel like the workspace settings are meant to enforce code styles and what not.

53jk1 (approved):

heclack:
Yes, but by tying settings into a specific editing application are you not forcing anyone who works on the codebase to use vscode? Is that an appropriate precedent to set, especially considering git version-control is used very often in open source codebase?

bdougie:

PR #3550, opened 2020-10-13
Add .textClipping to exclusions

nmdange2:
I ran into the issue referenced in https://github.com/github/gitignore/pull/3523 and I'm not sure how to fix it.  I want to add the .textClipping exclusion without breaking the other rule so this pull request should not be accepted until the other one is first.

Files with the .textClipping extension should be ignored.

Per https://en.wikipedia.org/wiki/TextClipping this file is created when text is dragged and dropped to a folder.  The file created can't be shared between computers.

### DIFF
--- a/Global/macOS.gitignore
+++ b/Global/macOS.gitignore
@@ -2,9 +2,11 @@
 .DS_Store
 .AppleDouble
 .LSOverride
+.textClipping
 
 # Icon must end with two \r
-Icon
+Icon
+
 
 # Thumbnails
 ._*


MonLagenda (approved):

i-to-r:
Did you edit the `macOS.gitignore` file [through a browser](https://docs.github.com/en/free-pro-team@latest/github/managing-files-in-a-repository/editing-files-in-another-users-repository) (https://github.com/github/gitignore/edit/master/Global/macOS.gitignore)?

And if so, did you see the message:
> We’ve detected the file has mixed line endings. When you commit changes we will normalize them to **Windows-style** (<kbd>CR</kbd><kbd>LF</kbd>).

?

53jk1 (approved):

i-to-r:
@nmdange2, unfortunately the fix #3523 won't help you directly. GitHub online editor (browser, [CodeMirror editor](https://codemirror.net/)) still breaks the `Icon[␍]` rule (i-to-r/gitignore@123480c339a4fe720145b219ffa85571b923ccb7), but the fix will preserve the line endings `␊` of the rest of rules (i.e. the message "_We’ve detected the file has mixed line endings._ […]" will not appear).

After the fix #3523 is ​​merged, you should make a commit using, for example, [GitHub Desktop](https://docs.github.com/en/free-pro-team@latest/desktop/contributing-and-collaborating-using-github-desktop/committing-and-reviewing-changes-to-your-project) and the standard [TextEdit.app](https://en.wikipedia.org/wiki/TextEdit) to edit the file (among web browser based editors: latest [Atom](https://atom.io/) works fine too, but old [Brackets](https://en.wikipedia.org/wiki/Brackets_\(text_editor\)) [1.8](https://github.com/adobe/brackets/releases/tag/release-1.8) don't — but in GitHub Desktop, [you can still select which changed lines to include in the commit](https://docs.github.com/en/free-pro-team@latest/desktop/contributing-and-collaborating-using-github-desktop/committing-and-reviewing-changes-to-your-project#creating-a-partial-commit) <sup>partial commit</sup> ).

nmdange2:
@i-to-r  Thanks I'll try redoing the commit in something other than a web browser

tempelmann:
Please do _not_ exclude .textClipping files by default.

Reason: Before macOS 10.12, textClippings were only stored in the rsrc fork, which isn't handled by git, and therefore would lead to a complete loss of information.
However, since 10.12, newly created .textClippings are _also_ stored in the data fork, in binary plist format, containing the same keys (types) and values of the rsrc entries of ID 256.
And macOS does, when opening (double clicking) these files, show the correct contents. Only quicklook does not work with the data fork but still (even in macOS 12.1) look only at the rsrc fork's data.
Which means, that if you restore a .textClipping file from git, the data fork will still be there, containing all the information (allowing to view the clipping by double clicking it), but which would also allow the restoring of the rsrc fork data to make quicklook work again.
In fact, I'm currently working on a little tool to do exactly that:
1. If a textclipping has data only in its rsrc fork (data fork being empty), it'll copy the clip rsrc content into the data fork. This would be performed on _old_ clips created pre-10.12, so that git will then preserve the contents in the data fork.
2.If a textclipping has no rsrc fork but a data fork, it restore the rsrc fork accordingly. This would be done after restoring from git, so that quicklook works again.

tempelmann:
OTOH, it would be much nicer if rsrc forks could also be handled by git, by using the ".. namedfork/rsrc" path method. That sub-path allows both reading and writing rsrc forks as regular files.

tempelmann:
There's some "new" behavior:

Recent macOS versions do actually use the data fork with .textClipping files. In fact, in my tests, the same information is stored both in the data and the resource fork. I even made a program (not made public yet) that can recreate the missing side.

So, with this knowledge, there's no need to exclude .textClipping files any more, because they _do_ now store content in the git repo, as long as the file have been created by a recent macOS version.

There remains one caveat, though: When restoring such a file from the repo onto a macOS system, some macOS features do not work, e.g. Quicklook can't show the textClipping's contents if it's stored only in the data fork while the data in the rsrc fork is missing. I've reported this as a bug to Apple in 2022 (FB9920034), and nothing has changed, of course. Still, someone could write code to restore the rsrc fork when restoring such a file, and then it would work fine.

If someone likes to tackle this, I'm happy to provide the code for the conversion from the data to the rsrc fork.

PR #3557, opened 2020-10-22
Ignore keystore.properties in Android.gitignore

OsinniyApps:

- _keystore.properties_ file is used for storing sensitive information such as java keystore password and key alias, which is used for signing Android apps. And this info should not be visible to others

- This is fully described in follow guide: https://developer.android.com/studio/publish/app-signing#secure-shared-keystore


53jk1 (approved):

willmeet:
do you lost your keystore of your android app ? 

Then the simple solution >
http://www.technologymanias.com/2020/10/missing-or-lost-keystore-of-your.html

OsinniyApps:
> do you lost your keystore of your android app ?
> 
> Then the simple solution >
> http://www.technologymanias.com/2020/10/missing-or-lost-keystore-of-your.html

It has nothing to do with this pull request

PR #3558, opened 2020-10-22
Remove Accio after official deprecation

Jeehut:

[Accio](https://github.com/JamitLabs/Accio) was oficially deprecated in favor of SwiftPM integration within Xcode. I don't think any new project will be opting for it, so it might also be a good idea to clean this up and remove the Accio related entries from the `.gitignore`.

https://github.com/JamitLabs/Accio/issues/88

### DIFF
--- a/Swift.gitignore
+++ b/Swift.gitignore
@@ -66,10 +66,6 @@ playground.xcworkspace
 
 Carthage/Build/
 
-# Accio dependency management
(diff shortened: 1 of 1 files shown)


ealeksandrov (approved):

wirecat (approved):
Thank you for contributing!

PR #3552, opened 2020-10-15
[SBT] Ignore .bsp directory

armanbilge:
As documented in the [release notes](https://github.com/sbt/sbt/releases/tag/v1.4.0), as of v1.4.0 sbt generates a .bsp directory for IDE support via the build server protocol.

PR #3561, opened 2020-10-28
Add .bsp directory to SBT.gitignore

aaabramov:

Since sbt 1.4.0 on every bootstrap sbt will create `.bsp/sbt.json` file that is environment specific.

Reference: https://contributors.scala-lang.org/t/build-server-protocol-in-sbt/4234

If this is a new template:

No.


armanbilge:
Duplicate of #3552 :)

PR #3562, opened 2020-10-29
feat(NuxtJS.gitignore): add support for NuxtJS

rlam3:
Currently no NuxtJS .gitignore file

https://github.com/nuxt/nuxtjs.org/blob/master/.gitignore

If this is a new template:

https://nuxtjs.org/


amirul473 (commented):
hello guys

rlam3:
@MattIPv4 I believe the original reason was to have a pure distinction between Node.js and Nuxt.js app. Similar to Vue.js

https://github.com/github/gitignore/blob/2ee4656dd865c162652d78806349da4fe173ec01/community/JavaScript/Vue.gitignore

Nuxt.js has certain modules we would like to ignore. Instead of having using a Node.js general .gitignore, we can use this as an extension of Node.gitignore following the Nuxt.js guidelines in their tutorials. Let me know what you think.

bdougie:
This is a duplicate of the Node.gitignore.

PR #3565, opened 2020-10-31
Add Django.gitignore

anurag-mandal042:

Missing Django gitignore template


bdougie:
This context already exists in https://github.com/github/gitignore/blob/master/Python.gitignore#L58

PR #3574, opened 2020-11-10
Added Exclusion for OS Thumbnails

ZoeDreams:
We should not included these, as they are autogenerated by Unity and the system.

Had to manually adjust the default .gitignore file which was generated by github. This creates issue when working between multiple system, and also triggers unnecessary indexing. Thank you.

https://superuser.com/questions/624897/stop-thumbs-db-files-in-certain-folders

### DIFF
--- a/Unity.gitignore
+++ b/Unity.gitignore
@@ -69,3 +69,7 @@ crashlytics-build.properties
 # Temporary auto-generated Android Assets
 /[Aa]ssets/[Ss]treamingAssets/aa.meta
 /[Aa]ssets/[Ss]treamingAssets/aa/*
+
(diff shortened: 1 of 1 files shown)


LiteLT:
The `.DS_Store` file is created on macOS only, which is already ignored in the global [gitignore file](https://github.com/github/gitignore/blob/master/Global/macOS.gitignore). Rather than ignoring it for all projects including this gitignore file, would it not make more sense to have a global `.gitignore_global` file in your home directory so it won't be added across all projects on your system?

ZoeDreams:
didn't know about that, thank you for the info. Yes that would be best.

PR #3575, opened 2020-11-12
Added supports JetBrains Rider.

kekyo:
Added supports JetBrains Rider.


kekyo:
It contains totally exclude .idea/ directory, because many open source repository makes better excluding many IDE related content/scratch files/directories.
I think it's better manner for commonly starting OSS project. And more useful template already contains into Global's JetBrains template if applicable.

elovelan:
this is already included and can be closed

kekyo:
I read and verified already inserted it, thx.

PR #3577, opened 2020-11-14
Added .DS_Store to Swift.gitignore

hoppsen:

The Swift.gitignore was not ignoring .DS_Store files.

n.a.

### DIFF
--- a/Swift.gitignore
+++ b/Swift.gitignore
@@ -2,6 +2,9 @@
 #
 # gitignore contributors: remember to update Global/Xcode.gitignore, Objective-C.gitignore & Swift.gitignore
 
+## General
+.DS_Store
+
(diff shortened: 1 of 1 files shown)


LiteLT:
The `.DS_Store` file is specific to macOS, which is already ignored as a [global gitignore file](https://github.com/github/gitignore/blob/master/Global/macOS.gitignore). I don't think it would be appropriate to ignore in all Swift projects.

hoppsen:
If you are not relying on a global `macOS.gitignore` file, it might be beneficial in here as well.

But I agree, that the ultimate goal might be to use a global `macOS.gitignore` file, plus the corresponding gitignore file based on the language you are working with.

PR #3578, opened 2020-11-17
adds VL cache to VVVV gitignore template

sebescudie:

VVVV now uses VL, a visual real-time programming language, that compiles dlls on the fly. Those dlls are added to `.vl/` folders and should not be versioned. 

This feature is not explicitly mentioned in the documentation. For more information on what VL is, see :

- [VL](https://betadocs.vvvv.org/using-vvvv/vl.html) in VVVV beta's documentation

### DIFF
--- a/VVVV.gitignore
+++ b/VVVV.gitignore
@@ -4,3 +4,6 @@
 
 # Dynamic plugins .dll
 bin/
+
+# VL cache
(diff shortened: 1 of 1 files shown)


LarsEck:
Would be helpful to have this merged.

PR #3513, opened 2020-08-30
Adding VS Code folders

ravenboilinux:
Reasons for making this change:

To add the folder and files that Unity generates for the VS Code IDE

**Links to documentation supporting these rule changes:**
[Link to Unity Doc about VS Code](https://code.visualstudio.com/docs/other/unity)

PR #3586, opened 2020-11-29
add vscode ignore

tasty0tomato:

add vscode ignore when using unity.

just like when I used vs.

### DIFF
--- a/Unity.gitignore
+++ b/Unity.gitignore
@@ -26,6 +26,9 @@
 # Visual Studio cache directory
 .vs/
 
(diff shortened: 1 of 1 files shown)


keveleigh (approved):

keveleigh:
I came here to make this change! It'd be great if this could be included.

keveleigh:
Looks like this change is also present in https://github.com/github/gitignore/pull/3467 and https://github.com/github/gitignore/pull/3513. Any will work!

PR #3587, opened 2020-12-01
Added .idea folder, to ignore AppCode's settings.

12q:
**Clean up Xcode project from .idea folder:**
When you switching beetween Xcode and AppCode, 
the last one updates original Xcode project with its own settings folder, that I don't want to see in my repo.

### DIFF
--- a/Global/Xcode.gitignore
+++ b/Global/Xcode.gitignore
@@ -22,5 +22,8 @@ DerivedData/
 *.perspectivev3
 !default.perspectivev3
 
+## AppCode's settings folder
+.idea/
+
 ## Gcc Patch
 /*.gcno
--- a/Objective-C.gitignore
+++ b/Objective-C.gitignore
@@ -30,6 +30,9 @@ DerivedData/
 *.dSYM.zip
 *.dSYM
 
+## AppCode's settings folder
+.idea/
+
 # CocoaPods
 #
 # We recommend against adding the Pods directory to your .gitignore. However
--- a/Swift.gitignore
+++ b/Swift.gitignore
@@ -30,6 +30,9 @@ DerivedData/
 *.dSYM.zip
 *.dSYM
 
+## AppCode's settings folder
+.idea/
+
 ## Playgrounds
 timeline.xctimeline
 playground.xcworkspace


LiteLT:
You're supposed to add the [JetBrains](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore) gitignore file's contents to your global gitignore file. It wouldn't appropriate to ignore it for all projects because not everyone uses AppCode or Xcode.

12q:
> You're supposed to add the [JetBrains](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore) gitignore file's contents to your global gitignore file. It wouldn't appropriate to ignore it for all projects because not everyone uses AppCode or Xcode.

It's only for the swift, obj-c & Xcode .gitignore profiles. Or I've missed something @LiteLT?

LiteLT:
I don't think it should be in the gitignore files for Swift, Objective-C, or Xcode because environment settings should be ignored globally. The `.idea` file is from IntelliJ-based IDEs, but IntelliJ is not the main purpose behind your project. So, global/community gitignore files exist for those kinds of files that aren't related to the project but are still good to ignore.

I believe it would be better to have a global `.gitignore_global` file in your user home directory to specify files you'd like to ignore across your system because not everyone will use an IntelliJ-based IDE to write Swift/Objective-C code, or in Xcode.

12q:
> I believe it would be better to have a global `.gitignore_global` file in your user home directory to specify files you'd like to ignore across your system because not everyone will use an IntelliJ-based IDE to write Swift/Objective-C code, or in Xcode.

So it will not affect them :) I guess there's a lot of people who're using both IDEs and it will be useful for community (coz AppCode is more flexible and Xcode you help you out to mange the xib and storyboards). Perhaps you have some ideas how to make it right. I'm open for any kind of suggestions, or you could just close this PR, if you think it will be useful.

LiteLT:
I see the benefit of not having to go add any other files that need disabling (by combining the two IDEs), but there are many instances where this can happen, hence there are two additional folders in the repository: `Global` and `community`. I believe, since not everyone is going to use AppCode (I know you have to have Xcode CLT), it shouldn't be added to this gitignore file because it's not going to have an effect on the project. Similar to how the `.vscode/` file isn't ignored for every file because you can use it almost everywhere.

dooleydevin:
We do not want to ignore the entirety of the `.idea` directory in any templates since there are many files in it that users may actually want to share in source control.

I'd recommend following @kyleerhabor's advice here and using [the JetBrains template](https://github.com/github/gitignore/blob/ceea7cab239eece5cb9fd9416e433a9497c2d747/Global/JetBrains.gitignore#L1) globally or pasting it into the other templates you are using.

Closing this out 👍

PR #3589, opened 2020-12-02
Unity.gitignore ignore meta files for already ignored unitypackages

Tymski:
Files that are not tracked should not have tracked metafiles.  
  
With this .gitignore I got a lot of   
image

### DIFF
--- a/Unity.gitignore
+++ b/Unity.gitignore
@@ -59,6 +59,7 @@ sysinfo.txt
 *.apk
 *.aab
 *.unitypackage
+*.unitypackage.meta
 
 # Crashlytics generated file
 crashlytics-build.properties


cscharfen:
Hi Tymski,

your observation is correct. Due to the blocking of *.unitypackage and because of some Unity Asset Store packages come with .unitypackage files, the related meta file needs to be ignored too. Else, they get added with the first commit and deleted with the next commit done by someone else or after cleaning the repository.

I see two options to solve this:
1) Either completely remove the "*.unitypackage" line from the gitignore (However, I guess this was already discussed at some point and the decision was made pro adding it to the gitignore at that time, which is fine.)
2) Or add the *.unitypackage.meta file as Tymski suggests (to provide consistency).

@shiftkey I recommend to accept this pull request or, in case you need more opinions, raise a discussion.

PR #3595, opened 2020-12-07
CMake: Add CMakeUserPresets.json

KyleFromKitware:

CMake 3.19 added CMakePresets.json, which is intended to be
version-controlled, and CMakeUserPresets.json, which should NOT
be version-controlled. Add CMakeUserPresets.json to the gitignore.

https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html

### DIFF
--- a/CMake.gitignore
+++ b/CMake.gitignore
@@ -9,3 +9,4 @@ install_manifest.txt
 compile_commands.json
 CTestTestfile.cmake
 _deps
+CMakeUserPresets.json


Maarrk:
Just dropping a comment from a user (me) that this seems fairly clear to review and I'd really like it, thanks in advance 👍 

Also, this file may get autogenerated by [conan](https://conan.io/), making the template more important for users who didn't expect it

Croydon:
The last PR was merged 2 years ago, does this repo need to get permanently forked or will it be maintained again?

Maarrk:
For future reference personally, I moved over to use https://github.com/toptal/gitignore

The issue in this PR is already resolved there, they also give reasoning behind the fork

petk:
+1 I was just browsing for some useful CMake gitignore template and it didn't have CMakeUserPresets.json in it. It seemed like a bit basic template until I've seen there are pull requests in the queue, so would be kind of good to merge this.

wirecat (approved):
Thank you for contributing! ❤️

PR #3601, opened 2020-12-19
added #self .gitignore in Python

aa-ag:

It's something I find myself doing over and over and over. Hopefully, it helps others.

N/A.

If this is a new template:

N/A.

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -1,3 +1,6 @@
+# Self
+.gitignore
+
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]


magnusbaeck:
> It's something I find myself doing over and over and over.

Okay, but why are you adding a .gitignore entry to the .gitignore file?

aa-ag:
I was looking @ https://stackoverflow.com/questions/767147/how-can-i-stop-gitignore-from-appearing-in-the-list-of-untracked-files and it seems like it is best practice to leave it there. However, I usually end up with a ton of junk files / cached stuff from my IDE. 

Would it be better to add this instead?

```
# VS Code
.vscode
.vscode/*
!.vscode/settings.json
!.vscode/tasks.json
!.vscode/launch.json
!.vscode/extensions.json
*.code-workspace
```

drothmaler:
As the top answer (with 900+ Upvotes) in your cited SO question states:

> The `.gitignore` file should be in your repository, so it should indeed be added and committed in, as `git status` suggests. It has to be a part of the repository tree, so that changes to it can be merged and so on.

So adding itself to a `.gitignore` file makes no sense at all. 
If you really want to ignore some files only locally, you should either put them in the `.git/info/exclude` file in your project, or place them into your [global gitignore](https://git-scm.com/docs/gitignore): 
https://docs.github.com/en/free-pro-team@latest/github/using-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer

aa-ag:
Sounds good.

PR #3602, opened 2020-12-24
Update VisualStudio.gitignore

brettrowberry:

- Explains that this .gitignore is used for both Visual Studio and Visual Studio Code. 
- Includes the date last updated to make it easier to determine when a local copy was made.


brettrowberry:
Ping!

martinwoodward:
The date idea is a good one. Keeping it up to date with each revision without causing lots of merge conflicts for PR's would be an issue though. Thanks for the suggestion however

PR #3609, opened 2020-12-30
Fixed editor specific folder exclusion for subfolders

xperiandri:

If an .sln file is located in a repository subfolder then editor specific folders are not covered by the current gitignore rules


wirecat (approved):

PR #3616, opened 2021-01-07
Update Android.gitignore

crazygo:

*.apks is generated by BundleTools

https://stackoverflow.com/questions/52059339/difference-between-apk-apk-and-app-bundle-aab


bdougie:

PR #3618, opened 2021-01-07
add IntelliJ IDE content

dotrinhdev:

_TODO_

_TODO_

If this is a new template:


simaki:
Thank you @dotrinhdev !

I'm afraid we don't have to add these files because they are already included or supposed to be included in [Global/JetBrains.gitignore](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore).

cf. #3451

PR #3619, opened 2021-01-07
Add direnv .envrc to python gitignore

fslds:

The direnv project (https://direnv.net/ ) uses `.envrc` files for dynamically loading environment variables (or other settings) per directory using the user's shell... As it can contain sensitive information, and is similar in goal as to dotenv etc, I think it is a good idea to exclude this by default as well into popular languages that already can use a 'dotenv' `.env` file as well. Adding this to the python `.gitignore`. 

 * https://direnv.net/
 * https://github.com/direnv/direnv
 * http://manpages.ubuntu.com/manpages/xenial/man1/direnv.1.html

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -106,6 +106,7 @@ celerybeat.pid
 
 # Environments
 .env
+.envrc
 .venv
 env/
 venv/


garrison:
One might want to intentionally add an `.envrc` file to the repository, e.g., one that includes [the single line](https://direnv.net/man/direnv-stdlib.1.html#codelayout-python-ltpythonexegtcode) `layout python3`.  In this case, `.direnv` should be added to `.gitignore`, but `.envrc` shouldn't be.

In fact, it is safe to exclude `.direnv` in all cases.  I came here to submit a pull request that adds `.direnv` to `.gitignore` for python (really it could be added for every single language).  But I don't believe that `.envrc` should be added.

garrison:
In other words, while `.envrc` *can* contain secrets, that is not the sole purpose of direnv.

fslds:
That is actually correct, but the likelihood is large that a secret ends up in the .envrc file,  and this is an effort to increase overall security. 
By adding it to the general .gitignore, - and yes it should be added for each language -  it needs to be uncommented explicitly, making it a conscious action...
That being said I would recommend using the `dotenv` command or `source .env`  in `.envrc` and store secrets in the more commonly used `.env`.
I switched to that workflow myself in the meanwhile 😇 .

garrison:
In case my 👍 was overlooked... I am in favor of both `.envrc` and `.direnv` being included in got ignore.

wirecat (approved):
Thank you for contributing!

PR #3621, opened 2021-01-12
Ignore vim swap files

mTvare6:
Vim is **still** used by many people, this ignores the swap files when using git

[reason for swap files](https://vi.stackexchange.com/questions/177/what-is-the-purpose-of-swap-files) say vim creates a swap file when it is running so during unexpected closing it doesn't lose stuff.


rugk:
This is likely better suited for the `Global` ignores…

PR #3623, opened 2021-01-17
Autotools.gitignore: ignore config.h

J-M0:

`config.h` is one of the files generated by `configure` and should not be tracked

https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.70/html_node/Configuration-Headers.html#Configuration-Headers

### DIFF
--- a/Autotools.gitignore
+++ b/Autotools.gitignore
@@ -17,6 +17,7 @@ autom4te.cache
 /aclocal.m4
 /compile
 /config.guess
+/config.h
 /config.h.in
 /config.log
 /config.status


bdougie:
The documentation does not state any concern about tracking or storing this in a gitignore. Can you provide a link specific to that?

J-M0:
I can't find a document that explicitly says `config.h` should not be committed, but `config.h` is a file generated from `config.h.in` and this link talks about the pitfalls of committing generated files into version control.

https://www.gnu.org/software/automake/manual/html_node/CVS.html

wirecat:
👋 This feels a bit specific. I've personally worked on projects where it was expected that you commit the config.h.
Closing for now as we want to keep this as broadly applicable as possible.

PR #3627, opened 2021-01-26
Add file from common packaging tool.

jwg4:
Poetry creates poetry.lock - typically it is not checked in.

This is a commonly used tool for creating Python packages.

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -136,3 +136,6 @@ dmypy.json
 
 # Cython debug symbols
 cython_debug/
+
+# poetry packaging tool
+poetry.lock


simaki:
Thank you @jwg4 .

I'm afraid `poetry.lock` should not be added because it is supposed to be checked in.
Please see [Poetry's Documentation](https://python-poetry.org/docs/basic-usage/#installing-without-poetrylock).

> You should commit the poetry.lock file to your project repo so that all people working on the project are locked to the same versions of dependencies (more below).

jwg4:
you are of course right, thanks for taking a look at the PR and sorry for the error!

PR #3629, opened 2021-01-27
Fix Rust.gitignore rules for build artifacts

clmay:

The changes made in #3436 are incorrect and should be reversed. Cargo, the de facto build tool for Rust, always outputs build / compilation artifacts to `target/` in the root of the crate (project), by default (see [the relevant portion](https://doc.rust-lang.org/cargo/guide/build-cache.html) of the Cargo book). We should target Cargo's defaults and allow other users to override them if they have customized the configuration of their builds.

The correct gitignore rule for excluding the `target` directory is

```
/target
```

since we want to ignore the top-level directory (only), as well as all of its children. Since the `debug` directory is itself always a child of `target` under the default Cargo configuration, excluding only `target` is sufficient for also excluding the `debug` directory that the previous committer added to the rules. Several other commenters in #3436 also pointed this out, but after the changes were already merged.

Importantly, the rules used in #3436 are **incorrect**, because they can exclude directories/files in `src`, _e.g._:

- `src/target/some_file.rs` or 
- `src/components/debug/`

which are _actual source files for the project_. Scoping the exclusion specifically to the top-level `target` directory generated by Cargo is a safer default.

---

tl;dr: We always know (absent custom configuration) that the `target` directory at the project root will contain all of Cargo's build and compilation artifacts, and that `/target` in the .gitignore is the sole safe default that will exclude them from git tracking correctly and safely; the same cannot be said for the current .gitignore rules, which could very well ignore actual source code in a project (and in a way that could be easy to miss / very unexpected).

Aside: the above rules will always be adequate for a "vanilla" Cargo project without custom config and should be considered the community defaults. Changes for customization or third-party tooling probably shouldn't be added here.

---

https://doc.rust-lang.org/cargo/guide/build-cache.html


clmay:
@shiftkey I don't seem to be able to add a reviewer, could you have a look at this when you've time?

clmay:
@shiftkey ?

gracehop (approved):

clmay:
Thanks, @gracehop !

clmay:
Ready to merge.

clmay:
P.S. just a note to project maintainers:

I was unable to

1. Tag anyone for review on this PR
2. Merge the PR after approval

Perhaps this is because I opened the PR from a fork of the repo, rather than cloning the repo directly and branching within it?

Certseeds:
> P.S. just a note to project maintainers:
> 
> I was unable to
> 
> 1. Tag anyone for review on this PR
> 2. Merge the PR after approval
> 
> Perhaps this is because I opened the PR from a fork of the repo, rather than cloning the repo directly and branching within it? Or perhaps it was desirable to have such restrictive settings on the repo, I'm not sure.
> 
> In any case, this could be merged, anytime, by anyone who has the ability to do so!

Anyone can modify a public repo is absolutely a bad idea.

PR #3636, opened 2021-01-29
Update Images.gitignore

joonykims:
Adds several extensions.

# JPEG 2000
*.jpc

# High Efficiency Image File Format
*.heif

# Adobe PhotoShop Document
*.psd

_TODO_

_TODO_

If this is a new template:

### DIFF
--- a/Global/Images.gitignore
+++ b/Global/Images.gitignore
@@ -13,6 +13,7 @@
 *.jpx
 *.jpm
 *.mj2
+*.jpc
 
 # JPEG XR
 *.jxr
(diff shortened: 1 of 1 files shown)


Global/Images.gitignore
```diff
@@ -61,3 +62,9 @@
 # 3D Images
 *.3dm
 *.max
+
+# High Efficiency Image File Format
+*.heif
+
+# Adobe PhotoShop Document
+*.psd
```
MattIPv4, line 70: Open Q: Is this the best place to exclude this? I'd consider this to be a "source" file for an image, and I notice that .ai files etc. aren't currently ignored here.

bdougie:
Should this be leveraged from the global gitignore instead? 

PR #3647, opened 2021-02-10
Delete fixed workaround for Gradle

thatrobotdev:

The workaround for https://youtrack.jetbrains.com/issue/IDEA-116898 is fixed now, so we can remove the workaround.

See: https://github.com/github/gitignore/pull/1856#issuecomment-453354946

* https://youtrack.jetbrains.com/issue/IDEA-116898


martinwoodward:
Thanks for looping back on this @thatrobotdev ✨

PR #3649, opened 2021-02-13
Update for PuppeteerSharp

binn:

Many developers using PuppeteerSharp may accidentally push their local chrome install up to their git repo (which is like 200mb of files that goes into the commit log).


martinwoodward:
Trying to decide between the additional maintenance issues this bring with the pain that this could potentially save for the small percentage of VisualStudio.gitignore users who would otherwise run into this.  Thoughts VS folks?

modnxs88 (changes requested):
Looks like bs

PR #3650, opened 2021-02-15
Updated Chef Cookbook with defaults and VSCode/Bundler exclusions

Taegost:

The Chef Cookbook .gitignore was out-of-date from the recommended default settings from Chef, and did not have exclusions for some of the more common temp/extraneous files that don't need to be included, such as VS Code and Bundler temp files.
I also included a starter OpenSCAD gitignore

I have been unable to locate documentation from Chef, but the majority of this comes straight from the default .gitignore created when generating a new cookbook through the Chef Workstation CLI.
These are standard exclusions to ensure that temporary files and "compiled" files are not checked into the repo as they are generated at runtime and should not be included.

If this is a new template:

 - **Link to application or project’s homepage**: [https://docs.chef.io/workstation/](url)


wirecat:
Submissions should only edit 1 template a time. Please open up changes individually if you'd still like to pursue them.

PR #3651, opened 2021-02-15
Ignore `.env.production` in Node.js

kachick:

See https://github.com/search?q=filename%3A.env.production

スクリーンショット 2021-02-16 3 02 47

Then I saw some configurations they contain actual DB passwords, AWS secrets... (So I sent 2 PRs and 1 repository is deleted immediately...)

* I don't know other languages use `.env.production` or not.
* As far as I know, JavaScript projects often have these files.  (e.g. https://github.com/gatsbyjs/gatsby/blob/10dfe011c368e28e8de19f0f5569748ebb9a6bc3/benchmarks/source-contentful/README.md#build-a-site)

I 🙏 this change reduces sad accidents. 

---

FYI: `Rails.gitignore` already has same config from another issue https://github.com/github/gitignore/pull/3065

### DIFF
--- a/Node.gitignore
+++ b/Node.gitignore
@@ -70,7 +70,7 @@ web_modules/
 
 # dotenv environment variables file
 .env
-.env.test
+.env.*
 
 # parcel-bundler cache (https://parceljs.org/)
 .cache
--- a/Node.gitignore
+++ b/Node.gitignore
@@ -70,7 +70,8 @@ web_modules/
 
 # dotenv environment variables file
 .env
-.env.*
+.env.test
+.env.production
 
 # parcel-bundler cache (https://parceljs.org/)
 .cache


Node.gitignore
```diff
@@ -70,7 +70,7 @@ web_modules/
 
 # dotenv environment variables file
 .env
-.env.test
+.env.*
```
kachick, line 73: This change ignores `.env.example`. So just adding `.env.production` might be better?

bl-ue: Yes, I think so.

bl-ue (approved):
Really, **really** good idea 😬 

I just opened 3 issues on repos where I found similar things.

BTW it's already 32,996 `.env.production`s now haha :smile:

Node.gitignore
bl-ue: I've also seen  `.env.sample`.  So @kachick please apply the above suggestion.  We only want to ignore `.env.production`.

bl-ue: Also, why not add `.env.test`?   Some people may use e.g AWS for testing as well as production and have secret keys for them as well.

kachick: Thanks for your review! Updated in https://github.com/github/gitignore/pull/3651/commits/b236e1717b90f1755af76900f5387bfa20636de3 !

MattIPv4 (approved):
Lgtm, seems sane

kachick:
Thank you all!

PR #3654, opened 2021-02-16
Added .vscode  in python.gitignore

abhi7585:
While using vs code when choosing an interpreter for python, a ".vscode" file is generated and I notice that this is not included in the gitignore file. But I'm recommending to add this to the list. There might be related cache files for other editors, but I just use vs code.

_TODO_

_TODO_

If this is a new template:

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -136,3 +136,6 @@ dmypy.json
 
 # Cython debug symbols
 cython_debug/
+
+# vs code cache
+.vscode 


bl-ue:
@abhi7585 - `.vscode` does not contain a cache.  It contains VS Code settings.  `.vscode/settings.json` is getting created when you set your Python interpreter because VS Code stores your choice in that file.  See https://github.com/github/gitignore/pull/3661#issuecomment-787093820

If you really want to ignore `.vscode`, add the line to your own `.gitignore`.  More users would prefer to keep it that to not.

PR #3659, opened 2021-02-20
Add ignoring editors files

Z4RX:

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -136,3 +136,7 @@ dmypy.json
 
 # Cython debug symbols
 cython_debug/
+
+# Editors
+.vscode
+.idea


bl-ue:
Please remove `.vscode`.  See https://github.com/github/gitignore/pull/3661#issuecomment-787093820

Z4RX:
Probably Adding only necessary files is better solution?
```
.vscode/*
!.vscode/settings.json
!.vscode/tasks.json
!.vscode/launch.json
```

bl-ue:
Hmm...well so far I've only seen `*.json` configuration files in `.vscode/`.  Are you seeing anything different?

simaki:
Thank you!

I'm afraid these files do not have to be added because they are already included or supposed to be included in
* [Global/JetBrains.gitignore](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore)
* [Global/VisualStudioCode.gitignore](https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore)

PR #3661, opened 2021-02-22
Added .vscode/ to unity gitignore

mikhailb1990:

VS code workspace settings not needed to be shared

### DIFF
--- a/Unity.gitignore
+++ b/Unity.gitignore
@@ -25,6 +25,7 @@
 
 # Visual Studio cache directory
 .vs/
+.vscode/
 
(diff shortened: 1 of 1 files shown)


bl-ue:
I don't recommend this.  `.vscode` _can_ contain many editor settings and is more and more checked in to version control.

simaki:
Thank you @Matin-B !

I'm afraid these files are already included or supposed to be included in [Global/VisualStudioCode.gitignore](https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore).

PR #3663, opened 2021-02-22
[R] Drop renv-defined library from R projects

russHyde:
Renv can be used to manage the environment for R projects. The ./renv/library directory that is produced when working with renv should not typically be added to version control. 
Please see: https://rstudio.github.io/renv/articles/renv.html#infrastructure-1

This removes the local library directory generated by the R-specific env-management tool 'renv'.

https://rstudio.github.io/renv/articles/renv.html#infrastructure-1

### DIFF
--- a/R.gitignore
+++ b/R.gitignore
@@ -41,5 +41,8 @@ vignettes/*.pdf
 # pkgdown site
 docs/
 
+# renv project-library directory
+renv/library/
+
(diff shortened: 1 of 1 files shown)


russHyde:
Sorry, just realised that renv adds it's own .gitignore into the ./renv/ directory. This changeset is of no value.

PR #3665, opened 2021-02-26
Update Unity.gitignore

Morgan-6Freedom:
Add .idea/ for the Rider cache directory. it's exactly the same interest as the visual studio cache directory.

Because Rider is used by a lot of Unity Developers and has a cache directory too. Like Visual Studio.


CyberFlameGO (approved):

bl-ue (approved):

dustingraham:
I always manually add this, would be a good addition.

thatrobotdev:

martinwoodward:
Duplicate of #3529

PR #3666, opened 2021-02-27
Ignore VSCode project settings

Matin-B:

Add VSCode project settings


simaki:
Thank you @Matin-B !

I'm afraid these files are already included or supposed to be included in [Global/VisualStudioCode.gitignore](https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore).

PR #3672, opened 2021-03-07
Update Unity.gitignore

AttalaKheireddine:
This is just adding the /Temp folder

It is very annoying to work without this addition as you always need to close unity before making commits, I was unaware of why that was a thing for the longest time and it made working with git and Unity quite the pain. Please  accept this change to make the life of some fellow git ignorants (or should I saw gitignore ignorants) easier

KH!

_TODO_

_TODO_

If this is a new template:

### DIFF
--- a/Unity.gitignore
+++ b/Unity.gitignore
@@ -9,6 +9,7 @@
 /[Bb]uilds/
 /[Ll]ogs/
 /[Uu]ser[Ss]ettings/
+/Temp
 
 # MemoryCaptures can get excessive in size.
 # They also could contain extremely sensitive data


18mkt (approved):
Duplicate of #

SahilMayenkar:
Line 6 of the existing template already specifies the Temp folder. See [here](https://github.com/github/gitignore/blob/master/Unity.gitignore#L6).

CertifiedRice (approved):

bl-ue:
Ah yes, @SahilMayenkar is correct.  This PR should be closed @AttalaKheireddine, because line 12, which you added, `/Temp`, is already encapsulated by line 6, `/[Tt]emp/` (which also takes care of `/temp`).

PR #3675, opened 2021-03-10
Update the file for ignoring .vscode folders

ToniIvars:

I have a project in which I setted up a virtuel environment, so I thought I could update this .gitignore file as I have edited my own .gitignore file.


ToniIvars:
I have just seen a pull request opened with the same objective as mine, so I cose this one.

PR #3680, opened 2021-03-15
Add Azure Functions localsettings file

matt40k:
Add Azure Functions localsettings file to the visualstudio gitignore.

This is included in the Azure Function project template.

Ignore local settings

Part of the Azure Function project template.

If this is a new template:

N\A

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -360,3 +360,6 @@ MigrationBackup/
 
 # Fody - auto-generated XML schema
 FodyWeavers.xsd
+
+# Azure Functions localsettings file
+local.settings.json


martinwoodward:
I see some debate about the issues of not checking this file in to source control. https://www.tomfaltesek.com/azure-functions-local-settings-json-and-source-control/

Would it be possible to send along more information about why this is definitely the way to go?

Thanks!

matt40k:
Configuration files shouldn’t be commit to version control. The argument for including it is to help with initial developer setup, this can be achieved with templates and info in the readme file rather than making assumptions

phocodes (approved):

MisinformedDNA:
By this same logic, we should block all appsettings*.json files. (NOTE: We shouldn't.) The ASP.NET team does not try to ignore these files, and the Azure Functions team is simply wrong. Part of the problem they were having was that some developers were pushing the `local.settings.json` file to Azure and expecting it to work. We should not merge this or at least merge it commented.

PR #3681, opened 2021-03-18
Update Unity.gitignore (Added rule for Rider cache directory)

Zeejfps:
Added rule for Rider

Rider generates a .idea/ directory which contains cached files that probably should not be committed

I don't have any specific documentation, but for the same reason, the .vs/ folder is excluded rider's .idea/ folder should be as well.


thatrobotdev:

PR #3684, opened 2021-03-24
Add Vlang Support By Vlang Community

AliChraghi:
**Vlang** Documention:
https://github.com/vlang/v/blob/master/doc/docs.md

**Vlang** Repo:
https://github.com/vlang/v


Larpon:
Could we remove the `*.obj` ? It's also a used by Wavefront OBJ files (3D model format)

Please see recent incident [here](https://github.com/vlang/v/pull/9456):
Screenshot_20210325-065459_Firefox

Larpon:
Thanks! No further issues as I see it!

Larpon (approved):

martinwoodward:
Looks good.  I've moved it into the community folder to begin with.

community/V.gitignore
```diff
@@ -0,0 +1,14 @@
+*
+!*/
+!*.*
```
leighmcculloch, line 3: These three lines are problematic and I don't think they should included in a default gitignore configuration. While they are somewhat effective at ignoring binary files on non-windows systems that don't have an extension, they will also ignore other files like Makefiles, and also override any users global gitignore configuration which can lead to engineers accidentally committing secrets or IDE files that they exclude using global gitignore configurations. I've opened a PR proposing they be removed: https://github.com/github/gitignore/pull/3727

PR #3685, opened 2021-03-24
Removed nonsense from VisualStudio.gitignore

CoufalikJaroslav:
Backup*/ Not sure it is a good idea to have Backup*/ in the ignorelist. E.g  Projects can have names as BackupJob. BackupMonitor, Backup....

_TODO_

_TODO_

If this is a new template:

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -254,7 +254,6 @@ Generated_Code/
 # to a newer Visual Studio version. Backup files are not needed,
 # because we have git ;-)
 _UpgradeReport_Files/
-Backup*/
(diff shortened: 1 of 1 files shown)


aruanoguate:
I do not recommend to completely remove it, because Visual Studio creates these `Backup` folders each time a project is migrated from an older version of Visual Studio, and these folders are not really needed in source control.

What about changing the rule to remove the `*` and just leave something like what I show below, this way the rule will no longer catch the examples you provided: BackupJob. BackupMonitor, etc.

```
[Bb]ackup/
```

CertifiedRice (approved):

PR #3687, opened 2021-03-28
Unity.gitignore update for Rider cache directory

CumaliSAH1N:

Rider creates cache directory called *.idea* in Unity project directory and it was not in the template. I thought it would be good idea to add include it in the template since it's being used by a lot of developers. 

_TODO_

If this is a new template:


thatrobotdev:

martinwoodward:
Dupe of #3529

PR #3691, opened 2021-03-31
update env files ignore only local

Aysnine:

`.env` is necessary.

https://cli.vuejs.org/guide/mode-and-env.html#environment-variables


Mandroide (approved):
It looks great!

bdougie:
I think you may be misinterpreting the documentation. It is common practice to not include .env files in git. They are indeed necessary but you should avoid saving them in version control.

PR #3692, opened 2021-04-01
Update Unity.gitignore

safaiyeh:

macOS generates a .DS_Store locally for OS level attributes.

https://www.jeffgeerling.com/blogs/jeff-geerling/stop-letting-dsstore-slow-you

### DIFF
--- a/Unity.gitignore
+++ b/Unity.gitignore
@@ -71,4 +71,4 @@ crashlytics-build.properties
 /[Aa]ssets/[Ss]treamingAssets/aa/*
 
 # macOS local file
-.DS_Store
+**/.DS_Store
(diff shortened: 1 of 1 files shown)


Almamu:
If you don't want .DS_Store files in your projects, then you should be using this gitignore in addition to the Unity one: https://github.com/github/gitignore/blob/master/Global/macOS.gitignore

Take a look at the global folder for more information.

safaiyeh:
> If you don't want .DS_Store files in your projects, then you should be using this gitignore in addition to the Unity one: https://github.com/github/gitignore/blob/master/Global/macOS.gitignore
> 
> Take a look at the global folder for more information.

Ah this is cool. Thanks! will close

PR #3693, opened 2021-04-01
Adding visual studio code config folder to the visual studio gitignore

iamnicoj:

I always end up adding this to my .gitignore file in all my vscode projects. You may argue that vscode is beyond the scope the visual studio realm. However it still has visual studio on its name.

A VS Code "workspace" is usually just your project root folder. Workspace settings as well as debugging and task configurations are stored at the root in a .vscode folder.

If this is a new template:

NA


martinwoodward:
Fixed in #3722

PR #3698, opened 2021-04-08
Add local configuration files to Vim.gitignore

Farzat07:
When using coc.nvim (https://github.com/neoclide/coc.nvim), sometimes a
local configuration file is created at .vim/coc-settings.json, so it was
ignored.

I am not sure whether this directory is used by coc only or other
packages as well, so for now I think it is better to just ignore the
whole .vim/ directory.

A link to the documentation of coc's local configuration file:
https://github.com/neoclide/coc.nvim/blob/ba262ef03029c3fd6a5062bebce7d8553e3b1a1f/doc/coc.txt#L2230

### DIFF
--- a/Global/Vim.gitignore
+++ b/Global/Vim.gitignore
@@ -17,3 +17,6 @@ Sessionx.vim
 tags
 # Persistent undo
 [._]*.un~
+
+# Local configuration files
+.vim/


QWERTYail (approved):

martinwoodward:
Are .vim folders commonly created inside directories too?  I could only find reference to them being used in your profile directory.

Farzat07:
I don't really get your question.

martinwoodward:
Trying to figure out why the entire .vim folder is being excluded.  Does this file exist outside of the root of a users profile folder or is it sometime stored in source folders?

CertifiedRice (approved):

Farzat07:
Actually thinking about it, ignoring the .vim/ directory can be very damaging for someone who uses git to keep track of his home directory.

The .vim/coc-settings.json is a specific case which does not necessarily apply to the normal user.

I'm thinking of closing the pull request, is it ok?

bl-ue:
>is it ok?

PR #3699, opened 2021-04-12
Create template for .NET

mu88:

A template for .NET does not exist at the moment. Since there are so many .NET projects on GitHub, there should be a template.

[This is one `.gitignore` file from the .NET Core team at Microsoft](https://github.com/dotnet/core/blob/main/.gitignore)

If this is a new template:

 - **Link to application or project’s homepage**: https://github.com/dotnet/core/blob/main/.gitignore

### DIFF
--- a/DotNet.gitignore
+++ b/DotNet.gitignore
@@ -1,0 +1,35 @@
+# Package management
+nupkg/
+packages/
+
+# Visual Studio Code
+.vscode
+
+# Rider
+.idea
+
+# Visual Studio
+.vs/
+
+# User-specific files
+*.suo
+*.user
+*.userosscache
+*.sln.docstates
+*.sln.DotSettings.user
+
+# Build results
+[Dd]ebug/
+[Dd]ebugPublic/
+[Rr]elease/
+[Rr]eleases/
+x64/
+x86/
+build/
+bld/
+[Bb]in/
+[Oo]bj/
+[Oo]ut/
+msbuild.log
+msbuild.err
+msbuild.wrn


martinwoodward:
Right now folks tend to use the VisualStudio one for .NET development which I know is a bit odd but what happens.  Thoughts on this @timheuer / @terrajobst?

timheuer:
I agree here, the VisualStudio.gitignore encompasses the .NET perspective here plus tools/ecosystem.

mu88:
So maybe we should rename the existing template from "Visual Studio" to ".NET"? Because it contains a lot of thing that are unrelated to Visual Studio. Furthermore, .NET nowadays is so much more than only Visual Studio.

martinwoodward:
Possibly, but then that would break the older clients who look for that file.

I do see a lot of value in breaking up the VS one into editor specific and then a different set of exclusions
per project type however I don't see any evidence of that happening. 

Closing for now but happy to revisit if there is a PR in the dotnet cli looking to change the built in command or if there are proposals to break up the VS one into something easier to maintain.

bl-ue:
>Possibly, but then that would break the older clients who look for that file.

Also, it would break a lot of existing links.  I use `VisualStudio.gitignore` every week and I keep a link to the original file in `.gitignore`. 🤔

timheuer:
> I do see a lot of value in breaking up the VS one into editor specific and then a different set of exclusions
> per project type however I don't see any evidence of that happening.

I think until we see tools enabling better 'merging' of existing .gitignore with additional values as new tools are used for projects.  Until that happens (doesn't appear any tool vendors lean this direction), we have stronger signals at least for .NET developers that they use a myriad of tools/CLI/git tools and across multiple people working on projects.

Consider the scenario of an OSS project that maybe used a "CLI only" .gitignore and then others fork but use Visual Studio, Rider, or VS Code.  The CLI version wouldn't be enough and these other tools might contribute back to the project garbage.  We have seen this happen, hence favoring more verbosity for a file that honestly once created should serve it's purpose and not bother anyone :-).  Just adding this comment as at least one of the tool vendors who has heard the feedback.

terrajobst:
Agreed. I used to hand author these files and was striving for minimalism as some kind of engineering pride. At first, using the standard VS .gitignore felt like letting a huge mess into my repo. But since I’ve used them in many repos I’ve come to appreciate that having a good ignore file means fewer mistakes with people committing a mess and much less baby sitting for chasing standard excludes. I prefer the single one stop shop now and I’ve never looked back.

PR #3700, opened 2021-04-13
Update ArchLinuxPackages.gitignore

adocampo:
Added zst package extension

_TODO_

_TODO_

If this is a new template:


adocampo:
Perhaps I was wrong. I guessed as other package files were included, the new arch linux package standard would be a good addition. But double checking the file, the old .xz format wasn't present either, so there is no reason to add this either.

PR #3701, opened 2021-04-13
Update Coq.gitignore

JasonGross:
Coq now uses .mlg rather than .ml4 (since https://github.com/coq/coq/pull/8763), so we have to ignore its generated dependency files.  The `native_compute_profile_*.data` files are generated by `Set NativeCompute Profiling` (see https://github.com/coq/coq/pull/950).  Finally `.coq-native` is a directory that may be generated in any subdirectory, not only at top level, so we should not use absolute paths for it.

### DIFF
--- a/Coq.gitignore
+++ b/Coq.gitignore
@@ -10,6 +10,7 @@
 *.glob
 *.ml.d
 *.ml4.d
+*.mlg.d
 *.mli.d
 *.mllib.d
 *.mlpack.d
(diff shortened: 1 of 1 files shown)


JasonGross:
Can someone merge this?  Is this blocked on something?

bdougie:
Sorry for the delay on this @JasonGross. Thanks for your contributions.

PR #3702, opened 2021-04-20
Windows: Exclude zone identifier files.

jtagcat:
https://docs.microsoft.com/en-us/previous-versions/windows/internet-explorer/ie-developer/platform-apis/ms537183(v=vs.85)

tl;dr 'where the heck did this file come from'.  
Note the files do not only apply to IE downloads.

Windows adds the files, they are unwanted to be synced/added.

### DIFF
--- a/Global/Windows.gitignore
+++ b/Global/Windows.gitignore
@@ -22,3 +22,6 @@ $RECYCLE.BIN/
 
 # Windows shortcuts
 *.lnk
+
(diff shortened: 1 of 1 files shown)


bdougie:
Thanks for the PR, please provide more context on why this is needed.

bl-ue:
I agree with @jtagcat.  I once committed one of these files a long while ago, to my disgust 😝

PR #3705, opened 2021-04-21
add scala jvm crash logs

hajlaoui-nader:
add in scala  jvm crash logs

http://www.java.com/en/download/help/error_hotspot.xml


bdougie (approved):
LGTM

PR #3706, opened 2021-04-21
Add nuget.config file

parsalotfy:
Add nuget.config file because it may contain secret data and should be kept localy

This file may contain tokens and passwords which are secret data. They shouldn't be pushed to github.

_TODO_

it can be seen here [in this tutorial](https://docs.github.com/en/packages/guides/configuring-dotnet-cli-for-use-with-github-packages#authenticating-with-a-personal-access-token) that this file may contain secret data.

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -205,6 +205,9 @@ PublishScripts/
 *.nuget.props
 *.nuget.targets
 
+# Nuget personal access tokens and Credentials
+nuget.config
+
 # Microsoft Azure Build Output
 csx/
 *.build.csdef


martinwoodward:
✨  Thanks

Ninja2:
Just broken some of my projects because of this today 😄. It contains far more important data than secrets, mainly which package sources you are using.

parsalotfy:
> Just broken some of my projects because of this today 😄. It contains far more important data than secrets, mainly which package sources you are using.

Maybe you are right, but you could insert those package sources in .csproj file hmmm ? 🤔
What kind of project is that ?

MisinformedDNA:
Package sources do not go into the csproj. The nuget.config is the correct place. All NuGet customizations go into this file. These customizations are then shared with your team. Secrets are not a required and are _optional_ if you are using a private package source (there are other ways and isn't terribly applicable for open source). Can we revert this?

parsalotfy:
> hi parsalotfy, can you give an example of secret data ?

Hi @oelderinkX , sorry for the late reply, I don't know why I didn't saw that sooner :/
In the link I provided with my Issue, there is a config sample containing secret key, so I taught it would be a good idea to add this file in gitignore file.

parsalotfy:
> Package sources do not go into the csproj. The nuget.config is the correct place. All NuGet customizations go into this file. These customizations are then shared with your team. Secrets are not a required and are _optional_ if you are using a private package source (there are other ways and isn't terribly applicable for open source). Can we revert this?

Hi @MisinformedDNA ,
Thank you for your comment, it seems the drawbacks of the change I've done overcome the profits.
Sorry for the inconvenient, I'll try to revert that.

robertlarkins:
As an added link, Microsoft's documentation says add the NuGet.Config: https://docs.microsoft.com/en-us/nuget/consume-packages/packages-and-source-control

parsalotfy:
> As an added link, Microsoft's documentation says add the NuGet.Config: https://docs.microsoft.com/en-us/nuget/consume-packages/packages-and-source-control

Thank you for your information, so I closed #3757.

PR #3708, opened 2021-04-22
Ignore Rider IDE files

MikD1:
Ignore internal Rider IDE files


Almamu:
Doing a blanket ignore on that folder is not a good idea. There's legitimate reasons you might want to include part of the .idea folder, like for example, ensuring all the developers have the proper coding style settings for it's linter. On top of that, there's already a JetBrains specific .gitignore that can take care of this: https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore

PR #3715, opened 2021-04-29
Added missing ignore on VisualStudio.gitignore.

AraHaan:

When opening an .NET Repository in Finder or even in JetBrains Rider on mac these generated files generate unneeded noise and should be ignored by every repository in the future that uses this .gitignore.

This feels like a common thing to me on this.

Bonus Points: for all the repositories that use this .gitignore for some bot that can autosync their .gitignore with the one from here.

No links to any docs are know, all I do know is that not ignoring .DS_Store generated on a mac recursively inside every directory on the repository is annoying as heck.

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -360,3 +360,8 @@ MigrationBackup/
 
 # Fody - auto-generated XML schema
 FodyWeavers.xsd
+
+# Created by Finder on Mac or JetBrains Rider on Mac.
+# I am sick and tired of cloning .NET Repositories that uses this .gitignore
+# but does not ignore these.
+.DS_Store


Almamu:
If you don't want .DS_Store files in your projects, then you should be using this gitignore in addition to the VisualStudio one: https://github.com/github/gitignore/blob/master/Global/macOS.gitignore

AraHaan:
The issue is even projects I do not own do not use that other gitignore and I feel like if we do that then all the other ones should have a special version that combines the macos one then.

I think something different has to be made. What if for those who do the ``I make my own gitignore`` they could put it inside of an repository named ``gitignore`` relative to where their repositories are and then use some yml file (under ``.github`` on the repository that could:
- url of the repository to get the ``.gitignore from (by ``author/gitignore``) and then specify the name of the gitignore.
- this repository but specify one of the gitignores here.
- ``VisualStudio+macOS`` (if you want to tell it to pull in both the VisualStudio & macOS gitignore files then pr them into the repository root and keep them in sync with changes to them).

Either way something needs done and some bot needs to be made to sync gitignore files from here so people do not have to manually update them themselves. Plus it would allow them to place their own gitignore somewhere else and it could be used by other projects as well too if they decide to make their own too or they got to ignore specific things.

Almamu:
You do not have to use the other gitignores yourself. They are designed to be put in the global .gitignore of the machine of the contributor, not on the project's. After all, not everyone is using a Mac, or a JetBrains editor, so why should they use the Mac-specific one project-wise if it does not affect them at all? What if the project maintainer doesn't feel like .DS_Store files are a problem? What If the .DS_Store is required for the project they're working on (aka any software targeted for Mac might want to customize the way the .dmg image generated looks)?

The repo here offers *templates* for .gitignores, not a blanket rule you have to follow. Most of the time your .gitignore won't change unless you either change the language you're working on, the versions of the toolset you're using or think you've missed a file/folder that should be ignored.

If you still feel like these .gitignores should be combined, you can easily set that up for your projects using the Actions feature of GitHub, there's actions that allow you to run custom scripts, so you should be able to setup an automatic process of downloading the latest version and merging them into the same, then auto-push the changes to the branch you desire.

AraHaan:
Usually most project maintainers use windows, linux,and mac to build their programs to ensure they work on all 3 OS's, likewise many like me use Visual Studio but also use Rider because it's more lightwieght and plus it offers more ways to make your code *more* efficient than what Visual Studio offers. As such they would need to combine multiple gitignores on the project level and enforces the need to use Visual Studio starting out on an update, then move to Rider for the second phase (simplification and optimization), and then the third phase testing on Windows, linux, and mac to ensure the change works on all 3 systems. For an .NET ClassLib this is critical for them to work on all platforms by default without issues.

With how .NET Core works and them advertising crossplatform the same compilation *should* work for windows, linux, and mac out-of-box. This part is what gets tested by them to verify that there is no critical bugs that are missed if it was tested only on windows. Not to mention most people (like me) do code sharing between my windows OS <-> my Mac where the code is not actually stored on my mac, just viewed on my mac from my windows pc by connecting them to each other as such files created on windows that *should* be ignored, and files created on my Mac that *should* be ignored move their way onto my Windows PC on the codebase. And yes it is possible to connect a windows pc and share files between each pc without actually storing them on the mac itself. That part just takes some work to get right.

bdougie:
Hey @AraHaan. Thanks for the PR and the info. Per the guidelines in the README, it encourages using the global mac .gitignore and combine that

If this is an issue you need a quick fix for, may I suggest setting up a global .gitignore in your project root? https://gist.github.com/subfuzion/db7f57fff2fb6998a16c#global-gitignore-contents

PR #3717, opened 2021-05-03
adds Flutter gitignore file

leomoon:
This adds Flutter gitignore file

### DIFF
--- a/Flutter
+++ b/Flutter
@@ -1,0 +1,118 @@
+# Miscellaneous
+*.class
+*.lock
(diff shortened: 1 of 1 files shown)


bdougie:
Thanks for the PR, please provide more context on why this is needed.

leomoon:
It is for Flutter with Dart to exclude Flutter binaries!

manoj398:
https://github.com/flutter/flutter/blob/master/.gitignore

PR #3722, opened 2021-05-07
Adding VS Code to global VS family

timheuer:
Our customer research supports that we have lots of developers using multiple tools, primarily VS and Code together.  Ensuring that we have one VS family that encapsulate them is convenient for the developers and saves them a step and has been requested.


martinwoodward:
But you wouldn't use this one by default on VS Code correct?  You are just wanting .vscode ignored for all Visual Studio projects?

timheuer:
Yes, "Visual Studio" is overloaded.  We understand that our developers rarely are one-tool customers.  Especially in x-plat development teams you may have one team using .NET and Visual Studio but other team members using VS Code for front-end development.  They are both working on the same repos, yet all sources to acquire .gitignore for "Visual Studio" are too myopic.  We think of VS as a family of products and something that represents VS, should be inclusive of this.

martinwoodward:
Fair enough.

PR #3723, opened 2021-05-07
add VS Code and Rider user specific files

Alex-G:


martinwoodward:
Thanks!  Can you provide more information in the description? See https://github.com/github/gitignore#contributing-guidelines

martinwoodward:
Current advice from JetBrains is not to ignore the whole .idea folder. See https://intellij-support.jetbrains.com/hc/en-us/articles/206544839-How-to-manage-projects-under-Version-Control-Systems

You can combine with the JetBrains.gitignore locally foe a full list of you would like.

PR #3727, opened 2021-05-09
Remove the all files matches from V.gitignore

leighmcculloch:

The three lines at the top of the file are problematic and I don't think they should included in a default gitignore configuration.

The first line matches all files, ignoring all files. The second line unmatches directories, unignoring all directories. The third line unmatches all files with any extension, unignoring all files with any extension.

The three lines appear to try and ignore any binary files on non-windows systems that get generated and end up in the root, and that don't have an extension.

While these three directives are somewhat effective at ignoring binary files on non-windows systems that don't have an extension, they create two problems:

1. They will also ignore other files like Makefiles.

2. They will also override any users global gitignore configuration. The overriding of users global gitignore configuration takes folks by surprise and can easily cause people to commit secrets, or IDE files, to new repositories that they exclude using the global gitignore.

There's not a ton of documentation that reinforces removing these directives because likewise there's not a ton of documentation supporting using them. These three directives are mostly encourages in forums like Stackoverflow as one way to ignore binary files, however, as noted on those same forums, using them is imperfect and creates the problems noted above.

https://stackoverflow.com/questions/5711120/gitignore-binary-files-that-have-no-extension


martinwoodward:
Great catch @leighmcculloch - sorry about that ✨

PR #3728, opened 2021-05-10
Update Java.gitignore

xu-song:

Add Eclipse Files

### DIFF
--- a/Java.gitignore
+++ b/Java.gitignore
@@ -19,5 +19,18 @@
 *.tar.gz
 *.rar
 
+# Maven generated target folders
+target
+out
+
+# Eclipse Files
+.project
+.classpath
+.settings
+.factorypath
+.metadata
+.idea
+*.iml
+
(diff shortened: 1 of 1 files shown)


martinwoodward:
These files has some uses in team settings therefore not including in the template for now.

xu-song:
In more cases, these setting are useless, and may cause conflict with differenct absolute path, such as path in  `.idea/workspace.xml`.

You can refer to the following projects 
- spring-framework: https://github.com/spring-projects/spring-framework/blob/main/.gitignore
- spring-boot: https://github.com/spring-projects/spring-boot/blob/main/.gitignore
- guava: https://github.com/google/guava/blob/master/.gitignore
- apache-common: https://github.com/apache/commons-lang/blob/master/.gitignore
- haddop: https://github.com/apache/hadoop/blob/trunk/.gitignore
- hive: https://github.com/apache/hive/blob/master/.gitignore

PR #3729, opened 2021-05-10
Update Unity.gitignore

baobao:

This is because many people use Rider when using Unity.


sufficients (approved):

martinwoodward:
The idea folder doesn't want to be excluded for everyone - suggest merging with https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore in your local repo is this is causing issues.

PR #3735, opened 2021-05-13
Updated .gitignore for Umbraco

richarth:
Umbraco 9 now stores its various auto generated files within the umbraco directory whereas older versions of Umbraco stored them in App_Data.  I've therefore added the Umbraco 9 paths.

### DIFF
--- a/Umbraco.gitignore
+++ b/Umbraco.gitignore
@@ -11,6 +11,13 @@
 **/App_Data/TEMP/
 **/App_Data/NuGetBackup/
 
+# Umbraco 9 now stores it's unimportant folders within the umbraco directory
+**/[Uu]mbraco/Data/*
+# we want to keep the packages directory
+!**/[Uu]mbraco/Data/[Pp]ackages/*
+**/[Uu]mbraco/Logs/*
+**/[Uu]mbraco/mediacache/*
+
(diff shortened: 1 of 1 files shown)


bl-ue (approved):

soreng:
I would suggest creating a new version for v9, since there are a lot of changes that can be very hard to cover with a single version.

martinwoodward (changes requested):
Can we provide more links to the documentations / discussions around this one.  See also #3739 which was suggesting a broader exclusion of `ood.flag` due to configurability settings.

phocodes (approved):

soreng:
> Can we provide more links to the documentations / discussions around this one. See also #3739 which was suggesting a broader exclusion of `ood.flag` due to configurability settings.

Hi,

It's getting closer to a release of Umbraco v9, so we not have a better understanding of what to include / exclude. A lot of folder have been removed since the move from .Net Framework to .Net 5, thus removing the App_Data completely.

I will make a PR with a new, and very limited, file, that we can build upon. Umbraco v8 will still be used for some time, so is there a way to (maybe with a tag) to make a version of the file, or should the file be just for the current version (v8 today, v9 later)?

PR #3739, opened 2021-05-16
Excluding ood.flag in all folders for Umbraco gitignore

callumbwhyte:

The `ood.flag` file is generated by Umbraco to indicate it's code-first models need regenerating.

This file is automatically created in the root of the folder where it has been designated Umbraco should generate code-first models. It should never be committed to source control.

By default Umbraco stores these files in `/App_Data/Models/*`, however this is [configurable](https://our.umbraco.com/documentation/reference/templating/modelsbuilder/configuration) via the `Umbraco.ModelsBuilder.ModelsDirectory` app setting. In most cases developers will modify this path as files in `App_Data` are not compiled by default by Visual Studio.

Additionally, the upcoming Umbraco V9 release moves the default location of these files to `/umbraco/models/*`.

This change adjusts the ignore rule to apply broadly regardless of where the `odd.flag` file lives.

### DIFF
--- a/Umbraco.gitignore
+++ b/Umbraco.gitignore
@@ -15,7 +15,7 @@
 **/App_Data/umbraco.config
 
 ## this [Uu]mbraco/ folder should be created by cmd like `Install-Package UmbracoCms -Version 8.5.3` 
-## you can find your umbraco version at your Web.config. (i.e. <add key="Umbraco.Core.ConfigurationStatus" value="8.5.3" />)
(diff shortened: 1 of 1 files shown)


bl-ue (approved):
The configurable bit is a good point.  Hopefully this can be merged soon.

initpwn (approved):

martinwoodward:
Looks good to me - thanks!

PR #3744, opened 2021-05-25
Restructuing README.md contents

shivaylamba:

Making README contents more logically positioned. ( Example : Placing Contributing guide with Contributing Workflow )


martinwoodward:
Thanks very much - agree that the readme file needs to be improved but I really need this guidance to be better read right now.  Therefore going to say no to this for now, but you remind me that we should make it easier to go get started with this project.

PR #3745, opened 2021-05-27
Add line to ignore JetBrains IDE cache directory

BBrown4:

To ignore the `.idea/` directory generated by the use of [JetBrains IDE (mainly JetBrains Rider)](https://www.jetbrains.com/rider/)

The `.idea/` directory is similar to the `.vs/` directory in that it's mainly for settings/configurations. The majority of users will likely ignore this directory entirely. My assumption would be that the number of people ignoring the directory outweighs the number including it. There are minimal benefits to including it in the first place.

I could be wrong, haha but I have used this gitignore for so long and every time, I find myself ignoring the `.idea/` directory and figured I can't be the only one.

[JetBrains Rider (also linked above)](https://www.jetbrains.com/rider/)
[What is the .idea/ folder?](https://rider-support.jetbrains.com/hc/en-us/articles/207097529-What-is-the-idea-folder-)

### DIFF
--- a/Unity.gitignore
+++ b/Unity.gitignore
@@ -26,6 +26,9 @@
 # Visual Studio cache directory
 .vs/
 
+# JetBrains cache directory
+.idea/
+
 # Gradle cache directory
 .gradle/
 


CertifiedRice (approved):

bl-ue:
PRs that do this are so common, maybe we should just merge it in.  True, it is similar to `.vs/` which _is_ already in. 🤔

wirecat:
@BBrown4  We get this quite a bit, but this change is not something that we can accept. The .idea folder is the workspace folder and is intended to house workspace configuration/settings. We don't want to globally ignore this folder for that reason.
Please read more about the workspace folder here: [Workspace settings documentation](https://rider-support.jetbrains.com/hc/en-us/articles/207097529-What-is-the-idea-folder) 
* Note: While the document linked is for the Rider product, the folder is used by all JetBrains products in the same fashion.

PR #3749, opened 2021-05-31
Limit Python.gitignore to 79-character line length

davevelasco:
Improves readability of notes, also matches Python line length standard

There are currently 2 optional lines in Python.gitignore that are accompanied
with commented notes explaining reasons why you might want to enable them.

However, these notes are somewhat long and may decrease readability
for persons using 80-column terminals that want to inspect the file.

Before, using nano:
before

After, the user can immediately read the note in its entirety:
after

Why specifically 79 instead of 80?

Because 80-column Emacs windows start wrapping at 79:
before2

This is also "supposedly" [why PEP8 enforces 79 chars instead of 80](https://www.reddit.com/r/learnpython/comments/1h2eug/pep_8_why_is_the_character_limit_79_and_not_80/)

Lastly, if you're OC like I am 😅, you might prefer all the project's source
files to follow the same max line length standard, not just python files 

[Some reasons why to keep the 80-column rule](https://stackoverflow.com/a/111009)
[Why 80-column rule is still worth sticking to even today](https://nickjanetakis.com/blog/80-characters-per-line-is-a-standard-worth-sticking-to-even-today)


bdougie:
Thank you for the contribution. I assume a macro or script is available to update line lengths for nano and eMacs. This change does not seem justifiable.

At least, to what end? The focus on this repo is source control and not on ergonomics. 

Let me know if there is a consensus or justification for 80 chars in the Python community.

bdougie:
After polling the community, it seems if most folks use 120 chars due to that being the VS Code default. This request does not cover a broad enough member of the Python/Dev community. 

Feel free to reopen if you can prove otherwise.

ayvazjet:
- [ ] **@_@[@]()_**

PR #3750, opened 2021-06-01
* DS_Store ignored for Unity.gitignore

mopsicus:


r-ush:
DS_Store is not something not related to unity but mac os specifically, so in my opinion, it isn't required here.

mopsicus:
Anyway all mac users add it to gitignore by themselves

bl-ue:
Mac users (like myself) can use https://github.com/github/gitignore/blob/master/Global/macOS.gitignore in conjuection with Unity.gitignore.

mopsicus:
Roger that

PR #3752, opened 2021-06-03
Gnuplot ignored in TeX.gitignore

lafeychine:

When using `TikZ` (`TeX` package), it is possible to plot function using `Gnuplot`.
During the process, `.gnuplot` and `.table` files will be created.
Same as: #1422 (was closed due to inactivity)

Chapter 22.6: "Plotting a Function Using Gnuplot" of the TikZ documentation.
Available here: [CTAN](http://mirrors.ctan.org/graphics/pgf/base/doc/pgfmanual.pdf)

### DIFF
--- a/TeX.gitignore
+++ b/TeX.gitignore
@@ -115,6 +115,10 @@ acs-*.bib
 # uncomment this for glossaries-extra (will ignore makeindex's style files!)
 # *.ist
 
+# gnuplot
+*.gnuplot
+*.table
+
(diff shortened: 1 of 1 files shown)


bdougie:
That is not a useful link, can you provide a screenshot to what you are referencing.

bdougie:
Linking the description for TeX that helps better understand these changes. https://github.com/github/gitignore/pull/3544#issuecomment-991354602

PR #3753, opened 2021-06-04
Remove `*.jar` from Java.gitignore

bl-ue:
Though it's not a great practice to store dependencies in version
control, it's very common among enterprises to keep jar files in their
source trees.

### DIFF
--- a/Java.gitignore
+++ b/Java.gitignore
@@ -11,7 +11,6 @@
 .mtj.tmp/
 
 # Package Files #
-*.jar
 *.war
 *.nar
 *.ear


drothmaler:
As you said it yourself, it’s not a good practice to add dependencies/binaries, to git - so why would you make it a default? If you really want to add them for a specific project, you could always modify the ignore-file in that case. E.g. by adding a negating pattern for your dep dir (something like `!deps/**/*.jar`). Or you could just [force-add](https://git-scm.com/docs/git-add#Documentation/git-add.txt--f) the jar files, you really want…

bl-ue:
True, these are all good points.  However, the enterprises that do this don't always respect common practices :). They're older and have more ancient systems.  The also do stuff like download dependencies and store them in-tree for offline use.  In my particular case, I'm processing an uploaded zip file, opening it up and removing files that match the patterns listed in a few gitignores here, which I download on-the-fly.  Most of them I like; however, `*.jar` are usually critical dependencies and should not be removed.  Since you think it's a bad idea, I'll just remove `*.jar` from the list after I compile it, like I do with the dangerous `*` (and `!*`). 👍🏻

bdougie:
I don't great justification for removing it. I am going to close this and request we not make decisions that affect all open users based on "some" enterprises.

PR #3754, opened 2021-06-04
Add `*.sar` to Java.gitignore

bl-ue:
I've seen them before.  References:
- https://docs.jboss.org/jbossas/docs/Server_Configuration_Guide/4/html/ch02s01.html
- https://forum.step.esa.int/t/sar-files-in-java-8/5074

### DIFF
--- a/Java.gitignore
+++ b/Java.gitignore
@@ -18,6 +18,7 @@
 *.zip
 *.tar.gz
 *.rar
+*.sar
 
 # virtual machine crash logs, see http://www.java.com/en/download/help/error_hotspot.xml
 hs_err_pid*


drothmaler:
While I‘ve never encountered them myself, It sounds reasonable to me :-)

bdougie:
Bey @bl-ue, can you provide more context on what is a .sar file and when it is created. Despite @drothmaler's approval, their response gives me low confidence that this is needed

bl-ue:
@bdougie, from that jboss.org page that I linked, it says:

>The SAR application archive (e.g., **myservice.sar**) packages a JBoss service in a JAR file. It is mostly used by JBoss internal services.

Is that not enough?

bl-ue:
@bdougie `.sar` is the only one of the `.*ar` Java ZIP-based archives that are not ignored in this file.

PR #3755, opened 2021-06-07
Add `.idea/jarRepositories.xml` to Android.gitignore

Abdulrahman-AlGhamdi:
The file `.idea/jarRepositories.xml` should be added to .gitignore. 
It is auto-generated and has only redundant information about remote jar repositories.
* https://www.jetbrains.com/help/idea/library.html#configure-custom-remote-repo

### DIFF
--- a/Android.gitignore
+++ b/Android.gitignore
@@ -44,6 +44,7 @@ captures/
 .idea/assetWizardSettings.xml
 .idea/dictionaries
 .idea/libraries
+.idea/jarRepositories.xml
 # Android Studio 3 in .gitignore file.
 .idea/caches
 .idea/modules.xml


bdougie:
> Global contains templates for various editors, tools and operating systems that can be used in different situations. It is recommended that you either add these to your global template or merge these rules into your project-specific templates if you want to use them permanently.

Per the guide, outline in the README, This should be added to the Jetbrains template

bdougie:
Upon further research this can be added. 

This is a better link, with an actual explanation and reasoning

https://stackoverflow.com/questions/63342780/how-can-i-decide-to-add-idea-jarrepositories-xml-in-gitignore/63348855#63348855

PR #3756, opened 2021-06-07
Added Django gitignore template

UlugbekMuslitdinov:

To help beginners to track only necessary files via Git while they are using the Django framework

If this is a new template:

 - **Link to application or project’s homepage**: [Official Website](https://www.djangoproject.com/)


bdougie:
Thanks @UlugbekMuslitdinov, but this context already exists in https://github.com/github/gitignore/blob/master/Python.gitignore#L58

PR #3760, opened 2021-06-15
Add media folder to umbraco.gitignore

erikjanwestendorp:

Excluding the media folder is best practice.

https://our.umbraco.com/Documentation/Fundamentals/Code/Source-Control/#media

### DIFF
--- a/Umbraco.gitignore
+++ b/Umbraco.gitignore
@@ -19,6 +19,10 @@
 ## Uncomment this line if you think it fits the way you work on your project.
 ## **/[Uu]mbraco/ 
 
+## This [Mm]edia/ folder contains content. Content may vary by environment and should therefore not be added to source control.
+## Uncomment this line if you think it fits the way you work on your project.
+## **/[Mm]edia/ 
+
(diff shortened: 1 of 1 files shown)


Umbraco.gitignore
```diff
@@ -19,6 +19,10 @@
 ## Uncomment this line if you think it fits the way you work on your project.
 ## **/[Uu]mbraco/ 
 
+## This [Mm]edia/ folder contains content. Content may vary by environment and should therefore not be added to source control.
```
bl-ue, line 22: I might say `The` rather than `This`, but otherwise LGTM!
```suggestion
## The [Mm]edia/ folder contains content. Content may vary by environment and should therefore not be added to source control.
```

bl-ue (approved):

Umbraco.gitignore
erikjanwestendorp: @bl-ue  That was blazing fast! Thanks for your suggestion.

bl-ue:
ping @bdougie @martinwoodward this looks correct to me, can we merge?

PR #1802, opened 2015-12-14
update license to CC0

shiftkey:
This replaces the original MIT license with [CC0](https://creativecommons.org/about/cc0) as it feels like a more suitable license for the `gitignore` templates we maintain here, and has more clarity than the changes proposed in #752.

cc @haacked who proposed the initial PR
cc @bkeepers @benbalter @cobyism @aroben @bhuga who were involved in that PR

PR #3761, opened 2021-06-16
Dual license (add MIT)

adamatan:

Licensing can be confusing. I'm not sure if I can use the `.gitignore` files in my MIT project, and Google searching is ambiguous. 
Dual licensing allows users to use the `.gitignore` files in their MIT-licensed repos without worrying about license conflicts.

[License_compatibility](https://en.wikipedia.org/wiki/License_compatibility) on Wikipedia. Long, cumbersome - let's avoid it.

### DIFF
--- a/LICENSE.mit
+++ b/LICENSE.mit
@@ -1,0 +1,7 @@
+Copyright 2021 Brian Douglas
+
+Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the "Software"), to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
--- a/LICENSE.mit
+++ b/LICENSE.mit
@@ -1,4 +1,4 @@
-Copyright 2021 Brian Douglas
+Copyright 2010-2021 GitHub, Inc. and contributors
 
 Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the "Software"), to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the following conditions:
 
--- a/README.md
+++ b/README.md
@@ -152,4 +152,6 @@ by automatically forking the project and prompting to send a pull request too.
 
 ## License
 
-[CC0-1.0](./LICENSE).
+This work is dual-licensed under [CC0-1.0](./LICENSE) and [MIT](./LICENSE.mit), or any later version of both licenses. You can choose between one of them if you use this work.
+
+`SPDX-License-Identifier: CC0-1.0 OR MIT`


LICENSE.mit
```diff
@@ -0,0 +1,7 @@
+Copyright 2021 Brian Douglas
```
bl-ue, line 1: This should probably be something like the following, because @bdougie didn't make the project, @defunkt [did](https://github.com/github/gitignore/commit/b7cc33a99b02fada900d0e4ba6b7bd38a142f064), and he works for GitHub (helped create it, made the first commit on it, etc.).
```suggestion
Copyright 2010-2021 GitHub, Inc. and contributors
```

adamatan:
Fixed.

Jayman2000:
I think that this change makes things more confusing, not less. It makes the `.gitignore` files seem copyrighted when they probably aren’t and it makes copyright law seem relevant when it probably isn’t.

### Does your .gitignore file even need to have a compatible license?

If you’re really concerned about it, you could just say the `.gitignore` file is available under CC0, and that the rest of the project is available under another license. Is there any situation where you couldn’t do that?

### Could CC0 or the MIT License even apply to what you’re using?

Let’s assume that you do need to have the .gitignore file under a compatible license.

Both CC0 and the MIT License leverage copyright law. In [the PR that made this repo use CC0 to begin with](https://github.com/github/gitignore/pull/1802), @shiftkey posted
>@revi as the previous public domain pull request #752 noted .gitignore files are too trivial to be subject to copyright. We're just making it clear now that no you don't have to copy around the MIT license and our copyright notice when using any of these .gitignore files in your projects, not that anyone ever did anyway. If we had any code in this repo you're right, getting agreement from all contributors would be necessary, if very painful.

If the gitignore file that you’re using could never have been subject to copyright, then you don’t need permission to incorporate it into a copyrightrighted work and you don’t need permission to license that work under the MIT License.

[Quote from Creative Commons’s FAQ:](https://creativecommons.org/faq/#do-i-always-have-to-comply-with-the-license-terms-if-not-what-are-the-exceptions)
>Additionally, if you are using an excerpt small enough to be uncopyrightable, the license does not apply to your use, and you do not need to comply with its terms.

### Is your use exempt from copyright law?

Let’s assume that
1. you do need to have the `.gitignore` file under a compatible license and
1. what you’re using is copyrightable.

Sometimes, you don’t need permission to use a copyrighted work. Your use might fall under [a copyright law exception](https://www.lib.purdue.edu/uco/CopyrightBasics/exceptions.html).

### Is what you’re trying to use in the public domain?

Let’s assume that
1. you do need to have the `.gitignore` file under a compatible license,
1. what you’re using is copyrightable and
1. your use isn’t exempt from copyright law.

[CC0 is a public domain dedication, not a license](https://github.com/github/gitignore/pull/3907). As [CC0’s deed](https://creativecommons.org/publicdomain/zero/1.0/) says,
>The person who associated a work with this deed has dedicated the work to the public domain by waiving all of his or her rights to the work worldwide under copyright law, including all related and neighboring rights, to the extent allowed by law.
>
>You can copy, modify, distribute and perform the work, even for commercial purposes, all without asking permission.

(That that text isn’t legally binding, but it’s still a good summary). The MIT License is a copyright license. [“A copyright license gives a licensee authorization to use a work from the copyright owner, usually in exchange for payment.”](https://www.legalzoom.com/articles/what-is-a-copyright-license). A copyright license couldn’t possibly do anything in this situation because
1. You don’t need authorization.
2. There is no copyright owner.

[Creative Commons’s FAQ even says:](https://creativecommons.org/faq/#how-do-cc-licenses-operate)
>When CC0 is applied to a work, copyright no longer applies to the work in most jurisdictions around the world. Therefore, references to **dual licensing arrangements like the one above are inapplicable to CC0**.

So, if these `.gitignore` files are in the public domain, then can they be incorporated into a copyrighted work? Can that copyrighted work use the MIT license? [The answer to both of those questions is yes.](https://www.quora.com/Can-my-work-be-copyrighted-if-it-is-based-on-a-public-domain-work)

### Is the Public License Fallback good enough?

Let’s assume that
1. you do need to have the `.gitignore` file under a compatible license,
1. what you’re using is copyrightable,
1. your use isn’t exempt from copyright law and
1. what you’re using is not public domain.

This situation is possible because not all jurisdictions allow CC0 to dedicate something to the public domain like it tries to. When copyright law doesn’t allow CC0 to dedicate a work to the public domain, CC0 provides a Public License Fallback:
><strong>3. Public License Fallback.</strong> Should any part of the Waiver for any reason be judged legally invalid or ineffective under applicable law, then the Waiver shall be preserved to the maximum extent permitted taking into account Affirmer's express Statement of Purpose. In addition, to the extent the Waiver is so judged Affirmer hereby grants to each affected person a royalty-free, non transferable, **non sublicensable**, non exclusive, irrevocable and unconditional license to exercise Affirmer's Copyright and Related Rights in the Work[…]

In this situation, the Public Fallback License might very well be compatible with the MIT License. I don’t really know.

Something that’s key here is the fact that the Public Fallback License is “non sublicensable”. In this situation, releasing old contributions under a new license could count as sublicensing. Or maybe not. I don’t really know.

### Would using an incompatible license even matter?

Let’s assume that
1. you do need to have the `.gitignore` file under a compatible license,
1. what you’re using is copyrightable,
1. your use isn’t exempt from copyright law,
1. what you’re using is available under the Public Fallback License and
1. the Public Fallback License is incompatible.

So what? There was no way you could have possibly known that you were breaking the law in this case. It would take a lot of effort to determine that whether or not all 5 of those assumptions are true, and it’s impossible to know if a court will agree that all 5 of those assumptions are true. It’s not worth being concerned about such a niche scenario, since [it’s very possible that you’re already unintentionally breaking the law in other, more important, ways](https://www.quora.com/Does-the-average-American-really-commit-an-average-of-three-felonies-a-day-If-not-where-did-the-idea-come-from). 

## Conclusion

[CC0’s Statement of Purpose says](https://creativecommons.org/publicdomain/zero/1.0/legalcode)
>Certain owners wish to permanently relinquish those rights to a Work for the purpose of contributing to a commons of creative, cultural and scientific works ("Commons") that the public can reliably and **without fear of later claims of infringement build upon, modify, incorporate in other works, reuse and redistribute as freely as possible** in any form whatsoever and for any purposes, including without limitation commercial purposes. These owners may contribute to the Commons to promote the ideal of a free culture and the further production of creative, cultural and scientific works, or to gain reputation or greater distribution for their Work in part through the use and efforts of others.

As someone who applies CC0 to some of my work, please, take that statement at face value.

bdougie:
Thanks for the contribution but at this time we are not considering new licenses.

PR #3763, opened 2021-06-17
Update VS Code file spec

Vivraan:

The current gitgnore doesn't account for VS Code files generated by the engine.

https://docs.unrealengine.com/4.26/en-US/WhatsNew/Builds/ReleaseNotes/4_18/
Granted it's not as refined as Rider or Visual Studio support, with the [right extensions](https://github.com/boocs/ue4-intellisense-fixes) it works like a charm.


bdougie:

PR #3764, opened 2021-06-17
Adding rider cache directory.

robotron2084:

The `.idea` folder contains various Rider artifacts that should not be checked in.


bdougie:

PR #3766, opened 2021-06-19
Create Hugo.gitignore

jfoshee:
Hugo is a popular static website generator written in Go.  
https://github.com/gohugoio/hugo (50k stars)
https://gohugo.io/

This proposal includes two common rules:
1. The `public` folder (rooted) is for build output 
2. The resource cache folder (rooted) (which, as noted, is generated but can be included in a repo as an optimization)

Hugo is a popular site generator where it's common practice for the site content to be managed in a Git repo.

(inline)

If this is a new template:

https://github.com/gohugoio/hugo (50k stars)

### DIFF
--- a/Hugo.gitignore
+++ b/Hugo.gitignore
@@ -1,0 +1,16 @@
+# HUGO
+# https://gohugo.io/
+# https://github.com/gohugoio/hugo
+# Hugo is a static HTML and CSS website generator written in Go. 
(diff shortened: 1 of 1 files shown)


bdougie:
This already exists, https://github.com/github/gitignore/blob/master/community/Golang/Hugo.gitignore

Can you make your updates there?

PR #3771, opened 2021-06-23
Adding in support for Nasa SpecsIntact

hecon5:

NASA SpecsIntact is a specifications editing tool which is used by nearly all US Government agencies for facilities, construction, and procurement. Specifically USACE, NAVFAC, and others.

When using git to manage SI files (which are mostly text-based XML files), SI auto generates a good number of clutter files; this causes confusion within users who erroneously commit these to the repository. Some of these are locking files which have significant consequences to other users (or themselves).

If this is a new template: YES

 - **Link to application or project’s homepage**: https://specsintact.ksc.nasa.gov/


bdougie:
Thanks for the contribution.

PR #3773, opened 2021-06-28
Update Drupal.gitignore

zhouhana:

This template is also valid for Drupal 9. As a matter of fact, when starting new projects you _should_ use Drupal 9. And 8 is only supported until November 2021.

https://www.drupal.org/project/drupal

### DIFF
--- a/Drupal.gitignore
+++ b/Drupal.gitignore
@@ -1,4 +1,4 @@
-# gitignore template for Drupal 8 projects
+# gitignore template for Drupal 8 and 9 projects
 #
 # earlier versions of Drupal are tracked in `community/PHP/`
 


bdougie:
Would it be preferred to remove the version number altogether? https://github.com/github/gitignore#versioned-templates

zhouhana:
Drupal 7 and earlier versions has different needs for the .gitignore than Drupal 8 and later. So maybe it should say

# gitignore template for Drupal 8 and later

or 

# gitignore template for version 8 and later of the Drupal 8 project

or similar.

bdougie:
> # gitignore template for Drupal 8 and later
This is preferred. Is it also safe to assume that new Drupal projects will be Drupal 8 or later?

bdougie:
Linking other Drupal PRs https://github.com/github/gitignore/pull/3386

lpalgarvio:
the structure is expected to be roughly the same in 8 vs 9, so the comment change should suffice

but perhaps #3914 should be reworked to consider #3914 and update paths is there are any official changes

PR #3774, opened 2021-06-29
Update C++.gitignore

mdisprgm:

*.recipe generated at build, and it contains User Path


bdougie:
Can you provide more context and links to examples

PR #3778, opened 2021-07-03
Os depend

bastie:

Operating systems create unwanted files in directory structures. For example macOS create a file .DS_Store when you browse with Finder. This file should not add to GitHub repository

See https://en.wikipedia.org/wiki/.DS_Store 

**Extensions needed**
It seems not a bad idea to integrate same for Windows Thumbs.db and also do it in all gitignore files

### DIFF
--- a/Java.gitignore
+++ b/Java.gitignore
@@ -1,3 +1,6 @@
+# OS depend
+.DS_Store
+
 # Compiled class file
 *.class
 
--- a/Swift.gitignore
+++ b/Swift.gitignore
@@ -1,3 +1,6 @@
+# OS depend
+.DS_Store
+
 # Xcode
 #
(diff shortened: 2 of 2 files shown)


bdougie:
This is leverage from the global gitignore.

https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer

bastie:
> This is leverage from the global gitignore.
> 
> https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer

Sorry, this is not a solution! 

Repositories can be cloned on other computers and at this time, so I hope, nobody have admin rights on all computers. If other developer do not your hint (or the nobody with rights on all computers), a merge bring waste to the master repo.
In result of git is a tool to bring developer together and share/merge code your solution idea is IMHO a problem that doesn‘t help.

with regards

PR #3780, opened 2021-07-09
add Umbraco excludes for version 9 / .Net 5 (Core)

idseefeld:

Umbraco CMS for .Net 5 (Core) has a different project / folder structure.

[https://our.umbraco.com/documentation/Umbraco9Articles](https://our.umbraco.com/documentation/Umbraco9Articles)


Umbraco.gitignore
```diff
@@ -24,3 +24,8 @@
 
 # Ignore the Models Builder models out of date flag
 **/App_Data/Models/ood.flag
+
+# NEW for version 9 .Net 5 (Core)
+**/umbraco/Data/*
```
soreng, line 35: I would suggest just ignoring the whole folder, but include umbraco/models, making it:

/umbraco/
!/umbraco/models

for the wwwroot, the full umbraco should also be ignore

/wwwroot/umbraco/

soreng: the full gitignore for umraco 9 would then be:

```
#ignore umbraco backoffice assest from wwwroot
/wwwroot/umbraco/

#ignore umbraco data/views/settings
/umbraco/

#include default location for modelsbuilder output
!/umbraco/models
```

idseefeld: Hi @soreng,
you might be right. Lately I realised the _Build > Clean Solution_ command deletes these folders.
But you are missing `**` when the .gitignore is one folder up from project, like it is in a common solution setup for multi project solutions.

```
#ignore umbraco backoffice assest from wwwroot
**/wwwroot/umbraco/

#ignore umbraco data/views/settings
**/umbraco/

#include default location for modelsbuilder output
!**/umbraco/models
```

idseefeld:
Looks like the `!**/umbraco/Data/packages` should not be ignored

soreng:
> Looks like the `!**/umbraco/Data/packages` should not be ignored

be aware of the casing. maybe it should be "[Dd]data" instead of "Data"

I would still stuggest removing the v8 stuff altogether

idseefeld:
@soreng removing V8 stuff is to early from my point of view. Either we create a version specific ignore file, but as far as I can see the v9 and v8 stuff does not interfere. Normally you have to merge this stuff into another ignore file for Visual Studio anyways. Thus I think a comment for version specifics is sufficient.

soreng:
@idseefeld I agree - it would be better to create a new version

idseefeld:
Sorry @soreng, but you got me wrong. My suggestion is to leave it as it is😉

raydix19813:
Well am being hack by someone that works here

idseefeld:
@raydix19813 what do you want to tell us?

bdougie:
Can you move this to the [DotNet folder](https://github.com/github/gitignore/tree/main/community/DotNet)

idseefeld:
Hm, looks like I cannot update this PR. I'll close it and create another PR.

PR #3781, opened 2021-07-11
Remove trailing whitespace

GrpeApple:

*Trailing whitespace* might annoy someone or add extra storage space, and I am surprised no one has gotten far into Removing *trailing whitespace* in **all** files.


bdougie (approved):
LGTM.

PR #3784, opened 2021-07-13
Added wildcard to crash.log because the newer versions of Terraform c…

msharma24:
 Added wild card to crash.log because the newer versions of Terraform creates crash log with random numbers in between. For example :  crash.105189318.log

### DIFF
--- a/Terraform.gitignore
+++ b/Terraform.gitignore
@@ -6,7 +6,7 @@
 *.tfstate.*
 
 # Crash log files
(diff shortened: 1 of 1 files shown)


z4id (approved):

msharma24:
Hey @z4id May I know when will this PR get merged ?

bdougie:
Thanks for the contribution.

PR #3786, opened 2021-07-18
Ignore .idea And .vscode Folders in Python.gitignore

ammar-oker:

Ignore the folders created by JetBrains IDEs and VS Code

[What is JetBrains .idea folder?](https://rider-support.jetbrains.com/hc/en-us/articles/207097529-What-is-the-idea-folder-)
[VS Code setting file location](https://code.visualstudio.com/docs/getstarted/settings#_settings-file-locations)

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -136,3 +136,7 @@ dmypy.json
 
 # Cython debug symbols
 cython_debug/
+
+# IDEs
+.idea/
+.vscode/


Python.gitignore
```diff
@@ -136,3 +136,7 @@ dmypy.json
 
 # Cython debug symbols
 cython_debug/
+
+# IDEs
```
z4id, line 140: Instead of writing IDEs (like combining all of them), it would be better to mention them separately so everyone should know which IDE a configuration file/directory belongs to. It will increase the clarity for readers.

**For Example**:
`# PyCharm Configuration Files`
`.idea/`
`# VS Code Configuration Files`
`.vscode/`

I purposed this change related to PyCharm configuration directory around 6 weeks ago in my PR. https://github.com/github/gitignore/pull/3769

bdougie:

PR #3790, opened 2021-07-21
Add Pycharm Project Files to Python .gitignore

DillonB07:

*[Pycharm](https://www.jetbrains.com/pycharm) is a popular IDE for Python development. It has built-in GitHub integration, but also has a `.idea` folder where it stores project configuration. This doesn't need to be uploaded to GitHub, so I am PRing it to be added to the Python `.gitignore`*

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -117,6 +117,10 @@ venv.bak/
 .spyderproject
 .spyproject
 
+
+# Pycharm
+/.idea
+
 # Rope project settings
 .ropeproject
 
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -119,7 +119,7 @@ venv.bak/
 
 
 # Pycharm
-/.idea
+.idea/
 
 # Rope project settings
 .ropeproject


Python.gitignore
```diff
@@ -117,6 +117,10 @@ venv.bak/
 .spyderproject
 .spyproject
 
+
+# Pycharm
+/.idea
```
z4id, line 122: As .idea is a directory, it should be treated as **.idea/** instead of **/.idea**, like you added in this commit. Correct me if I am wrong.

I purposed this change related to PyCharm configuration directory around 6 weeks ago in my PR. https://github.com/github/gitignore/pull/3769

z4id (changes requested):

Python.gitignore
Thunya154712: Leavea a commint

dbarnes18: Looks like your PR does what I wanted. *But will GitHub ever merge any of these PRs?*

I don't really modify gitignores, so I didn't know the formatting. Thanks for pointing that out!

bdougie:
In this case, a PyCharm is preferred. Similar [JetBrains](https://github.com/github/gitignore/blob/main/Global/JetBrains.gitignore).

bdougie:

DillonB07:
@bdougie Sure, but that doesn't have the Python stuff in it. For example venvs, pycache etc.

bdougie:
It should still live in a PyCharm.gitignore in the community folder

wirecat:
Closing for now as I believe this has gone stale and @bdougie had the right of it.
We have the PyCharm.gitignore that can be used instead.

PR #3791, opened 2021-07-21
Add Replit Project Files to Python `.gitignore`

DillonB07:

*[Replit](https://replit.com) is a popular online IDE that has GitHub integration. This will remove the Replit project files from the GitHub repository.*


bdougie:
Add IDE specific files to the ignore will lead to bloat. My assumption is that Replit projects originate from there and not GitHub as well. Feel free to correct me if I am wrong.

PR #3797, opened 2021-07-30
Actually "ignore everything in the root"

rinogo:

The existing Wordpress `.gitignore` doesn't "ignore everything in the root". In order to do so, we have to ignore `/*`.

None. Umm, it just doesn't work as currently implemented.


bdougie:

PR #3800, opened 2021-08-04
Unity.gitignore: Do not ignore .sln and .csproj

marcospgp:

Some issues, such as Omnisharp linting in VSCode, can be fixed by regenerating project files in Unity.
By keeping these files in the repo, only one team member has to refresh them to fix any issues.

It also seems generally agreed upon that .csproj and .sln files should be kept in code versioning: https://stackoverflow.com/a/65400320/2037431


marcospgp:
Since Unity auto generates C# project files based on user configuration and possibly other context-dependent values (OS, version, and so on) it does make sense to not keep these in the repo for a Unity project.

PR #3802, opened 2021-08-08
update yarn v2 to reflect documentation

deriando:
This PR assumes yarn users are **NOT** utilizing Zero-Installs.
Reason for not defaulting to Zero-Installs is due to link [2]

[1] https://yarnpkg.com/getting-started/qa#which-files-should-be-gitignored
[2] https://next.yarnpkg.com/features/zero-installs#does-it-have-security-implications

### DIFF
--- a/Node.gitignore
+++ b/Node.gitignore
@@ -111,8 +111,10 @@ dist
 .vscode-test
 
 # yarn v2
-.yarn/cache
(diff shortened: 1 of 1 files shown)


bdougie:
Can you provide data on which is the most common use case with yarn? zero-installs or non zero-installs

PR #3811, opened 2021-08-26
added cypress directories to node ignore

chrisenitan:

1. I've seen a lot like mine and my teams projects testing with [Cypress io](https://www.cypress.io/) and thought best to have that included in the Node ignore list. 
    - [>17k Repo](https://github.com/search?q=cypress&type=repositories)
2. I also look around other similar test frameworks like Test cafe and Cypress seems to be the only one creating custom files at runtime. 

- According to the Cypress [official documentation](https://docs.cypress.io/guides/core-concepts/writing-and-organizing-tests), cypress generates sample tests in the integration folder
- `cypress.env.json` is an alternative to cypress.json which acts as similar to dotenv [Documentation](https://docs.cypress.io/guides/guides/environment-variables#Option-1-configuration-file)
- Other asset files "report, videos, screenshots" are mostly ignored but optional. 

If this is a new template:

- Not a new template but there are standalone Cypress projects that may require a Cypress template. Please let me know what you think.

### DIFF
--- a/Node.gitignore
+++ b/Node.gitignore
@@ -26,6 +26,14 @@ coverage
 # nyc test coverage
 .nyc_output
 
+#cypress
+cypress/integration/1-getting-started/
+cypress/integration/2-advanced-examples/
+cypress/videos/
+cypress/screenshots/
+cypress/reports/
+cypress.env.json
+
 # Grunt intermediate storage (https://gruntjs.com/creating-plugins#storing-task-files)
 .grunt
 


Node.gitignore
```diff
@@ -26,6 +26,14 @@ coverage
 # nyc test coverage
 .nyc_output
 
+#cypress
+cypress/integration/1-getting-started/
+cypress/integration/2-advanced-examples/
```
bdougie, line 31: These files seem specific to a tutorial or guide. They are not listed in the documentation you provided.

bdougie (changes requested):
Remove the tutorial files.

Node.gitignore
chrisenitan: - Could not find any documentation on Cypress seeding the project with these files but they are scaffolded once you run Cypress initially. 
- You do have options to remove them, so adding the dir to the ignore list might not always be useful. 

What do you think? 
Screenshot 2021-12-05 at 12 50 16

chrisenitan: - Could not find any documentation on Cypress seeding the project with these files but they are scaffolded once you run Cypress initially. 
- You do have options to remove them, so adding the dir to the ignore list might not always be useful. 
- However I don't assume anyone cares to stage those. 

What do you think? 
Screenshot 2021-12-05 at 12 50 16

bdougie: I still question the number of files added here. Could this instead be `cypress/` and `!cypress/whatever` the files needed?

cc @bahmutov in there is some context missing.

wirecat:
I believe this PR has gone stale and I'm going to close for now @kannanputhuval.

PR #3812, opened 2021-08-26
Remove ignore on *.DotSettings.user since it's covered in *.user and keep publi…

adameska:
…sh settings since private info is in the pubxml.user

Everytime i add a publish profile to a new solution i always get tripped up because pubxml is ignored... this is a huge pain since i want the server name etc in there. Passwords are already stored hidden in the pubxml.user which is ignored.

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -130,7 +130,6 @@ $tf/
 # ReSharper is a .NET coding add-in
 _ReSharper*/
 *.[Rr]e[Ss]harper
-*.DotSettings.user
 
 # TeamCity is a build add-in
 _TeamCity*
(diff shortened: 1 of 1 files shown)


bdougie:
Is this a problem just for you? Can you point to any public discussions on this issue?

adameska:
No, this is a problem for anyone I've talked with, they general don't use the VS .gitignore for this issue and just re-use their own. I did not look at any public discussion but if you create a new repo with this gitignore, then create a publish profile to push to azure or w/e server you want, check your code in, then try and reference that publish profile on your build server it will fail (because publish profiles being checked in is ignored by default). I don't see how that's a good thing?

adameska:
I guess the main reason this was ignored was because connection strings to databases fully get stored in the pubxml? Is that the reason?

PR #3814, opened 2021-08-26
Added Yarn template for the latest version of yarn.

jhessin:
If this is a new template:

https://yarnpkg.com/advanced/qa#which-files-should-be-gitignored


bdougie:
Per the comments, Node.gitignore is preferred here.

PR #3815, opened 2021-08-28
adding: webstorm ignore item .idea and mocha item

BirajMainali:

_TODO_

_TODO_

If this is a new template:


bdougie:

PR #3821, opened 2021-09-04
add GH Action for checking whitespace

jimustafa:

Automatically check for trailing whitespace and proper EOF termination with newline character.

Uses the [`pre-commit` package](https://pre-commit.com/) and [action](https://github.com/pre-commit/action).

Trim trailing whitespace and fix EOF newlines for all files in the repo with `pre-commit run --all-files`.


bdougie:
Thank you for the contribution but at this time we cannot accept PRs with automation included. If you would like to make a separate PR for the removal of whitespace, that would be appreciated.

jimustafa:
Appreciate the feedback @bdougie! Will open a new PR with the whitespace fixes. Would it be OK to keep the `pre-commit` config? Other developers could use it to run the checks locally...

PR #3825, opened 2021-09-09
Remove nuget.config from VisualStudio.gitignore

asherber:

`nuget.config` was added to this file a few months ago; I think this was a mistake. It's true that `nuget.config` **may** contain credentials, but in my experience it's more likely to contain just other config settings and repo URLs. The onus should be on anyone storing credentials in this file to manually add it to .gitignore.

I recently wasted time trying to figure out why a Jenkins pipeline couldn't find the NuGet packages from the feed in `nuget.config`; it didn't occur to me that .gitgnore had prevented the file from being added to the repo.


bdougie:
There are some new merge conflicts. Once fixed, I can merge.

bdougie (approved):

PR #3826, opened 2021-09-09
Added "next.config.js" to Node.gitignore

AryaAnish121:
Added "next.config.js" to .gitignore

Because it contains env

https://nextjs.org/docs/api-reference/next.config.js/environment-variables

### DIFF
--- a/Node.gitignore
+++ b/Node.gitignore
@@ -80,6 +80,7 @@ web_modules/
 
 # Next.js build output
 .next
+next.config.js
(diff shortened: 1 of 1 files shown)


omBratteng:
Not a good idea to add that to gitignore, as it can contain custom configs for stuff like webpack.
Instead, you should use a `.env` file, like described in https://nextjs.org/docs/basic-features/environment-variables

PR #3839, opened 2021-09-27
Fixed .vscode not being ignored

tropicalraisel:
The `.vscode` directory was not being fully ignored.

🌴


mattfbacon:
This should be in your global gitignore.

PR #3840, opened 2021-09-28
Added Thumbs.db to all files that did not have it.

SteveBarnes-BH:

The file `Thumbs.db` is automatically created, _or updated_, on Windows machines whenever a user browses into a directory
 containing images using file explorer.

This leads to repository clones being spuriously marked as "dirty" which is undesirable on just about any project.

_TODO_


drothmaler:
Thank you for your contribution... Yes your are right this file is created by Windows and we have `Global/Windows.gitignore` to help with that... Please avoid spreading such global rules over all ignore-files and keep the repo modular instead; otherwise it will become a(n even bigger) maintenance hell... :-( You can instead combine the multiple existing gitignore-templates to fit your project needs...

PR #3844, opened 2021-09-30
Added .DS_Store for MacOS

Michael-Allgeier:

MacOS creates a hidden .DS_Store file that contains attributes and metadata about the application that could be a danger from hackers if pushed to a repo. It's a simple of fix of just putting .DS_Store into a gitignore, but if people are unaware they have to do that, it can cause problems. Adding it to the template would also be nice so MacOS developers don't have to add it to a gitignore for every project. 

https://buildthis.com/ds_store-files-and-why-you-should-know-about-them/

If this is a new template:

### DIFF
--- a/Node.gitignore
+++ b/Node.gitignore
@@ -116,3 +116,6 @@ dist
 .yarn/build-state.yml
 .yarn/install-state.gz
 .pnp.*
+
+# MacOS Temporary Files
+.DS_Store


Habetdin:
You should include [Global/macOS.gitignore](https://github.com/github/gitignore/blob/master/Global/macOS.gitignore) to the your .gitignore instead of adding that line to the Node template.

bdougie:
Should this be leverage from the global gitignore instead? 

PR #3855, opened 2021-10-12
Update Node.gitignore to exclude ./idea folder

Sinsstranger:
Exclude IDE project config folder

### DIFF
--- a/Node.gitignore
+++ b/Node.gitignore
@@ -1,3 +1,5 @@
+# IDE Project Config
+./idea
 # Logs
 logs
 *.log


AnilSeervi:
Refer Global gitignore - https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer
And JetBrains gitignore - https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore

bdougie:

PR #3857, opened 2021-10-13
Add .ptc to TeX.gitignore

janniklasrose:

The titletoc package creates an auxiliary `.ptc` file to store information about partial ToC's.

See **6.3. Partial ToC's** in [the manual](http://tug.ctan.org/tex-archive/macros/latex/contrib/titlesec/titlesec.pdf):

image

### DIFF
--- a/TeX.gitignore
+++ b/TeX.gitignore
@@ -213,6 +213,9 @@ pythontex-files-*/
 *.md5
 *.auxlock
 
+# titletoc
+*.ptc
+
 # todonotes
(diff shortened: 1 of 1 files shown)


bdougie:
Please provide more context on why the .ptc is not needed in the git history.

bdougie:
Linking the description for TeX that helps better understand these changes. https://github.com/github/gitignore/pull/3544#issuecomment-991354602

PR #3858, opened 2021-10-16
Fix Unreal Engine plugin binaries on subfolders

maximetinu:

When I installed the Rider plugin for UE5, I noticed that some subfolders with binaries were not being git ignored

This is the project with which I noticed the issue

https://github.com/Maximetinu/batterycollector-ue5

This is the commit I made to fix it

https://github.com/Maximetinu/batterycollector-ue5/commit/71dad496dd7a15c53103fb2a95be7442bf438feb

I've slightly modified the rule for this PR to make it more specific to this case

**Steps to reproduce**
- Pull down that repository prior to the commit I made to fix it
- Open the project with UE5, enable Rider plugin on the project and open the IDE to see the issue happening

### DIFF
--- a/UnrealEngine.gitignore
+++ b/UnrealEngine.gitignore
@@ -47,7 +47,7 @@ SourceArt/**/*.tga
 
 # Binary Files
 Binaries/*
-Plugins/*/Binaries/*
+Plugins/**/Binaries/*
 
 # Builds
 Build/*
@@ -68,7 +68,7 @@ Saved/*
 
 # Compiled source files for the engine to use
 Intermediate/*
-Plugins/*/Intermediate/*
+Plugins/**/Intermediate/*
 
 # Cache files for the editor to use
 DerivedDataCache/*


Maximetinu:
I'm participating in hacktoberfest, so if you can add the `hacktoberfest-accepted` label to the PR, I'll appreaciate it 😄

bdougie:
Unfortunately, we could not make this an Hacktoberfest contribution.

Can you clarify this is an issue happening to other UE5/Rider users?

Maximetinu:
Hi 

Yes, other users must also be encountering this. 

The original ignore rule is, for example, Plugins/*/Binaries/*. The single * between Plugins and Binaries only contemplates one level of subfolders.

The issue was that the Rider plugin doesn't output its binaries directly to Plugins/Rider/Binaries/, like the original rule assumed. It had the binaries on a deeper subfolder, like Plugins/Rider/whatever/Binaries/ which was not being git ignored 

So I changed the rule to ignore any Intermediate/ and Binaries/ folder, doesn't matter it's depth in the folder hierarchy. 

For example, these are the changes that Git detects when I open Rider with the original .gitignore:

image

As you can see, that Plugins/Developers/RiderLink/Binaries/ were not being .gitignored by Plugins/*/Binaries/*. Hence my change to make it just Binaries/

Another valid alternative would have been Plugins/**/Binaries/*. If you prefer a more explicit one like that, we can also change it.

filipechagas (approved):

LiterallyFabian:
Also having issues with RiderLink and other plugins binary files not being ignored. Is there any reason why it isn't merged yet?

image

Maximetinu:
I'd merge it if I had the rights!

jackfranzen:
Bump. This change will only cause good. I've closed my dupe #4213

wirecat (approved):

wirecat:
Thank you for contributing!

PR #3860, opened 2021-10-17
Ignore Pickle files in Python Projects

gcakir:

Pickle files are created while the Python project is executed. They aren't needed for the development or the execution of the project.


gcakir:
Wrong code. My bad. :-)

PR #3861, opened 2021-10-17
Ignore Pickle files in Python Projects

gcakir:

~~Pickle files are created while the Python project is executed. They aren't needed for the development or the execution of the project.~~


bdougie:

gcakir:
I am just thinking that I was mistaken and that there could be cases where pickle files are needed in a repo. My bad.

PR #2743, opened 2018-06-21
AL Template for Dynamics 365 Business Central

acyvas:
Support for AL language projects for Microsoft Dynamics 365 Business Central
link [https://github.com/Microsoft/AL/](https://github.com/Microsoft/AL/)

PR #3871, opened 2021-11-02
Add .gitignore for Microsoft Business Central

GuillemPM:

Dynamics 365 Business Central is a Microsoft solution used by thousands of companies.
We as developers have to use external services in order to get a .gitignore template which is outdated or quite simple.

https://docs.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/devenv-reference-overview
https://gitignore.io/api/AL

If this is a new template:

https://dynamics.microsoft.com/en-us/business-central/overview/
https://github.com/microsoft/AL

### DIFF
--- a/AL.gitignore
+++ b/AL.gitignore
@@ -1,0 +1,22 @@
+### AL ###
+#Template for AL projects for Dynamics 365 Business Central
+#launch.json folder
+.vscode/
+#Cache folder
+.alcache/
+#Symbols folder
+.alpackages/
+#Snapshots folder
+.snapshots/
+#Testing Output folder
+.output/
+#Extension App-file
+*.app
+#Rapid Application Development File
+rad.json
+#Translation Base-file
+*.g.xlf
+#License-file
+*.flf
+#Test results file
+TestResults.xml


bdougie:

bdougie (changes requested):
please move to the community folder

JeremyVyska:
For what it's worth, this version is much better than #2743 for general use.

@bdougie - question for you, if you'd be so kind:  The reason folks are trying to support AL language gitignore files is because that will add them to the "Add a .gitignore file" dropdown.  I notice that the community folder ones are not in that list.   Is there a smart way that one could create new AL repos using the gitignore from the community folder?

GuillemPM:
I agree with @JeremyVyska as AL is a language that repositories are totally based on. If it's possible we would like to have direct access to the template in Add .gitignore at repository creation time

GuillemPM:
> For what it's worth, this version is much better than #2743 for general use.
> 
> 
> 
> @bdougie - question for you, if you'd be so kind:  The reason folks are trying to support AL language gitignore files is because that will add them to the "Add a .gitignore file" dropdown.  I notice that the community folder ones are not in that list.   Is there a smart way that one could create new AL repos using the gitignore from the community folder? 

I agree with @JeremyVyska as AL is a language that repositories are totally based on. If it's possible we would like to have direct access to the template in Add .gitignore at repository creation time

bdougie:
@JeremyVyska thanks for the input and the context @GuillemPM. I will go ahead and merge as-is.

bdougie (approved):

PR #3882, opened 2021-11-11
Added tpr extension

Roald87:

After refactoring a *.tmcRefac gets created. Then after a build it gets converted into a *.tpr file. No need to have this file under source control.

There is no Beckhoff documentation on this file extension.

### DIFF
--- a/TwinCAT3.gitignore
+++ b/TwinCAT3.gitignore
@@ -13,6 +13,7 @@
 #   2. You've created a standalone PLC-project and added events to it, as these are stored in the TMC-file.
 *.tmc
 *.tmcRefac
+*.tpr
 *.library
 *.project.~u
 *.tsproj.bak


sagatowski:
Have you asked Beckhoff support of whether this file should be handled in version control or not? That's the second best thing to do after checking the documentation in this case I'd say.

Roald87:
@sagatowski  No I haven't yet. So far I haven't noticed issues when removing it. I'll check with support next week.

Roald87:
closing files are needed.

sagatowski:
@Roald87 
What are the files needed for?

Roald87:
I'm still discussing with support. I will update [the stackoverflow question](https://stackoverflow.com/questions/69926304/can-i-gitignore-the-tmcrefac-and-tpr-files-in-twincat) later when I have the full answer.

PR #3893, opened 2021-11-17
Azure Functions local settings should be excluded.

drumsta:
During local development of a function app, settings required by your functions are stored in the local.settings.json file. The local.settings.json file also stores settings used by local development tools.

Because the local.settings.json may contain secrets, such as connection strings, you should never store it in a remote repository.

The original article can be found [here](https://docs.microsoft.com/en-us/azure/azure-functions/functions-develop-vs?tabs=in-process#local-settings).

_TODO_

_TODO_

If this is a new template:

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -3,6 +3,9 @@
 ##
 ## Get latest from https://github.com/github/gitignore/blob/master/VisualStudio.gitignore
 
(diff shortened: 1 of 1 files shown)


bdougie:
Thanks for the PR, but you failed to follow the template provided. See the project's [README](https://github.com/github/gitignore#readme) for help.

PR #3895, opened 2021-11-24
Remove .idea folder from VisualStudio

cmeeren:
Conflicts with JetBrains.gitignore, which is a lot more detailed. Some files in this directory should be checked in.

Not sure about *.sln.iml, so I'm leaving that. The JetBrains.gitignore file contains commented-out IML stuff with instructions on when to uncomment.


bdougie:
I feel like this change makes sense, this change belongs in the JetBrains.ignore

PR #3897, opened 2021-11-25
Ignore Rider files

Bram-Reuling:
# Added ignore statements for .idea/ and the RiderLink plugin.

## Reasons for making this change:

Rider automatically creates the .idea/ folder which contains settings for the IDE. These files are not supposed to be pushed to the repository since everyone uses a different IDE to program in Unreal Engine. (Basically the same reason why .vs/ is in this .gitignore)

For complete support between Rider and Unreal Engine, Rider requires you to install the RiderLink plugin. This plugin contains 500+ files. These files are not supposed to be pushed to the repository since everyone uses a different IDE to program in Unreal Engine. Including these files and getting the project from GitHub also caused some compile errors for me, that is also why I am including this.

## Links to documentation supporting these rule changes:

Link to Rider for Unreal Engine: https://www.jetbrains.com/lp/rider-unreal/ (No documentation)

### DIFF
--- a/UnrealEngine.gitignore
+++ b/UnrealEngine.gitignore
@@ -1,6 +1,12 @@
 # Visual Studio 2015 user specific files
 .vs/
 
+# Rider for Unreal Engine specific files
+.idea/
+
+# RiderLink plugin
+Plugins/Developer/RiderLink/*
+
 # Compiled Object files
 *.slo
 *.lo


bdougie:
We have an existing [Global\JetBrains.gitignore](https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore) that covers more than just this rule.

bdougie:
Example of a conflicting PR removing this from a non JetBrains https://github.com/github/gitignore/pull/3895

UnrealEngine.gitignore
```diff
@@ -1,6 +1,12 @@
 # Visual Studio 2015 user specific files
 .vs/
 
+# Rider for Unreal Engine specific files
+.idea/
```
bdougie, line 5: This is jetbrains specific.

PR #3898, opened 2021-11-28
AL Language gitignore template

faludigabor:

This is a new template for (Microsoft's AL Language)
https://github.com/microsoft/AL


bdougie (approved):
Thanks for this contribution @faludigabor

PR #3904, opened 2021-12-04
added vscode ignore for unity

designerjason:
I use gitignore for unity projects. I used vscode and noticed there was no ignore for the .vscode specific folder that's created

### DIFF
--- a/Unity.gitignore
+++ b/Unity.gitignore
@@ -25,6 +25,7 @@
 
 # Visual Studio cache directory
 .vs/
+.vscode/
 
 # Gradle cache directory
 .gradle/


bdougie:

designerjason:
It's needed because you have .vs as an ignore (for visual studio) but you do not have .vscode. VS Code is different to visual studio in that it creates a different folder (.vscode). I'm not sure how much more context I can describe? There's a use case for unity, where you can use either visual studio or visual studio code as the code editor. the gitignore you have for unity only seems to ignore the generated folder for visual studio.

PR #3853, opened 2021-10-08
Add `poetry.toml` to `Python.gitignore`

FaresAhmedb:
**Reasons for making this change:**

Poetry is a Python dependency management and packaging tool.

`poetry.toml` is a local configuration file for projects powered by poetry, The settings
commonly added to this file are user-specific settings like 'cache-dir' and
'virtualenvs.path' and shouldn't be committed for everyone .

Even the poetry repository includes it in it's `.gitignore` file:
https://github.com/python-poetry/poetry/blob/0d7b16f7f6f5bb63a7a507291064dd18f1ff2d93/.gitignore#L40

**Links to documentation supporting these rule changes:**

The documentation for `poetry.toml`: https://python-poetry.org/docs/configuration/

**Link to application or project’s homepage**:
- https://github.com/python-poetry/poetry
- https://python-poetry.org/

PR #3911, opened 2021-12-06
[Python] Add poetry.lock

JP-Ellis:
Similar to Pipfile.lock, it is generally recommended to include poetry.lock in version control.  This is especially recommended for binary packages to ensure reproducibility, and is more commonly ignored for libraries.

This is taken from https://python-poetry.org/docs/basic-usage/#commit-your-poetrylock-file-to-version-control

This is very similar reasoning to the existing Pipenv entry and Poetry is quite popular too (17.4k stars on Github).  As the benefits/cons of including the lock file can vary for each project, the default is left commented out (just as the Pipenv entry).

- https://python-poetry.org/docs/basic-usage/#commit-your-poetrylock-file-to-version-control

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -94,6 +94,13 @@ ipython_config.py
 #   install all needed dependencies.
 #Pipfile.lock
 
+# poetry
+#   Similar to Pipfile.lock, it is generally recommended to include poetry.lock in version control.
+#   This is especially recommended for binary packages to ensure reproducibility, and is more
+#   commonly ignored for libraries.
+#   https://python-poetry.org/docs/basic-usage/#commit-your-poetrylock-file-to-version-control
+#poetry.lock
+
 # PEP 582; used by e.g. github.com/David-OConnor/pyflow
 __pypackages__/
 


bdougie:
I understand the use of the lockfile, but I am curious if you explain the reason for the poetry toml file and if this is related? https://github.com/github/gitignore/pull/3853

JP-Ellis:
So Poetry is a general tool (like `npm`, `yarn`, `cargo`, etc.) and relies on a couple of files:

- `pyproject.toml` specifies the project's dependencies (among other project-specific things)
- `poetry.toml` specifies general configuration options (and is not typically stored in the project, but instead belong in the user's config directory such as `$XDG_CONFIG_HOME`
- `poetry.lock` which contains the specific versions of dependencies installed by Poetry in order to be able to reproduce an installation exactly (specifically, it contains exact versions and hashes which is in contrast to `pyproject.toml` which will contain dependency ranges).

Now to address the question on whether `poetry.toml` should be included or not into `.gitignore`, I am leaning towards a 'yes' on this.  Of the available settings, all of them are clearly user specific instead of project specific with the exception of [`repositories.<name>`](https://python-poetry.org/docs/configuration/#repositoriesname) which allows for alternative repositories to be specified (e.g. private ones).  Having said that, there is [an equivalent option with `pyproject.toml`](https://python-poetry.org/docs/pyproject/#dependencies-and-dev-dependencies) and that file is definitely project specific thus I do not see why one would prefer `poetry.toml` over `pyproject.toml` for this setting and arguably, one should discourage using `poetry.toml` for this as `pyproject.toml` is intended to be agnostic of which tool is being used.

Learnersintraining:
gh pr checkout 3911

PR #3912, opened 2021-12-07
Adding AllowList for Go

kuritka:

_Allowlisting with .gitignore is a technique for dealing with source trees that can have various different untracked local files such as generated output, packages installed through package managers, “working” files, config for individual developers, etc. Rather than trying to come up with some master list of all possible untracked files and add them to .gitignore, it can be easier to start by ignoring everything and then add specific directories back._
_I think the requirements for software development are changing and there is a need to offer an alternative to the denylist._

- https://jasonstitt.com/gitignore-whitelisting-patterns
- https://github.com/golang/go

Signed-off-by: kuritka <kuritka@gmail.com>

### DIFF
--- a/community/Golang/Go.AllowList.gitignore
+++ b/community/Golang/Go.AllowList.gitignore
@@ -1,0 +1,23 @@
+# Allowlisting gitignore template for GO projects prevents us
+# from adding various unwanted local files, such as generated
+# files, developer configurations or IDE-specific files etc.
+#
+# Recommended: Go.AllowList.gitignore
+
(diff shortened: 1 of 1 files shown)


sexjun:
【自动回复】
你好，感谢来信，我是陈东升，会尽快阅读并回信。Hello, thank you for your letter. I am Chen Dongsheng. I will read and reply as soon as possible。

bdougie:
Linking previous PR https://github.com/github/gitignore/pull/3870

PR #3914, opened 2021-12-08
Add drupal 8 and above ignore file

mourad-zitouni:
The existing file is compatible with drupal 7, this file is for later versions (8/9/10...)


bdougie:
Thanks for the contribution, but we will need more context about this change. Please explain the reason for divergence in Drupal and why the existing file does not support Drupal 8.

mourad-zitouni:
In fact there have been changes in the folder structure of Drupal in version 8, also they introduced composer to manage dependencies and because of that Drupal core and contributed modules files (and all their dependencies) are no longer part of the repository (they are downloaded by composer).

bdougie:
I believe that is all covered in this PR https://github.com/github/gitignore/pull/3386 which I will go ahead merge now.

lpalgarvio:
> I believe that is all covered in this PR #3386 which I will go ahead merge now.

correct, plus there are wrong paths here in this PR.
contributed modules, themes, etc go in /sites/all/* or /sites/[sitename]/*

lpalgarvio:
> > I believe that is all covered in this PR #3386 which I will go ahead merge now.
> 
> correct, plus there are wrong paths here in this PR. contributed modules, themes, etc go in [root]/sites/all/* or [root]/sites/[sitename]/* for drupal7 and [root]/* or [root]/sites/all/* or [root]/sites/[sitename]/* for drupal 8

after all might need future review for paths, im seeing new references in some places

PR #3507, opened 2020-08-24
Update to Global/VisualStudioCode

corrupt952:
**Reasons for making this change:**

VSCode has a snippet feature that allows for
It would be very useful if snippets could be shared per repository.

**Links to documentation supporting these rule changes:**

https://code.visualstudio.com/docs/editor/userdefinedsnippets#_project-snippet-scope

If this is a new template:

N/A

PR #3916, opened 2021-12-10
Update VS Code to support *.code-snippets

selenehyun:
Code snippets are very useful because they make it easy to manage code snippets for each project.

https://github.com/github/gitignore/pull/3507 is the same PR over there, but I suggest it again because it is too old and corrupt.

- https://code.visualstudio.com/docs/editor/userdefinedsnippets#_project-snippet-scope


bdougie (approved):
Thanks for the PR. Closed the previous due to a merge conflict.

PR #3917, opened 2021-12-10
Create seo.gitignore

Pik-1:
Common SEO files that need to be excluded on prototype websites or for demo purposes. Since things like domains, page structure and page linking might be changed a lot and frequently. Serving proper SEO related files with a prototype website risks being indexed too early and with wrong information thus making it harder later to improve search results or even to get flagged by browsers or services since the website structure doesnt fit anymore with whats in the google database. Serving no robots.txt and sitemap.xml is better practice than serving intentionaly bad files. See Google's documents about Crawling for further information.

see above + I do this automatically when creating a new repo/website

https://developers.google.com/search/docs/beginner/seo-starter-guide

### DIFF
--- a/seo.gitignore
+++ b/seo.gitignore
@@ -1,0 +1,5 @@
+# Common SEO files that would jeopardise indexing if unfinished or subject to change repo is made online for demo purposes via github pages or services like netlify
+# Delete this gitignore as the last step and in order to enable proper crawling for your finished website 
+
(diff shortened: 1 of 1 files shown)


bdougie:
I understand your use case, but I don't see this as needed for a default template for all users. Your link documentation explains SEO but does not reference adding these files to the gitignore.

PR #3920, opened 2021-12-11
Ignore Eclipse .project and .classpath when using build tools Maven or Gradle

baincd:

[Eclipse's documentation](http://wiki.eclipse.org/FAQ_How_do_I_set_up_a_Java_project_to_share_in_a_repository%3F) confusingly recommends `.classpath` and `.project` being under version control. That recommendation makes sense if neither maven or gradle is used. 

However, when using either maven or gradle those files are generated each time based on the maven or gradle configuration - which causes issues when those files are version controlled. That is why I recommended ignoring these files _only_ if using maven or gradle, but _not_ for all Eclipse Java projects.

See #3710 for discussion and additional documentation regarding this change

### DIFF
--- a/Gradle.gitignore
+++ b/Gradle.gitignore
@@ -10,3 +10,9 @@ gradle-app.setting
 
 # Cache of project
 .gradletasknamecache
+
+# Eclipse Gradle plugin generated files
+# Eclipse Core
+.project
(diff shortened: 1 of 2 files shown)


bdougie (approved):
Thanks for the links and clarification

PR #3923, opened 2021-12-14
Ignore files in directory, not the directory itself

eightsixnine:

When directory is ignored using `dir/` pattern, then everything under it will **always** be excluded.

Negation pattern underneath don't work.

When written as `dir/*`, git excludes files from the directory, not the directory itself and the negation patterns work.

https://stackoverflow.com/questions/5533050/gitignore-exclude-folder-but-include-specific-subfolder

### DIFF
--- a/Global/VisualStudioCode.gitignore
+++ b/Global/VisualStudioCode.gitignore
@@ -1,4 +1,4 @@
-.vscode/
+.vscode/*
 !.vscode/settings.json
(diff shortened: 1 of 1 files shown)


bdougie:
Linking to the last change for context. https://github.com/github/gitignore/pull/3839

Thanks for the link to the answered s/o question.

PR #3925, opened 2021-12-15
Ignore .bundle and vendor per official Jekyll documentation

BillRaymond:
Per the [Jekyll documentation](https://jekyllrb.com/tutorials/using-jekyll-with-bundler/), you should add the `vendor/` and `.bundle/` paths to the .gitignore file, so you are not managing all dependencies within your repo.

I am a regular user of Jekyll and find all the packages get synced in my git repo when that should not be the case (you can just run `bundle install` based on the gemfile.

_TODO_
[Jekyll documentation](https://jekyllrb.com/tutorials/using-jekyll-with-bundler/)


bdougie (approved):
LGTM

PR #3926, opened 2021-12-15
Added HashiCorp Packer variables file

rrmistry:
### Reasons for making this change:

HashiCorp Packer uses a variables file that acts like secrets file containing variable values.

This file is usually not intended to be version controlled.

### Links to documentation supporting these rule changes:

The file naming conventions are documented in the Packer website under:
https://www.packer.io/guides/hcl/variables#from-a-file

Below is excerpt from Packer website:

> Given a file named `variables.pkrvars.hcl` Packer can be instructed to use the variable definitions file by using the `-var-file` command line flag.
> 
> ```bash
> $ packer build -var-file="variables.pkrvars.hcl" .
> ```
> 
> Packer will automatically load any var file that matches the name `*.auto.pkrvars.hcl`, without the need to pass the file via the command line. If we rename the above variable definitions file from `variables.pkrvars.hcl` to `variables.auto.pkrvars.hcl`, then we can run our build simply by calling
>
> ```bash
> $ packer build .
> ```

__________

This is similar in pattern to another HashiCorp product: `Terraform`
https://github.com/github/gitignore/blob/e89e04390435f1e7628740d7b4ce1fe65ac218fd/Terraform.gitignore#L12-L17


bdougie:

Deshun3350 (commented):
217661

PR #3927, opened 2021-12-15
Create UTAU.gitignore

oxygen-dioxide:

UTAU is a singing voice synthesizer. A UTAU voicebank containes audio files (.wav)、voicebank configurations (oto.ini, presamp.ini, character.txt, prefix.map, etc.). When developing UTAU voicebanks on Github, compiled data files (llsm, frq, mrq, etc.) will be generated. We don't want these compiled data files to be uploaded to Github, because they are large and they can be regenerated using the wav files.

_TODO_

If this is a new template:

 Here is an example repo: https://github.com/oxygen-dioxide/hitsuboku-kumi-chn/blob/main/.gitignore


bdougie:
Can you add a link to the documentation of the project?

PR #3928, opened 2021-12-15
Addition of .DS_Store

bhavenjain:

Always had to add .DS_Store to the gitignore file myself even after adding this template in every single project. So addition of this to the template would be a nice.


eightsixnine:
Hi @bhavenjain,

there is a separate gitignore for macOS.

Here is the link:
https://github.com/github/gitignore/blob/main/Global/macOS.gitignore

PR #3929, opened 2021-12-15
Add .vscode folder

DillonB07:

_VSC is, in my opinion, the best editor. However, the `.vscode` doesn't need to be added into a repo._

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -144,6 +144,9 @@ dmypy.json
 # Cython debug symbols
 cython_debug/
 
+# Visual Studio Code
+.vscode/
+
 # PyCharm
 #  JetBrains specific template is maintainted in a separate JetBrains.gitignore that can
 #  be found at https://github.com/github/gitignore/blob/main/Global/JetBrains.gitignore


eightsixnine:
Hi  @DillonB07 

There is a separate gitignore for VS Code.

Here's a link:
https://github.com/github/gitignore/blob/main/Global/VisualStudioCode.gitignore

DillonB07:
I understand that, but separate gitignores for specific IDEs aren't helpful. If you're just cloning a repo to make a quick fix you don't want IDE specific stuff in there.

Also, if I know most developers don't really want to be messing around with combining gitignores in a project.

eightsixnine:
I've personally used several gitignore files for different languages and they are all language-specific, there are no OS or IDE files.

Look at Node.js, Java, C and so on.

I guess there is a reason why they make it like this, but who knows :)

DillonB07:
Wait, you can have multiple gitignores in one repl? I thought you only had one called `.gitignore`. That makes a lot more sense now, thanks!

eightsixnine:
Well, not multiple files, a single file, but nothing stops you from combining multiple files into one.

Or making your own gitignore file. This is just a plain-text file with your preferences. And this repo is just a template with those preferences, you may use it as you wish :)

PR #3933, opened 2021-12-22
Create TypeScript.gitignore

Tyler887:
This PR contains code under the [MIT license](https://choosealicense.com/licenses/mit). To comply with the terms of the license, I must provide the source of the code. The source of the code is from https://github.com/microsoft/TypeScript-Node-Starter/blob/master/.gitignore.

Just because I need TS Gitignore!

https://www.typescriptlang.org/docs/

If this is a new template:

 - **Link to application or project’s homepage**: https://typescriptlang.org

### DIFF
--- a/TypeScript.gitignore
+++ b/TypeScript.gitignore
@@ -1,0 +1,8 @@
+#Files for the code
+
+*.ts
+*.typescript
+
+#Executables
+*.exe
+*.app
--- a/TypeScript.gitignore
+++ b/TypeScript.gitignore
@@ -1,8 +1,63 @@
-#Files for the code
+lib-cov
 
-*.ts
-*.typescript
+*.seed
 
-#Executables
-*.exe
-*.app
+*.log
+
+*.csv
+
+*.dat
+
+*.out
+
+*.pid
+
+*.gz
+
+*.swp
+
+pids
+
+logs
+
+results
+
+tmp
+
+# Build
+
+public/css/main.css
+
+# Coverage reports
+
+coverage
+
+# API keys and secrets
+
+.env
+
+# Dependency directory
+
+node_modules
+
+bower_components
+
+# Editors
+
+.idea
+
+*.iml
+
+# OS metadata
+
+.DS_Store
+
+Thumbs.db
+
+# Ignore built ts files
+
+dist/**/*
+
+# ignore yarn.lock
+
+yarn.lock


SuperZooper3:
There seems to be a more comprehensive gitinore here: https://github.com/microsoft/TypeScript-Node-Starter/blob/master/.gitignore . I sugest we use this one instead?

Polda18:
Are you sure you understand what `.gitignore` means? Do you understand how it works? I think that you may want Typescript files to remain instead of being ignored, right? Typescript is the source code you want to be included in the project, you don't want the resulting Javascript files to be included in the project. Unfortunately, that's something that isn't quite easy to be set up via a single universal `.gitignore` file, it needs to be set up individually for each specific project and kept up to date.

Tyler887:
> Are you sure you understand what `.gitignore` means? Do you understand how it works? I think that you may want Typescript files to remain instead of being ignored, right? Typescript is the source code you want to be included in the project, you don't want the resulting Javascript files to be included in the project. Unfortunately, that's something that isn't quite easy to be set up via a single universal `.gitignore` file, it needs to be set up individually for each specific project and kept up to date.

Changed everything up to match the other repo.

Tyler887:
> There seems to be a more comprehensive gitinore here: https://github.com/microsoft/TypeScript-Node-Starter/blob/master/.gitignore . I sugest we use this one instead?

We can make it easier to find it after the most recent [commit](https://github.com/github/gitignore/pull/3933/commits) to the `TypeScript.gitignore` file.

SuperZooper3:
Should also add the source of the gitignore to the top. It's under MIT as well so that should be added.

Tyler887:
> Should also add the source of the gitignore to the top. It's under MIT as well so that should be added.

Thanks! Addressed and noted source. ;)

TypeScript.gitignore
```diff
@@ -0,0 +1,63 @@
...
+
+.idea
+
+*.iml
+
+# OS metadata
+
+.DS_Store
+
+Thumbs.db
```
dooleydevin, line 55: Per our contribution guidelines, we don't include OS or editor specific files in our gitignore templates.

Do you mind removing these lines?

dooleydevin (changes requested):
Apologies for the late feedback! Everything here looks good other than the lines I called out. We don't want to have overlap within the templates, and ignoring the entirety of directories like `.idea` can have unintended consequences 🙂

wirecat:
@Tyler887 Closing for now since we haven't heard anything after @dooleydevin's review. Please feel free to re-open once you're ready to engage with us on the change.

PR #3935, opened 2021-12-23
Update Xcode.gitignore

berikv:
- Remove a reminder to look at files that don't exist anymore
- Remove ignore rules for an IDE from 2007
- Remove an ignore rule for Gcc which is no longer used since 2013

New projects can safely not ignore the files that Xcode before v5 created. 

[Wikipedia](https://en.wikipedia.org/wiki/Xcode#Xcode_3.0_-_Xcode_4.x)


bdougie (approved):
I suppose Wikipedia is a valid link to documentation

PR #3939, opened 2021-12-30
chore: Remove empty Travis-CI config

nschonni:
Looks like the CI hook isn't running anymore, and the actual config was never actually testing anything


bdougie (approved):
Most likely added as part of a repo template years ago

PR #3941, opened 2022-01-05
Ignore the .vsconfig file generated by VS

BoyceLig:

The .vsconfig file is automatically generated by VS, it does not need version control, and it will be automatically generated when VS is opened.

https://devblogs.microsoft.com/setup/configure-visual-studio-across-your-organization-with-vsconfig/

### DIFF
--- a/Unity.gitignore
+++ b/Unity.gitignore
@@ -46,6 +46,7 @@ ExportedObj/
 *.mdb
 *.opendb
 *.VC.db
+.vsconfig
 
 # Unity3D generated meta files
 *.pidb.meta


ayvazjet:
saimakin@saimakin.info

bdougie:

BoyceLig:
> We have an existing [Global\VisualStudioCode.gitignore](https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore) that covers more than just this rule.

This is just a file generated by unity when you open VS to edit the code, and does not contain those files that you lock indicate.

Youssef1313:
I don't think this should be in gitignore.

PR #3955, opened 2022-01-22
support for `env` and `.idea`

ixxiv:
Added `.gitignore` support for `.env` - Environment Variables files and `.idea` - Jetbrains.ignore files. 

[https://go.dev/doc/](https://go.dev/doc/)

### DIFF
--- a/Go.gitignore
+++ b/Go.gitignore
@@ -19,3 +19,13 @@
 
 # Go workspace file
 go.work
+
+# Environment Variable
+*.env
+
+#  GoLand
+#  JetBrains specific template is maintainted in a separate JetBrains.gitignore that can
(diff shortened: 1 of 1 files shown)


thorrsson:
We get this quite a bit, but this change is not something that we can accept. The .idea folder is the workspace folder and is intended to house workspace configuration/settings. We don't want to globally ignore this folder for that reason.
Please read more about the workspace folder here: [Workspace settings documentation](https://rider-support.jetbrains.com/hc/en-us/articles/207097529-What-is-the-idea-folder) 
* Note: While the document linked is for the Rider product, the folder is used by all JetBrains products in the same fashion.

PR #3958, opened 2022-01-23
VisualStudio.gitignore - Include JetBrains temp directory

FriedrichRehren:
added '.idea' folder to include JetBrains cache/options directory
updated Visual Studio Versions in comment

I am proposing this change, since I am using Visual Studio 2022 and JetBrains Rider on a regular basis.

My only relation to this project is that I am using the VisualStudio.gitignore file in almost all my C# projects.

[JetBrains.gitignore](https://github.com/github/gitignore/blob/main/Global/JetBrains.gitignore)

[JetBrains Rider support page](https://rider-support.jetbrains.com/hc/en-us/articles/207097529-What-is-the-idea-folder-)

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -32,8 +32,11 @@ bld/
 [Ll]og/
 [Ll]ogs/
 
-# Visual Studio 2015/2017 cache/options directory
+# Visual Studio 2015/2017/2019/2022 cache/options directory
 .vs/
+# JetBrains cache/options directory
+.idea/
+
 # Uncomment if you have tasks that create the project's static files in wwwroot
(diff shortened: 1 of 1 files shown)


LiterallyFabian:
First of all you're ignoring a directory that has nothing to do with Visual Studio (out of scope), and secondly you're completely ignoring the two documents that you linked yourself, which states that some files in the `.idea` directory *should* be included in your VCS - yet you ignored the entire folder.

If you don't want your JetBrains settings on Git that's on you, but adding it to a gitignore made for Visual Studio is a bad idea.

bdougie:
I am just as confused as @LiterallyFabian

PR #3960, opened 2022-01-25
add .esbuild directory to Node.gitignore

SiriusDely:

I am not an esbuild maintainer. However it would be nice when people are not missing for it on their .gitignore.

https://github.com/evanw/esbuild/blob/d4a7cd3/.gitignore#L5

If this is a new template:

 - **Link to application or project’s homepage**: https://esbuild.github.io

### DIFF
--- a/Node.gitignore
+++ b/Node.gitignore
@@ -128,3 +128,6 @@ dist
 .yarn/build-state.yml
 .yarn/install-state.gz
 .pnp.*
+
+# esbuild directories
+.esbuild


bdougie:
You are referencing the gitignore for the project. From my understanding, this project is leveraged by Node devs, but written in Go and includes Go binaries. 

Those files are not related and are not a justifiable reason to ignore them here. I would need a better reference for why this is needed.

SiriusDely:
@bdougie I think you are correct, this is not actually the convention from `esbuild`, but more like a configuration of another project. Which in this case is https://github.com/floydspace/serverless-esbuild or https://github.com/serverless/serverless . I will redirect this issue to those projects and close this one. Thank you.

PR #3962, opened 2022-01-27
Update KiCad.gitignore to ignore KiCad 6.0 zip backups

nopunin10did:

I like having github automatically populate gitignore files for my KiCad projects.

[KiCad 6 Common Preferences](https://docs.kicad.org/6.0/en/kicad/kicad.html#common_preferences)
Note the "Automatically backup projects" option.  In practice, this creates a subfolder with a name similar to the current project, just with "-backups" appended to it.  ZIP files are populated there.

It may be reasonable to more widely ignore ZIP files in KiCad, but it may be useful for people to maintain snapshots of stable manufacturer export files (which are usually packaged as ZIPs).

If this is a new template:

N/A

### DIFF
--- a/KiCad.gitignore
+++ b/KiCad.gitignore
@@ -27,3 +27,6 @@ fp-info-cache
 # Exported BOM files
 *.xml
 *.csv
+
+# Archived Backups (KiCad 6.0)
+**/*-backups/*.zip


bdougie:
Can you help me understand how this is beneficial to the rest of the KiCad community?

nopunin10did:
@bdougie 
Because KiCad 6 is constantly updating these backups, it means that a repository will get filled with version after version of binary files that hold onto information already present in git in other files.

Such files should be ignored by default for any KiCad user working in git.

wirecat (approved):
Thank you for contributing!

PR #3963, opened 2022-01-28
In Jekyll, ignore folders generated by bundler only at the top-level directory

kannanputhuval:
I am a user of https://github.com/jekyll/jekyll

When building a new project, I noticed that my CI/CD build failed, because it was missing a required file. The file should have been located at `foo/bar/vendor/foobar`.

 The file was missing, because any folders matching `vendor/` were ignored **in all directories**.

We should ignore `vendor/` and `.bundle/` in the top-level directory only.

https://jekyllrb.com/tutorials/using-jekyll-with-bundler/

Note that the sample `.gitignore` file at this location is ambiguous.

Connects https://github.com/jekyll/jekyll/pull/8950


wirecat:
I believe this PR has gone stale and I'm going to close for now @kannanputhuval.

PR #3967, opened 2022-02-01
Add Jabref temp files

pourmand1376:
Hi, I use JabRef for my reference management a lot. I understood that sometimes it creates some temporary files that I need to ignore from my commit. So, here's the change that makes annoying files disappear. 

I haven't referenced any docs, since I couldn't find any documentation to support my pull request.

### DIFF
--- a/TeX.gitignore
+++ b/TeX.gitignore
@@ -295,7 +295,12 @@ TSWLatexianTemp*
 # xwatermark package
 *.xwm
 
+# Jabref temp files 
+*.tmp
+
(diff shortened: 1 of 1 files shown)


bdougie:
Can you point to others in the community needing this?

pourmand1376:
I'm probably the first one using JabRef and git together! (Note: Jabref is new. not so many people know about it)

bdougie:
If you are the only one using it, it doesn't seem beneficial to add to a public template. Going to close for now and suggest providing links or documentation in the future.

If none exists, considering writing some.

PR #3972, opened 2022-02-05
Created Pylint section

anshtiwatne:

Pylint is one of the most popular Python linters, I found that pylint configuration files (pylintrc) were missing from gitignore so added them.

- https://docs.pylint.org/en/1.6.0/run.html


bdougie:
My understanding is that flake8 along with a few others are a bit more popular. Why this and not the others? With something as specific as linting, not sure if that opinion is needed in the template.

Curious to hear more thoughts on this.

anshunderscore:
Sure, that makes sense not to only include Pylint in this template. Maybe a collective linting section could be added?

PR #3975, opened 2022-02-16
Create Zig.gitignore

svc-user:

Zig is a new, modern take on C with a still growing userbase and active development. The growing userbase means more reporitories for which a Zig gitignore template is a little thing that makes sense in a busy day. 

The Zig compiler includes its own build system, which generates two folders in the project root. This is the recommended system for building Zig projects. 

[Zig build system docs](https://ziglang.org/documentation/master/#Zig-Build-System) 

[Zig language homepage](https://ziglang.org/)

### DIFF
--- a/Zig.gitignore
+++ b/Zig.gitignore
@@ -1,0 +1,4 @@
+# This file is for zig-specific build artifacts.
+
+zig-cache/
+zig-out/


its0x08 (approved):

AliChraghi (approved):

r00ster91 (approved):

ominitay (approved):

matu3ba (approved):

der-teufel-programming (approved):

leecannon (approved):

Zig.gitignore
```diff
@@ -0,0 +1,4 @@
+# This file is for zig-specific build artifacts.
+
+zig-cache/
+zig-out/
```
nektro, line 4: `zig-*` would be better

nektro (changes requested):

Zig.gitignore
bdoner: I think explicitness, in this case, is better than being implicit. I can see a scenario where someone would include a `zig-something` directory to use as a package or maybe create a `zig-build` directory to hold helpers for `build.zig`.

leecannon:
There are loads of zig libraries which are `zig-*`, you even have a few @nektro

Also if you ignore a submodule clones of the repo will have an empty submodule https://stackoverflow.com/a/62808998

Zig.gitignore
its0x08: Or could be `zig-{cache,out}` 🤔🤔

ominitay: It's fine as it is...

nektro: > I can see a scenario where someone would include a `zig-something` directory to use as a package

should be in a `libs` or `deps` folder. additionally people should not be venturing to begin with.

> or maybe create a `zig-build` directory to hold helpers for `build.zig`.

do you have even one example of this? feels very contrived

leecannon: > should be in a `libs` or `deps` folder
Why should something as simple as a gitignore constrain project structure?

> additionally people should not be vendoring to begin with.
While I would not vendor, vendoring is a completely valid thing to do. And having the default gitignore shoot people in the foot is bad.

bdoner: > do you have even one example of this? feels very contrived

No I don't have any examples - I'm not sure such an example exists. Yes, both my points are contrived. Nonetheless it's not technically difficult to end up with either of my examples, nor super far fetched. Edge cases matter. And in this case, edge cases are covered by being explicing, ignoring both paths. 

I'm not trying to prove you wrong, nektro, I just don't think the few saved bytes does anything good should someone deviate (greatly) from "how things are done".

ominitay: Agreeing with Lee here. The gitignore shouldn't dictate project structure. It should ignore files generated by the compiler, and that is all it should do. We shouldn't lock people out of using submodules or vendoring their dependencies.

BratishkaErik (approved):

Hanaasagi (approved):

Zig.gitignore
```diff
@@ -0,0 +1,4 @@
+# This file is for zig-specific build artifacts.
```
lumynou5, line 1: Do we really need this comment?

r00ster91: I don't think so. All directories included in this `.gitignore` even have a `zig-` prefix.
I would just make it
```
zig-cache/
zig-out/
```
without anything else.

bdoner: Does it really matter? Probably not. 
Do I care to setup my machine to make the change? Nope. 
Why not? Because GitHub obviously doesn't give a f**k :) 

Maybe go vote for this PR #4294 in case he gets lucky.

I'm just a bitter old man by now.

r00ster91: That PR is missing a trailing newline.

But I mean who uses these templates anyway? I definitely won't go here to copy two lines for a file I could write in 10 seconds. I'm guessing the main reason people want to see it merged is for publicity for Zig.

Hanaasagi: @r00ster91 
When creating a project on GitHub, you can choose the programming language used for the project, and this is automatically generated, similar to your LICENSE file. `.gitignore` is used for that. :)

bdoner: > That PR is missing a trailing newline.
> 
> But I mean who uses these templates anyway? I definitely won't go here to copy two lines for a file I could write in 10 seconds. I'm guessing the main reason people want to see it merged is for publicity for Zig.

I'm surely not here to copy gitignore files, but as @Hanaasagi mentions, if it's in a dropdown I'd rather just select it there than writing the three-line (including newline) file! 😅

ewrogers: As someone who is entirely new to Zig (like a lot of people), one of the first questions I ask is "what do I need to ignore?"

It is definitely nice to have that documented somewhere and I usually do pull it off GitHub if the built-in tooling/project template does not generate one.

Tythos:
There are three open PRs for the same change (#4354 , #4294 , and this one). Recommend this PR be accepted and others closed. Changes are identical (+/- a comment line) and build artifact structures have mostly stabilized in zig compiler passes (as of v0.12).

der-teufel-programming:
As of Zig 0.13.0 `zig-cache` should now be `.zig-cache`

Zig.gitignore
```diff
@@ -0,0 +1,4 @@
...
+
+zig-cache/
```
bschaatsbergen, line 3: As per the recent 0.13.0 release of Zig, `zig-cache/` has been renamed to `.zig-cache/`. See https://ziglang.org/download/0.13.0/release-notes.html#codezig-cachecode-renamed-to-codezig-cachecode.

I've opened a PR that proposes an up-to-date `Zig.gitignore`: https://github.com/github/gitignore/pull/4503

dooleydevin (approved):
Appreciate the contribution (and the patience)! This seems to be the most active PR for this change, so I'll start with merging this one and work towards integrating any other additions separately 😄

PR #3976, opened 2022-02-16
Ignore PDM project-wide config

astrojuanlu:

Will make it easier for new users of pdm to get the configuration right from the start. Among other reasons (see below), `pdm init` creates a `.pdm.toml` storing the absolute path of the Python interpreter, which is machine-dependent and therefore should not be tracked in version control.

See https://pdm.fming.dev/#use-with-ide

See also https://github.com/pdm-project/pdm/issues/699

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -101,7 +101,13 @@ ipython_config.py
 #   https://python-poetry.org/docs/basic-usage/#commit-your-poetrylock-file-to-version-control
 #poetry.lock
 
-# PEP 582; used by e.g. github.com/David-OConnor/pyflow
+# pdm
+#   pdm stores project-wide configurations in .pdm.toml, and it is recommended to not include it
+#   in version control.
+#   https://pdm.fming.dev/#use-with-ide
+.pdm.toml
+
+# PEP 582; used by e.g. github.com/David-OConnor/pyflow and github.com/pdm-project/pdm
 __pypackages__/
 
 # Celery stuff
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -102,6 +102,14 @@ ipython_config.py
 #poetry.lock
 
 # pdm
+#   Similar to Pipfile.lock, it is generally recommended to include pdm.lock in version control.
+#pdm.lock
+#   pdm stores project-wide configurations in .pdm.toml, but it is recommended to not include it
+#   in version control.
+#   https://pdm.fming.dev/#use-with-ide
+.pdm.toml
+
+# PEP 582; used by e.g. github.com/David-OConnor/pyflow and github.com/pdm-project/pdm
 #   pdm stores project-wide configurations in .pdm.toml, and it is recommended to not include it
 #   in version control.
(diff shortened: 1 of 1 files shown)


Python.gitignore
```diff
@@ -101,7 +101,13 @@ ipython_config.py
 #   https://python-poetry.org/docs/basic-usage/#commit-your-poetrylock-file-to-version-control
 #poetry.lock
 
-# PEP 582; used by e.g. github.com/David-OConnor/pyflow
+# pdm
```
pypros, line 104: I think it would be worth adding info about pdm.lock also.
```suggestion
# pdm
#   Similar to Pipfile.lock, it is generally recommended to include pdm.lock in version control.
#pdm.lock
#   pdm stores project-wide configurations in .pdm.toml, but it is recommended to not include it
#   in version control.
#   https://pdm.fming.dev/#use-with-ide
.pdm.toml

# PEP 582; used by e.g. github.com/David-OConnor/pyflow and github.com/pdm-project/pdm
```

Python.gitignore
```diff
@@ -101,7 +101,21 @@ ipython_config.py
...
+#   Similar to Pipfile.lock, it is generally recommended to include pdm.lock in version control.
+#pdm.lock
+#   pdm stores project-wide configurations in .pdm.toml, but it is recommended to not include it
+#   in version control.
+#   https://pdm.fming.dev/#use-with-ide
+.pdm.toml
+
+# PEP 582; used by e.g. github.com/David-OConnor/pyflow and github.com/pdm-project/pdm
```
pypros, line 112: I think something went wrong. Now we have duplicate lines. Lines 112 to 117 are not needed, you should delete them. Apparently, it's my fault as I added suggestions. I'm sorry.

astrojuanlu: Oops, you're right! I didn't see it before applying, sorry. Now it's fixed

pypros:
Hello [bdougie](https://github.com/bdougie) 

Could you have a look? Could you add some comments or approve it?

Br,
Łukasz

astrojuanlu:
Friendly ping @pypros @bdougie

pypros:
How can I help you?

astrojuanlu:
Sorry, I assumed you had write access here.

pypros:
No, I found the same problem as you and help in the review.

bdougie (approved):

Python.gitignore
WILLIAMJOESPHBAUGHMANJR: Is this good for use

astrojuanlu:
@Iceberg-Dev that's an individual commit, please look at the whole diff. This is the reason https://github.com/github/gitignore/pull/3976#discussion_r810583909

thisnameisme:
no idea what I am doing, but I am having fun doing it.

Python.gitignore
2021119: 8e67b9420cb6796e5eeca72682babdb06627ec8c

PR #4004, opened 2022-03-16
Added LaTeX gitignore file

devceline:

- With more people using latex to write their research papers, books, etc.
  Having a quick gitignore that includes all the intermediary & build files will
  make backing things up easier.
- I've recently started backing up my latex work on github and came here looking
  for a template like this, didn't find it so I made one.
- This includes all the different file formats used by the different build
  systems like `latexmk`, `pdflatex`, etc.

- Couldn't find documentation listing all the different build files, but here's 
  a github issue mentioning them: https://github.com/lervag/vimtex/issues/1297
- All these files are generated when `vimtex` runs the underlying latex build system
- And here's a link from `latexmk` docs mentioning it creates a bunch of temporary files: 
  https://mg.readthedocs.io/latexmk.html
- You can also verify this by building latex files locally.

If this is a new template:

 - **Link to application or project’s homepage**: https://www.latex-project.org/


jorgepz:
Hi! great idea! I was just about to create the PR and found this. May I suggest not including .pdf extension? ... many images are in pdf formats so that might be a problem, wouldn't it?

jorgepz:
I proposed some additional changes to the @devceline 's PR in https://github.com/devceline/gitignore/pull/1

devceline:
> Hi! great idea! I was just about to create the PR and found this. May I suggest not including .pdf extension? ... many images are in pdf formats so that might be a problem, wouldn't it?

Yep, I did consider that. Will be removing the `*.pdf` from the template

jorgepz:
If capitals are relevant....it should be LaTeX.gitignore...

devceline:
> If capitals are relevant....it should be LaTeX.gitignore...

 Yep, considering `CMake` is spelled with proper capitalization, I'll quickly rename the file.

PR #4039, opened 2022-04-14
Update Swift.gitignore // Add Pods/ line to add Pods Generated

efemazlumoglu:
I add Pods cause if you are for example add RealmSwift as a pod you can not push it to the git cause big file error appears so it is need to be there.

_TODO_

_TODO_

If this is a new template:

### DIFF
--- a/Swift.gitignore
+++ b/Swift.gitignore
@@ -12,6 +12,7 @@ xcuserdata/
 ## compatibility with Xcode 3 and earlier (ignoring not required starting Xcode 4)
 build/
 DerivedData/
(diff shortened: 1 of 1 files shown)


ealeksandrov (changes requested):
CocoaPods related section already exists in this file.

wirecat:
Thank you for pointing that out @ealeksandrov! ❤️ 

@efemazlumoglu we won't be accepting this at this time as we already have a section documented for this and I would prefer to have this be an optional configuration for now.

PR #4040, opened 2022-04-15
Add *.trx to ignored files

kkostrzewa-coupa:
As a user of .gitignore files, I wanted to include the ephemeral .trx (Test Result XML) files, as they are frequently built under project root, but (IMHO) not of value to commit to SCM.

No links available. This is my own opinion, from a project that I'm working on.

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -43,6 +43,7 @@ Generated\ Files/
 # MSTest test Results
 [Tt]est[Rr]esult*/
 [Bb]uild[Ll]og.*
+*.trx
 
 # NUnit
 *.VisualState.xml
@@ -396,6 +397,3 @@ FodyWeavers.xsd
 
 # JetBrains Rider
 *.sln.iml
-
-# Microsoft Test Results XML file
-*.trx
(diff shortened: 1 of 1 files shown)


ggirard07:
Bumping this

**Test results from MSTest and NUnit are already excluded**, so it seems logical to do the same regardless of which test runner is being used! Usually test reports are instead published as part of the build process as an independent artifact/metadata.

Maybe updating the change to place this exclusion next to the ones for MSTest and NUnit could help get this approved?

ggirard07:
Actually just realizing that `trx` is the MSTest output format, not XUnit output.

So current support for MSTest does not support override of default output location. Excluding `*.trx` as OP is proposing would fix the issue. This also confirms that the change should be moved in [the existing MSTest section](https://github.com/github/gitignore/blob/4488915eec0b3a45b5c63ead28f286819c0917de/VisualStudio.gitignore#L43).

kkostrzewa-coupa:
Moved to MSTest section, thx @ggirard07

ggirard07:
Thanks @kkostrzewa-coupa !
Now let's pray for a repo owner to look back into this as it seems a justified issue to fix... maybe @bdougie could help with this?

wirecat (approved):

PR #4041, opened 2022-04-15
Added DS_Store

NetsaiMeklit:

When programming on mac computers the .DS_Store file is very common. This should be added to the .gitignore so that it does not accidentally get pushed. When i have worked with my many friends it is a common issue that they push this file for no reason. I therefore think it would be good to have this included so that this mistake is not carried though. Especially since the .DS_Store is a hidden file and many people do not think about its existence and therefore it can be hard to miss. Oh wait i meant it can be easy to miss.

_TODO_

[_TODO_](https://en.wikipedia.org/wiki/.DS_Store)

If this is a new template:

### DIFF
--- a/Swift.gitignore
+++ b/Swift.gitignore
@@ -21,6 +21,7 @@ DerivedData/
 !default.mode2v3
 *.perspectivev3
 !default.perspectivev3
+.DS_Store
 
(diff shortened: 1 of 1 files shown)


bdougie:
Should this be leveraged from the global gitignore instead? 

PR #4051, opened 2022-04-26
Fix: do not ignore "*target/" subdirectories.

voidpointer0x00:

This change fixes ignoring `*.target.*` java packages.

### DIFF
--- a/Maven.gitignore
+++ b/Maven.gitignore
@@ -1,4 +1,4 @@
-target/
+/target/
 pom.xml.tag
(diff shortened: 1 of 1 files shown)


dgruntz:
I just wanted to create the exact same pull request in order to prevent that e.g. Java source packages with the name `*.target.* ` are excluded.

voidpointer0x00:
@dgruntz dunno why it's not merged yet, they don't even answer 😁

PR #4052, opened 2022-04-27
adding Intellisense precompiled headers

Murazaki:

ipch files can appear in Unreal projects folder with VS 2022.
I personally use Unreal Engine and git/perforce, and I would like to have Intellisense precompiled headers ignored on version control in the first place. I think other devs would like it as well. 

For more details on ipch files :
https://stackoverflow.com/questions/19398287/ipch-files-on-a-visual-studio-project

### DIFF
--- a/UnrealEngine.gitignore
+++ b/UnrealEngine.gitignore
@@ -10,6 +10,7 @@
 # Precompiled Headers
 *.gch
 *.pch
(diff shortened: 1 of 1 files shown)


Murazaki:
Any review ? Do you plan on merging that ?

PR #4056, opened 2022-04-30
Added CMakeUserProfiles.json to CMake template

rvodden:
CMake 3.19 introduced a feature which allows project maintainers to distribute CMake settings with their code. In addition, there is a facility for users to maintain a local set of CMake settings for convenience.

[CMake Profiles Documentation](https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html)

>CMakePresets.json may be checked into a version control system, and CMakeUserPresets.json should NOT be checked in. For example, if a project is using Git, CMakePresets.json may be tracked, and CMakeUserPresets.json should be added to the .gitignore.

### DIFF
--- a/CMake.gitignore
+++ b/CMake.gitignore
@@ -8,4 +8,5 @@ cmake_install.cmake
 install_manifest.txt
 compile_commands.json
 CTestTestfile.cmake
+CMakeUserPresets.json
 _deps


aloisklink:
For the maintainers of this repo: this PR is a duplicate of #3595. That one seems to have been opened by an employee of Kitware 
([and CMake contributor](https://gitlab.kitware.com/kyle.edwards)), so it's probably worth sending questions that way.

PR #4065, opened 2022-05-09
Update Perl .gitignore to include Carton

lathropd:
Ignore /local and cpanfile.snapshot which really shouldn't be in VCS

Carton has become the go-to dependency management for Perl and really should be reflected here.

[Carton documentation](https://metacpan.org/pod/Carton)


Perl.gitignore
```diff
@@ -33,3 +33,7 @@ inc/
 /MANIFEST.bak
 /pm_to_blib
 /*.zip
+
+# Carton
+/local/
+cpanfile.snapshot
```
daenney, line 39: All the examples in https://metacpan.org/pod/Carton seem to explicitly add and commit `cpanfile.snapshot`. That would suggest it should not be in `.gitignore` by default?

oalders: I don't think `cpanfile.snapshot` should be ignored by default. @miyagawa could confirm.

miyagawa: It depends. For a library i.e. CPAN module, the `cpanfile.snapshot` file should be excluded, like this PR does, since people would run it on different platforms with different perl versions.

For an "application" repository, i.e. your work app, `cpanfile.snapshot` is definitely intended to be committed.

miyagawa:
You should add `/.carmel/` as well while you're at it :) 
https://metacpan.org/pod/Carmel#Development

lathropd (commented):
OK, this appears to reflect feedback from @oalders @daenney @miyagawa

oalders (approved):

rwp0 (approved):

oalders:
This PR should be merged, not closed.

wirecat:
Thank you @oalders for the poke, @lathropd for the changes, and everyone for their contributions and review on this one!

PR #3758, opened 2021-06-14
Ignore misc files added by VSCode for Maven

NishkarshRaj:
**Reasons for making this change:**

VSCode adds some misc files which are not meant to be pushed to source control.

Screenshot 2021-06-14 at 1 19 50 PM

**Links to documentation supporting these rule changes:**

_TODO_

If this is a new template:

 - **Link to application or project’s homepage**: _TODO_

PR #4070, opened 2022-05-17
Update Maven.gitignore

claubermartins:

It's good to take it out on legacy projects that can't be compiled with the latest and greatest java compiler or a source code generated by a 3rd party produces a lot of warnings that are benign but pollute your view. Otherwise keep it in .gitignore

[https://www.ibm.com/docs/en/spm/7.0.4?topic=files-settings-directory](url)
[https://wiki.lyrasis.org/pages/viewpage.action?pageId=30221028&focusedCommentId=30221030](url)

### DIFF
--- a/Maven.gitignore
+++ b/Maven.gitignore
@@ -15,3 +15,6 @@ buildNumber.properties
 .project
 # JDT-specific (Eclipse Java Development Tools)
 .classpath
+
+#This file records project specific settings vs. workspace preferences.
+.settings/


Maven.gitignore
```diff
@@ -15,3 +15,6 @@ buildNumber.properties
 .project
 # JDT-specific (Eclipse Java Development Tools)
 .classpath
+
+#This file records project specific settings vs. workspace preferences.
```
stonar96, line 19: Missing space `# This ...`.

stonar96 (changes requested):
Duplicate of #3758. There are also comments on why `.settings/` wasn't added: (https://github.com/github/gitignore/pull/3710#issuecomment-834650616), https://github.com/github/gitignore/pull/3710#issuecomment-912181502

Maven.gitignore
claubermartins: I used other sources of clarification, but I observed that the discussions you mentioned are mature and clear.

claubermartins:
I have used other sources of clarification, but I have observed that the discussions mentioned to me are mature and clear.

PR #4081, opened 2022-05-31
Ignore go.work.sum

katexochen:

As the `go.work` file is excluded, the `go.work.sum` file, which is generated based on the `go.work` file, should be excluded, too.
Also mentioned in https://github.com/github/gitignore/commit/fbc053fe49d7f3b4a882ddf9651fc60f8954db21#commitcomment-73352144

Currently there isn't much documentation regarding `go.work.sum` (see https://github.com/golang/go/issues/51941), the only thing I can refer to is the [proposal](https://go.googlesource.com/proposal/+/master/design/45713-workspace.md#preventing-files-from-being-checked-in-to-repositories), which suggests to not check in the `go.work` file.

### DIFF
--- a/Go.gitignore
+++ b/Go.gitignore
@@ -19,3 +19,4 @@
 
 # Go workspace file
 go.work
+go.work.sum


avelino (approved):
LGTM

These files shouldn't be commited since they are specific to a **local dev environment**.

Nirusu (approved):

StLeoX:
Why not ` **go.sum `?
The sum file is used to check locally, I think.

katexochen:
> Why not `**go.sum`? The sum file is used to check locally, I think.

> [ Typically your module's go.sum file should be committed along with your go.mod file.](https://github.com/golang/go/wiki/Modules#should-i-commit-my-gosum-file-as-well-as-my-gomod-file)

rubenpoppe:
@avelino, @Nirusu will this get merged?

smoyer64:
I'm in favor of this change - if `go.work` is ignored, then `go.work.sum` should also be ignored.  Note that there's not actually a recommendation or even consensus on whether these files should be ignored.  My read on the situation is that for cases like a monorepo, you probably do want to commit these files but in many other situations, you do not. See the discussion at https://github.com/golang/go/issues/53502.

goto1134:
@bdougie, could you please take a look at this PR since you already merged #3884? The file mentioned here is a checksum file that is autogenerated, and present iif `go.work` file is present in the local repository.  Since `go.work` is ignored, we should ignore `go.work.sum` files as well.

Omarabdul3ziz:
Hey there, I noticed that the PR is already approved. is there an estimated timeframe for when it might be merged? Just curious and looking forward to it

smoyer64:
I think those gray check marks indicate that the approvals are by developers who are not CODEOWNERS.  I don't even see a merge button!

wirecat (approved):
Thank you for following our contribution guide and helping improve our templates!
I just finished reviewing and this looks good to me.

PR #4088, opened 2022-06-12
Update Unity gitignore to ignore personal layout changes

smitdylan2001:
Unity 2021+ has a folder called Layouts (with default-2021.dwlt file for info) which stores all layout changes. This should not be pushed to git

New layout folder/file are user generated files storing the Unity window layout. This is different per person and will be edited every layout change. This should not be included in a repo because of the per-user based nature of the layout. This is added in Unity 2021, which now has an LTS release, so it will be used often now for new projects

### DIFF
--- a/Unity.gitignore
+++ b/Unity.gitignore
@@ -8,6 +8,7 @@
 /[Bb]uild/
 /[Bb]uilds/
 /[Ll]ogs/
+/[Ll]ayouts/
 /[Uu]ser[Ss]ettings/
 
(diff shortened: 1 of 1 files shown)


smitdylan2001:
Closed with https://github.com/github/gitignore/pull/4088

RasoulMotaghi35:
سلام 🙏

در تاریخ دوشنبه ۲۰ مهٔ ۲۰۲۴،‏ ۱۳:۴۳ Dylan Smit ***@***.***>
نوشت:

PR #4095, opened 2022-06-26
Add .idea folder and .DS_Store file to Unity gitignore.

avinashmx7:


lewismunday (approved):

Casey-Hofland:
This is a more complete list of what you may ignore from mac
```
# Mac general files
.DS_Store
.AppleDouble
.LSOverride

# Mac Icon must end with two \r
Icon

# Mac thumbnails
._*

# Mac files that might appear in the root of a volume
.DocumentRevisions-V100
.fseventsd
.Spotlight-V100
.TemporaryItems
.Trashes
.VolumeIcon.icns
.com.apple.timemachine.donotpresent

# Mac directories potentially created on remote AFP share
.AppleDB
.AppleDesktop
Network Trash Folder
Temporary Items
.apdisk
```

Probably a little much, but up to Mac thumbnails would be very reasonable to add next to the .DS_Store

wirecat:

🙇

PR #4103, opened 2022-07-05
gitignore for Visual Studio Code ( vscode)

abhay-h:

As of  05-07-2022 , there is no official** .gitignore **for  **Visual Studio Code ( vscode).**
I'm contributing my version of **vscode .gitignore which cover **

1. All unwanted vscode files.
2. All unwanted node files.


editicalu:
https://github.com/github/gitignore/blob/main/Global/VisualStudioCode.gitignore

PR #3567, opened 2020-10-31
Add .godot/ folder to Godot.gitignore for Godot 4

aaronfranke:
**Reasons for making this change:**

Add `.godot/` folder to the Godot gitignore because in Godot 4.0 all generated files that should be gitignored will go in `.godot/`. This PR also updates the comments, let me know what you think.

Ignoring `.godot/` is harmless in Godot 3.x, and ignoring the rest will be harmless in Godot 4.x, so we can include both.

**Links to documentation supporting these rule changes:**

https://github.com/godotengine/godot/pull/38607 and https://github.com/godotengine/godot/pull/38704

 - **Link to application or project’s homepage**: https://godotengine.org/ and https://github.com/godotengine/godot

PR #4107, opened 2022-07-08
add .godot/

ygypt:
.godot/ is a new folder in godot that tracks shader chaching and editor layouts. should not be shared between contributors


Calinou:
Duplicate of https://github.com/github/gitignore/pull/3567.

PR #4110, opened 2022-07-10
add ds_store to ignore files at android

youssefshaaban:
Mac OS generation this file  .DS_Store I add it to ignore file


wirecat:

🙇

PR #4115, opened 2022-07-19
Add Hyperf gitignore template

polynds:

Prevent the cache and log files in the project from being uploaded to the warehouse.

https://github.com/hyperf/hyperf/blob/master/.gitignore

If this is a new template:

 - **Link to application or project’s homepage**: https://github.com/hyperf/hyperf


polynds:
Add Hyperf gitignore template

PR #4116, opened 2022-07-20
Create .gitignore for Zig

yurisi0212:

Exclude log and cache files when creating a project in Zig

https://github.com/ziglang/zig/blob/master/.gitignore

If this is a new template:

 - **Link to application or project’s homepage**: https://github.com/ziglang/zig/


yurisi0212:
sorry already existed

PR #4121, opened 2022-07-29
Update Scala template to include `.tasty` files introduced in Scala 3

KacperFKorban:

Scala 3 introduced a new type of IR file that is produced when compiling Scala sources with the new compiler - `.tasty` files. When version control is concerned, those files should be treated the same way `.class` files are.

My relationship to the project: I'm one of the maintainers [see here](https://github.com/lampepfl/dotty/blob/main/MAINTENANCE.md).

[https://docs.scala-lang.org/scala3/guides/tasty-overview.html](https://docs.scala-lang.org/scala3/guides/tasty-overview.html)

### DIFF
--- a/Scala.gitignore
+++ b/Scala.gitignore
@@ -1,4 +1,5 @@
 *.class
+*.tasty
 *.log
 
 # virtual machine crash logs, see http://www.java.com/en/download/help/error_hotspot.xml


michaelimasih:
> 
> Scala 3 introduced a new type of IR file that is produced when compiling Scala sources with the new compiler - `.tasty` files. When version control is concerned, those files should be treated the same way `.class` files are.
> 
> My relationship to the project: I'm one of the maintainers [see here](https://github.com/lampepfl/dotty/blob/main/MAINTENANCE.md).
> 
> **Links to documentation supporting these rule changes:**
> 
> https://docs.scala-lang.org/scala3/guides/tasty-overview.html

KacperFKorban:
I would like this PR to be merged. It will be a usability improvement for people using Scala 3.

KacperFKorban:
I want to keep this open.

beegriff2:
Please leave Open

Chrissymars (approved):

KacperFKorban:
I want to keep this open.

wirecat (approved):
Sorry for the long delay and thank you very much for the contribution!

PR #4128, opened 2022-08-17
Update outdated link and add TEMP folder to ignore

nul800sebastiaan:

I'm on the maintainer team at Umbraco.


nul800sebastiaan:
I initially only changed the link but found that the temp folder is not being ignored when you use this ignore.

When you create a new Umbraco project using our dotnet templates, you get this ignore file which does include the temp folder: https://github.com/umbraco/Umbraco-CMS/blob/v10/contrib/templates/UmbracoProject/.gitignore#L469

PR #4142, opened 2022-08-30
ignore ncrunch solution file for visual studio

AChehre:
Ignore NCrunch solution file for visual studio

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -155,6 +155,7 @@ coverage*.info
 _NCrunch_*
 .*crunch*.local.xml
 nCrunchTemp_*
+*.ncrunchsolution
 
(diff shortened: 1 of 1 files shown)


fuzzybair:
What is the justification for ignoring the ncrunchsolution file? We use that as a way for the team to have constant configuration when the NCrunch defaults do not work. For example are solution does not work with AnyCPU due to external dependencies the ncrunchsolution file is how you tell NCrunch to use x64. True we could have everyone configure it manually but that is an onboarding/fresh clone challenge, or we could add !*.ncrunchsolution to all of our repositories, but I do not see why you would not want a consistent experience on your team. Do you have some supporting documentation that that file should not be tracked with your solution.

PR #4149, opened 2022-09-10
Update Unity.gitignore

FozerG:
Ignore proguard mapping files that appear on build

### DIFF
--- a/Unity.gitignore
+++ b/Unity.gitignore
@@ -61,6 +61,9 @@ sysinfo.txt
 *.unitypackage
 *.app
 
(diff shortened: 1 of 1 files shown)


proJM-Coding:
What about .meta files?

FozerZ:
> What about .meta files?

.meta files form a critical part of your project and hence should be committed to your Version Control System.

Meta files are used to store the settings for each Asset along with a Globally Unique Identifier (GUID) that is used to track references between Assets, for instance when a Material Asset references a Texture Asset.

PR #4156, opened 2022-09-18
Update Python.gitignore

MaxBui95:

_TODO_

_TODO_

If this is a new template:

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -1,4 +1,9 @@
-# Byte-compiled / optimized / DLL files
+ .idea/
+ .pytest_cache
+ *.__pycache__/
+ env/
+ 
+ Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]
 *$py.class


prashantk01 (approved):

Python.gitignore
```diff
@@ -1,4 +1,9 @@
-# Byte-compiled / optimized / DLL files
+ .idea/
+ .pytest_cache
+ *.__pycache__/
+ env/
+ 
+ Byte-compiled / optimized / DLL files
```
mvolfik, line 6:
```suggestion
# Byte-compiled / optimized / DLL files
```

This is a comment, so keep the `#` at the beginning

Python.gitignore
```diff
@@ -1,4 +1,9 @@
-# Byte-compiled / optimized / DLL files
+ .idea/
```
mvolfik, line 1: This isn't from Python, but from Jetbrains editors/IntelliJ

Python.gitignore
```diff
@@ -1,4 +1,9 @@
...
+ .pytest_cache
+ *.__pycache__/
+ env/
```
mvolfik, line 4: Is this created by default or has some special meaning for Python? If it's just a convention or your preference to create this folder, it shouldn't be included here

mvolfik (changes requested):

Gatito1010:
Gracias error mio noe avía dado cuenta lo arreglo gracias

PR #4173, opened 2022-10-17
ignore meta files of unity packages

jmarsh411:
Unity will automatically generate [meta files](https://docs.unity3d.com/Manual/AssetMetadata.html) for just about any file added to its directory structure. This is in order to maintain references between things that may be moving around or getting renamed in the project.

It's very common that an asset from the Unity Asset Store comes with `.unitypackage` files for optional features or compatibility add-ons. In such a case, the importer's machine will contain a `.unitypackage` and `.unitypackage.meta` file. 

The `.unitypackage` file is ignored by this template, but the `.unitypackage.meta` file is not. If the meta file isn't also ignored, the person who imports an asset containing a `.unitypackage` file could accidentally commit the meta files. The next person will pull in those meta files, which Unity will consider orphaned without the accompanying `.unitypackage` file. Unity will automatically delete the orphaned meta files, and those deletes also tend to get committed. The importer auto-adds and commits them again, and the cycle goes on and on until someone like me gets fed up.

This change fixes that inconsistency by adding the `*.unitypackage.meta` to the ignore list.

### DIFF
--- a/Unity.gitignore
+++ b/Unity.gitignore
@@ -59,6 +59,7 @@ sysinfo.txt
 *.apk
 *.aab
 *.unitypackage
+*.unitypackage.meta
 *.app
 
 # Crashlytics generated file


Dstoney (approved):

Sambo3975:
Why was this approved but never pulled?

jmarsh411:
I would merge it myself, but I don't have the access.

@Dstoney can you pull it?

danliukuri:
Considering the analogy with `.unitypackage`, to my mind, it would be logical to extend the ignore list to cover all `.meta` files for already ignored files under the `Assets` folder. Also, it would be ideal if it were possible to use some generic approach and not write an additional line for each file type.

jmarsh411:
> Considering the analogy with `.unitypackage`, to my mind, it would be logical to extend the ignore list to cover all `.meta` files for already ignored files under the `Assets` folder. Also, it would be ideal if it were possible to use some generic approach and not write an additional line for each file extension.

It looks like someone once upon a time had a generic approach that someone else removed for reasons I'm not really understanding.
https://github.com/github/gitignore/commit/7bcea281916d32d480abfe41a78b204c9935d5b4

I think the discussion of a more generic approach can be opened up in a new PR. But I still believe .unitypackage files cover the most common frustration.

In any case, I'll try to find someone responsive that has write access to merge this.

jmarsh411:
@github @shiftkey @bdougie @aroben

Sorry to blast you all. This PR has been approved and sitting for a while. I read the workflow part of the README, but it's not really clear how to proceed in this case.

RyanBorkBork:
Can someone please merge this? Having to track this down on all new projects is a bit of a drag!

Thanitsak:
Will this also do the trick? Instead of just ignoring a single file ".unitypackage.meta"

```
# Asset meta data should only be ignored when the corresponding asset is also ignored
!/[Aa]ssets/**/*.meta
```
Any ideas guys?

Kottakji:
@Dstoney /ping

It would be nice to have this merged.

jmarsh411:
Tagging a few more people who have merged recent pull requests.

@thorrsson 
@dooleydevin 
@wirecat 
@Dstoney 

This PR has been approved and just sitting for years. Can someone please merge it so Unity developers can stop dealing with unnecessary noise in their commits?

thorrsson (approved):

jmarsh411:
Thank you! We all appreciate it!!

PR #4182, opened 2022-10-31
Remove extra .cache pattern

brajnacs:
pattern ".cache" is duplicate and already defined in parcel-bundler section

".cache" patter is extra and already defined in parcel-bundler section at line number 87

https://github.com/github/gitignore/blob/main/Node.gitignore#L83

If this is a new template:

 - **Link to application or project’s homepage**: NA

### DIFF
--- a/Node.gitignore
+++ b/Node.gitignore
@@ -102,7 +102,6 @@ dist
 
 # vuepress v2.x temp and cache directory
 .temp
-.cache
 
 # Docusaurus cache and generated files
 .docusaurus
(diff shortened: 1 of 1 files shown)


Node.gitignore
```diff
@@ -102,7 +102,6 @@ dist
 
 # vuepress v2.x temp and cache directory
```
spirit-x64, line 103: You should update the comment and remove "and cache directory"

wirecat (approved):

PR #4184, opened 2022-11-06
Adding the dependency lock file the template.

16c7x:
The lock file ties to an external provider dependency, it's something you'd want the option to not have under git control. https://developer.hashicorp.com/terraform/language/files/dependency-lock

It's a file that many would probably want to ignore.

_TODO_

_TODO_

If this is a new template:

### DIFF
--- a/Terraform.gitignore
+++ b/Terraform.gitignore
@@ -32,3 +32,6 @@ override.tf.json
 # Ignore CLI configuration files
 .terraformrc
 terraform.rc
+
+# Ignore Dependancy lock files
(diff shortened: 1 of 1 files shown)


lvjp:
The [Lock File Location](https://developer.hashicorp.com/terraform/language/files/dependency-lock#lock-file-location) section says totally the opposite of your arguments.

> You should include this file in your version control repository so that you can discuss potential changes to your external dependencies via code review, just as you would discuss potential changes to your configuration itself.

PR #4198, opened 2022-12-01
Update Unity.gitignore

Upsidedrop:
Fixes bug that stops the .gitignore from working on Unity projects.

_TODO_

_TODO_

If this is a new template:

### DIFF
--- a/Unity.gitignore
+++ b/Unity.gitignore
@@ -2,13 +2,13 @@
 #
 # Get latest from https://github.com/github/gitignore/blob/main/Unity.gitignore
 #
-/[Ll]ibrary/
-/[Tt]emp/
-/[Oo]bj/
-/[Bb]uild/
-/[Bb]uilds/
-/[Ll]ogs/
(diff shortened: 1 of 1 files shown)


DamianSuess:
@Upsidedrop  Yes, 100% yes! I've been making this change every time I used the GitHub template

For newer devs or anyone who uses these pre-baked templates, this update helps save repos from adding 15,000+ files in the `Library/` folder alone which are not needed to be maintained by source control.

**Reason for making this change:**

The leading slash does not ignore the folders when the project is located in a subfolder such as, `src/` or `ProjectName/`.

andresmpajaro:
Hi. That's a good way to solve the problem but you must have in mind that if you put the slash at the end of the name (Name/) git won't track sub folders with the name "Name". That's the function of slash at the end of the folder's name. So, let's suppose you have a project where you must have an asset call Library, then you won't track that file if you add to the gitignore Library/

PR #4201, opened 2022-12-04
Update Godot.gitignore for Godot 4+

digiur:
Godot 4 Beta was recently released. It adds a .godot folder that should be ignored in git

I am an end user that used this template but it was missing the .godot folder

.gitignore file from a official godot 4 demo project:
https://github.com/godotengine/godot-demo-projects/blob/master/.gitignore


digiur:
I didn't check for duplicates before creating this

PR #4202, opened 2022-12-04
Add Mac OS .DS_Store files ignoring

kbarkati:

I use to use LaTeX for several projects, mainly publishing projects.
When working on Mac OS, adding files or folders to a TeX repository may yield .DS_Store files. Those files are useless regarding versioning, hence they add unnecessary noise to the repository, and would better be ignored automatically.

[.DS-Store files Wikipedia page](https://en.wikipedia.org/wiki/.DS_Store)

### DIFF
--- a/TeX.gitignore
+++ b/TeX.gitignore
@@ -299,3 +299,7 @@ TSWLatexianTemp*
 # option is specified. Footnotes are the stored in a file with suffix Notes.bib.
 # Uncomment the next line to have this generated file ignored.
 #*Notes.bib
+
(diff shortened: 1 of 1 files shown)


suhailskhan:
This wouldn't be necessary since this is already present in the global .gitignore for macOS in this same repository, see [Line 2](https://github.com/github/gitignore/blob/4488915eec0b3a45b5c63ead28f286819c0917de/Global/macOS.gitignore#L2).

wirecat:

🙇

PR #4209, opened 2022-12-11
Update Paths in JetBrains.gitignore File

ondergormez:
Update paths to the relative path.

I have a [repository](https://github.com/ondergormez/BLM5127_Big_Data_Analytics/tree/master/02-Homeworks/Homework1/Question2) hierarchy that has multiple projects inside one repository.
I did this change because;
* The expectation is not just that the files in the root folders are ignored.
* For supporting multiple sub-projects inside the same repository.
* Fully backward compatible.

* Tested on the repo [my repo](https://github.com/ondergormez/BLM5127_Big_Data_Analytics/blob/master/.gitignore) and works fine.
* Inside the site [git-scm.com](https://git-scm.com/docs/gitignore#_pattern_format), you can see that the rule starting with

> Two consecutive asterisks ...

### DIFF
--- a/Global/JetBrains.gitignore
+++ b/Global/JetBrains.gitignore
@@ -2,41 +2,41 @@
 # Reference: https://intellij-support.jetbrains.com/hc/en-us/articles/206544839
 
 # User-specific stuff
-.idea/**/workspace.xml
-.idea/**/tasks.xml
-.idea/**/usage.statistics.xml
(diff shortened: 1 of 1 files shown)


ondergormez:
Hello there,

Is there any update or thoughts on this?

FYI @drothmaler, @bdougie, @shiftkey

PR #4213, opened 2022-12-20
Updated Unreal GitIgnore to ignore nested  Binaries/Intermediate folders

jackfranzen:
I.E Plugins/MyCompaniesPlugins/Plugin1/Intermediate/ would not have been ignored previously.

Usually, plugins in unreal are not nested, but there is no reason they can't be!

I have no relationship to this project.


TheThief:
This is a dupe of #3858, which really should have been merged by now

jackfranzen:
True and real. Please Accept #3858.

PR #4214, opened 2022-12-20
Ignore .vscode folder

skywalkerSam:
- The folder (.vscode) sometimes gets committed for no reason. Some people might need it but, most people don't.

- I use Visual Studio Code & this file for almost every python project to ignore certain files and folders.

_TODO_
- If someone needs to ignore the (.vscode) folder, they can just uncomment it and they're good to go.

There are no official documentations. If there are, I just couldn't find them. It's more of a convention, some people want to ignore this folder and some don't  :)

_TODO_
- If someone finds some documentation on this, help is appreciated.

If this is a new template:

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -158,3 +158,7 @@ cython_debug/
 #  and can be added to the global gitignore or merged into this file.  For a more nuclear
 #  option (not recommended) you can uncomment the following to ignore the entire idea folder.
 #.idea/
+
+# Visual Studio Code (VS Code)
+# If you do not want to share settings, tasks and debug configurations, you can uncomment the following to ignore the entire (.vscode) folder.
+#.vscode/


stijnb1234:
https://github.com/github/gitignore/blob/main/Global/VisualStudioCode.gitignore exists with a reason?

skywalkerSam:
Hello,

I said that "some people need it but, some don't". At least not in Python
projects, most of the time we try to ignore the (.vscode) folder. If
someone wants to keep it, not a problem, it stays that way. If someone
wants to ignore this folder, they can just uncomment the line and be good
to go.

The reason for this pull request is that the repository gitignore is pretty
big, with a lot of files. If someone starts a Python project and wants to
ignore unnecessary files through git, all they need to do is copy the
"Python.gitignore" file and be good to go.

I think that all the available options should be on the table in one place.
Otherwise, We will need to navigate through multiple files to get that one
task done.


Thanks,
Sam

muffinhydra:
> https://github.com/github/gitignore/blob/main/Global/VisualStudioCode.gitignore exists with a reason?

If I want to include the VS Code settings in my commit I wouldn't need a .gitignore in order to do so. I would just not include it in the .gitignore. If I create/append a .gitignore for VS code I do expect that the VS code specific things will be ignored. The file as it is makes simply no sense. So no that file has no reason to exist as it is.

And just to be clear the .gitignore file for VScode does NOT ignore VS code files. Exactly the opposites it specifies VS Code files to not be ignored.

stijnb1234:
> > https://github.com/github/gitignore/blob/main/Global/VisualStudioCode.gitignore exists with a reason?
> 
> If I want to include the VS Code settings in my commit I wouldn't need a .gitignore in order to do so. I would just not include it in the .gitignore. If I create/append a .gitignore for VS code I do expect that the VS code specific things will be ignored. The file as it is makes simply no sense. So no that file has no reason to exist as it is.
> 
> And just to be clear the .gitignore file for VScode does NOT ignore VS code files. Exactly the opposites it specifies VS Code files to not be ignored.

Incorrect. It excludes all vscode files expect a few important ones.

muffinhydra:
But the exclusions are defacto all the relevant files in a .vscode directory. As in it says ignore the content of the .vscode directory but then goes and excludes all contents from ignoring manually.

What happens is, if you want your local settings not to be tracked by git that .gitignore does not do it. It specifically allows git to track your local editor/ide settings. 

Because that's what .vscode is, local settings for you IDE/editor/plugins.

skywalkerSam:
I get it, It's already too complicated. Some need it, and some don't. And here we're trying to build a standard. At least you have a starting point for your .gitignore file, you can customize it if you like because different people & projects have different needs. 

P.S. Guys, there are a lot of programming languages, frameworks, code editors, and IDEs. If we try to do this with every possible combination of projects, tools, and development environments. It's just not feasible...

muffinhydra:
I don't see it being complicated at all. If you want to share your IDE setting, your DB secrets and other sensitive data ( which all get stored in those files by the diverse extension of VS code ) just don't use  a gitignore for those files. If you don't want to push secrets to a possibly public repository use the gitignore. That functionality is currently not there. There is nothing complicated about it. 

Like why does this need to be argued at all? The counter  argument is synonymous to saying that the gitignore for .env files  should exclude .env from ignoring and as a standard such files should be pushed to repositories.

skywalkerSam:
> I don't see it being complicated at all. If you want to share your IDE setting, your DB secrets and other sensitive data ( which all get stored in those files by the diverse extension of VS code ) just don't use a gitignore for those files. If you don't want to push secrets to a possibly public repository use the gitignore. That functionality is currently not there. There is nothing complicated about it.
> 
> Like why does this need to be argued at all? The counter argument is synonymous to saying that the gitignore for .env files should exclude .env from ignoring and as a standard such files should be pushed to repositories.


Not everyone needs this, look at how many code editors, IDEs, and tools are for python. If you wanna protect your confidential information, I get it. Use the [VisualStudioCode.gitignore](https://github.com/github/gitignore/blob/main/Global/VisualStudioCode.gitignore) file with your [Python.gitignore](https://github.com/github/gitignore/blob/main/Python.gitignore) file.

 Hope you find this helpful :)

PR #4219, opened 2022-12-27
filter Windows thumbs and Mac Finder parameters

DeveloppeurPascal:
Hi
I have added the default Delphi compilation folders (but disabled for compatibility with previous .gitignore filters) even if their content has nothing to do in a code repository because it's regenerated each time a project is compiled.
I also added the Finder (macOS) and Windows folder settings files which has nothing to do with source code and should not be store anywhere.
Best regards
Patrick Prémartin

### DIFF
--- a/Delphi.gitignore
+++ b/Delphi.gitignore
@@ -26,6 +26,18 @@
 #*.obj
 #
 
+# Default Delphi compiler directories
+# Content of this directories are generated with each Compile/Construct of a project.
+# Most of the time, files here have not there place in a code repository.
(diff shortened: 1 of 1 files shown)


wirecat:
@DeveloppeurPascal I'm closing this one as I have merged https://github.com/github/gitignore/pull/4220 which was a similar change, but didn't contain the OS specific configuration.

For the time being, we're trying to keep OS specific configuration out of language/tool/IDE ignore templates. That may change in the future though.

PR #4220, opened 2022-12-29
Added Delphi default compile/construct directories

DeveloppeurPascal:
Files in the default compile/construct directories are generated each time Delphi need them. Most of them are binaries. There is no reason to put them in the code repository.

Filter added as comment, but it would be better to uncomment them.


wirecat (approved):
Thank you for contributing!

PR #4221, opened 2022-12-29
Add V language .gitignore template. 

nikeedev:

GitHub has already support for V's syntax, and now we need a .gitignore file template for new V repos which get created.


nikeedev:
> This PR is stale because there have been no updates in 90 days. It will close after 180 days of inactivity. Leave a comment if you want to keep it open 😄

of course I want

PR #4234, opened 2023-01-25
Ignore all `.env` files in `Node.gitignore` and unignore `.env.example`

cyber-amr:
## Reasons for making this change:

1. Removed ignores that uses the conventions from https://github.com/bkeepers/dotenv#what-other-env-files-can-i-use
Even if it used by some modern JS frameworks it shouldn't be used here since this is a general node.js template and these conventions are specific to some frameworks.

2. Added `.env.*` as in #3651 it's safer to ignore all `.env` files since people names it differently (e.g. `.env.production`, `.env.prod`)
see this comment https://github.com/github/gitignore/pull/3316#issuecomment-991413112

3. Unignored `.env.example` as what @kachick tried to do in b236e1717b90f1755af76900f5387bfa20636de3 

## Links to documentation supporting these rule changes:

`dotenv` node.js package: https://github.com/motdotla/dotenv#should-i-commit-my-env-file

### DIFF
--- a/Node.gitignore
+++ b/Node.gitignore
@@ -74,10 +74,6 @@ web_modules/
 
 # dotenv environment variable files
 .env
-.env.development.local
-.env.test.local
-.env.production.local
-.env.local
 
 # parcel-bundler cache (https://parceljs.org/)
 .cache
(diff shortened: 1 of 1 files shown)


amitbeck (approved):

kachick (approved):
:+1: for fail-safe config by default

wirecat (approved):
Thank you for contributing!

PR #4237, opened 2023-02-01
Fix Metals gitignore

ZachTylerPruitt:
* Bring the gitignore inline with current recommendations 
* Comments had leading spaces so they weren't treated as comments
* Change Reference to recommended gitignore configuration documentation

* [metals-vscode/README.md](https://github.com/scalameta/metals-vscode/blob/09380cda3f7529f05c9ec3766075ba450efeda65/README.md?plain=1#L215)


wirecat (approved):

PR #4239, opened 2023-02-01
Add delve debug binary

jollyrogue:

Delve is the major Go debugger, and it produces a `__debug_bin` binary by default, which isn't in the Go gitignore template.

I use delve to debug Go binaries, and I'd like to have delve's `__debug_bin` ignored by default.

`--output string   Output path for the binary. (default "./__debug_bin")`

https://github.com/go-delve/delve/blob/4c1f8b4b2536beb8b06c6a6ef0b0383a7928e188/Documentation/usage/dlv_debug.md

### DIFF
--- a/Go.gitignore
+++ b/Go.gitignore
@@ -11,6 +11,9 @@
 # Test binary, built with `go test -c`
 *.test
 
+# Default Delve debug binary
+__debug_bin
+
 # Output of the go coverage tool, specifically when used with LiteIDE
 *.out
 


Go.gitignore
```diff
@@ -11,6 +11,9 @@
 # Test binary, built with `go test -c`
 *.test
 
+# Default Delve debug binary
+__debug_bin
```
BananZG, line 15:
```suggestion
__debug_bin*
```

There's a new change to delve to include timestamp in output
source: https://github.com/go-delve/delve/commit/801a9109c7fd7be8b060b466cc041a3695455c77

PR #3732, opened 2021-05-12
add pnpm debug file to Node.gitignore

sakurayang:
**Reasons for making this change:**

add pnpm debug log file to ignore

**Links to documentation supporting these rule changes:**

https://pnpm.io/git

PR #4250, opened 2023-03-09
Remove pnpm debug log

karlhorky:

- it is contained within `node_modules/`, which is already ignored
- the previous versions, which were not in `node_modules/`, did not have a period at the beginning of the filename

Changelog with proof here:

https://github.com/pnpm/pnpm/blob/ba4b2db1f201640dad5d7ee9fe68a2d0defd6047/pnpm/CHANGELOG.md?plain=1#L3330

**History**

First introduced in https://github.com/github/gitignore/pull/3732 by @sakurayang (merged by @martinwoodward)

### DIFF
--- a/Node.gitignore
+++ b/Node.gitignore
@@ -5,7 +5,6 @@ npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
 lerna-debug.log*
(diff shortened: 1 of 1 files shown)


karlhorky:
Not stale

PR #4275, opened 2023-04-05
Add .vscode/ to Unity.gitignore

rubenbuelvas:
Add .vscode/

VSCode can be used as an alternative to VS, so I included the cache directory of VSCode in the .gitignore file.

### DIFF
--- a/Unity.gitignore
+++ b/Unity.gitignore
@@ -26,6 +26,9 @@
 # Visual Studio cache directory
 .vs/
 
+# Visual Studio Code cache directory
+.vscode/
+
 # Gradle cache directory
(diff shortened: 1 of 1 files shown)


TetieWasTaken (commented):
The [`.vscode/` directory](https://code.visualstudio.com/docs/getstarted/settings) is meant for workspace settings, and is not a cache directory. (Examples include code snippets, settings and extensions)

LiterallyFabian:
Yeah I don't see why this should be in the Unity gitignore. It's not cache files, and pretty much everyone _does_ version control their editor/workspace settings.

PR #4279, opened 2023-04-09
Add `.pdm-python` to python gitignore

SRv6d:

Starting with [v2.5.0b0](https://github.com/pdm-project/pdm/releases/tag/2.5.0b0) `pdm` requires a `.pdm-python` file to track the path of the local python installation.

[v2.5.0b0](https://github.com/pdm-project/pdm/releases/tag/2.5.0b0)

If this is a new template:

 - **Link to application or project’s homepage**: https://pdm.fming.dev

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -108,6 +108,7 @@ ipython_config.py
 #   in version control.
 #   https://pdm.fming.dev/#use-with-ide
 .pdm.toml
+.pdm-python
 
 # PEP 582; used by e.g. github.com/David-OConnor/pyflow and github.com/pdm-project/pdm
 __pypackages__/
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -109,6 +109,7 @@ ipython_config.py
 #   https://pdm.fming.dev/latest/usage/project/#working-with-version-control
 .pdm.toml
 .pdm-python
+.pdm-build/
 
 # PEP 582; used by e.g. github.com/David-OConnor/pyflow and github.com/pdm-project/pdm
 __pypackages__/
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -106,7 +106,7 @@ ipython_config.py
 #pdm.lock
 #   pdm stores project-wide configurations in .pdm.toml, but it is recommended to not include it
 #   in version control.
-#   https://pdm.fming.dev/#use-with-ide
+#   https://pdm.fming.dev/latest/usage/project/#working-with-version-control
 .pdm.toml
 .pdm-python
 


Python.gitignore
```diff
@@ -108,6 +108,7 @@ ipython_config.py
 #   in version control.
 #   https://pdm.fming.dev/#use-with-ide
 .pdm.toml
```
liggest, line 110: They also renamed their config file from `.pdm.toml` to `pdm.toml`
Maybe both two names should be at here for backward compatibility?

liggest:
```suggestion
.pdm.toml
pdm.toml
```

SRv6d: Quoting the release notes:

> Move the project python path to its own file, and rename the project config file as pdm.toml which can be committed to the VCS. https://github.com/pdm-project/pdm/issues/1742

As far as I understand this, the sole purpose of `.pdm-python` is to remove environment specific configuration from the old `.pdm.toml`, creating a newly named `pdm.toml` which can be commited to VCS, so it should not be ignored by default anymore.

YDX-2147483647 (approved):

thiagola92:
Can this be merged? Everything looks good to me.


Reference: https://pdm.fming.dev/dev/usage/project/#working-with-version-control  
> You **must** commit the `pyproject.toml` file. You **should** commit the `pdm.lock` and `pdm.toml` file. **Do not** commit the `.pdm-python` file.

Python.gitignore
```diff
@@ -108,6 +108,7 @@ ipython_config.py
 #   in version control.
 #   https://pdm.fming.dev/#use-with-ide
```
leonaascimento, line 109: Just updating the site reference.

```suggestion
#   https://python-poetry.org/docs/basic-usage/#commit-your-poetrylock-file-to-version-control
```

Python.gitignore
```diff
@@ -106,8 +106,9 @@ ipython_config.py
 #pdm.lock
 #   pdm stores project-wide configurations in .pdm.toml, but it is recommended to not include it
 #   in version control.
-#   https://pdm.fming.dev/#use-with-ide
+#   https://pdm.fming.dev/latest/usage/project/#working-with-version-control
 .pdm.toml
+.pdm-python
```
huxuan, line 111: Suggest to include `.pdm-build/` as referenced here: https://github.com/pdm-project/pdm/blob/main/src/pdm/cli/templates/default/.gitignore#L112

SRv6d: Added although it is not mentioned in the docs.

huxuan: It should be OK as the template is used when generating a new project when using `pdm init`.

DavidArmendariz:
Merge this already please!

wirecat (approved):

PR #4282, opened 2023-04-14
added gitignore for dotnet core web app

zakk616:

_TODO_

_TODO_

If this is a new template:


LiterallyFabian:
I don't think adding other IDEs like Jetbrains Rider is appropriate here, and the Rider ignored files aren't even correct; the entire `.idea` should _not_ be ignored according to Jetbrains themselves. https://intellij-support.jetbrains.com/hc/en-us/articles/206544839

zakk616:
@salmonslay yes correct, modified.

PR #4287, opened 2023-04-29
Update TeX.gitignore to capture .pygtex and .pygstyle

daniel-jakob:

When ``minted`` is ran with caching enabled, two file types are created used in the caching process, ``.pgytex`` and ``.pygstyle``. These were not being captured before pull request.

https://tug.ctan.org/macros/latex/contrib/minted/minted.pdf 
``.pygtex`` files are discussed in the section 8.3 of the implementation portion of the  documentation and ``.pygstyle`` is mentioned in section 8.5.

### DIFF
--- a/TeX.gitignore
+++ b/TeX.gitignore
@@ -169,7 +169,7 @@ acs-*.bib
 
 # minted
 _minted*
-*.pyg
+*.pyg*
 
 # morewrites
 *.mw


whybe7:
I think this is what they called the magic of open source :octocat:, it's putting effort into your contribution 💯wich will make it more spicy

daniel-jakob:
Just wondering if there is any work being done on this repo? Don't see too many pull requests being merged.

PR #4274, opened 2023-04-03
Add Ruff to Python gitignore

janheindejong:
Since [ruff ](https://github.com/charliermarsh/ruff) is becoming more and more popular in the Python ecosystem, I figured it would be nice to include it here.

PR #4199, opened 2022-12-02
Add ruff_cache

Benoitdw:
[Ruff](https://github.com/charliermarsh/ruff) is a Python linter which generate ruff_cache directory.

PR #4289, opened 2023-05-01
Update Python.gitignore for .ruff_cache/

csauge:
Add Ruff cache in gitignore. 
Ruff is becoming a very popular tool often run in pre-commit and aim to replace Flake8 (plus dozens of plugins), isort, pydocstyle, yesqa, eradicate, pyupgrade, and autoflake, all while executing tens or hundreds of times faster than any individual tool.

Ruff is run in pre-commit thus .ruff_cache/ could be included accidently in git.
Actually this is similar to .mypy_cache/ which is also run in pre-commit.

https://github.com/charliermarsh/ruff in section configuration we can see the exclude of .ruff_cache and .mypy_cache


Diegiwg (approved):

Grvzard (approved):

covracer:
See also #4199

JasperJuergensen:
I think this is not required, as ruff will add a `.gitignore` file to the `.ruff_cache` folder with `*` as content effectively ignoring the folder.

See https://github.com/astral-sh/ruff/issues/204 and https://github.com/astral-sh/ruff/pull/208

stdedos:
It is nicer to have it there.

e.g. PyCharm's excluded folders (and https://plugins.jetbrains.com/plugin/17288-foldable-projectview) work based on these ignore

paduszyk:
@JasperJuergensen `pytest` does the same - it also adds `.gitignore` with `*` to its cache dir. But `.pytest_cache` it's included in the template... The same regards `nox`, for instance.

stankudrow:
I agree that it is nice to have `.ruff_cache` directory similarly to `.mypy_cache` and `.pytest_cache` directories. The [ruff](https://github.com/astral-sh/ruff) project is steadily developing and becoming widespread, so it is good to have already the section for it.

Also was mentioned in https://github.com/github/gitignore/pull/4199 and https://github.com/github/gitignore/pull/4274 PRs.

For instance, without the `.ruff_cache` line in **.gitignore** file this directory is not grayed:
ruff_cache_not_ignored

When added:
ruff_cache_ignored

abelcheung (approved):

abelcheung:
Other tools like `mypy` and `pytest` indeed do add their own .gitignore as a fallback measure, in case top level `.gitignore` doesn't handle them. But in terms of performance it is always better to add those dirs to top level `.gitigore`, rather than descending into cache folders and read `.gitignore` there. And there is visual cue differences [noted above](#issuecomment-1891039030).

stdedos (approved):

stdedos:
Can we get some maintainer focus? e.g. @dooleydevin?

dooleydevin (approved):
Sorry for the late feedback! LGTM and will :shipit: with the next update of our templates 😄 

I'll handle the merge conflicts here too just to move things along since this PR is a bit old.

PR #4290, opened 2023-05-02
Ignore the new .pdm-python file that is now used by pdm to determine the path to the python interpreter used by the project.

taranlu-houzz:
PDM now writes a new `.pdm-python` file that is specific to the user's machine. It should be ignored.

https://pdm.fming.dev/latest/usage/project/#working-with-version-control


thiagola92:
Same as #4279

wirecat:
Closing since we've already pulled in https://github.com/github/gitignore/pull/4279 which includes these changes.
Thank you for contributing!

PR #4293, opened 2023-05-12
Add _CompileInfo_Upload/ to twincat3 gitignore

WolfThomas:
New additional compileinfo will be used with 3..1.4026.x

### DIFF
--- a/TwinCAT3.gitignore
+++ b/TwinCAT3.gitignore
@@ -21,5 +21,6 @@ LineIDs.dbg
 LineIDs.dbg.bak
 _Boot/
 _CompileInfo/
+_CompileInfo_Upload/
 _Libraries/
 _ModuleInstall/


WolfThomas:
@sagatowski Can you verify this change after 4026 release.

sagatowski:
Is there any information from Beckhoff that this should indeed be ignored? What does this folder do? What is the difference compared to the other CompileInfo?

WolfThomas:
As soon as the documentation is updated I will post the link. The folder "_CompileInfo" contains the compileinfo, this is not changed. The new folder "_CompileInfo_Upload" does also contains compileinfo  files but only to improve onlinechange behavior.

sagatowski:
@WolfThomas All clear. I'll look into this as soon as 4026 is released and do a review changes.

PR #4298, opened 2023-05-23
Node.gitignore include support for SvelteKit, Vite, Firebase, and .env.example

taronaeo:

1. SvelteKit - Upon running `npm run build` / `yarn build`, Vite generates a .svelte-kit folder that should not be part of the Git repository.
2. .env.example - Examples of environment files should be part of the Git repository for setup and reference.
3. Vite - Vite may generate temporary log files in the project that may be accidentally committed.
4. Firebase - `firebase-tools` generates cache directory `.firebase` to speed up build processes. Not strictly needed in the Git repository.

1. Firebase - https://stackoverflow.com/a/52131191


spirit-x64:
Second case (`.env.example`) should be fixed by #4234

Node.gitignore
```diff
@@ -104,6 +105,9 @@ dist
 .temp
 .cache
 
+# Sveltekit cache directory
+.svelte-kit
```
dooleydevin, line 109: A small nitpick: adding a trailing slash here will make sure this only matches directories with this name.

```suggestion
.svelte-kit/
```

Node.gitignore
```diff
@@ -122,6 +126,9 @@ dist
 # DynamoDB Local files
 .dynamodb/
 
+# Firebase cache directory
+.firebase
```
dooleydevin, line 130: Similar request to the one above:


```suggestion
.firebase/
```

dooleydevin (changes requested):
@taronaeo it looks like there are merge conflicts here based on upstream changes to the `.env` patterns and yarn v3 patterns. Once those are resolved, this looks good to merge to me 😄 

Also left a couple small suggestions, though I don't consider them blocking for this PR to be merged.

taronaeo:
@dooleydevin I've resolved all conflicts and merged your suggested changes, PTAL again :)

Node.gitignore
```diff
@@ -67,7 +67,10 @@ web_modules/
 
 # dotenv environment variable files
 .env
-.env.*
+.env.development.local
+.env.test.local
+.env.production.local
+.env.local
```
dooleydevin, line 73: Do you mind reverting this so that we take the upstream change here?

I've seen patterns like `.env.development` and `.env.production` (without the `.local` suffix). Looking around online that seems pretty common. I'd prefer we exclude all `.env.*` patterns except for `.env.example` for that reason 👍

dooleydevin (changes requested):
@taronaeo sorry for the back-and-forth here. One more request regarding the dotenv file matching before this is ready to merge 👍

Node.gitignore
taronaeo: Reverted! Feel free to review again! :)

dooleydevin (approved):
Thanks for the contribution and for addressing feedback quickly! And apologies for the delay on initial review, us new maintainers are still trying to get on top of the backlog.

Sum1cares (approved):

PR #4301, opened 2023-05-28
Rename Octave.gitignore to Octave.gitignore.

1ohn-Brown1:

_TODO_

_TODO_

If this is a new template:


ElBe-Plaq:
Why is there an extra dot at the end of the name? There shouldn't be.

PR #4303, opened 2023-05-29
Update Node.gitignore

lwabish:

Is it a little arbitrary to ignore logs directory?
In golang ,some common packages name themselves as logs, for example, k8s.io/component-base/logs

### DIFF
--- a/Node.gitignore
+++ b/Node.gitignore
@@ -1,5 +1,4 @@
 # Logs
-logs
 *.log
 npm-debug.log*
 yarn-debug.log*
--- a/Node.gitignore
+++ b/Node.gitignore
@@ -1,5 +1,6 @@
 # Logs
 *.log
+/logs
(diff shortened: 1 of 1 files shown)


spirit-x64:
Ignoring the `logs` dir in the root dir instead of ignoring it globally won't be better?

as:
```gitignore
/logs
```

lwabish:
> Ignoring the `logs` dir in the root dir instead of ignoring it globally won't be better?
> 
> 
> 
> as:
> 
> ```gitignore
> 
> /logs
> 
> ```

Seems good

lwabish:
> Ignoring the `logs` dir in the root dir instead of ignoring it globally won't be better?
> 
> as:
> 
> ```gitignore
> /logs
> ```

I have updated this PR according to this suggestion, which finally change `logs` to `/logs`

PR #4307, opened 2023-06-08
Add Vim swap file

GangWang01:
As a common template [.gitignore](https://github.com/dotnet/sdk/blob/main/template_feed/Microsoft.DotNet.Common.ItemTemplates/content/Gitignore/.gitignore) maintained by https://github.com/dotnet/templating, the customer requested to add Vim swap files to it. To keep both sides aligned, request to add Vim swap files to this template.

https://github.com/dotnet/templating/issues/5048

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -396,3 +396,6 @@ FodyWeavers.xsd
 
 # JetBrains Rider
 *.sln.iml
+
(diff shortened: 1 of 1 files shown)


kkirsche:
Is there a reason that this cannot leverage https://github.com/github/gitignore/blob/main/Global/Vim.gitignore?

There are a few tools like gitignorer and whatnot to help with concatenating multiple gitignore files together

PR #3154, opened 2019-08-20
Reference why to exclude binary jar, fixes #111

fulldecent:
**Reasons for making this change:**

Improve documentation

**Links to documentation supporting these rule changes:**

https://github.com/takari/maven-wrapper#usage-without-binary-jar

PR #4308, opened 2023-06-09
Update reference to Maven Wrapper exclusion

chabala:

Improve documentation: Since this line was introduced in #3154, Takari has given this code to Apache Maven, the URL referencing the Takari repo is no longer valid, and is replaced with the official Apache Maven Wrapper documentation URL.

https://maven.apache.org/wrapper/#usage-without-binary-jar


chabala:
Yes, I still want it. @github/gitignore-maintainers

PR #4313, opened 2023-06-17
[ADD] Added *.pyc and *.DS_Store

altela:
I would like to simply add *.pyc and *.DS_Store into the python .gitignore

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -132,7 +132,6 @@ venv.bak/
 # Spyder project settings
 .spyderproject
 .spyproject
-*.DS_Store
 
 # Rope project settings
 .ropeproject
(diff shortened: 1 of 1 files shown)


movatica:
.DS_Store files are not Python-related.

altela:
Ok i will update it and remove

khr1stopher (approved):

zonuexe:
Why add `*.pyc`? It looks like this pattern already has you covered.

https://github.com/github/gitignore/blob/a60d59238afb22523038b549b9b27f15734f7563/Python.gitignore#L3

altela:
yes but it doesn't seem to work, perhaps this happen on my local so i guess i will close it

PR #4314, opened 2023-06-17
Updated ignore rule for "umbraco" folder with trailing *

christophnz:
Updated ignore rule for "umbraco" folder with trailing * to ensure include folders below are actually included

The current version ignores the folder `umbraco`, but there are include rules below it that don't work.

I use the Umbraco CMS and every time we set up a new project we have to remember to make this change to the gitignore file.

Without this change, the include `!**/umbraco/models` isn't working and the models folder is missing from the repo causing compilation issues.

### DIFF
--- a/community/DotNet/Umbraco.gitignore
+++ b/community/DotNet/Umbraco.gitignore
@@ -43,7 +43,7 @@
 *.sqlite.db*
 
 #ignore umbraco data/views/settings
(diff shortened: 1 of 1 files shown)


christophnz:
This is still relevant.

wirecat (approved):
Thank you for contributing!

PR #4315, opened 2023-06-21
JetBrains.gitignore: add ignore for the HTTP Client's private environment

kaznovac:

prevent accidental push of private variables

documentation regarding env config files: https://www.jetbrains.com/help/idea/exploring-http-syntax.html#example-working-with-environment-files

also renamed in the comment `Rest Client` to `HTTP Client` as it is current JetBrains name for this feature


kaznovac:
I still believe this would be useful to add, can someone review?

wirecat (approved):
Thank you for contributing!

PR #4321, opened 2023-06-30
Update swift, Xcode and Objective-C

KevinRaney:

Update headers for clarity and remove `.DS_Store` files since they do not make sense in the context of a source control repo.

[.DS_Store Files](https://buildthis.com/ds_store-files-and-why-you-should-know-about-them/)


Timmysmart (approved):

ealeksandrov:
It's already a part of [macOS.gitignore](https://github.com/github/gitignore/blob/main/Global/macOS.gitignore), no need to add OS-specific stuff in language templates.

wirecat:

🙇

PR #4328, opened 2023-07-20
Update VisualStudio.gitignore

matthewblott:
Added the ".idea" entry. This template is used with the .NET CLI and anyone using the JetBrains platform currently has to add this entry manually.


FurkanKambay:
it should have a `/`

matthewblott:
It isn't necessary, it works without.

FurkanKambay:
I meant that the rule is too broad since it also ignores any files like `test.idea`, so only the directory should be ignored.

PR #4332, opened 2023-07-28
Use yarn v3 instead of v2

Quantalabs:
Updates the `gitignore` for Node to use yarn v3 instead of yarn v2. Ignores the files recommended by yarn on the docs.

Yarn v3 has a different ruleset than yarn v2, ignoring everything except the patches, plugins, releases, sdks, and versions in the `.yarn/` folder. My changes update the `gitignore` to support these changes.

- ["Which files should be gitignored?"](https://yarnpkg.com/getting-started/qa#which-files-should-be-gitignored)

### DIFF
--- a/Node.gitignore
+++ b/Node.gitignore
@@ -122,9 +122,11 @@ dist
 # Stores VSCode versions used for testing VSCode extensions
 .vscode-test
 
-# yarn v2
-.yarn/cache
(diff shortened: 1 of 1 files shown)


fulldecent:
bump

... to yarn v4

wirecat (approved):
Thank you for contributing!

PR #4333, opened 2023-07-28
Update VS Code (remove a path and add a missing one)

FurkanKambay:

- Removes a 3rd party extension directory (`.history/`) because of reasons described below.
- Adds the `*.code-workspace` path that exists in https://github.com/github/gitignore/blob/main/VisualStudio.gitignore#L385 and https://github.com/github/gitignore/blob/main/Global/AL.gitignore#L6. (They should also be updated with the missing `!.vscode/*.code-snippets` that exists here. I can open PRs when this is merged)

The `.history/` directory is only used by [this third party extension](https://marketplace.visualstudio.com/items?itemName=xyz.local-history) which has already been stated in https://github.com/github/gitignore/pull/3008#issue-423274609.
Moreso, VS Code has since officially released their own 'local history' feature which doesn't use such a directory: [release notes](https://code.visualstudio.com/updates/v1_66#_local-history)

### DIFF
--- a/Global/VisualStudioCode.gitignore
+++ b/Global/VisualStudioCode.gitignore
@@ -4,9 +4,7 @@
 !.vscode/launch.json
 !.vscode/extensions.json
 !.vscode/*.code-snippets
-
-# Local History for Visual Studio Code
-.history/
+!*.code-workspace
 
 # Built Visual Studio Code Extensions
 *.vsix


Imen-Frigui (approved):

BinToss:
Visual Studio Code's `*.code-workspace` is intended to contain workspace-specific configurations for VSCode and its extensions. It *can* contain user-specific values, but is intended for sharing settings, tasks, and more between contributors—just like the JSON files already whitelisted. It's comparable to Visual Studio's `*.sln` files. https://code.visualstudio.com/docs/editor/workspaces

Why does `VisualStudio.gitignore` contain Visual Studio _Code_ rules, anyway?

FurkanKambay:
@BinToss you're right, I guess I just blindly copied it.

For your second question, the answer is in #3722:

> Our customer research supports that we have lots of developers using multiple tools, primarily VS and Code together. Ensuring that we have one VS family that encapsulate them is convenient for the developers and saves them a step and has been requested.

> We think of VS as a family of products and something that represents VS, should be inclusive of this.

BinToss:
Does this need to be rebased?

wirecat (approved):
Thank you for contributing!

Sum1cares (approved):

PR #4340, opened 2023-08-14
Update VisualStudio.gitignore for ENV files

timheuer:
Adding .env exclusion

.env file support is starting to happen more and more in .NET ecosystem and VS users.


FurkanKambay:
it should have a `/`

PR #4346, opened 2023-08-30
ignore split-dwarf debugging-related auxilary .dwo files for C and C+…

krishnakumarg1984:
Adds `.dwo` file to the ignore templates of C and C++. `.dwo` is the file extension for auxiliary debug information written to disk when using the `-gsplit-dwarf` compile flag.

- https://undo.io/resources/gdb-watchpoint/reduce-binary-size-compile-time-split-dwarf/
- https://www.productive-cpp.com/improving-cpp-builds-with-split-dwarf/

### DIFF
--- a/C++.gitignore
+++ b/C++.gitignore
@@ -30,3 +30,6 @@
 *.exe
 *.out
 *.app
+
+# debug information files
+*.dwo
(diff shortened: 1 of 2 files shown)


wirecat (approved):
Thank you for contributing!

PR #4117, opened 2022-07-22
Update Global/MATLAB.gitignore to include broader set of transitory files

rpurser47:
**Reasons for making this change:**
Our customers have asked for a more complete `.gitignore` for the [MATLAB family of products](https://www.mathworks.com/products/matlab.html).  I've worked with developers at MathWorks to identify a number of transitory file types that should be in the `.gitignore`.

**Links to documentation supporting these rule changes:**
* Expanded the list of autosave file extensions and caching files based on developer feedback
* Added [content-obscured file extension](https://www.mathworks.com/help/matlab/matlab_prog/building-a-content-obscured-format-with-p-code.html)
* Added [deployable archive file extension](https://www.mathworks.com/help/compiler/mcc.html)
* Added [Cloud based storage](https://www.mathworks.com/products/matlab-drive.html) dotfile

PR #4348, opened 2023-08-31
Update MATLAB.gitignore

signorettae:
Added Matlab Drive to .gitignore


Adamou02 (commented):
Duplicate of #4117 with fewer features

PR #4351, opened 2023-09-08
Update TwinCAT3.gitignore

Tc-KGu:
with the existing .gitignore in this repo, which can be select directly in the VS 2022 if you create a new repo, our customers have a lot of problems. thats why we suggest this update

Link to documentation from Beckhoff TwinCAT3
https://infosys.beckhoff.com/english.php?content=../content/1033/tc3_sourcecontrol/14604066827.html&id=

### DIFF
--- a/TwinCAT3.gitignore
+++ b/TwinCAT3.gitignore
@@ -1,25 +1,46 @@
-# gitignore template for TwinCAT3
+### TwinCAT3 ###
 # website: https://www.beckhoff.com/twincat3/
-#
-# Recommended: VisualStudio.gitignore
 
-# TwinCAT files
+# TwinCAT PLC
+*.plcproj.bak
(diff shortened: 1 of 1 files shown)


TwinCAT3.gitignore
```diff
@@ -29,8 +29,9 @@ LineIDs.dbg.bak
 *.xti.bk?
 *.xti.orig
 *.xtv
-*.xtv.bak
```
dooleydevin, line 32: Can you explain the removal of this line?

The example you linked includes it, and the patterns you are including do not cover this line as far as I can tell 🤔

dooleydevin (changes requested):

wirecat (approved):
Thank you for contributing!

dooleydevin (approved):

PR #4358, opened 2023-09-28
Update Unity.gitignore

theDrake:
Add WebGL build-related directory and files plus `.vscode/` and a new crash-related file (also gather crash-related files together for clarity and simplicity).

These additions are based on my experience working with Unity and VS Code in a Linux environment over the past year.

### DIFF
--- a/Unity.gitignore
+++ b/Unity.gitignore
@@ -7,6 +7,7 @@
 /[Oo]bj/
 /[Bb]uild/
 /[Bb]uilds/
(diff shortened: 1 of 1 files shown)


wirecat:

Would you please remove the .vscode entry?

Additionally, could you please provide documentation justifying the new entries?

theDrake:
It's been two years and, well, I don't use Unity as much lately. 🙂 So, I might revisit this at some point, but for now I'll just close it.

PR #4363, opened 2023-10-25
Ignore mono_crash files

rhys-vdw:
These can be generated by Unity on crash

- Unity can generate mono_crash.* files in some crash conditions.
- This happened to us recently and we ended up with these files in version control.

Honestly I can't find it documented in Unity docs. I asked in official Unity discord and was told it was created by Unity.
A mention of it here: https://forum.unity.com/threads/debugging-editor-crashes.1457002/

I assume they're created by Mono itself, which is embedded in Unity.

### DIFF
--- a/Unity.gitignore
+++ b/Unity.gitignore
@@ -55,6 +55,9 @@ ExportedObj/
 # Unity3D generated file on crash reports
 sysinfo.txt
 
+# Mono auto generated files
+mono_crash.*
+
 # Builds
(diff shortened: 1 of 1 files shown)


rhys-vdw:
This is also covered by #4358

rhys-vdw:
afaik this is still an appropriate change

wirecat (approved):
Thank you for contributing!

PR #4367, opened 2023-11-02
ignore vscode

yuyang9119:

Visual Studio Code now supports Unity, so need ignore the folder created by it

https://marketplace.visualstudio.com/items?itemName=VisualStudioToolsForUnity.vstuc


loveisreal01214 (changes requested):
Sim card not working

karol-gro:
You don't want to do it. Dir `.vscode` may contain several files you definitely want to check in. E.g. settings, workspaces, tasks

If you want to ignore some particular cache file from this dir, you should write more specific rule

wirecat:

Closing for now.

PR #4369, opened 2023-11-04
Update Python.gitignore for PEP 441

hspaans:
PEP 441 introduces Python ZIP Applications to bundle a Python Application into a ZIP file with file extension `pyz`.

Python ZIP Applications, described in [PIP 441](https://peps.python.org/pep-0441/), are an artifact from the build process and shouldn't be added to the repository. These files have their own file extension `pyz` making it easy to exclude them.

- [PEP 441](https://peps.python.org/pep-0441/) describing Python ZIP Applications

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -1,6 +1,6 @@
 # Byte-compiled / optimized / DLL files
 __pycache__/
-*.py[cod]
+*.py[codz]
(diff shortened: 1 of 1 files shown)


hspaans:
Hopefully this can still be merged as PEP-441 is still active.

PR #4376, opened 2023-11-22
feat: Ignore `.pdm-python` file.

laipz8200:

[PDM](https://github.com/pdm-project/pdm) saves the Python interpreter path in the `.pdm-python` file, and it should not contained in the git repository because it is based on a local file path. I think `.pdm-python` should be ignored by default.

Reference [this document](https://github.com/pdm-project/pdm/blob/a339412867d94ad3c2fa99325dc90ce6c6d79c84/docs/docs/usage/project.md#choose-a-python-interpreter)

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -108,6 +108,7 @@ ipython_config.py
 #   in version control.
 #   https://pdm.fming.dev/#use-with-ide
 .pdm.toml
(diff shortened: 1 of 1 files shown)


Roang-zero1:
Duplicate of #4279

PR #4377, opened 2023-11-25
UC: Added ignore template for BFH workflows

BFH-hga3:
This set of ignore patterns is used at BFH in combination with other gitignore templates.

I work at BFH. We have several projects either in education or research activities. We use git a lot and therefore gitignore pattern templates from this repositories. To build a specific gitignore file we use APIs like gitignore.io/api. In terms of maintainability and ease of use, a bfh template covering some specific properties would be very handy.

Unfortunately, not possible.

If this is a new template:

 - **Link to application or project’s homepage**:  https://www.bfh.ch/ti/en/

### DIFF
--- a/community/bfh.gitignore
+++ b/community/bfh.gitignore
@@ -1,0 +1,19 @@
+# gitignore template for Bern University of Applied Sciences (BFH) workflows
+# website: https://www.ti.bfh.ch
+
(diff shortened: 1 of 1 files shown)


Lassana21:
Grazie mille

PR #4382, opened 2023-12-06
Add `.DS_Store` to `Python.gitignore`

heyodai:

As a Python developer, I always manually add `.DS_Store` to `.gitignore` in my projects. These files are created by macOS to store folder attributes: https://en.wikipedia.org/wiki/.DS_Store

These files are not relevant to the project's codebase and can clutter the repo if committed. I think it makes sense for `.DS_Store` exclusion to be the default.

- The `.gitignore` documentation from Git supports the idea of excluding system-specific files which are not relevant to the project: https://git-scm.com/docs/gitignore
- Stack Overflow discussions on the topic, such as [this thread](https://stackoverflow.com/questions/107701/how-can-i-remove-ds-store-files-from-a-git-repository) which highlights the common practice of ignoring `.DS_Store` in git repos.

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -158,3 +158,6 @@ cython_debug/
 #  and can be added to the global gitignore or merged into this file.  For a more nuclear
 #  option (not recommended) you can uncomment the following to ignore the entire idea folder.
 #.idea/
+
+# System files
+.DS_Store


hspaans:
The gitignore repository is split into a section of templates that you can use for specific languages and/or frameworks and should be in your repository. The second section of templates are for global usage like setting them for a user of the system as they're working on Windows or MacOS for example. Putting the content in `$XDG_CONFIG_HOME/git/ignore` will enable filtering for all git repositories in use by that user as that is default location.

You can also choose to use `$HOME/.gitignore` for example, but then this location has to be enabled with the command below:

```shell
git config --global core.excludesfile ~/.gitignore
```

People who are using GitHub Codespaces or Devcontainers for example the `.gitignore` and `.gitconfig` files can also be set in `<username>/dotfiles` repository so it will be part of every environment created.

Hopefully this helps to improve your experience with git and gitignore without extra modifications.

heyodai:
Thank you, @hspaans. I actually wasn't aware of these options. I'll look into them.

However, these workarounds have to be configured. New Mac users won't know about them and will end up committing the `.DS_Store` file at least some of the time. 

Unless there are legitimate situations where a `.DS_Store` file should be included in a repo, I don't see any reason not to have it be part of the default `.gitignore`. It's similar to `.venv` in this sense.

Sil-P:
> It's similar to .venv in this sense.

Not really, .venv is python-specific.

Maybe consider this suggestion instead: https://github.com/github/gitignore/pull/4401#issuecomment-1923723409

wirecat:

🙇

PR #4388, opened 2023-12-27
feat: add KiCad lock/autosave files

loozhengyuan:

Add `~*.lck` and `#auto_saved_files#` files for [KiCad](https://www.kicad.org)

These are _working files_ so they are not well-documented but we are able to see their usage from the [`KiCad/kicad-source-mirror`](https://github.com/KiCad/kicad-source-mirror).

**`~*.lck`**

https://github.com/KiCad/kicad-source-mirror/blob/69c95acc579e09195410f713ca42c79610cc4e05/common/wildcards_and_files_ext.cpp#L126-L127

https://github.com/KiCad/kicad-source-mirror/blob/69c95acc579e09195410f713ca42c79610cc4e05/include/lockfile.h#L54-L55

**`#auto_saved_files#`**

https://github.com/KiCad/kicad-source-mirror/blob/69c95acc579e09195410f713ca42c79610cc4e05/eeschema/files-io.cpp#L1688

### DIFF
--- a/KiCad.gitignore
+++ b/KiCad.gitignore
@@ -16,6 +16,8 @@ _autosave-*
 *-save.pro
 *-save.kicad_pcb
 fp-info-cache
+~*.lck
+\#auto_saved_files#
 
 # Netlist files (exported from Eeschema)
 *.net


wirecat:
No worries @loozhengyuan, thank you for providing the code references! That's very helpful when trying to decide what should be included. Thank you for contributing to our ignore templates!

wirecat (approved):

PR #4390, opened 2023-12-30
Add Carton local to Perl.gitignore

leonyu:

Carton is the most popular per-project dependency manager for Perl. ([Ref Mojolicious blog](https://mojolicious.io/blog/2018/12/22/use-carton-for-your-mojolicious-app-deployment/)) Carton is similar to Bundler for Ruby and NPM/Yarn for Node.js. In Python, Carton is analogous to Pipenv/Poetry (modify local pyenv), while CPAN/cpanminus analogous to PIP (modify system). 

This PR adds required ignore directory for Carton. 

I have no relationship with this project. I simply have a few Perl projects that use this tool. 

* https://metacpan.org/pod/Carton#Initializing-the-environment
* https://manpages.debian.org/testing/carton/Carton.3pm.en.html#Initializing_the_environment

### DIFF
--- a/Perl.gitignore
+++ b/Perl.gitignore
@@ -33,3 +33,6 @@ inc/
 /MANIFEST.bak
 /pm_to_blib
 /*.zip
+
+# Carton
+local/


leonyu:
It is still relevant if someone would approve it

PR #4394, opened 2024-01-11
Fix invalid comments

nielsbasjes:

While working on my gitignore reader https://github.com/nielsbasjes/codeowners I ran into problems while using the templates that originated from here as a testing reference.

There are several places where the author of the template intended to create a comment but did not. The root cause is that in a `.gitignore` file a comment starts with a `#` as the first character on a line. In all other places it is NOT interpreted as a comment.

I have verified this with the Dart.gitignore which has

    *.js             # When generated by dart2js. Don't specify *.js if your

If I create a `test.js` it is not ignored by git.

I have checked all cases found by this:

    find . -type f -name '*.gitignore' | xargs grep -v '^#' | fgrep '#'

https://git-scm.com/docs/gitignore#_pattern_format states

    A line starting with # serves as a comment. 
    Put a backslash ("\") in front of the first hash for patterns that begin with a hash.


dooleydevin (approved):
Good catch! Thanks for this!

Sum1cares (approved):

PR #4398, opened 2024-01-19
Create Nix.gitignore

cdmistman:
Support for common Nix output paths. I use Nix Flakes in all of my projects and this would be awesome for automatically ignoring the related Nix output directories

- for `result/`: https://nixos.org/manual/nix/stable/command-ref/nix-build#description
- for `result-<N>-<outname>/`: https://nixos.org/manual/nix/stable/command-ref/new-cli/nix3-build#description

If this is a new template:

 - **Link to application or project’s homepage**: https://nixos.org/

### DIFF
--- a/Nix.gitignore
+++ b/Nix.gitignore
@@ -1,0 +1,2 @@
+/result
+/result-*


Nix.gitignore
```diff
@@ -0,0 +1,2 @@
+/result
+/result-*
```
trueNAHO, line 2: The resulting output paths are not always at the root of the source tree:

> Build the default package from the flake in the current directory:
>
> (Source: nix3-build(1))

```suggestion
result
result-*
```

Related:

- https://github.com/github/gitignore/pull/616
- https://stackoverflow.com/questions/17888695
- https://stackoverflow.com/questions/24139478

trueNAHO (changes requested):

trueNAHO (approved):

Nix.gitignore
```diff
@@ -0,0 +1,2 @@
+result
+result-*
```
kishorekishoregit, line 2: 11

kishorekishoregit, line 2: COMMENTS

wirecat:
We already have these rules, but in our community folder. I'm moving the community template to root [here](https://github.com/github/gitignore/pull/4644).

Closing this guy as we'll have these rules in root soon.

PR #4401, opened 2024-01-26
Need to ignore .DS_Store from macOS filesystem

as1605:

In macOS, there is a hidden system-generated file `.DS_Store` which is not important for git, but always gets added to commits making them dirtier.

[WikiPedia](https://en.wikipedia.org/wiki/.DS_Store):
>.DS_Store files may impose additional burdens on a [revision control](https://en.wikipedia.org/wiki/Revision_control) process, since they are frequently changed and can therefore appear in commits, unless specifically excluded.

[Microsoft](https://support.microsoft.com/en-us/office/remove-ds-store-files-on-macos-d2f8dca0-740a-4f7e-89b9-5b2cbbc50386):
> Corrupted .DS_Store files can prevent OneDrive from syncing correctly. Deleting the corrupted files can help restore normal sync operation.
No data is lost by deleting the corrupted files as Mac Finder only creates .DS_Store files to store viewing options, such as the positions of icons, size of the Finder window, and window backgrounds.

[Adobe](https://helpx.adobe.com/dreamweaver/kb/remove-ds-store-files-mac.html): 
> These files contain information about system configuration. If you upload them along with other files, the files can be misused to obtain information about your computer.

[Apple](https://support.apple.com/en-in/102497): 
> By enticing a user to open a directory containing a maliciously crafted .DS_Store file, an attacker may cause arbitrary code execution. This update addresses the issue through improved bounds checking.

If this is a new template:

@shiftkey @bdougie

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -158,3 +158,6 @@ cython_debug/
 #  and can be added to the global gitignore or merged into this file.  For a more nuclear
 #  option (not recommended) you can uncomment the following to ignore the entire idea folder.
 #.idea/
+
+# macOS
+.DS_Store


Sil-P:
The global ignore should be used for OS-specific files: https://github.com/github/gitignore/tree/main/Global#globally-useful-gitignores

A comment like [the one for PyCharm](https://github.com/github/gitignore/pull/4401/files#diff-97bde3afc9623b8be1bb50eb1c85a420507a84d64f1c38f9a107f1cbda975806L155-L160) may be added instead, if maintainers find this to be a common request.
As https://github.com/github/gitignore/blob/main/Global/macOS.gitignore has many rules, I'd rather not include the whole thing.

PR #4407, opened 2024-02-03
Update Python.gitignore

azasada:
Added .ruff_cache directory (a cache directory used by [ruff linter](https://github.com/astral-sh/ruff)).


alexei:
`.ruff_cache` has it's own `.gitignore` set up to ignore everything inside i.e. `*`

See also:

https://github.com/github/gitignore/pull/4289

https://github.com/astral-sh/ruff/pull/208

PR #4419, opened 2024-03-06
added .vsconfig file to ignore in Unity's .gitignore file

am1goo:

`.vsconfig` is a part of Visual Studio 2022 Installer that able to exists in Unity's project folder.

https://learn.microsoft.com/en-us/visualstudio/install/import-export-installation-configurations?view=vs-2022

### DIFF
--- a/Unity.gitignore
+++ b/Unity.gitignore
@@ -25,6 +25,7 @@
 
 # Visual Studio cache directory
 .vs/
+.vsconfig
 
 # Gradle cache directory
 .gradle/


Unity.gitignore
```diff
@@ -25,6 +25,7 @@
 
 # Visual Studio cache directory
```
gapspt, line 26: I'd probably change this comment as well to simply `Visual Studio`

am1goo: Yeah, sure, go ahead!

am1goo:
That's strange coz every file and folder that started with . (dot) has external usage and not related to a project.

wirecat:
I don't have enough justification to add this rule. Closing for now.
Feel free to re-open if you find documentation and sources to back this change.

PR #3092, opened 2019-06-22
Added gitignore-file for Beckhoff TwinCAT3.

sagatowski:
**Provide a link to the application or project’s homepage:**
https://www.beckhoff.com/english.asp?twincat/twincat-3.htm
TwinCAT3 is a realtime kernel and development environment for PLCs (programmable logic controller). The IDE is based on Visual Studio, which TwinCAT is integrated into. It's one of the more popular development environments for industrial automation.

**Provide links to documentation:**
The official one:
https://download.beckhoff.com/download/document/automation/twincat3/TC3_SourceControl_EN.pdf
This only covers part of the files (specifying that TMC and TPY-files are not necessary to version control). Unfortunately, even Beckhoff (the manufacturer of TwinCAT) have failed to properly document which files should be version controlled and not.

Regarding the other files included in the gitignore-file of the pull request:
- *.tclrs - These are license response files, basically license files that are residing on the target PLC device
- *.compiled-library and *.library - which are libraries of which the project depends on. These are put here when they are referenced in the project, but the actual libraries needed for development are installed in "C:\TwinCAT\3.1\Components\Plc\Managed Libraries", which need to be installed there prior of compilation. The ones residing in the project must not be version controlled.
- *.compileinfo - Self explaining, file created everytime a compilation is made. Beckhoff documentation says: "The compile information was created during the last download of the PLC project and stored in a file".
- _Boot/ - Folder generated when project is compiled with all the binaries. This is what is transferred to the actual target device (PLC) for execution.
- _CompileInfo/ - Self explaining. Folder created during compilation. Holding the *.compileinfo-files.
- _Libraries/ - Folder holding the *.compiled-library and *.library in the local development folder. 

There are various github repos for TwinCAT that use this list for gitignore, one being TcUnit, a unit testing framework for TwinCAT3:
https://github.com/tcunit/TcUnit/blob/master/.gitignore

Note that because TwinCAT is using visual studio as development environment (it integrates itself into VS during installation), all visual studio files that are normally ignored shall be ignored as well (therefore the comment # Recommended: VisualStudio.gitignore).

Update 2019-12-05: I've published a blog post that goes through these files and the reason for having a gitignore-file for TwinCAT3 in a much greater detail now which is accessible here:
https://alltwincat.com/2019/12/02/gitignore-for-twincat/

**Explain why you’re making a change:**
[...]

PR #4424, opened 2024-03-25
Add gitignore template for Automation Studio Projects

shanereetz:
B&R Automation Studio is an IDE for developing industrial software for Programmable Logic Controllers (PLCs). [B&R](http://www.br-automation.com) is part of the ABB group and is one of the major names in industrial automation software. I work for a company,[ Loupe]( https://github.com/loupeteam), that develops open-source libraries and custom software for this platform, and other automation platforms like [Beckhoff TwinCAT](https://github.com/github/gitignore/pull/3092) that support modern software development practices.

https://help.br-automation.com/#/en/4/projectmanagement%2Fprojectorganisation_datastructure.html
- Binaries - contains generated data
- Diagnosis - contains generated files relevant for diagnostics and monitoring
- Temp and TempObjects - more temporary files

https://help.br-automation.com/#/en/4/automationruntime%2Fenhancedtransfer%2Fgeneral-description%2F04-x-pip-til.html
- *artransfer.br, *arTrsfMode.nv - files related to the transfer of programs to PLCs

https://help.br-automation.com/#/en/4/ncsoftware%2Farnc0ext%2Farnc0ext.html
- *arnc0ext.br - special library files subject to special terms of use that must be obtained from B&R

AS file types
- *.bak
- *.isopen - marks if a project file is open
- *.orig
- *.log
- *.asar
- *.set - user settings files

Library outputs
- RevInfo.var - variable files with embedded Git commit information generated by RevInfo
- *.csvlog* - log files generated when using CSVFileLib

Misc
- **/MPLogs/*
- **/DLFiles/*
- *.DS_Store - Mac files to ignore

 https://www.br-automation.com/en-us/products/software/automation-software/automation-studio/
 https://github.com/br-automation-com
 https://github.com/loupeteam
 https://github.com/loupeteam/StarterAsProject


AutomationStudio4.gitignore
```diff
@@ -0,0 +1,36 @@
...
+# 'ignored' directory
+ignored/
+
+# ARNC0ext
+*arnc0ext.br
+
+# AS File types
+*.bak
+*.isopen
+*.DS_Store
```
wirecat, line 23: We don't accept OS specific rules in template files unless they're added to the OS specific template. See [macOS.gitignore](https://github.com/github/gitignore/blob/7b22f8ab6c85b4ef1469d72a8ba96462e2a44853/Global/macOS.gitignore) for an example. Could you remove the OS specific rules?

AutomationStudio4.gitignore
```diff
@@ -0,0 +1,36 @@
...
+*.csvlog*
+*.set
+!**/Physical/**/*.set
+
+# RevInfo variables
+*RevInfo.var
+
+# Misc
+**/MPLogs/*
+**/DLFiles/*
```
wirecat, line 36: Please see other comment about OS Specific files.

AutomationStudio4.gitignore
wirecat, line 1: We don't accept template files for specific versions in the root. Please see [our docs](https://github.com/github/gitignore?tab=readme-ov-file#versioned-templates) for more details and adjust accordingly.

wirecat, line 1: I did some checking around and I'm not sure Autmoation Studio quite reaches the degree of widespread use and popularity we're looking for to be included in our root.

Would you be willing to move it to the community folder?

wirecat (changes requested):
Thank you for contributing!

I think this is a template we want to pull in, but there are some changes I would like to see first. 🙇

AutomationStudio4.gitignore
shanereetz: To clarify, this version number is not for the .gitignore template itself, but is referencing a specific Automation Studio version. An [existing .gitignore template for automation software, TwinCAT3](https://github.com/github/gitignore/pull/3092), follows this format.

Would it still be best to remove the version number from the filename?

More context:  I'm not sure which would be most appropriate as "current supported version." Automation Studio 4 was released in 2013 and is the current standard. But later this year, Automation Studio 6 will be released (5 was skipped) so this distinction seemed useful, as I'm unsure if there will be similarities with regards to gitignore. The automation space moves somewhat slowly, so most users will still be using version 4 for years to come, because so many machines were developed with it and it is more proven.

shanereetz: Is there data I could provide that might be useful in that decision? Otherwise I defer to you on what would be most appropriate.

According to B&R, Automation Studio has programmed machines for 30,000 factories worldwide, and is used by about 4,000 machine builders worldwide. They are unique in the industrial automation space as their PLC code is stored in plaintext, and can be effectively used with source control. I know it might be surprising to hear that as a unique metric, [but it is.] (https://github.com/SASE-Space/ot-openness-comparison)

AutomationStudio4.gitignore
shanereetz: Yes! Removed.

AutomationStudio4.gitignore
shanereetz: Completed.

AutomationStudio4.gitignore
wirecat: Sorry for the long delay @shanereetz. I think the best thing to do is to have the most recently supported version of the language at the root here with no versioning information in the name.

We can create ignore files for older versions under `community/` with versioning in their name.

If AutomationStudio4 is currently the latest, I think dropping the `4` from this gitignore is good enough for now.

shanereetz: Thank you @wirecat. I've made the changes requested, please let me know if there is anything else you would like to discuss.

shanereetz: Please let me know if any of these statistics move the needle in regards to the community folder or root folder decision. Thank you!

wirecat (approved):

wirecat:
Thank you for your contribution, patience, and working through this with me @shanereetz!

kazuki23887 (commented):
レビュー

dishu262626 (commented):
Thanq

PR #4350, opened 2023-09-03
Remove Cargo.lock from Rust

figsoda:
**Reasons for making this change:**

The Cargo team no longer have definitive suggestion on whether you should ignore `Cargo.lock`, and the previous description is now outdated.

**Links to documentation supporting these rule changes:**

https://blog.rust-lang.org/2023/08/29/committing-lockfiles.html
https://doc.rust-lang.org/nightly/cargo/faq.html#why-have-cargolock-in-version-control

PR #4427, opened 2024-03-28
Remove lock from rust gitignore

khionu:
As per updated guidelines, the lockfile should be committed by default.

https://doc.rust-lang.org/cargo/faq.html#why-have-cargolock-in-version-control


bensheldon:
Closed by https://github.com/github/gitignore/pull/4350

PR #4428, opened 2024-03-31
Update Go.gitignore -> Added .env

LombardiDaniel:

I use the default Go gitignore and it should have .env in it by default

_TODO_

_TODO_

If this is a new template:

### DIFF
--- a/Go.gitignore
+++ b/Go.gitignore
@@ -19,3 +19,6 @@
 
 # Go workspace file
 go.work
+
+# env file
+.env


philfreshman (approved):
LGTM

wirecat (changes requested):
Could you please provide a bit more context around why you believe this should be added?

Also, we currently have some merge conflicts that will need resolved.

LombardiDaniel:
@wirecat resolved! The main reason for this change is just that so many apps do need some sort of .env file to keep secrets while on development, me and a couple of colleagues have accidently uploaded our .env files with API keys to git in go projects because we did not check the .gitignore (just assumed that `.env` was in there, as is the case with lots of other default .gitignores files)

wirecat (approved):
@LombardiDaniel Thank you for resolving the merge conflict and adding clarification!

PR #4429, opened 2024-04-03
Add rubber cache files to TeX.gitignore

JasonGross:
Generated by [rubber](https://tex-talk.net/2011/12/building-documents-with-rubber/).  [The docs](https://github.com/petrhosek/rubber/blob/8ec18fd096b186901f197d26c8e1060b42f0b34f/doc/rubber.texi#L158) claim to use `rubber.cache`, but I've seen $filename.rubbercache in the wild, so I'm including both.

_TODO_

_TODO_

If this is a new template:

### DIFF
--- a/TeX.gitignore
+++ b/TeX.gitignore
@@ -39,6 +39,8 @@
 *.synctex.gz
 *.synctex.gz(busy)
 *.pdfsync
+*.rubbercache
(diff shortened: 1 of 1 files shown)


wirecat (approved):
Thank you for contributing!

PR #4430, opened 2024-04-07
Add .gitignore for C Sharp

suhaibjanjua:
I just added a .gitignore file made especially for C# projects. It's there to make sure we don't end up including unnecessary files and folders in our version control.


madebylydia:
I'm sad to see this is probably never gonna get committed. Even tho, it was truly helpful for me to ignore files I wasn't aware I should ignore, as a beginner in C#.
Thank you very much for your `.gitignore`!

wirecat:
👋 Hi @suhaibjanjua, this is primarily covering what is already in our [VisualStudio template](https://github.com/github/gitignore/blob/main/VisualStudio.gitignore).

Could you please look at that ignore file and consider extending it if needed?

MattPlum:
The problem is when people search for a C# gitignore, there is none. Better to be redundant than not have a repo for C#

wirecat:
> @wirecat The problem is when people search for a C# gitignore template here, there is none. Better to be redundant than not have a repo for C#

If we were to include it, this is very targeted at Visual Studio currently, but there are a growing number of developers working with C# using other IDEs.

stevefan1999-personal:
The template for Visual Studio is too general, so we should be more specific with C#, despite I personally think ".NET / C#" would be better naming, but consider that Kotlin also have its own gitignore option while it is not called "JVM / Kotlin" in retrospect.

ericnewton76:
Realistically the visualstudio.gitignore needs to start versioning.  VS2022 gets rid of a lot of crap that the current gitignore file includes that ignores old <VS2022 stuff.  Granted people are still using VS2019.  I dont believe they (and I wouldnt myself) expect visualstudio.gitignore to always work well for them. 

VisualStudio brings big enough changes iwth new versions that IMO there SHOULD BE separate gitignores for it.  visualstudio.gitignore really should be visualstudio-latest.gitignore and have a visualstudio-2022.gitignore and a vistualstudio-2019.gitignore and so on.

wirecat:
@suhaibjanjua AI haven't heard from you in response to my questions. I'm going to close this one for now as it isn't something we want to merge down as-is.

Thank you for contributing and I encourage you to look into more generic C# ignore rules we might accept.

PR #4434, opened 2024-04-22
Ignore lock info files in terraform `.gitignore`

brainplot:
This is recommended by the official Style Guide here: https://developer.hashicorp.com/terraform/language/style#gitignore.


wirecat (approved):
Thank you for contributing!

PR #4437, opened 2024-04-30
Added Carton's local directory

jonasbn:
I am a user of Carton. Carton is a dependency manager for Perl, much like Bundler for Ruby.

Carton creates a local directory with the dependency file tree named `local/` just as for example: `node_modules`.

Carton installation documentation recommends adding this particular directory to: `.gitignore`, see: https://metacpan.org/pod/Carton#Initializing-the-environment

> echo local/ >> .gitignore

Carton is widely used for Perl development and on the same level as the other tools listed in the Perl gitignore file.

### DIFF
--- a/Perl.gitignore
+++ b/Perl.gitignore
@@ -33,3 +33,6 @@ inc/
 /MANIFEST.bak
 /pm_to_blib
 /*.zip
+
+# Carton
+local/


wirecat:
👋 Hi @jonasbn. We typically create tool specific templates for things like this. Would you consider creating a Carton template in `Global`?

I'm a bit uncomfortable including `local/` for all of perl. While not common, we do have examples of [people committing files in that directory.](https://github.com/search?q=language%3APerl+path%3A%2Flocal%2F*&type=code)

jonasbn:
Hi @wirecat 

I will do that, thanks for the feedback

PR #4438, opened 2024-05-01
Update Python.gitignore to keep .idea in gitignore by default

githubofkrishnadhas:
uncommenting .idea in line 160 python gitignore file to make it by default

The .idea folder (hidden on OS X) in the solution root contains IntelliJ’s project specific settings files. These include per-project details such as VCS mapping and run and debug configurations, as well as per-user details, such as currently open files, navigation history and currently selected configuration.

_TODO_

https://github.com/github/gitignore/blob/main/Global/JetBrains.gitignore

we can see entirely dedicated gitignore for it. but for python folks selecting python as language .idea is commented by default
_TODO_

If this is a new template:
Not a new template. just updated existing

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -157,4 +157,4 @@ cython_debug/
 #  be found at https://github.com/github/gitignore/blob/main/Global/JetBrains.gitignore
 #  and can be added to the global gitignore or merged into this file.  For a more nuclear
 #  option (not recommended) you can uncomment the following to ignore the entire idea folder.
-#.idea/
+.idea/


Redulation:
Can someone please explain to me what this is I am not a technology smart
person at all I'm sorry I noticed this was a thread or whatever you would
call it as attached to my phone when I was looking at things I have access
to my information and my Google account every time I erase it it's on there
again thank you in advance I'm sorry

githubofkrishnadhas:
When you work with a Python project and use JetBrain IDEs (pycharm), it keeps track using some files created inside a hidden folder named .idea. These are some xml files which don't have anything to do with real project.

Project settings are stored with each specific project as a set of xml files under the .idea folder. If you specify the default project settings, these settings will be automatically used for each newly created project.

image

wirecat:
👋 Hi @githubofkrishnadhas, thank you for contributing to our repo. I've reviewed your changes and I don't feel comfortable pulling them in.

The comment itself gives good reasoning for why this is commented out by default and attempt to refer people to the [JetBrains specific](https://github.com/github/gitignore/blob/main/Global/JetBrains.gitignore) ignore. I don't see a good reason to turn this on by default for everyone.

I'm going to close for now, but please free to re-open if you think this needs additional discussion.

githubofkrishnadhas:
:)

githubofkrishnadhas:
What would be the demerit if .idea is enabled by default or uncommented?

It's not going to affect in any negative manner.

wirecat:
@githubofkrishnadhas The preference is to use [global](https://github.com/github/gitignore/blob/main/Global/README.md) .gitignores for things like IDE and OS specific files we don't want to include.

PR #4441, opened 2024-05-05
Add hypdoc temporary files

schtandard:

The `hypdoc` package is used in the documentation of most LaTeX packages. It creates a `.hd` auxiliary file upon compilation that should be ignored by Git.

- [Package page on CTAN](https://ctan.org/pkg/hypdoc)
- [Package documentation (pdf)](http://mirrors.ctan.org/macros/latex/contrib/hypdoc/hypdoc.pdf)

### DIFF
--- a/TeX.gitignore
+++ b/TeX.gitignore
@@ -138,6 +138,9 @@ acs-*.bib
 *.trc
 *.xref
 
+# hypdoc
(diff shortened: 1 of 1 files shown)


wirecat (approved):
Thank you for contributing!

PR #2860, opened 2018-10-28
[Xcode, Obj-C, Swift] Remove legacy settings files 

ealeksandrov:
**Reasons for making this change:**

If you look for popular Xcode gitignore templates and its descriptions, you'll see that `pbxuser`, `mode1v3`, `mode2v3` and `perspectivev3` are ignored everywhere with comment that it is "user's preferences for the project, e.g. window sizes, active build style, file bookmarks, and a few other things". Original reference about its purpose comes back to Apple mailing lists from 2004-2007.

But for last 3+ years I never saw those files in Xcode projects. For my own use I removed them (lines in gitignore for my projects) and had no issues. Xcode still saves breakpoints and windows sizes, but they don't go inside repository. Easy to check - open your local `xcodeproj` folder (not just fresh git clone, but opened in Xcode before this) and try to find them. There are none.

We can safely exclude it to make templates cleaner.

PR #4444, opened 2024-05-12
[Xcode, Objective-C, Swift] Remove legacy Xcode compatibility

ealeksandrov:
This is a rebase of https://github.com/github/gitignore/pull/2860. Check it for earlier discussion and approvals.

- ebee5bfe87a2799d26d346798f7364eb0dccdcaf is a cleanup of Xcode 3 compatibility in `ObjectiveC` and `Swift` templates (already merged in https://github.com/github/gitignore/pull/3935 for `Xcode`)
  - related code is unnecessary since ~2011
- a4043487860cd48d0bdb54aa41e21dbd7fc91afc is a cleanup of Xcode 8 compatibility in `Xcode`, `ObjectiveC` and `Swift` templates
  - related code is unnecessary since ~2017

> Please note that as of April 2024 all iOS and iPadOS apps submitted to the App Store must be built with a minimum of Xcode 15 and the iOS 17 SDK.
(https://developer.apple.com/ios/submit/)

Also see https://stackoverflow.com/a/53039267

### DIFF
--- a/Global/Xcode.gitignore
+++ b/Global/Xcode.gitignore
@@ -1,6 +1,2 @@
 ## User settings
 xcuserdata/
-
-## Xcode 8 and earlier
-*.xcscmblueprint
-*.xccheckout
--- a/Objective-C.gitignore
+++ b/Objective-C.gitignore
@@ -9,19 +9,6 @@ xcuserdata/
 *.xcscmblueprint
 *.xccheckout
 
-## compatibility with Xcode 3 and earlier (ignoring not required starting Xcode 4)
-build/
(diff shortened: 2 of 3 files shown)


Coeur (approved):
Xcode 9 was released on 12 Sep 2017.
It's been an eternity that Xcode 8 isn't allowed anymore to publish any app on AppStoreConnect.
We could have deleted those lines since 2018-2019 without second thoughts.

wirecat (approved):
@ealeksandrov Thank you for dusting off that PR!

PR #4445, opened 2024-05-13
Ignore terragrunt cache

costaparas:
This excludes the common/default cache location from source control

Terragrunt is a convenient tool used when using terraform. The cache directory is not needed in source control; as per the docs, it can be safely deleted as it will be re-created automatically when needed.

Here is a sample .gitignore from the terragrunt project itself:
https://github.com/gruntwork-io/terragrunt/blob/master/.gitignore

Reference docs about the cache directory:
https://terragrunt.gruntwork.io/docs/features/caching/

### DIFF
--- a/Terraform.gitignore
+++ b/Terraform.gitignore
@@ -32,7 +32,3 @@ override.tf.json
 # Ignore CLI configuration files
 .terraformrc
 terraform.rc
-
-# Ignore the default terragrunt cache directory
-# https://terragrunt.gruntwork.io/docs/features/caching/
-.terragrunt-cache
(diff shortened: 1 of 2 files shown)


wirecat:
👋 Hi @costaparas. We don't typically add tooling specific ignores into the more generic templates.
Would you be open to creating a template in `community/` for terragrunt instead?

costaparas:
> 👋 Hi @costaparas. We don't typically add tooling specific ignores into the more generic templates. Would you be open to creating a template in `community/` for terragrunt instead?

Updated accordingly @wirecat

wirecat:
@costaparas Thank you for the change and contributing to gitignore!

PR #4447, opened 2024-05-19
Update Terraform.gitignore to ignore .hcl file to be committed

githubofkrishnadhas:

Making this change to ignore .terraform.lock.hcl which is an internal file generated by terraform and not needed in our remote repos as they are automatically generated on each terraform init runs
image

_TODO_

NA
_TODO_

If this is a new template:
existing template, just modified.

 NA

### DIFF
--- a/Terraform.gitignore
+++ b/Terraform.gitignore
@@ -32,3 +32,6 @@ override.tf.json
 # Ignore CLI configuration files
 .terraformrc
 terraform.rc
+
+# Ignore hcl file
+*.hcl


Terraform.gitignore
```diff
@@ -32,3 +32,6 @@ override.tf.json
 # Ignore CLI configuration files
 .terraformrc
 terraform.rc
+
+# Ignore hcl file
+*.hcl
```
wirecat, line 37: Could we call out the `.terraform.lock.hcl` file specifically?

If not, could you provide more external references to justify the more broad rule?

wirecat (changes requested):

Terraform.gitignore
githubofkrishnadhas: yes we can call it as `.terraform.lock.hcl` instead of wildcard expresssion `*.hcl`.

shall i raise a new PR? please advice

wirecat: Updating this PR with the change is preferred in this situation. Thank you!

PR #4449, opened 2024-05-25
Create ReScript.gitignore

brettcannon:
There is a lack of support for the [ReScript programming language](https://rescript-lang.org/).

There was a [discussion](https://forum.rescript-lang.org/t/what-would-belong-in-a-gitignore-template-for-rescript/5263) about what to include in the file. What is being proposed matches what a [new project](https://rescript-lang.org/docs/manual/latest/installation#new-project) sets in its `.gitignore` file.

If this is a new template:

 - **Link to application or project’s homepage**: https://rescript-lang.org/

### DIFF
--- a/ReScript.gitignore
+++ b/ReScript.gitignore
@@ -1,0 +1,4 @@
+.DS_Store
+/node_modules/
+/lib/
+.bsb.lock
--- a/ReScript.gitignore
+++ b/ReScript.gitignore
@@ -1,4 +1,3 @@
-.DS_Store
 /node_modules/
 /lib/
 .bsb.lock


ReScript.gitignore
```diff
@@ -0,0 +1,4 @@
+.DS_Store
```
wirecat, line 1: Thank you for contributing, but we don't accept OS specific rules in template files unless they're added to the OS specific template in [Global](https://github.com/github/gitignore/blob/main/Global/README.md).

Would you consider dropping `.DS_Store`?

wirecat (changes requested):
Thank you for contributing! Only 1 comment from me and then I think this is ready to merge.

ReScript.gitignore
brettcannon: Done!

brettcannon:
I addressed the review comment (and apologies on the delay; new parent 😅).

wirecat:
@brettcannon Thank you for the update! I can't throw stones about any delay here 😅.
Thank you for contributing to the project!

PR #4450, opened 2024-05-27
Update Unity.gitignore

Cadons:
vscode and jetbrains folders added

### DIFF
--- a/Unity.gitignore
+++ b/Unity.gitignore
@@ -25,6 +25,8 @@
 
 # Visual Studio cache directory
 .vs/
+.vscode/
+.idea/
 
 # Gradle cache directory
 .gradle/


offensive-vk (approved):

Unity.gitignore
```diff
@@ -25,6 +25,8 @@
 
 # Visual Studio cache directory
 .vs/
+.vscode/
+.idea/
```
wirecat, line 30: 👋 Hello @Cadons we are moving away from applying ignores like this for IDEs as it is desirable to commit them for some.

Please see [this example](https://github.com/github/gitignore/blob/7b22f8ab6c85b4ef1469d72a8ba96462e2a44853/Python.gitignore#L157-L162).

Would you be willing to take a similar approach here?

Cadons: I looked at the example, but I didn't understand if I need to include these global ignores in the gitignore, I don't know if it's possible, or if I need to add the global ignores in my repo settings. 
I like the idea of splitting the ignore file into separate files

wirecat: Sorry about that, I don't think my original ask was very clear.

The idea is we don't have lines for IDEs in ignore files that aren't specifically related to them. We have begun to have a bit more flexibility by having them in ignore files, but commented out with a comment explaining why one might chose to enable them (like in [this example](https://github.com/github/gitignore/blob/7b22f8ab6c85b4ef1469d72a8ba96462e2a44853/Python.gitignore#L157-L162)). That's an option available here if you'd like to go that route.

Otherwise, our advice use to create and apply [global gitignores](https://github.com/github/gitignore/blob/main/Global/README.md) that apply to things like OS and IDE specific configuration.

wirecat:
@Cadons I haven't heard anything on this PR in a while. Could you review the comments I've made?

Cadons:
> @Cadons I haven't heard anything on this PR in a while. Could you review the comments I've made?

I have read your comment, what do you expect that I do? undo the edits?

wirecat:
> > @Cadons I haven't heard anything on this PR in a while. Could you review the comments I've made?
> 
> I have read your comment, what do you expect that I do? undo the edits?

You could create an entry similar to [this one](https://github.com/github/gitignore/blob/7b22f8ab6c85b4ef1469d72a8ba96462e2a44853/Python.gitignore#L157-L162) if you like. Otherwise I believe we should close this PR.

PR #4452, opened 2024-05-30
Update Java.gitignore 

ayush8914:
specified few files and directory name

### DIFF
--- a/Java.gitignore
+++ b/Java.gitignore
@@ -22,3 +22,12 @@
 # virtual machine crash logs, see http://www.java.com/en/download/help/error_hotspot.xml
 hs_err_pid*
 replay_pid*
+
+# Gradle specific files
+.gradle/
+build/
+
+# Eclipse specific files
+.settings/
+.project
+.classpath


Java.gitignore
```diff
@@ -22,3 +22,12 @@
...
 replay_pid*
+
+# Gradle specific files
+.gradle/
+build/
+
+# Eclipse specific files
+.settings/
+.project
+.classpath
```
wirecat, line 33: We have an [eclipse specific](https://github.com/github/gitignore/blob/main/Global/Eclipse.gitignore) template and would prefer to not include IDE specific config in language templates.

Java.gitignore
```diff
@@ -22,3 +22,12 @@
 # virtual machine crash logs, see http://www.java.com/en/download/help/error_hotspot.xml
 hs_err_pid*
 replay_pid*
+
+# Gradle specific files
+.gradle/
+build/
```
wirecat, line 28: See the comment about the eclipse specific template. I would prefer we create a new Global entry for Gradle.

In addition, we could also consider adding a descriptive comment and leaving these entries, but having them commented out by default.
[Example of this in the Python template.](https://github.com/github/gitignore/blob/main/Python.gitignore#L157-L162)

wirecat (changes requested):
Thank you for contributing! There's a few points I'd like to discuss further.

Also, can you please provide more details about your changes? Please see our [contribution guidelines](https://github.com/github/gitignore/blob/main/README.md#contributing-guidelines) for more details.

ayush8914:
yaa, that's right.

PR #4455, opened 2024-06-02
Update Terraform.gitignore to ignore graph files

RScrafted:
Added line to ignore graph file('s).

I am doing a cloud deployment project. .gitignore is critical and relevant for me to avoid uploading unnecessary files. I find the template helpful and feel the necessity to ignore graph files. These are saved as .dot extension which is standard with terraform. I noticed the default template was missing graph ignore files which are optional in most cases as they serve as a local function. As a result, I have added the same.

https://developer.hashicorp.com/terraform/cli/commands/graph

If this is a new template:


RScrafted (commented):
Added other formats to be ignored

Terraform.gitignore
```diff
@@ -35,3 +35,22 @@ override.tf.json
...
+
+# Ignore hcl files
+.terraform.lock.hcl
+*.terraform.*
+
+# Ignore graph files
+*.dot
+
+# to skip MacOS custom attributes. It appears when creating a .bicep file.
+.DS_Store
```
thorrsson, line 47:
```suggestion
```

Could you please remove this OS specific line? We encourage people to instead use global gitignore configs and have a collection of them that can be used. Please see our [Global/README](https://github.com/github/gitignore/blob/main/Global/README.md) for more information and [Global/macOS.gitignore](https://github.com/github/gitignore/blob/main/Global/macOS.gitignore) for an example.

Terraform.gitignore
```diff
@@ -35,3 +35,22 @@ override.tf.json
 # Ignore CLI configuration files
 .terraformrc
 terraform.rc
+
+# Ignore hcl files
+.terraform.lock.hcl
```
thorrsson, line 40: https://developer.hashicorp.com/terraform/language/files/dependency-lock#lock-file-location, specifically this section is contrary to your proposed change:

You should include this file in your version control repository so that you can discuss potential changes to your external dependencies via code review, just as you would discuss potential changes to your configuration itself.

```suggestion
```

Terraform.gitignore
```diff
@@ -35,3 +35,22 @@ override.tf.json
...
+*.terraform.*
```
thorrsson, line 41: This is a bit dangerous
if for some reason I name a module `thorrsson.terraform.tf` I won't be able to commit it.

```suggestion
```

Terraform.gitignore
```diff
@@ -35,3 +35,22 @@ override.tf.json
...
+
+# Ignore graph files
+*.dot
+
+# to skip MacOS custom attributes. It appears when creating a .bicep file.
+.DS_Store
+
+# Ignore plan files (optional) as some might save plan before destroy the terraform configuration
```
thorrsson, line 49: If this is optional (per the comment) let's just comment out the line and let folks uncomment if they need to

Terraform.gitignore
```diff
@@ -35,3 +35,22 @@ override.tf.json
...
+planout
+
+# Ignore tfplan files, saved during terraform plan execution
+*.out
```
thorrsson, line 53: Given the naming of the plan files is entirely user-driven I don't think we can assume that every user of terraform will name their plan outputs $something.out

thorrsson (changes requested):
Some feedback on this PR, I think there are a number of items we need to remove from this PR before it could be merged and consumed by the community

Terraform.gitignore
RScrafted: Thanks @thorrsson! You're right — I've removed .terraform.lock.hcl from the ignore list since it's recommended to include it in version control.

Terraform.gitignore
RScrafted: Thank you so much for the helpful feedback! I really appreciate the guidance on best practices. I’ve removed `.DS_Store` from the `.gitignore` since I understand it’s better handled via global gitignore configs. Looking forward to learning more from the team!

thorrsson (approved):

thorrsson:
Thanks @RScrafted

yehtikezc4732-rgb:
https://github.com/github/gitignore.git

PR #4479, opened 2024-07-03
Update Swift.gitignore

zohairhadi:

.DS_Store don’t add to your project and [could leak data from other file names you didn’t stage for checkin](https://metacpan.org/dist/Mac-Finder-DSStore/view/DSStoreFormat.pod).

https://apple.stackexchange.com/questions/471585/is-it-dangerous-to-leave-ds-store-files-left-in-git-commit-history#:~:text=If%20you're%20really%20worried,be%20necessary%20for%20your%20protection.&text=I%20would%20remove%20them%20git,didn't%20stage%20for%20checkin.

### DIFF
--- a/Swift.gitignore
+++ b/Swift.gitignore
@@ -2,6 +2,8 @@
 #
 # gitignore contributors: remember to update Global/Xcode.gitignore, Objective-C.gitignore & Swift.gitignore
 
+.DS_Store
+
 ## User settings
 xcuserdata/
 


KosFM:
I think this is not a Swift related rule. I think it needs to be added to `macOS.gitignore` and it already exists in there.
https://github.com/github/gitignore/blob/8779ee73af62c669e7ca371aaab8399d87127693/Global/macOS.gitignore#L2

PR #4481, opened 2024-07-04
Update VisualStudio.gitignore - Add Verify test results

karol-gro:

These are files generated by Verify test library for approval tests. They shouldn't be checked in.

https://github.com/VerifyTests/Verify?tab=readme-ov-file#source-control-received-and-verified-files


karol-gro:
@thorrsson thanks. TBH, the `*.received/` is also relevant if one use Verify in [UseSplitModeForUniqueDirectory](https://github.com/VerifyTests/Verify/blob/main/docs/naming.md#usesplitmodeforuniquedirectory) mode, but that's definitely more rare

PR #4460, opened 2024-06-05
Update Android.gitignore and Java.gitignore with kotlin directory

GianfrancoMS:
**Reasons for making this change:**
Starting with Kotlin 2.0, Jetbrains suggests adding the `.kotlin` directory to the project's `gitignore` file

**Documentation**
https://kotlinlang.org/docs/whatsnew20.html#new-directory-for-kotlin-data-in-gradle-projects

PR #4482, opened 2024-07-10
Update Gradle.gitignore

DRSchlaubi:
Kotlin 2.0 added a new data directory called `.kotlin` for Gradle projects. I am not sure whether this should be in the Kotlin or the Gradle template as it's only for Gradle projects and the Kotlin template is a symlink to Java

https://kotlinlang.org/docs/whatsnew20.html#new-directory-for-kotlin-data-in-gradle-projects

### DIFF
--- a/Gradle.gitignore
+++ b/Gradle.gitignore
@@ -14,6 +14,9 @@ gradle-app.setting
 # Cache of project
 .gradletasknamecache
 
+# Kotlin compilation data
+.kotlin
+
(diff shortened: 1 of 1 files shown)


kyhule:
This should probably be combined with #4460.

DRSchlaubi:
Didn't see that one, but should we really extend the Java template

ykws:
I also think the Java template is unnecessary. However, I believe it should be applied to Android together with the Gradle template.

PR #4483, opened 2024-07-13
Fixed Godot 4+ .gitignore file

GJMCharles:
A follow-up to these that were left open since LAST YEAR:
https://github.com/github/gitignore/pull/4414 & https://github.com/github/gitignore/pull/4385

As a game dev on Godot 4+, I have to constantly include these small changes in the .gitignore file.
To prevent errors while working with a team: such as class_name conflict when renamed, because local cache does not understand the changes made.

All I can say, is that files such as "icon.svg.import" or "gam3E1D.tmp" are now ignored from git commits.
It would greatly help the community as well.


NexusSfan (approved):
I approve of this PR. As a godot dev the .import files seem very useless and take up space

PowerfulBacon (changes requested):
You didn't include a link to documentation supporting the idea that the .import files are useless, after doing some digging there are documentation pages about the .import files: https://docs.godotengine.org/en/4.0/tutorials/assets_pipeline/import_process.html#files-generated

> Importing will add an extra <asset>.import file next to the source file, containing the import configuration.
> 
> **Make sure to commit these files to your version control system, as these files contain important metadata.**

NexusSfan:
> You didn't include a link to documentation supporting the idea that the .import files are useless, after doing some digging there are documentation pages about the .import files: https://docs.godotengine.org/en/4.0/tutorials/assets_pipeline/import_process.html#files-generated
> 
> > Importing will add an extra .import file next to the source file, containing the import configuration.
> > **Make sure to commit these files to your version control system, as these files contain important metadata.**

Only for custom assets. Normal assets that are unmodified will most likely work normally.

PR #4484, opened 2024-07-15
Create onlycode.gitignore

hackpro4005:

The .gitignore i added only downloads stuff the project may need

_TODO_

I have no documentation
_TODO_

If this is a new template:

  It isn't for a project

### DIFF
--- a/onlycode.gitignore
+++ b/onlycode.gitignore
@@ -1,0 +1,5 @@
+.md
+.txt
+.
+.blend
+.gitignore


dooleydevin (changes requested):
Hey there 👋 if this file is not specific to a project that you can link to, then we do not see a reason for including it. With that in mind, I'm going to close out this PR for now.

If you disagree and can provide a link to a project that this applies to, feel free to open it back up with more information 🙂

PR #4485, opened 2024-07-17
Create Hexo.gitignore

SunnyCloudYang:
Add .gitignore file for the popular static site generator Hexo.

[.gitignore at Hexo example](https://github.com/hexojs/hexo-starter/blob/master/.gitignore)
[.gitignore at hexo/site](https://github.com/hexojs/site/blob/master/.gitignore)

If this is a new template: 

 - **Link to application or project’s homepage**: [Hexo](https://hexo.io/)

### DIFF
--- a/community/Hexo.gitignore
+++ b/community/Hexo.gitignore
@@ -2,16 +2,13 @@
 # website: https://hexo.io/
 # Recommended: Node.gitignore
 
-# Ignore generated files and directories
-.tmp*
+# Ignore generated directory
 public/
 
-# Ignore temp files and lock files
+# Ignore temp files
 tmp/
-yarn.lock
-package-lock.json
-pnpm-lock.yaml
+.tmp*
 
-# Add any additional patterns here
+# additional files
 db.json
 .deploy*/
--- a/community/Hexo.gitignore
+++ b/community/Hexo.gitignore
@@ -1,24 +1,18 @@
 # gitignore template for Hexo sites
+# website: https://hexo.io/
+# Recommended: Node.gitignore
+
 # Ignore generated files and directories
-.DS_Store
 .tmp*
 dist/
 public/
 coverage/
 
-# Ignore dependencies and lock files
-node_modules/
+# Ignore temp files and lock files
 tmp/
 yarn.lock
 package-lock.json
 pnpm-lock.yaml
-
-# Ignore IDE specific files and directories
-.idea/
-.vscode
-
-# Ignore code coverage output
-.nyc_output/
 
 # Add any additional patterns here
 db.json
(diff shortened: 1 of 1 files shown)


community/Hexo.gitignore
```diff
@@ -0,0 +1,19 @@
...
+.tmp*
+dist/
+public/
+coverage/
+
+# Ignore temp files and lock files
+tmp/
+yarn.lock
+package-lock.json
+pnpm-lock.yaml
```
dooleydevin, line 15: Generally, package lock files should be included in source control. You can see that yarn's guidance on this [here](https://yarnpkg.com/getting-started/qa#which-files-should-be-gitignored).

If for some reason this is different for hexo, please let me know (references are helpful here)! Otherwise, could you remove these lock files from the template?

community/Hexo.gitignore
```diff
@@ -0,0 +1,19 @@
...
+
+# Add any additional patterns here
+db.json
+.deploy_git/
```
dooleydevin, line 19: Based on the example `.gitignore` file from the hexo-starter repo you shared, it may make sense to change this to `.deploy*` [like they do](https://github.com/hexojs/hexo-starter/blob/master/.gitignore#L7). Let me know what you think!

community/Hexo.gitignore
```diff
@@ -0,0 +1,19 @@
+# gitignore template for Hexo sites
+# website: https://hexo.io/
+# Recommended: Node.gitignore
+
+# Ignore generated files and directories
+.tmp*
+dist/
+public/
+coverage/
```
dooleydevin, line 9: I'm seeing reference to the `public/` directory in the Hexo docs, but not the `dist/` and `coverage/` directories 🤔 can you give a bit more context on what they are used for? I want to make sure we aren't excluding anything that may be useful to commit.

dooleydevin (changes requested):
Thanks for your contribution! Just a few questions/concerns before we can move forward with merging this 😄

community/Hexo.gitignore
SunnyCloudYang: Apologize for not providing a full reference, in fact I referenced [.gitignore at hexojs/site](https://github.com/hexojs/site/blob/master/.gitignore) as well, which include the lock files

community/Hexo.gitignore
SunnyCloudYang: ok, I'll change that

community/Hexo.gitignore
SunnyCloudYang: Well I found them [here](https://github.com/hexojs/hexo/blob/master/.gitignore), maybe they're for the hexo project, not a hexo-built project, I think it's better to remove them.

community/Hexo.gitignore
SunnyCloudYang: I read the guidance of yarn carefully and I agree with removing them from the template

dooleydevin:
@SunnyCloudYang thanks for making the suggested changes! Merged!

PR #4462, opened 2024-06-07
Updated to KiCAD v8

sethkaz:
**Reasons for making this change:**
I'm a KiCAD user, and the gitignore that keeps popping up as the first google hit is 3-4 major versions old.  And many of the temporary and lock files have changed namings.

PR #4487, opened 2024-07-21
Ignore lock file in KiCad

akshaybabloo:

Ignore all lock files. These are only used when one of the editors are opened and are deleted when closed.

Doxygen link - https://docs.kicad.org/doxygen/lockfile_8h.html


settinger:
I believe this PR is already covered by #4462 , which has several other changes for modern KiCad versions as well.

PR #4488, opened 2024-07-23
Update Unity.gitignore

whisper3zzz:
PlasticSCM is the official version control tool provided by Unity, it is popular for its fast and convenient interactive experience in the editor, and many developers are using it, but it should be recognised that while using PlasticSCM many developers also have the need to upload their projects to Github.

JetBrains Rider is the world's most loved .NET and game dev IDE，more and more developers are using it.

So I've updated the file for the useless cache files generated by these two popular software.

[Plastic SCM](https://www.plasticscm.com/)
[JetBrains Rider](https://www.jetbrains.com/rider/)


Unity.gitignore
```diff
@@ -23,9 +23,15 @@
 # Autogenerated Jetbrains Rider plugin
 /[Aa]ssets/Plugins/Editor/JetBrains*
 
+# Plastic SCM meta and ache directory
```
Almenon, line 26:
```suggestion
# Plastic SCM meta and cache directory
```

Unity.gitignore
```diff
@@ -23,9 +23,15 @@
...
+.plastic/
+
 # Visual Studio cache directory
 .vs/
 
+#JetBrains Rider cache directory
```
PowerfulBacon, line 32:
```suggestion
# JetBrains Rider cache directory
```

Unity.gitignore
```diff
@@ -23,9 +23,15 @@
 # Autogenerated Jetbrains Rider plugin
 /[Aa]ssets/Plugins/Editor/JetBrains*
 
+# Plastic SCM meta and ache directory
+.plastic/
```
dooleydevin, line 27: @whisper3zzz do you believe that it is desirable to always ignore this directory? Would this hinder developers from contributing to Unity projects they source from GitHub?

Searching through public GitHub repositories, I see many projects that are including this directory. It is difficult to determine whether this is desirable in all cases, which would make this undesirable for us to include here.

Unity.gitignore
```diff
@@ -23,9 +23,15 @@
...
+
 # Visual Studio cache directory
 .vs/
 
+#JetBrains Rider cache directory
+.idea/
```
dooleydevin, line 33: @whisper3zzz per the contribution guidelines, we do not accept editor-specific paths to ignore since they are included in the [global gitignore directory](https://github.com/github/gitignore/blob/main/Global/README.md). However, we are open to having them commented out with an explanation like you see [here](https://github.com/github/gitignore/blob/main/Python.gitignore#L157-L162) 👍

dooleydevin (changes requested):

PR #4489, opened 2024-07-31
Update Linux.gitignore

mattalxndr:

I was comparing my global gitignore with the offerings here and I noticed some holes that could be filled.

- [`man nohup`](https://man7.org/linux/man-pages/man1/nohup.1.html)

If this is a new template:

### DIFF
--- a/Global/Linux.gitignore
+++ b/Global/Linux.gitignore
@@ -11,3 +11,9 @@
 
 # .nfs files are created when an open file is removed but is still being accessed
 .nfs*
+
+# Log files created by default by the nohup command
+nohup.out
+
+# Metadata left by Dolphin file manager, which comes with KDE Plasma
+.directory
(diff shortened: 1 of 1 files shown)


Global/Linux.gitignore
```diff
@@ -11,3 +11,9 @@
 
 # .nfs files are created when an open file is removed but is still being accessed
 .nfs*
+
+# Log files created by default by the nohup command
+nohup.out
+
+# Metadata left by Dolphin file manager, which comes with KDE Plasma
+.directory
```
dooleydevin, line 19: This is actually already included on [lines 6-7 above](https://github.com/github/gitignore/pull/4489/files#diff-8704ae78f5e916819383a0c34fab3c7d7acdc2b6fc15dd189184780821f734edR6-R7)!

dooleydevin (changes requested):
Thanks for this contribution! Just one small request to remove the duplicate `.directory` path 😄

asummerford (approved):

Global/Linux.gitignore
wirecat: @mattalxndr Are you available and willing to make that change?

PR #4490, opened 2024-08-02
Updated NCrunch cache path for v4.16+

fuzzybair:
NCrunch updated the default location of the cache storage directory in version 4.16

> "Adjusted the default path for NCrunch's cache storage directory so that this now starts with a period instead of an underscore. For example, '_NCrunch_Solution' will now be '.NCrunch_Solution'. Solutions using the old path will continue to use the old path unless this is manually renamed. This change has been made to prevent VS from automatically including the files from NCrunch's cache storage directory in projects that may be stored adjacent to the solution file."

https://www.ncrunch.net/download/showChangeLog?version=4.16

### DIFF
--- a/Objective-C.gitignore
+++ b/Objective-C.gitignore
@@ -42,10 +42,3 @@ fastlane/report.xml
 fastlane/Preview.html
 fastlane/screenshots/**/*.png
 fastlane/test_output
-
-# Code Injection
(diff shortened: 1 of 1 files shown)


fuzzybair:
What is needed to move forward with this?

PR #4498, opened 2024-09-01
Ignore cargo-mutants output

katediadechko:

This change is to ignore mutation testing output in Rust projects using [cargo-mutants](https://mutants.rs/).

[The mutants.out directory](https://mutants.rs/mutants-out.html).


wirecat (approved):

PR #4501, opened 2024-09-05
Handle source folders inside subprojects in Gradle.gitignore

oSumAtrIX:

Subprojects, traditionally in a Gradle project, can also have a build directory inside the default source set directory. 

https://docs.gradle.org/current/userguide/multi_project_builds.html


wirecat (approved):
Thank you for contributing!

PR #4502, opened 2024-09-06
Update Unity.gitignore .utmp

Thaina:

It seem unity build process will generate this folder in new android build for cmake


thorrsson:
Merged as part of https://github.com/github/gitignore/pull/4525

PR #4506, opened 2024-09-09
Updated UnrealEngine.gitignore with common IDE cache/user files

sudotman:
Unreal Engine supports a variety of IDEs for it's development. The current gitignore includes common Visual Studio files in the gitignore, the file doesn't include arguably the second most common choice for development i.e. [JetBrains Rider](https://www.jetbrains.com/lp/rider-unreal/).


MalikuMane:
Should ignore Makefile too, used with Null Source Code Access.

PR #4508, opened 2024-09-12
Update Cake comment line to match other comment lines

Lamparter:

I was looking at the `.gitignore` for a .NET project when I noticed that the "uncomment these" lines for Cake had an extra space in comparison to that of other tools and exclusion recommendations. This change is hardly anything but nevertheless it's something I noticed 😁

_This is not a major change and so does not require docs, just something I noticed_ 😅


Lamparter:
Oops, looks like the Gitmoji extension got a little ahead of itself 💩

PR #4509, opened 2024-09-18
Add vscode cache folder

mcgodamn:
Add vscode cache folder for vscode users

### DIFF
--- a/Unity.gitignore
+++ b/Unity.gitignore
@@ -25,6 +25,7 @@
 
 # Visual Studio cache directory
 .vs/
+.vscode/
 
 # Gradle cache directory
 .gradle/


Unity.gitignore
```diff
@@ -25,6 +25,7 @@
 
 # Visual Studio cache directory
 .vs/
+.vscode/
```
tbeu, line 28: It's not a cache directory. We have VS code settings and extensions recommendations commited and not ignored. I wonder what the recommandation is though.

koppor:
Do not mix up Visual Studio and Visual Studio Code. These are very different products.

See https://github.com/github/gitignore/blob/main/Global/VisualStudioCode.gitignore for the VS Code ignore list.

PR #4517, opened 2024-10-11
add libogc.gitignore (wii homebrew development)

sctech-tr:
as a dev who recently worked with libogc on wii projects, i’ve noticed that the existing .gitignore file for plain C is minimal and does not account for several common temporary system files generated during the build process, nor the compiled executable for the Wii. By creating a brand new .gitignore file just for libogc, this issue can be fixed. And wii homebrew development is not common, so it's in the community directory. that's all

https://wiibrew.org/wiki/DevkitPPC
https://wiibrew.org/wiki/Libogc
not so good but it works

If this is a new template:

https://github.com/devkitPro/libogc

let me know if you have any suggestions

### DIFF
--- a/community/libogc.gitignore
+++ b/community/libogc.gitignore
@@ -1,0 +1,40 @@
+# Ignore build directories and files
+build/
+*.dol
+*.elf
+
+# Ignore Wii-specific metadata files
+meta.xml
+icon.png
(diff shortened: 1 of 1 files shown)


wirecat (approved):

PR #4520, opened 2024-10-16
Add OpenTofu to community

mediowen:

I'm a heavy user - basically just easier addition of OpenTofu-specific extensions, now that the project has extended that part of Terraform.
Theoretically these could be added to Terraform.gitignore, but as the OpenTofu project progresses there may be more differences that aren't worth having both sides. Open to suggestion though.

* https://opentofu.org/docs/intro/whats-new/#override-files-for-opentofu-keeping-compatibility
* https://opentofu.org/docs/language/files/#file-extension
* https://opentofu.org/docs/language/files/override/

If this is a new template:

 - **Link to application or project’s homepage**: https://opentofu.org/

### DIFF
--- a/community/OpenTofu.gitignore
+++ b/community/OpenTofu.gitignore
@@ -1,0 +1,42 @@
+# Local .terraform directories
+**/.terraform/*
+
+# .tfstate files
+*.tfstate
+*.tfstate.*
+
+# Crash log files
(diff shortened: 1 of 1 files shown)


thorrsson (approved):
❤️ Thanks for the contribution @mediowen!

PR #4524, opened 2024-10-19
Update Node.gitignore with ignore for bun files

indprtech:
Bun is an new JavaScript runtime with an new package manger and has a lock file named bun.lockb which is not use case for small projects.


zumbiepig:
`bun.lockb` is the equivalent of `package-lock.json` in node. It is not meant to be ignored by source control.

PR #4533, opened 2024-11-06
Add rye to Python.gitignore

s-smits:

Adding Rye-specific entries to ensure proper version control behavior when using Rye as the Python package manager. This prevents committing virtual environments, lock files, and version specifications that should be managed locally.

- [Rye Documentation](https://rye.astral.sh/guide//)

If this is a new template:

 - **Link to application or project's homepage**: [Rye - An Experimental Package Management Solution](https://rye.astral.sh)

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -101,6 +101,15 @@ ipython_config.py
 #   https://python-poetry.org/docs/basic-usage/#commit-your-poetrylock-file-to-version-control
 #poetry.lock
 
+# Rye - Python packaging and dependency management tool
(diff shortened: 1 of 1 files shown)


Roxyfundz1 (approved):

wirecat:
@s-smits would you be comfortable with commenting out the entries? I think we would prefer these entries to be an opt-in operation for users if they are use rye.

s-smits:
All good as is, thank you. _rye_ usage has been replaced by _uv_ for most.

s-smits:
Closing.

PR #4535, opened 2024-11-09
Fix "build" source folder being ignored (Gradle)

niklaswimmer:

I used the template and my files in `buildSrc/src/**/build/` (my package name includes `build`) were being ignored. I tried around and found out that with this change they are no longer ignored. This was originally introduced in #3370, but the previous configuration only works if `src` is a top level directory.

> If there is a separator at the beginning or **middle** (or both) of the pattern, then the pattern is relative to the directory level of the particular .gitignore file itself.

From https://git-scm.com/docs/gitignore, emphasizes mine.

### DIFF
--- a/Gradle.gitignore
+++ b/Gradle.gitignore
@@ -1,6 +1,6 @@
 .gradle
 **/build/
-!src/**/build/
+!**/src/**/build/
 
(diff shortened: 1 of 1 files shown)


wirecat (approved):

PR #4538, opened 2024-11-13
Enhance C.gitignore template by adding IDE files

Husam-Altamimi:
Added additional IDE files and cross-platform build directories to the C.gitignore template

### DIFF
--- a/C.gitignore
+++ b/C.gitignore
@@ -1,3 +1,14 @@
+# IDE and Editor Files
+.vs/
+.vscode/
+.idea/
+.project
+.cproject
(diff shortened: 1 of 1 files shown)


thorrsson:
As @tbeu mentioned we try to keep editor specific changes in their own files.
https://github.com/github/gitignore?tab=readme-ov-file#contributing-guidelines

```
Please consider the scope of your change. If your change is specific to a certain language or framework, then make sure the change is made to the template for that language or framework, rather than to the template for an editor, tool, or operating system.
```

PR #4545, opened 2024-12-03
Added uv.lock alongside pdm and poetry

KalleDK:

Just spotted that pdm and poetry was there, so added uv

https://docs.astral.sh/uv/pip/compile/


wirecat:
Thank you for the contribution @KalleDK! ❤️

PR #4546, opened 2024-12-06
Ignore vscode cache directory

N3VERS4YDIE:
VSCode is widely used with Unity, but it generates a `.vscode/` folder containing user-specific settings (e.g., extensions, debug configurations). Including this in version control can cause unnecessary conflicts and bloat. Ignoring it keeps the repository clean and avoids sharing editor preferences.

### DIFF
--- a/Unity.gitignore
+++ b/Unity.gitignore
@@ -26,6 +26,9 @@
 # Visual Studio cache directory
 .vs/
 
+# Visual Studio Code cache directory
+.vscode/
+
 # Gradle cache directory
 .gradle/
 


wirecat:
👋 Hello @N3VERS4YDIE, I don't think we want to ignore this folder.

We have too many examples of teams that commit this folder intentionally in order to share things like launch configurations. That folder is not meant to hold user configuration, but instead [workspace configuration](https://code.visualstudio.com/docs/getstarted/settings#_workspace-settings).

PR #4547, opened 2024-12-11
SECURITY ISSUE: JENKINS_HOME.gitignore must ignore credentials.xml

cowlinator:
JENKINS_HOME.gitignore must ignore credentials.xml, as that file may contain encrypted secrets.

Also, job config should be included.

I am a Jenkins user. I decided to use this JENKINS_HOME.gitignore as I version controlled my jenkins configuration.

1. Then I discovered that credentials.xml, which contains encrypted passwords, private ssh keys, and other secrets. In general, it is bad practice and a security issue to include any secrets or credentials-- encrypted or not-- in git repositories. (And in this case, they can be [decrypted very easily](https://devops.stackexchange.com/questions/2191/how-to-decrypt-jenkins-passwords-from-credentials-xml)).
2. I also discovered that the job configurations were being ignored. Job config.xml files are important, sometimes containing a lot of intricate configuration, and in fact sometimes contain `<script>{code}</script>` which literally contains code. Code should be included in version control.

1. Not documented. Just enter any credential into a jenkins instance, and look at the credentials.xml file. You will see your encrypted credential. The need for this change is self-evident.
2. https://wiki.jenkins-ci.org/display/JENKINS/Administering+Jenkins

### DIFF
--- a/JENKINS_HOME.gitignore
+++ b/JENKINS_HOME.gitignore
@@ -18,10 +18,11 @@
 !/.gitignore
 !/*.xml
 
-# Ignore all files in jobs subdirectories except for folders.
+# Ignore all files in jobs subdirectories except for folders and job config.
 # Note: git doesn't track folders, only file content.
 jobs/**
 !jobs/**/
+!jobs/**/config.xml
 
 # Uncomment the following line to save next build numbers with config.
 
--- a/JENKINS_HOME.gitignore
+++ b/JENKINS_HOME.gitignore
@@ -4,11 +4,12 @@
 #  http://jenkins-ci.org/
 #  https://wiki.jenkins-ci.org/display/JENKINS/Administering+Jenkins
 #
-# Note: secret.key is purposefully not tracked by git. This should be backed up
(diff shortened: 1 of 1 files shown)


cowlinator:
@samrocketman , as original submitter of the JENKINS_HOME.gitignore, could you comment on this PR?

samrocketman:
Hi there, credentials.xml being tracked is intentional in the backup and not a security issue.  The secrets are encrypted with AES-128-CBC and are encrypted at rest like other configurations with sensitive information.

Jenkins uses the secret.key paired with config.xml to decrypt secrets.  I intentionally included it because a Jenkins configuration without credentials,xml is arguably a useless backup since you wouldn’t be able to restore Jenkins home from a backup.

To protect your configuration be sure to backup separately (and store separately) files `$JENKINS_HOME/secret*` which would include secret.key.

cowlinator:
Thank you for taking a look and for the explanation.

samrocketman:
No worries, it doesn’t hurt to review and second guess.  Good looking out.

PR #4562, opened 2025-01-07
Added .DS_Store Node.gitignore

sebazelonka:
Added macOS auto-generated file

This is a generally ignored file on all projects I've been working on.

### DIFF
--- a/Node.gitignore
+++ b/Node.gitignore
@@ -134,3 +134,6 @@ dist
 .yarn/build-state.yml
 .yarn/install-state.gz
 .pnp.*
+
+# macos
+.DS_Store


Scutua (approved):

koppor:
Please use https://github.com/github/gitignore/blob/main/Global/macOS.gitignore

sebazelonka:
Maybe I'm missing something but that file doesn't ignore most of the files/folders a node project requires to ignore.

wirecat:
We are trying to avoid a lot of OS specific contents in language/editor specific templates.
Instead we encourage people to use global gitignore templates. You can see ours [here](https://github.com/github/gitignore/tree/main/Global).

Closing for now.

PR #4563, opened 2025-01-10
modifies lib exclude, adds example lib include

mark-lazarides:
explicitly excludes the `_Libraries` directory, and provides example of adding sub directories back in. Should be used for including custom and third party (non -Beckhoff shipped) libraries, improving maintainability

The _Libraries directory was blanket ignored before. If using a project with non-Beckhoff libraries, it can become hard to maintain the project, as sourcing those libraries is challenging. The ignore file was modified to explicitly ignore the _Libraries directory (adding wildcarding to ensure it is expllicitly selected).

A commented out example is provided guiding users on how to include specific library directories.


mark-lazarides:
@sagatowski - as original submitter of the TwinCAT3.gitignore, could you comment on this PR?

sagatowski:
@mark-lazarides Is the idea here that I can always have a separate library with me all the time and version control it? In other words, libraries that are default not installed through the TwinCAT installer, like in your example TcUnit? (so that I can install it on a clean and fresh system without having to download it separately)?

mark-lazarides:
Yes. As there is no default package manager for versions less than 4024, this also seems the quickest way to enable specific recreation of solutions/deployments, particularly when managing multiple sources of libs

sagatowski:
@mark-lazarides I like the idea! I sometimes include the additional libraries in a separate folder in the repo if the user doesn't want to download them separately. Many of these end up as Miscellaneous (the default folder for everything unless you are using [library categories](https://alltwincat.com/2018/08/16/library-categories/)), and in that case you could even put the whole miscellaneous folder as an exception I guess (I at least think that there is a Misc folder created for all 3rd party libraries that are not using library categories).

wirecat:
@mark-lazarides These changes look good to me, but we do have a merge conflict.
Could you please resolve it? 🙇

dooleydevin (approved):
This looks great to me 😄 Thanks for the contribution!

PR #4442, opened 2024-05-06
Add `*.bbl-SAVE-ERROR` TeX.gitignore

zhiyuanzhai:
This PR is proposing to add `*.bbl-SAVE-ERROR` to `TeX.gitignore` for BibLaTeX.

**Reasons for making this change:**

Recently I found that `*.bbl-SAVE-ERROR` files are generated in my LaTeX working directory, and I would like to exclude them from my git repository.

**Links to documentation supporting these rule changes:**

I do not recall when such files first appeared, but they appear in my working directory.

If this is a new template:

 - **Link to application or project’s homepage**: Not a new template.

PR #4567, opened 2025-01-15
Added "*.bcf-SAVE-ERROR" to TeX.gitignore

krewast:

".bcf" stands for "biblatex control file".
Files with this suffix are already getting ignored.
If there is an error, another file with the suffix "*.bcf-SAVE-ERROR" is created, which should also be ignored.

This kind of error file seems to be a "cousin" of files with the suffix "*.bbl-SAVE-ERROR" which has been added fairly recently: https://github.com/github/gitignore/pull/4442

I've seen files with both suffixes, so I'm adding the rule for the one that is currently still missing.

### DIFF
--- a/TeX.gitignore
+++ b/TeX.gitignore
@@ -28,6 +28,7 @@
 *.bbl
 *.bbl-SAVE-ERROR
 *.bcf
+*.bcf-SAVE-ERROR
 *.blg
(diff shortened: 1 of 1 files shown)


thorrsson (approved):
Thanks!

PR #4570, opened 2025-01-18
Add .vscode/ to Maven.gitignore file

BlackBossX:
What I added:
Added the following to Maven.gitignore to ignore VS Code-specific files:
```
### VS Code ###
.vscode/
```
Why:
This repository is very helpful, and I use `Maven.gitignore` often. Every time I use it with VS Code, I need to manually add `.vscode/`to the file. I thought it would be useful to fork the repo and make this change so it can help others as well.

Thank you for considering this update!

### DIFF
--- a/Maven.gitignore
+++ b/Maven.gitignore
@@ -15,3 +15,7 @@ buildNumber.properties
 .project
 # JDT-specific (Eclipse Java Development Tools)
 .classpath
+
+### VS Code ###
+.vscode/
+


koppor:
@BlackBossX VS Code is another applicatoin than Maven. Try to use https://github.com/github/gitignore/blob/main/Global/VisualStudioCode.gitignore.

wirecat:

Closing for now.

PR #4573, opened 2025-01-20
TeX: Add support for spelling package

koppor:

I am using TeX, lualatex in particular.  The LaTeX package [spelling](https://ctan.org/pkg/spelling) offers to highlight mis-spelled words. The list of the mis-spelled words is put into a file `<jobname>.spell.bad`. These are generated by spell checker - and thus should not be versioned.

Moreover, all words of the document are put into `<jobname>.spell.txt`, which also should be ignored.

http://mirrors.ctan.org/macros/luatex/generic/spelling/spelling-doc.pdf -> pages 3 and 8

image

image

### DIFF
--- a/TeX.gitignore
+++ b/TeX.gitignore
@@ -204,6 +204,7 @@ _minted*
 
 # spelling
 *.spell.bad
+*.spell.txt
 
 # svg
 svg-inkscape/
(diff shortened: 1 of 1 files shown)


wirecat (approved):

PR #4575, opened 2025-01-24
Add Rider, VS Code, Plastic cache directory

StarrySky-skyler:

Add Rider, vscode, plastic scm configuration and cache folders to unity.gitignore

_TODO_

https://rider-support.jetbrains.com/hc/en-us/articles/207097529-What-is-the-idea-folder

### DIFF
--- a/community/Nix.gitignore
+++ b/community/Nix.gitignore
@@ -1,3 +1,6 @@
 # Ignore build outputs from performing a nix-build or `nix build` command
 result
 result-*
+
+# Ignore automatically generated direnv output
+.direnv


thorrsson:
Thanks for the contribution @StarrySky-skyler 

Per our [contribution guidelines](https://github.com/github/gitignore?tab=readme-ov-file#contributing-guidelines)

```
Please consider the scope of your change. If your change is specific to a certain language or framework, then make sure the change is made to the template for that language or framework, rather than to the template for an editor, tool, or operating system.
```

The VScode and IntelliJ exclusions shouldn't be in this file, and I understand that it appears a `.vs` entry made it into a previous PR I don't think we can accept this change as is

thorrsson (changes requested):
^

PR #4579, opened 2025-01-25
Create monadçay

klckenan2147:
Maybe reprents a value that may or may not be present. This is natural when a value, or the result of a function, may or may not result in something meaningful. Traditionally, null or undefined is used for such cases, but these are ticking time bombs.

_TODO_

_TODO_

If this is a new template:


klckenan2147:
var maybe = monad.maybe(null);

thorrsson:
Definitely not a .gitignore change. Closing

PR #4525, opened 2024-10-19
Update Unity.gitignore

StephenHodgson:
**Reasons for making this change:**

Ignore `.utmp` directory generated by unity 6 projects.

PR #4580, opened 2025-01-28
Update Unity.gitignore

camille-avatarmedical:
With the new release of Unity 6 it appears a new temporary folder is created when building for Android. It can be safely deleted.

Unfortunately it is not documented. This [stackoverflow link talk about it](https://github.com/Unity-Technologies/a11y-public-sample/blob/main/.gitignore)
Note that even [Unity samples project](https://github.com/Unity-Technologies/a11y-public-sample/blob/main/.gitignore) ignore this as well


MGaetan89:
There's already a PR for that: #4502

thorrsson:
merged in https://github.com/github/gitignore/pull/4525

PR #4582, opened 2025-02-01
Unity Visual Scripting auto generated files

WybrenKoelmans:
These files are re-generated frequently and do not have to be committed according to documentation.

https://docs.unity3d.com/Packages/com.unity.visualscripting@1.9/manual/vs-version-control.html


Shannja (approved):
this should get approved as these files aren't needed

PR #4585, opened 2025-02-05
Aded /.utemp/ in Unity.gitignore

overmet15:
This folder is created when using Unity 6, i personaly added it in my projects.

This folder contains folders like "Debug", "RelWithDebInfo" and "tools".

It doesn appear on every single unity version but it happens only in Unity 6.

I added this folder because it doesnt affect project in any way and it creates temporary files.

I didnt found any. It just happened to me multiple times and i had to manualy add it in .gitignore

### DIFF
--- a/Unity.gitignore
+++ b/Unity.gitignore
@@ -9,6 +9,7 @@
 /[Bb]uilds/
 /[Ll]ogs/
 /[Uu]ser[Ss]ettings/
+/.utmp/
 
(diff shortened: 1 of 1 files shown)


MGaetan89:
This seems like a duplicate of #4502

thorrsson:
merged in https://github.com/github/gitignore/pull/4525

PR #4586, opened 2025-02-17
Ignore .vsconfig when outside of .vs dir

JonesWebConsulting:
Sometimes .vsconfig is located outside the .vs/ directory and won't be picked up by the ignore pattern above. Confirmed on Unity 6.


wirecat:
Could you please provide some more documentation and justification on why this specific file should be ignored?

PR #4590, opened 2025-02-21
Add .gitignore for Kirby CMS

chrfickinger:
Add a new .gitignore file for Kirby CMS ([getkirby.com](https://getkirby.com))


wirecat:
Could you please provide more documentation to justify the contents of the gitignore template?

Also, we prefer not to include OS specific rules like .DS_Store or Icon. We encourage people to use global gitignore templates and have our collection [here](https://github.com/github/gitignore/tree/main/Global) for OS specific rules.

wirecat:
@chrfickinger do they have official docs that discuss what a gitignore template should be? Or examples? Or perhaps some examples from people using Kirby?

PR #4591, opened 2025-02-22
Unity: MacOS files ignore

satelllte:

When Unity projects are being opened through MacOS's Finder file exporer, it's often the case that these files appear. It's pretty common to ignore them for code bases.


wirecat:
We are trying to avoid a lot of OS specific contents in language/editor specific templates.
Instead we encourage people to use global gitignore templates. You can see ours [here](https://github.com/github/gitignore/tree/main/Global).

Closing for now.

PR #4594, opened 2025-02-26
Update MATLAB.gitignore to exclude .MATLABDriveTag

guohaiping:
This change adds .MATLABDriveTag to the .gitignore file.
.MATLABDriveTag is a system file generated by MATLAB Drive for internal synchronization purposes.
Excluding it from Git ensures a cleaner repository and avoids unnecessary version control of system-generated files.

### DIFF
--- a/Global/MATLAB.gitignore
+++ b/Global/MATLAB.gitignore
@@ -29,3 +29,6 @@ codegen/
 
 # Octave session info
 octave-workspace
+
+# MATLAB Drive tag file
+.MATLABDriveTag


originalsigbe (commented):
Nice

wirecat:
@guohaiping In the future, please include references to backup the claim. 🙇 
In this case, I was able to find a [staffer suggesting these can be added to the .gitignore](https://www.mathworks.com/matlabcentral/answers/1627090-unseen-file-matlabdrivetag-is-seen). When we can't easily find justification, we'll close the PR.

PR #4595, opened 2025-02-27
Adding uv tool folders to python ignore

johny22:

When creating a project with python [uv](https://github.com/astral-sh/uv) tool this folders aren't be ignored


offbyone (commented):
I'm not sure I've ever seen any of these directories craeted in the base of a Python package using uv. Can you elaborate on how you get them to appear?

wirecat:
No dialogue from the author after feedback was given, closing for now.

PR #4597, opened 2025-03-04
Create Luau.gitignore

m4fh:
As far as I can tell, there is no official documentation outlining which files should be ignored when using the [Luau](https://github.com/luau-lang/luau) language. However, there are a few files that can be generated by Luau (or by `perfgraph`, a recommended tool included in the language repository), namely coverage.out, [profile.out, profile.svg](https://luau.org/profile), and trace.json—the latter specifically requiring the compiler to be built with `LUAU_ENABLE_TIME_TRACE` enabled. All of these files have been included in this PR.

Luau is also commonly used alongside a tool called [Rojo](https://github.com/rojo-rbx/rojo) and an LSP like JohnnyMorganz's [luau-lsp](https://github.com/JohnnyMorganz/luau-lsp). Developers generally choose to ignore files like *.rbxlx.lock, *.rbxl.lock, and sourcemap.json. But I think this would require a separate PR (perhaps a Rojo.gitignore?).

I’m not affiliated with the Luau project, but I believe this .gitignore file could be quite helpful to others, just as it has been to me.

### DIFF
--- a/Luau.gitignore
+++ b/Luau.gitignore
@@ -1,0 +1,9 @@
+# Code coverage
+coverage.out
+
+# Profiling
+profile.out
+profile.svg
+
+# Time trace
+trace.json


wirecat:
@m4fh Good instincts. We would prefer a seperate template for Rojo.

Luau.gitignore
wirecat, line 1: Could you add a small header to the top that points to the Luau repo and optionally have a short description?

That's my only note. Once we have that, I'm ready to merge 🚀.

m4fh: Done! Let me know what you think.

wirecat:
@m4fh Thank you for the contributions! Looks good. :)

PR #4619, opened 2025-04-23
Ignore all .terraform directories

rux616:
Changes to using the `foo/` trailing separator naming pattern to ensure that the directory itself will be ignored at all depths.

The current pattern ignores files inside the directory, but not the directory itself. This change makes it so the directory itself is ignored, and as a bonus, is cleaner as well.

I got my inspiration from the [Python.gitignore template](https://github.com/github/gitignore/blob/main/Python.gitignore) in this repository; it ignores the "\_\_pycache\_\_" directory (and others) by using the trailing separator method as well.

### Merge and Approval Steps
- [X] Confirm that you've read the [contribution guidelines](https://github.com/github/gitignore/tree/main?tab=readme-ov-file#contributing-guidelines) and ensured your PR aligns
- [x] Ensure CI is passing

### DIFF
--- a/Terraform.gitignore
+++ b/Terraform.gitignore
@@ -1,5 +1,5 @@
 # Local .terraform directories
-**/.terraform/*
+.terraform/
 
 # .tfstate files
 *.tfstate


wirecat:
@rux616 The prefix `**/` was important for ensuring we ignore the `.terraform` folder at any level within the repo.
Was there a reason you removed it? I'm under the impression we would like to keep it.

rux616:
> @rux616 The prefix `**/` was important for ensuring we ignore the `.terraform` folder at any level within the repo. Was there a reason you removed it? I'm under the impression we would like to keep it.

@wirecat Oh no, we absolutely want to continue ignoring the `.terraform` directory at all levels. My proposed change continues to do that, but specifically targets the directory _itself_, as opposed to the current pattern which targets just the contents of said directory. I am aware this is semantically the same thing, as git ignores directories without any contents to track, but my argument is that the proposed change is a slightly cleaner and more focused way of doing it. Additionally, in VSCode without this change, the `.terraform` directory does not show as ignored (dimmed) by source control in the Explorer view.

See these two bullet points in the [gitignore documentation's "pattern format" section](https://git-scm.com/docs/gitignore#_pattern_format):
> * If there is a separator at the end of the pattern then the pattern will only match directories, otherwise the pattern can match both files and directories.
> * For example, a pattern `doc/frotz/` matches `doc/frotz` directory, but not `a/doc/frotz` directory; however `frotz/` matches `frotz` and `a/frotz` that is a directory (all paths are relative from the `.gitignore` file).

wirecat:
> > @rux616 The prefix `**/` was important for ensuring we ignore the `.terraform` folder at any level within the repo. Was there a reason you removed it? I'm under the impression we would like to keep it.
> 
> @wirecat Oh no, we absolutely want to continue ignoring the `.terraform` directory at all levels. And my proposed change continues to do that, but specifically targets the directory _itself_, as opposed to the current pattern which targets just the contents of said directory. I am aware this is semantically the same thing, as git ignores directories without any contents to track, but my argument is that the proposed change is a slightly cleaner and more focused way of doing it. Additionally, in VSCode without this change, the `.terraform` directory does not show as ignored (dimmed) by source control in the Explorer view.
> 
> See these two bullet points in the [gitignore documentation's "pattern format" section](https://git-scm.com/docs/gitignore#_pattern_format):
> 
> > * If there is a separator at the end of the pattern then the pattern will only match directories, otherwise the pattern can match both files and directories.
> > * For example, a pattern `doc/frotz/` matches `doc/frotz` directory, but not `a/doc/frotz` directory; however `frotz/` matches `frotz` and `a/frotz` that is a directory (all paths are relative from the `.gitignore` file).

This appears to be a misunderstanding on my part. I also tested the change and confirmed any `.terraform/` directory is correctly ignored. Thank you for clarifying and contributing!

wirecat (approved):

rux616:
Thanks for merging! My apologies for any confusion, I should have included the relevant link to the gitignore documentation in my initial ticket.

PR #4669, opened 2025-06-15
Add .DS_Store to Swift, ObjC & Xcode gitignores

ChrisMash:

I've spent years manually adding `.DS_Store` to every `.gitignore` I've created from `Swift.gitignore` so those (seemingly) meaningless files don't get included, maybe I'm missing a reason that they've been left in, though they're excluded from the `Global/macOS.gitignore`, so assume they should also be excluded in these templates?

I assume it's ok to "ignore" the contribution guideline regarding only one template change per PR, seeing as the templates being changed say to also change the other ones? If not I can raise them separately, but it seems relevant to keep them all in one PR for a single change to each like this.

Hope this suffices, let me know if not!

Screenshot 2025-06-15 at 16 18 56

### Merge and Approval Steps
- [ ] Ensure CI is passing

### DIFF
--- a/Swift.gitignore
+++ b/Swift.gitignore
@@ -60,3 +60,6 @@ fastlane/report.xml
 fastlane/Preview.html
 fastlane/screenshots/**/*.png
 fastlane/test_output
+
+# macOS
+.DS_Store


ChrisMash:
Not sure how CI gets run, it doesn't seem to be happening automatically and I can't find anywhere that says how to trigger it or see any actions that obviously relate to it 👀

PR #4711, opened 2025-08-18
Move ModelSim out of Global

sebastianw:

The ignore expression `[_@]*` matches a broad range of files, for example Python `__init__.py` files.

If we include all Global gitignore files (as the documentation suggests) this will break Python projects.

### Merge and Approval Steps
- [X] Confirm that you've read the [contribution guidelines](https://github.com/github/gitignore/tree/main?tab=readme-ov-file#contributing-guidelines) and ensured your PR aligns
- [x] Ensure CI is passing


dooleydevin (changes requested):
I'm conflicted about this change. You're right that using this globally will ignore a bunch of files that should be in version control. My concern is that it's not just this line that can lead to problems -- line 12 (`*.html`) is another one that would cause a lot of issues.

I don't think the solution here should be to comment out this line. Rather, we should move this file out of `/Global` and into the root of the repository 👍

dooleydevin (approved):
Thanks for making this change!

Sum1cares (approved):

PR #4724, opened 2025-08-31
Add *.lcov to Python gitignore

elliotwutingfeng:
When pytest is run with `--cov-report=lcov`, `*.lcov` files will be created. These should be gitignored.

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -47,6 +47,7 @@ nosetests.xml
 coverage.xml
 *.cover
 *.py.cover
+*.lcov
 .hypothesis/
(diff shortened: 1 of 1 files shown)


elliotwutingfeng:
Random quote to keep this PR alive

> Many that live deserve death. And some that die deserve life. Can you give it to them? Then do not be too eager to deal out death in judgement. For even the very wise cannot see all ends.

elliotwutingfeng:
Nooo.

dooleydevin (approved):
Thank you for this! And sorry for the delay on the review 🙇

PR #4729, opened 2025-09-11
Node: Add react-router

deniz-blue:

react-router is a widely used library/framework for making routing based react applications. this PR adds react-router v7 framework mode's gitignore folders

- [reactrouter.com](https://reactrouter.com/)
- [.gitignore in default template](https://github.com/remix-run/react-router-templates/blob/4de6220f247744bee45112af027c7fc4437b128e/default/.gitignore)

React router has added **framework mode** on v7 which is based on Vite.

React router documentation unfortunately does not mention both of these:

- `.react-router` folder houses autogenerated `.ts` files (per-route) for type checking
- React router puts build output into the `build` folder (as opposed to `dist` with vite)

I have added `.react-router` into the gitignore but I am unsure about including the `build` folder. Should I keep it like this, include it or remove it? Currently its commented out (so if anyone needs to, they can uncomment it)

### Merge and Approval Steps
- [ ] Ensure CI is passing

### DIFF
--- a/Node.gitignore
+++ b/Node.gitignore
@@ -139,3 +139,7 @@ dist
 vite.config.js.timestamp-*
 vite.config.ts.timestamp-*
 .vite/
+
+# React Router
+.react-router/
+# build/


deniz-blue:
pending yes/no

dooleydevin (changes requested):
Appreciate the contribution! I think commenting out the `build/` rule is the right approach.

My only ask for this is that we explain the usage of the build folder and why it might be ignored in a comment as well, since it conflicts somewhat with line 38 where we ignore only `build/Release` 👍

PR #4706, opened 2025-08-10
Refactor .gitignore for MQL5: Clean up file extensions and improve clarity

softwaredevelop:
### Reasons for making this change

* I refined the `.gitignore` file to improve clarity and ensure it accurately ignores the relevant files for MetaTrader 5 (MQL5) projects.
* I removed unnecessary comments and example file extensions, which helps keep the file cleaner and easier to maintain.
* I made adjustments to the ignored files to make the `.gitignore` more precise and appropriate for future MQL5 projects.

### Links to documentation supporting these rule changes

* [MetaTrader 5 Documentation](https://www.metatrader5.com/en)
* [GitHub's Official Gitignore Templates](https://github.com/github/gitignore/blob/main/community/MetaTrader5.gitignore)

### If this is a new template

Link to application or project’s homepage: [MetaTrader 5 Homepage](https://www.metatrader5.com/en)

### Merge and Approval Steps

* [X] Confirm that you've read the [contribution guidelines](https://github.com/github/gitignore/tree/main?tab=readme-ov-file#contributing-guidelines) and ensured your PR aligns
* [X] Ensure CI is passing
* [ ] Get a review and approval from one of the maintainers

PR #4735, opened 2025-09-23
feat: Add mcp.json to Visual Studio gitignore exceptions

BenjaminMichaelis:

As mcp's are becoming more common, sharing a workspace configuration of mcp servers is becoming more and more common with the mcp.json file. (https://code.visualstudio.com/docs/copilot/customization/mcp-servers#_add-an-mcp-server)

Adding this file as an exclusion to the default template so that it can be utilized by the editors and shared via source control.

This file can also be picked up by Visual Studio as seen [here](https://learn.microsoft.com/visualstudio/ide/mcp-servers?view=vs-2022&WT.mc_id=8B97120A00B57354#file-locations-for-automatic-discovery-of-mcp-configuration)

- https://code.visualstudio.com/docs/copilot/customization/mcp-servers#_add-an-mcp-server
- https://learn.microsoft.com/visualstudio/ide/mcp-servers?view=vs-2022&WT.mc_id=8B97120A00B57354#file-locations-for-automatic-discovery-of-mcp-configuration

VSCode itself commits this file: https://github.com/microsoft/vscode/blob/main/.vscode/mcp.json

### If this is a new template

Link to application or project’s homepage: TODO

### Merge and Approval Steps
- [ ] Ensure CI is passing

### DIFF
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -412,6 +412,7 @@ FodyWeavers.xsd
 !.vscode/tasks.json
 !.vscode/launch.json
 !.vscode/extensions.json
+!.vscode/mcp.json
 !.vscode/*.code-snippets
 
 # Local History for Visual Studio Code
--- a/VisualStudio.gitignore
+++ b/VisualStudio.gitignore
@@ -407,13 +407,7 @@ MigrationBackup/
 FodyWeavers.xsd
 
 # VS Code files for those working on multiple tools
-.vscode/*
-!.vscode/settings.json
-!.vscode/tasks.json
-!.vscode/launch.json
-!.vscode/extensions.json
-!.vscode/mcp.json
-!.vscode/*.code-snippets
+!.vscode/*
 
 # Local History for Visual Studio Code
 .history/


baronfel:
TBH I'd go as far as removing the `.vscode` exclusion entirely - that file is meant for sharing/checking-in on a per-repo basis.

BenjaminMichaelis:
@baronfel updated in https://github.com/github/gitignore/pull/4735/commits/6d24007ad2d7683d49456a33d3a890d01e74681b - decided to make it a specific include to be clear rather than an omit, but open to opinion

mdkhakonmdkhakon12-afk:
#4706

BenjaminMichaelis:
Keep open please

dooleydevin (changes requested):
There are many valid reasons to include files like `settings.json` and `launch.json`, so we can't ignore the whole directory like this.

By our contribution guidelines, these rules really belong in the `Global/VisualStudioCode.gitignore` file, but they are already here and I'd rather not remove them as a part of this PR.

I'm also not sure that the `mcp.json` file really should be included in source control. From the VSCode docs [here](https://code.visualstudio.com/docs/copilot/reference/mcp-configuration#_configuration-structure) it seems they include an `inputs` parameter that is intended to include secrets:

> "inputs": []: an optional array of input variable definitions for sensitive information like API keys.

If we want to include this, I'd ask for a few changes:
1. Revert your last commit that ignores the whole `vscode` directory
2. Include the rule as a comment like so `#!.vscode/mcp.json`
3. Include a comment above it that explains why someone might choose to include it in their repository, and the risks of including secrets
4. Duplicate this change in `Global/VisualStudioCode.gitignore`

PR #4741, opened 2025-10-04
Julia: ignore *.jl.*.mem and CondaPkg's project-local environment

sundowatch:
Two rules for `Julia.gitignore`. The PR has been rescoped since it was opened — the earlier editor/OS and shared-library rules have been dropped, as those belong in `Global/` under the contributing guidelines.

**1. `*.jl.*.mem` — the existing `--track-allocation` rule matches nothing**

Julia writes allocation logs with the process ID embedded in the name: `<source>.jl.<pid>.mem`. The template currently only lists `*.jl.mem`, which does not match that. So the rule as written never ignores the files Julia actually produces.

The adjacent `--code-coverage` section already gets this right — it lists both `*.jl.cov` and `*.jl.*.cov`, because coverage output is named the same way. This change makes the allocation section symmetric with it.

Both filenames come from the same function in Julia's runtime, [`src/coverage.c`](https://github.com/JuliaLang/julia/blob/master/src/coverage.c):

```c
// jl_write_coverage_data
snprintf(stm, sizeof(stm), ".%d.cov", uv_os_getpid());
write_log_data(&coverageData, stm);

// jl_write_malloc_log
snprintf(stm, sizeof(stm), ".%d.mem", uv_os_getpid());
write_log_data(&mallocData, stm);
```

`write_log_data` appends that extension to the source path, so `foo.jl` becomes `foo.jl.51234.mem`. `jl_write_malloc_log` is the only place in the runtime that emits `.mem` files, and it always embeds the PID.

Reproducing it against the current template:

```console
$ curl -sO https://raw.githubusercontent.com/github/gitignore/main/Julia.gitignore
$ mv Julia.gitignore .gitignore
$ touch foo.jl.mem foo.jl.51234.mem foo.jl.cov foo.jl.51234.cov
$ git check-ignore -v foo.jl.51234.cov
.gitignore:3:*.jl.*.cov    foo.jl.51234.cov
$ git check-ignore -v foo.jl.51234.mem
$ echo $?
1     # not ignored
```

`foo.jl.51234.mem` is the file `julia --track-allocation=user` actually leaves behind, and it is currently untracked-but-not-ignored, so it shows up in `git status` and gets committed by accident.

**2. `.CondaPkg/` — a machine-specific Conda environment inside the project directory**

CondaPkg creates this directory in the active project and installs a full Conda environment into it. From [`src/resolve.jl`](https://github.com/JuliaPy/CondaPkg.jl/blob/main/src/resolve.jl):

```julia
# find the topmost env in the load_path which depends on CondaPkg
top_env = _resolve_top_env(load_path)
STATE.meta_dir = meta_dir = joinpath(top_env, ".CondaPkg")
...
conda_env = joinpath(meta_dir, "env")
```

The contents are downloaded binaries and absolute host paths — the same category as the `deps/usr/` and `deps/downloads/` rules already in this template. It is never correct to commit, and it is reconstructed by `CondaPkg.resolve()` on any machine.

This is not a niche package: 89 packages in the General registry depend on CondaPkg directly, and another 148 depend on PythonCall, which pulls it in. It is the standard mechanism for Python interop in Julia. CondaPkg's own repository [ignores it](https://github.com/JuliaPy/CondaPkg.jl/blob/main/.gitignore), as do [Flux.jl](https://github.com/FluxML/Flux.jl/blob/master/.gitignore), [Plots.jl](https://github.com/JuliaPlots/Plots.jl/blob/master/.gitignore) and [PythonCall.jl](https://github.com/JuliaPy/PythonCall.jl/blob/main/.gitignore). Because nothing else produces a `.CondaPkg` directory, the rule is inert for projects that don't use it.

- Julia manual, [`--track-allocation` / command-line options](https://docs.julialang.org/en/v1/manual/command-line-interface/#command-line-interface)
- Julia manual, [Profiling — memory allocation analysis](https://docs.julialang.org/en/v1/manual/profile/#Memory-allocation-analysis)
- Julia runtime source for both filenames: [`src/coverage.c`](https://github.com/JuliaLang/julia/blob/master/src/coverage.c)
- [`JuliaLang/julia/.gitignore`](https://github.com/JuliaLang/julia/blob/master/.gitignore) — carries both `*.jl.mem` and `*.jl.*.mem`
- CondaPkg.jl: [repository](https://github.com/JuliaPy/CondaPkg.jl) and [`.CondaPkg` creation in `src/resolve.jl`](https://github.com/JuliaPy/CondaPkg.jl/blob/main/src/resolve.jl)

### Scope

- One template only (`Julia.gitignore`), 4 added lines.
- Both rules go into a section describing the tool that generates them: `*.jl.*.mem` into the existing `--track-allocation` section, `.CondaPkg/` into a new section alongside the other package-generated rules.
- No duplicate rules. Neither pattern is matched by anything already in the template — verified with `git check-ignore` above.
- No editor, OS, or shared-library rules.

Happy to split this into two PRs, or to drop `.CondaPkg/` and land only the `*.jl.*.mem` fix, if either would be easier to review.

### Merge and Approval Steps
- [x] Ensure CI is passing


Sum1cares (approved):

sundowatch:
Still interested in this one — and I've rescoped it substantially, so it's worth a fresh look rather than reading the original diff.

**What changed:** the original version added `.history`, `.julia/`, `*.so`, `*.dylib` and `*.dll`. Re-reading CONTRIBUTING.md, those were wrong for this template — editor and OS artifacts belong in `Global/`, and the shared-library patterns were both mislabelled and risky for Julia projects that vendor binaries. All five are gone.

What's left is 4 lines in `Julia.gitignore`:

```diff
 # Files generated by invoking Julia with --track-allocation
 *.jl.mem
+*.jl.*.mem

 ...

+# Conda environment and metadata created in the project directory by the CondaPkg package
+.CondaPkg/
```

The first one is arguably a bug fix rather than a new rule. `--track-allocation` writes `<source>.jl.<pid>.mem`, so the existing `*.jl.mem` pattern doesn't match anything Julia actually produces — the real output files are still showing up in `git status`. The `--code-coverage` section right above it already handles this correctly with both `*.jl.cov` and `*.jl.*.cov`, since coverage output is named the same way. Both names come from the same helper in `src/coverage.c`. Upstream `JuliaLang/julia` carries both `.mem` patterns in its own `.gitignore`.

It reproduces in three commands against the template as it stands today:

```console
$ curl -s https://raw.githubusercontent.com/github/gitignore/main/Julia.gitignore -o .gitignore
$ touch foo.jl.51234.cov foo.jl.51234.mem
$ git check-ignore -v foo.jl.51234.cov foo.jl.51234.mem
.gitignore:3:*.jl.*.cov    foo.jl.51234.cov
                           # .mem file: no match
```

The second, `.CondaPkg/`, is a machine-specific Conda environment that CondaPkg installs into the project directory — same category as the `deps/usr/` and `deps/downloads/` rules already in the template, and never correct to commit. Full source references and prevalence numbers are in the PR description.

I've kept it to one template with no duplicate rules, and both patterns sit in a section that names the tool generating them. If it's easier to review, I'm glad to split this into two PRs, or drop `.CondaPkg/` and land just the `*.jl.*.mem` fix — that one stands on its own.

sundowatch:
@dooleydevin — apologies for the direct ping; you've handled most of the recent merges here, so you seemed like the right person to ask. Would you be able to take a look at this one when you have a moment?

It's down to 4 lines in a single template. The short version: `*.jl.mem` in the current template doesn't match what `--track-allocation` actually writes — Julia embeds the PID, so the real files are `<source>.jl.<pid>.mem` and that rule has been silently matching nothing. `*.jl.*.mem` fixes it and mirrors the `*.jl.*.cov` line already sitting in the coverage section directly above. The other line ignores `.CondaPkg/`, the Conda environment CondaPkg installs into the project directory.

Glad to drop the `.CondaPkg/` line and land just the one-line fix if that's an easier call. No urgency — flagging it mainly because it's been open a while and the stale bot has come through twice.

dooleydevin (approved):
LGTM and thank you for the patience on the review 🙇

PR #4759, opened 2025-11-18
Fix missing wildcard for xcuserdata folder

anisselbd:
Add a missing wildcard to the xcuserdata folder pattern.

This ensures that all user-specific Xcode data folders inside `xcuserdata` are properly ignored, preventing accidental commits of local workspace data.

This change aligns with common .gitignore patterns used in Xcode projects.

### DIFF
--- a/Global/Xcode.gitignore
+++ b/Global/Xcode.gitignore
@@ -1,2 +1,2 @@
 ## User settings
-xcuserdata/
+xcuserdata/*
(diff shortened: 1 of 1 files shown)


anisselbd:
Let me know if you'd like any adjustments , happy to update the PR.

anisselbd:
Thanks for the clarification , that makes sense.  
Reverted the change to keep the pattern consistent with the standard Xcode .gitignore behavior.

PR #4780, opened 2026-01-03
Update Unity.gitignore to add new ignore patterns

SzymonHalucha:
Added entries to ignore `.slnx` files and scene dependency cache.


dooleydevin (approved):
Thanks for this!

PR #4786, opened 2026-01-18
[AL] Ignore .bclicense file

christianbraeunlich:

Microsoft deprecated the License files in the .flf format: https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/upgrade/deprecated-features-platform#flfformat. It will be unsupoorted in Business Central 2023 release wave 1. The format remains supported in older versions.

Transition to using the newer .bclicense format instead.

--> we should ignore the `.bclicense` files aswell so that they don't get committed by accident.

- https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/cside/cside-upload-license-file#flf-and-bclicense-file-types: `So we recommend that you use the .bclicense file if you've gotten one.`


dooleydevin (approved):

PR #4808, opened 2026-03-03
Add gitignore to exclude sisyphus directory files (oh-my-openagent formerly known as oh-my-opencode)

malheleco:
oh-my-opencode excludes the sisyphus working directory in their own repo, so it seems to be best practice. Also, the directory captures agent/runtime state, so it should not be shared.

Add support for the opencode extension "oh-my-opencode"

https://github.com/code-yeongyu/oh-my-opencode/blob/dev/.gitignore

### If this is a new template

Link to application or project’s homepage: https://github.com/code-yeongyu/oh-my-opencode
This project is **not** affiliated with the website at ohmyopencode.[com]

### Merge and Approval Steps
- [ ] Ensure CI is passing


dooleydevin (changes requested):
These rules seem valid to me. Just a couple requests before this is ready to merge:
1. Can you move this under the `Global` directory? That is where we typically include templates like these.
2. It looks like OhMyOpenCode is now OhMyOpenAgent? I believe this file should be renamed to match the project renaming.

dooleydevin (approved):
Thanks for the contribution and the quick turnaround on the requested changes!

PR #4826, opened 2026-03-28
TeX: add ConTeXt utility cache file (.tuc, .tui & tuo)

gxono:
ConTeXt generates .tuc files (utility cache) during each run. 

These can be removed using the command line [--purgeall](https://wiki.contextgarden.net/Input_and_compilation/Executable_scripts_and_auxiliary_files/context_script#--purgeall).

ConTeXt MKII generates .tui and .tuo with the generation of [ToC](https://wiki.contextgarden.net/Document_structure_and_headlines/Table_of_contents).

These files store document metadata between compilations and are not part of the source, so they should be ignored.

### DIFF
--- a/TeX.gitignore
+++ b/TeX.gitignore
@@ -75,6 +75,10 @@ acs-*.bib
 # comment
 *.cut
 
+# context
+*.tua
+*.tuc
+
 # cprotect
 *.cpt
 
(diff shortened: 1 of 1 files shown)


dooleydevin:
@gxono can you provide documentation for this change? I don't doubt these files should be ignored, but I have not been able to verify their purpose from a quick search through the ConTeXt documentation.

dooleydevin (approved):
Appreciate the updated documentation! ✅

PR #4828, opened 2026-03-29
feat: create Lean.gitignore

quantum9Innovation:

This adds a standard gitignore template for the [Lean programming language](https://lean-lang.org/).

[This page](https://lean-lang.org/doc/reference/latest/Build-Tools-and-Distribution/Lake/) describes how Lake, Lean's package manager, works. Currently the template only ignores `.lake` which is where cache and other data is stored. Others are welcome to add/suggest other useful things to add to the gitignore template, though most Lean projects typically only have this in their gitignore.

### If this is a new template

Lean project homepage: <https://lean-lang.org/>

### Merge and Approval Steps
- [ ] Ensure CI is passing

### DIFF
--- a/Lean.gitignore
+++ b/Lean.gitignore
@@ -1,2 +1,2 @@
 # cache and other artifacts produced by Lean's Lake build tool
-.lake
+.lake/
--- a/Lean.gitignore
+++ b/Lean.gitignore
@@ -1,0 +1,2 @@
+# cache and other artifacts produced by Lean's Lake build tool
+.lake


Lean.gitignore
```diff
@@ -0,0 +1,2 @@
+# cache and other artifacts produced by Lean's Lake build tool
+.lake
```
dooleydevin: Since this expected to be a directory, I'd suggest making it explicit so that we don't match on files named `.lake` too:


```suggestion
.lake/
```

dooleydevin (changes requested):
Appreciate the contribution! Just one small request before this is good to merge 😄

Lean.gitignore
quantum9Innovation: Thanks for the catch, just updated!

quantum9Innovation:
Should be fixed now!

dooleydevin (approved):

PR #4833, opened 2026-04-03
Created macOS gitignore template.

wispgitgecko9:
Added a .gitignore template for unwanted macOS filesystem files.

There are certain files used by the macOS filesystem to define localized folder data that tend to clutter macOS users' repositories.

The following repositories have manually added these files to their gitignore:
- https://github.com/wispnet-org/eitkek/blob/main/.gitignore
- https://github.com/wispnet-org/wispnet/blob/main/.gitignore
- https://github.com/Maxelweb/ds-store-patrol
- https://github.com/tjharrop/test-DS_store-remover
- https://github.com/StefanoDeVuono/sturdy-potato

### If this is a new template

macOS is the widely used Apple desktop operating system.
https://apple.com/macos/

### DIFF
--- a/macOS.gitignore
+++ b/macOS.gitignore
@@ -1,0 +1,5 @@
+*.DS_Store
+.DS_Store
+.DS_Store?
+.localized
+.AppleDouble


dooleydevin:
Hey @wispgitgecko9! Appreciate the contribution, but there is already a macOS template over in https://github.com/github/gitignore/blob/main/Global/macOS.gitignore

I do see you included the `.localized` file which is missing in the existing template. I'm going to close out this PR, but if the existing one doesn't meet your needs because of that file feel free to open a PR that adds that update 😄

wispgitgecko9:
Hi @dooleydevin, thanks for the info! I'll go ahead and do that. The main reason I created this PR is because I don't see a macOS template option when creating a gitignore for a repository. Is there a way to implement that? Thanks.

dooleydevin:
@wispgitgecko9 [we recommend](https://github.com/github/gitignore?tab=readme-ov-file#folder-structure) copying relevant templates from `Global` into your global ignore file rather than configuring them on a per-repository basis. For now that's the best solution we have for this use-case.

Hope that helps!

wispgitgecko9:
@dooleydevin thanks, I understand now. I'll work on adding the `.localized` to the `Global` `macOS.gitignore` in a bit.

PR #4834, opened 2026-04-04
Update macOS.gitignore

wispgitgecko9:
Added .localized file to the gitignore.

Found another macOS filesystem file that may clutter repositories.

### DIFF
--- a/Global/macOS.gitignore
+++ b/Global/macOS.gitignore
@@ -1,9 +1,10 @@
 # General
 .DS_Store
+.localized
 __MACOSX/
 .AppleDouble
 .LSOverride
-Icon[]
+Icon[^M]
 
 # Thumbnails
 ._*
--- a/Global/macOS.gitignore
+++ b/Global/macOS.gitignore
@@ -4,7 +4,7 @@
 __MACOSX/
 .AppleDouble
 .LSOverride
-Icon[^M]
+Icon[]
 
 # Thumbnails
 ._*


Global/macOS.gitignore
```diff
@@ -1,9 +1,10 @@
 # General
 .DS_Store
+.localized
 __MACOSX/
 .AppleDouble
 .LSOverride
-Icon[]
+Icon[^M]
```
dooleydevin: Are you able to revert this change? The way this renders in the GitHub UI is a little odd, but I believe it should be a literal carriage return, not the literal `^M` characters that got put here instead.

Some context I found: https://superuser.com/questions/298785/icon-file-on-os-x-desktop

dooleydevin (changes requested):
Thanks for getting back to this `.localized` rule! Are you able to revert the change made to the Icon file rule? I'm hesitant to merge this PR until that is reverted.

Global/macOS.gitignore
wispgitgecko9: @dooleydevin I will try my best to do so. Regardless of how I have tried to edit the fie so far, it has turned into a newline whenever the file is edited in any way. Any suggestions on how to resolve that?

dooleydevin: I went ahead and reverted the line on my machine and pushed it 😄 hope that's alright!

dooleydevin (approved):
Thanks for the contribution!!!

Global/macOS.gitignore
wispgitgecko9: Yes, thanks for helping resolve that! Were you able to modify the special character via VSCode, or was it a different editor if you don't mind me asking?

wispgitgecko9:
Thanks for your help resolving it, @dooleydevin! Glad to help, and happy to contribute.

Global/macOS.gitignore
dooleydevin: I made the change in `vim` 👍 I just reset the file to the version in `main` then re-added the `.localized` rule. That carriage return character was tricky to deal with 😅

PR #4837, opened 2026-04-05
feat(nix): ignore nixos test history

haras-unicorn:

When testing NixOS modules, the interactive test driver leaves us with the `.nixos-test-history` file. This is common for anyone making NixOS module tests.

This file doesn't seem to be official but here is the line of code which creates it:

- https://github.com/NixOS/nixpkgs/blob/e347ac28905f77edcd1e9855dedcfb61e517f265/nixos/lib/test-driver/src/test_driver/__init__.py#L205

### If this is a new template

This is not a new template.

### Merge and Approval Steps
- [ ] Ensure CI is passing

### DIFF
--- a/Nix.gitignore
+++ b/Nix.gitignore
@@ -4,3 +4,6 @@ result-*
 
 # Ignore automatically generated direnv output
 .direnv
+
+# Ignore NixOS interactive test driver history
+**/.nixos-test-history


Nix.gitignore
```diff
@@ -4,3 +4,6 @@ result-*
 
 # Ignore automatically generated direnv output
 .direnv
+
+# Ignore NixOS interactive test driver history
+.nixos-test-history
```
dooleydevin: Based on the snippet you provided, it looks like this file shows up in the working directory of wherever the command is run. I'm not too familiar with the tooling, but if that's common you might want this to match in subdirectories too:


```suggestion
**/.nixos-test-history
```

dooleydevin (commented):
Thanks for the contribution @haras-unicorn! I left one suggestion that may apply here. I'll check back on this tomorrow and approve/merge it whether or not that update is committed 👍

Nix.gitignore
haras-unicorn: oh right thx :) - i force pushed ur suggestion

dooleydevin (approved):

PR #4840, opened 2026-04-10
Update Delphi.gitignore

chuacw:
Added RAD Studio 13.1 WinARM64EC directory and DUnitX results

### DIFF
--- a/Delphi.gitignore
+++ b/Delphi.gitignore
@@ -28,15 +28,16 @@
 
 # Default Delphi compiler directories
 # Content of this directories are generated with each Compile/Construct of a project.
-# Most of the time, files here have not there place in a code repository.
-#Win32/
-#Win64/
-#OSX64/
-#OSXARM64/
-#Android/
-#Android64/
-#iOSDevice64/
-#Linux64/
+# Most of the time, files here do not have a place in a code repository.
+Win32/
+Win64/
+OSX64/
+OSXARM64/
+Android/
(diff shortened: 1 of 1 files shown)


Delphi.gitignore
```diff
@@ -28,15 +28,16 @@
...
+# Most of the time, files here do not have a place in a code repository.
+Win32/
+Win64/
+OSX64/
+OSXARM64/
+Android/
+Android64/
+iOSDevice64/
+Linux64/
+WinARM64EC/
```
dooleydevin, line 40: Can you add the comments back to the start of these lines, or provide a justification for why you are removing them?

dooleydevin (changes requested):
One request before this is ready 👍

Delphi.gitignore
chuacw: Most of the time, files are generated in these directories, that should not be placed in a repo.
So, if files were generated in these directories, the default of having comments at the start of these lines would cause the files to be added to GitHub.
So, removing the comments would prevent this situation.

Sum1cares (approved):

nk6364261-ctrl (approved):

Delphi.gitignore
wirecat: >Most of the time

Not being universal is the operative key thing here. We endeavor to only accept rules that are truly universal.

PR #4890, opened 2026-08-05
Anchor MANIFEST to repo root to avoid matching subdirectories

mahyarmirrashed:

N/A

`MANIFEST` on it's own matches files and directories named "manifest" (case-insensitive) in the entire project. However, `/MANIFEST` matches for only the root, which is where 99% of all `MANIFEST` files that are being targeted by this rule are living. Therefore, I see this as an improvement.

N/A

### Merge and Approval Steps

### DIFF
--- a/Python.gitignore
+++ b/Python.gitignore
@@ -24,7 +24,7 @@ share/python-wheels/
 *.egg-info/
 .installed.cfg
 *.egg
-MANIFEST
+/MANIFEST
 
 # PyInstaller
 #   Usually these files are written by a python script from a template


Python.gitignore
```diff
@@ -24,7 +24,7 @@ share/python-wheels/
 *.egg-info/
 .installed.cfg
 *.egg
-MANIFEST
+/MANIFEST
```
wirecat: @mahyarmirrashed It'd be nice to have this apply to more than just the root for those repos that have multiple projects.

```suggestion
MANIFEST/
```

PR #4892, opened 2026-08-05
Add db-forest-config.xml to JetBrains gitignore

DKroot:

https://www.jetbrains.com/idea/

- db-forest-config.xml stores the IntelliJ Database tool window tree state (which data source  nodes are expanded, and in what order) — per-developer UI state that churns on  every expand/collapse, exactly like the already-ignored workspace.xml.
- Its payload tracks developer's local-only  connections, so committing the file might push dangling references into every teammate's IDE.

https://www.jetbrains.com/help/clion/database-tool-window.html#use_folders_in_the_tree

### Merge and Approval Steps

### DIFF
--- a/Global/JetBrains.gitignore
+++ b/Global/JetBrains.gitignore
@@ -3,6 +3,7 @@
 
 # User-specific stuff
 .idea/**/workspace.xml
+.idea/**/db-forest-config.xml
 .idea/**/tasks.xml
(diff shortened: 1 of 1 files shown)


wirecat:
JetBrains doesn't provide concrete guidance on whether this should be ignored by VCS or not. We have to be as universal as possible when accepting changes. Closing for now.

